PR #1, opened 2019-06-24
Add Parent POM

ppalaga:


pom.xml
```diff
@@ -0,0 +1,91 @@
...
+        <version>21</version>
+    </parent>
+
+    <groupId>org.camel.quarkus</groupId>
+    <artifactId>quarkus-camel-parent</artifactId>
+    <version>999-SNAPSHOT</version>
+    <packaging>pom</packaging>
+
+    <name>Camel Quarkus :: Parent POM</name>
+    <description>Camel extensions for Quarkus</description>
```
lburgazzoli: Wonder if it should be just Camel Quarkus

pom.xml
```diff
@@ -0,0 +1,91 @@
...
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache</groupId>
+        <artifactId>apache</artifactId>
+        <version>21</version>
+    </parent>
+
+    <groupId>org.camel.quarkus</groupId>
```
lburgazzoli: org.apache.camel

pom.xml
```diff
@@ -0,0 +1,91 @@
...
+    <artifactId>quarkus-camel-parent</artifactId>
```
lburgazzoli: camel-quarkus-parent

pom.xml
```diff
@@ -0,0 +1,91 @@
...
+    <version>999-SNAPSHOT</version>
```
lburgazzoli: Wondering if we should use a better version strategy rather than 999-SNAPSHOT

//cc @davsclaus

pom.xml
ppalaga: No problem with Camel Quarkus

pom.xml
ppalaga: +1

pom.xml
ppalaga: +1

pom.xml
ppalaga: I do not have a strong opinion on this. This is taken from Quarkus. IIRC their reasoning is that 999-SNAPSHOT will be always considered (by Maven) newer than any other version. Not sure this is motivated by CI or desktop development.

ppalaga:
8f07e29 resolves all comments by @lburgazzoli except for the  999-SNAPSHOT one.

pom.xml
ppalaga: Maybe one more advantage is that using single 999-SNAPSHOT version forever reduces the amount of clutter in the local Maven repo. This may save some space if the project is large.

A clear disadvantage is that this makes parallel working in older braches hard. Switching between 0.x and 1.x when reproducing bugs would always require full rebuild.

So I am quite ready to switch to 0.0.1-SNAPSHOT if you find it less controversial.

lburgazzoli: I don't have any strong opinion either, I would go for x.y.z-SNAPSHOT just for consistency with other camel projects

ppalaga: 0ca5a6f has 0.0.1-SNAPSHOT

ppalaga:
0ca5a6f resolves all comments by @lburgazzoli 
f50e786 is new, please review.

PR #2, opened 2019-06-25
Add Maven wrapper and .travis.yml

ppalaga:
@lburgazzoli and/or @davsclaus could you please review?

I find license-maven-plugin and .editorconfig very practical. They are best to add now, while the repo is quite empty. However, I admit the topic might be seen as controversial and I am ready to remove the commits for now so that we can move on with more important stuff. We can eventually discuss the two plugins later.


lburgazzoli:
@ppalaga I think we should cope with what we have on apache camel, so same checkstyle rules and license configuration.

ppalaga:
OK, I removed the two commits for now.

PR #12, opened 2019-07-01
Fix #11 Test the native mode on TravisCI

### LINKED ISSUE #11: Test the native mode on TravisCI

ppalaga:


.travis.yml
```diff
@@ -1,9 +1,19 @@
...
+    - GRAALVM_VERSION=19.0.2
+    - GRAALVM_HOME=/opt/graalvm-ce-$GRAALVM_VERSION
 cache:
   directories:
     - $HOME/.m2/repository
     - $HOME/.m2/wrapper
+    - $GRAALVM_HOME
+before_script:
+  - if [ ! -f "$GRAALVM_HOME/bin/java" ]; then
+      curl https://github.com/oracle/graal/releases/download/vm-${GRAALVM_VERSION}/graalvm-ce-linux-amd64-${GRAALVM_VERSION}.tar.gz | tar xvz -C /opt;
```
lburgazzoli: that's fine but we can eventually use the docker build provided by the quarkus maven plugin

ppalaga: > that's fine but we can eventually use the docker build provided by the quarkus maven plugin

What would be the advantage of that?

lburgazzoli: To always use the right version of Substrate VM for the version of quarkus we target

ppalaga: That sounds good and I'd be all for using it, but I am not finding any single piece of docs how to set it up. Do you happen to have an example at hand?

ppalaga:
Apache org's Travis has a long queue and the job has not started yet. But the job attached to my own account passed successfully already. Note the commitId 5cd0952  in the job: https://travis-ci.org/ppalaga/camel-quarkus/builds/553174383?utm_source=github_status&utm_medium=notification

.travis.yml
lburgazzoli: You can find some info here: https://quarkus.io/guides/building-native-image-guide.html

ppalaga:
3c106ef rebased on top of the current master. It is passing on Travis attached to my fork: https://travis-ci.org/ppalaga/camel-quarkus/builds/553207744?utm_source=github_status&utm_medium=notification

I propose to merge this as is (because it is ready and working) and eventually investigate the multi-stage Docker build later. I'd like to send the package rename PR just after the current PR gets merged.

PR #35, opened 2019-07-04
Fix #10 Produce test-jars of the integration tests

### LINKED ISSUE #10: Produce test-jars of the integration tests
This is to allow the (planned) Quarkus platform repo to run our tests.

ppalaga:
This is what I hope would make it possible to run our integration tests inside the planned Quarkus Platform. @aloubyansky could you please review, and eventually test with your Quarkus Platform branch?

I personally do not like the requirement to avoid `<scope>test</scope>` because this makes our integration tests non-idiomatic and thus non-usable as quickstarts. It would be nice if you @aloubyansky  could find a way to solve this on your side so that we may keep using `<scope>test</scope>`.

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -176,6 +176,11 @@
                     <artifactId>maven-resources-plugin</artifactId>
                     <version>3.1.0</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.codehaus.gmaven</groupId>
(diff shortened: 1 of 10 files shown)


aloubyansky:
I will look into that.

ppalaga:
Is this good to merge, @aloubyansky ?

aloubyansky:
I think so.

PR #49, opened 2019-07-09
chore(build): add Jenkinsfile

lburgazzoli:


ppalaga:
The motivation for this is because we have hit the memory limits of TravisCI in https://github.com/apache/camel-quarkus/pull/47 , right?

Anybody knows how much memory do the ASF Jenkins slaves have?

oscerd:
This was just an attempt. With the standard ASF infra approach we may need to use a personal access token of a personal account to be able to scan the github repo and check for PR. So it's not really feasible. I'm testing a different approach and probably we may remove the jenkinsfile later.

PR #52, opened 2019-07-09
Removed Jenkinsfile not used

oscerd:


ppalaga:
So the PR job on ASF Jenkins is now configured via Jenkins UI, right? What can I do to get read access to those settings?

oscerd:
You should already have read access to the ui with your ASF account. To modify the stuff you need to be committer.

ppalaga:
Yes, I can see a bit more when I log in, but https://builds.apache.org/job/camel-quarkus-pr/configure/ still returns 404.

But actually, there is a workaround: in the full build log I can see the commands that Jenkins uses to build the repo, so I am fine.

PR #77, opened 2019-07-19
Move test packages to org.apache.camel

lburgazzoli:


lburgazzoli:
@ppalaga added snapshot profile active by default for the moment as there were some changes in quarkus like moving to GraalVM 19.1.1 that have introduced some regressions, will disable this profile after quarkus 0.20.0

integration-tests/aws/src/main/java/org/apache/camel/quarkus/component/aws/CamelRoute.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.quarkus.it.camel.aws;
+package org.apache.camel.quarkus.component.aws;
```
ppalaga: I vote for keeping `it` somewhere in the integration test package names. `org.apache.camel.quarkus.component.aws.it` or `org.apache.camel.quarkus.it.component.aws` or some other variant. This is to avoid any potential naming clashes with the the platform code.

lburgazzoli: The reason I do usually keep the same name is to be able to access package private methods/classes for testing purpose, I can change it but I propose to do it later

ppalaga: > be able to access package private methods/classes for testing purpose

Yes, that's normal and OK for unit tests. OTOH integration tests should simulate how our end users use camel quarkus. We should not assume that they access non-public APIs.

lburgazzoli: That's partially true as we do not have any unit test for the native image side of things

ppalaga:
> added snapshot profile active by default for the moment as there were some changes in quarkus like moving to GraalVM 19.1.1 that have introduced some regressions, will disable this profile after quarkus 0.20.0

If there is any chance to re-think this proposal for the sake of build reproducibility and portability please consider using srcdeps instead of Maven SNAPSHOTs. I have sent a PR against your branch: https://github.com/lburgazzoli/apache-camel-quarkus/pull/1

lburgazzoli:
@ppalaga 0.20.0 is expected to be released in 5 days (week-end included) and after that release the snapshot profile won't be active by default as already stated above.

We need to be sure the code compile against the latest quarkus bits to avoid last minute bugs that would delay a first official release even further. Unfortunately srcdeps stick with a specific commit and won't give us an option to spot such problems.

PR #90, opened 2019-07-24
chore(it): cleanup it poms

lburgazzoli:


integration-tests/jdbc/pom.xml
```diff
@@ -31,6 +31,10 @@
     <name>Camel Quarkus :: Integration Tests :: JDBC</name>
     <description>Integration tests for Camel JDBC extension</description>
 
+    <properties>
+        <!-- Skip enforcer plugin as we want to use jboss marshalling for test class -->
+        <enforcer.skip>true</enforcer.skip>
```
ppalaga: Skipping the enforcer was not necessary in the JDBC itest, IIRC.

integration-tests/servlet/pom.xml
```diff
@@ -31,6 +31,11 @@
     <name>Camel Quarkus :: Integration Tests :: Servlet</name>
     <description>Integration tests for Camel Servlet component</description>
 
+    <properties>
+        <!-- Skip enforcer plugin as we want to use jboss marshalling for test class -->
+        <enforcer.skip>true</enforcer.skip>
```
ppalaga: Same here: it was passing with the enforcer on, IIRC. Are there other reasons for disabling it?

lburgazzoli: I found this comment on other poms so maybe it is not needed at all

lburgazzoli:
@ppalaga ok, I've disabled enforcer only were it make sense

PR #115, opened 2019-08-08
Fix #112 Make camel-quarkus-bom usable as a parent for user applications

### LINKED ISSUE #112: Make camel-quarkus-bom usable as a parent for user applications
The background story is that I am trying to write a user guide which is very hard without working examples. We could actually use our existing integration tests as examples but for that purpose, their pom hierarchy is not what the end user would need. So I am hereby proposing (1) to introduce a new POM (let's call it `camel-quarkus-application-parent` ) that could serve as a parent for end user applications and (2) use that POM as a parent in our integration tests. 

With (1) and (2) in place, the user could just copy an integration test out of our source tree to get a standalone example project (that is also granted to work because we run the tests regularly). We could further improve the user experience in the future by having per-itest source ZIPs on Maven Central that the user could just download and unpack instead of cloning the whole git repo. And even further, we could perhaps offer a maven plugin that would list, download and unpack the examples for a given Camel Quarkus version.

WDYT?

ppalaga:
@lburgazzoli this adopts your proposal to fix the BOM so that it can serve as parent for user apps. I moved just a single itest to `examples` making the BOM its parent to make sure that it actually works.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -323,6 +323,14 @@
                 <artifactId>camel-quarkus-xml-common</artifactId>
                 <version>${project.version}</version>
             </dependency>
+
+            <!-- Other third party dependencies -->
+            <dependency>
+                <groupId>org.eclipse.jetty</groupId>
+                <artifactId>jetty-client</artifactId>
+                <version>${jetty.version}</version>
+            </dependency>
+
         </dependencies>
     </dependencyManagement>
 
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -31,6 +31,10 @@
 
     <name>Camel Quarkus :: BOM</name>
 
+    <properties>
+        <camel-quarkus.version>0.1.1-SNAPSHOT</camel-quarkus.version><!-- kept in sync with project.version by the release plugin -->
+    </properties>
+
     <dependencyManagement>
         <dependencies>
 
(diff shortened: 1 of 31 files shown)


lburgazzoli:
can we keep the servlet test in the integration-tests and create an example in the examples folder ?

ppalaga:
I do not mind taking another one (which?) but having two close to identical apps, one under examples and one under itests does not sound right. 

Generally, I agree with your point that the itests may test aspects that have no didactic value for the app developers but as long as a given itest does not go too deep, it can well live under examples.

I believe examples should have tests. First, end users need to learn how to write those tests. And second, those tests should be run by the CI because otherwise there is no certainty that we will not accidentally break them at some point.

lburgazzoli:
> I do not mind taking another one (which?) but having two close to identical apps, one under examples and one under itests does not sound right.
> 
> Generally, I agree with your point that the itests may test aspects that have no didactic value for the app developers but as long as a given itest does not go too deep, it can well live under examples.

The problem is that when I run itest I have to remember that for some extensions I have also to build all the examples because my memory certainly fail and I won't remember what examples are also test.

> 
> I believe examples should have tests. First, end users need to learn how to write those tests. And second, those tests should be run by the CI because otherwise there is no certainty that we will not accidentally break them at some point.

It is good to have tests for examples but I would build examples i.e. once a day otherwise the build will take a very long time

ppalaga:
> > I believe examples should have tests. First, end users need to learn how to write those tests. And second, those tests should be run by the CI because otherwise there is no certainty that we will not accidentally break them at some point.
> 
> It is good to have tests for examples but I would build examples i.e. once a day otherwise the build will take a very long time

If we have no duplicates between itests and examples there will be no change in the execution time.

Daily jobs break the feedback and responsibility loop. If some breaking change happens, the author of the change cannot be notified automatically (and forced to deliver a fix) because it is generally not clear which change was the breaking one.

> > I do not mind taking another one (which?) but having two close to identical apps, one under examples and one under itests does not sound right.
> > Generally, I agree with your point that the itests may test aspects that have no didactic value for the app developers but as long as a given itest does not go too deep, it can well live under examples.
> 
> The problem is that when I run itest I have to remember that for some extensions I have also to build all the examples because my memory certainly fail and I won't remember what examples are also test.

If this is really a substantial issue, here is my compromise proposal:

Let's postpone the discussion about examples by moving it to a separate issue. In the meantime, before we agree on how we are going to maintain the examples, lets do the following: one of the itests (is servlet a good choice?) will have `camel-quarkus-bom` as its parent. The rest of the itests will inherit from `camel-quarkus-build-parent`.

WDYT?

lburgazzoli:
@ppalaga that's fine

ppalaga:
d3fb195 does things agreed upon with @lburgazzoli above.

ppalaga:
Is this good to merge?

lburgazzoli:
there's a conflict :)

PR #119, opened 2019-08-08
chore: replace custom properties binding implementation with PropertyBindingSupport from camel support

lburgazzoli:


extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/FastCamelContext.java
```diff
@@ -331,26 +331,27 @@ protected ReactiveExecutor createReactiveExecutor() {
...
-            }
+
+        CamelContextAware.trySetCamelContext(result, context);
+
+        Properties props = getPropertiesComponent().loadProperties();
+        if (props != null && !props.isEmpty()) {
+            PropertyBindingSupport.bindProperties(
+                this,
+                result,
+                new HashMap<String, Object>((Map) props),
```
ppalaga: Is there any good reason to allocate a new `HashMap` ? Simply `(Map) props` would compile too.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/FastCamelRuntime.java
```diff
@@ -77,21 +79,17 @@ public void stop() throws Exception {
...
         try {
             this.context = createContext();
 
-            // Configure the camel context using properties in the form:
-            //
-            //     camel.context.${name} = ${value}
-            //
-            RuntimeSupport.bindProperties(properties, context, PFX_CAMEL_CONTEXT);
+            if (properties != null) {
+                PropertyBindingSupport.bindProperties(context, context, new HashMap<String, Object>((Map) properties), PFX_CAMEL_CONTEXT);
```
ppalaga: Is there any good reason to allocate a new `HashMap` ? Simply `(Map) props` would compile too.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/FastCamelContext.java
lburgazzoli: no, leftover from another solution that I had before

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/FastCamelRuntime.java
lburgazzoli: fixed

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/FastCamelContext.java
lburgazzoli: fixed

lburgazzoli: @ppalaga 

ah no sorry, now I recall the reason: `PropertyBindingSupport.bindProperties` consumes the map so it removes every element that has been bound to the target leading to unexpected results.

pushed a new version that restores hash map, I need to work on the camel side to have a better solution but in the meantime I would merge the PR as it is as `PropertyBindingSupport` to support OGNL like property setting.

oscerd:
There is a conflict.

PR #121, opened 2019-08-09
create extension for camel's core cloud impl

lburgazzoli:


extensions/core-cloud/runtime/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-core-cloud-parent</artifactId>
+        <version>0.1.1-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-core-cloud</artifactId>
```
ppalaga: Why the name `camel-quarkus-core-cloud`? Are we going to have a separate `camel-quarkus-cloud` one day? 

The current one depends on `camel-cloud` which I assume is a Camel component, so why is this not called simply `camel-quarkus-cloud`?

integration-tests/core-impl/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-integration-tests</artifactId>
+        <version>0.1.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>camel-quarkus-integration-test-core-impl</artifactId>
```
ppalaga: Could you please explain what is this integration test supposed to do? There is no application inside, there are no assertions.

extensions/core-cloud/runtime/pom.xml
lburgazzoli: No, this is not a camel component but the implementation of the `org.apache.camel.cloud` interfaces that are part of camel-core-api

integration-tests/core-impl/pom.xml
lburgazzoli: test that it can build natively as by default camel-cloud won't build as native image without the substitutions

ppalaga: Are there any good reasons not to have an app and a scenario that asserts that the compiled code actually works?

lburgazzoli: we don not have yet a full story for the cloud bits in camel-quarkus as we need to bring some components such as the consul one. When we'll have it, I'm all to get rid of this integration "test"

ppalaga: Sounds good. Could you please add this info to a README?

PR #130, opened 2019-08-12
Fix #129 Properies evaluation broken after the introduction of Proper…

### LINKED ISSUE #129: Properies evaluation broken after the introduction of PropertyBindingSupport
The evaluation of properties does not work after a76c7526960ecabb1a1ebafc18c211a361bcf998 "chore: replace custom properties binding implementation with PropertyBindingSupport from camel support".

Reproducible through building the Twitter itest and running the jar manually. Clearly the twitter credentials need to be set as documented in the itest README. The scenario succeeds before a76c7526960ecabb1a1ebafc18c211a361bcf998 and fails after a76c7526960ecabb1a1ebafc18c211a361bcf998.

Please assign to me.

ppalaga:
…tyBindingSupport


ppalaga:
d4e5f6b uses a dedicated endpoint to query the resolvePropertyPlaceholders value of the timer component. There is no error, but it fails in the native mode. Will investigate tomorrow.

lburgazzoli:
thx @ppalaga

PR #134, opened 2019-08-13
Fix #133 Test netty4-http as a producer

### LINKED ISSUE #133: Test netty4-http as a producer
Required by https://github.com/apache/camel-quarkus/issues/132

Plz assign to me.

ppalaga:


integration-tests/netty4-http/src/main/java/org/apache/camel/quarkus/component/netty4/http/CamelRoute.java
```diff
@@ -19,10 +19,16 @@
 import org.apache.camel.builder.RouteBuilder;
 
 public class CamelRoute extends RouteBuilder {
+
+    public static final int PRODUCER_TARGET_PORT = 10102;
```
oscerd: Probably we'll need to re-use the AvailablePortFinder from Camel

integration-tests/netty4-http/src/test/java/org/apache/camel/quarkus/component/netty4/http/CamelTest.java
```diff
@@ -16,18 +16,119 @@
...
+            RestAssured.when().get(new URI("http://localhost:8999/producer")).then()
+                    .statusCode(200)
+                    .body(is("Hello from the mockserver!"));
+        }
+    }
+
+    /**
+     * An HTTP server always sending the same response.
+     */
+    static class Server implements Closeable, Runnable {
```
lburgazzoli: can't we just use something like undertow or the simple http server provided by openjdk ?

ppalaga: I did not want to expand the class path while we cannot use the test scope. I did not like `com.sun.net.httpserver.HttpServer` for not being portable accross Java vendors. I have not thought of Undertow (or vert.x) though. Let me have a look if it is there in the CP already.

lburgazzoli: Quarkus is not portable too and I think we need to introduce the test scope and push quarkus team to support it otherwise it is a mess (imho)

integration-tests/netty4-http/src/main/java/org/apache/camel/quarkus/component/netty4/http/CamelRoute.java
ppalaga: Not sure how? The test app AND the test classes need to agree on the same port number while they live in distinct JVM processes. We'd have to pass the random port via an endpoint running on some constant port - which I think is too complicated for what it brings. Or you maybe have a better idea?

integration-tests/netty4-http/src/test/java/org/apache/camel/quarkus/component/netty4/http/CamelTest.java
```diff
@@ -16,18 +16,119 @@
...
+
+        private volatile AtomicBoolean stop = new AtomicBoolean(false);
```
ppalaga: `volatile AtomicBoolean` - what a nice nonsense! :)

integration-tests/netty4-http/src/test/java/org/apache/camel/quarkus/component/netty4/http/CamelTest.java
lburgazzoli: Another option would be to reuse an existing netty http endpoint and ditch the server at all

ppalaga: > reuse an existing netty http endpoint

What an idea!

ppalaga: > reuse an existing netty http endpoint

done in b8ae8a1

lburgazzoli:
@ppalaga can you rebase ?

ppalaga:
3821fca rebased and fixed.

PR #138, opened 2019-08-15
Add tests to the twitter itest project

ppalaga:
The tests are run conditionally using `@EnabledIfEnvironmentVariable(named = "TWITTER_CONSUMER_KEY", matches = "[a-zA-Z0-9]+")` so they won't fail if the env is not set.

@oscerd do we have a read-write twitter account we could setup on Jenkins? `cameltweet` is read only, IIRC.


ppalaga:
Anybody knows what the Jenkins message `git clean -fdx" returned status code 143:` means?

lburgazzoli:
I think this can be greatly simplified by using consumer/producer templates instead i.e.  putting thing on a queue and polling.  I'd also go for using resteasy over servlet + camel rest so we can reduce components intra dependencies

integration-tests/twitter/src/main/java/org/apache/camel/quarkus/component/twitter/CamelRoute.java
```diff
@@ -16,25 +16,118 @@
...
+                    }
+                })
+                .to("twitter-search://_keywords_passed_via_CamelTwitterKeywords_header_")
+                .to("log:search?showAll=true")
+                .endRest()
+
+            .post("/directmessage")
+                .route()
+                .id("post-directmessage")
+                .to("twitter-directmessage://{{camel.component.twitter-directmessage.user}}")
```
lburgazzoli: `camel.component.twitter-directmessage` is a syntax used to automatically bind properties to a component so I would use a property like `twitter.user` to avoid confusion

ppalaga: Tried the original `twitter.user.name` and it did not work. IIRC the PropertyBindingSupport was complaining that the prop is not available. The JavaDoc of `PropertyBindingSupport.bindProperties()` says "Binds the properties with the given prefix to the target object" so it seems to consciously ignore all other props.

ppalaga:
> I think this can be greatly simplified by using consumer/producer templates instead i.e. putting thing on a queue and polling. 

Thanks for the tip. I am a Camel noob. 

> I'd also go for using resteasy over servlet + camel rest so we can reduce components intra dependencies.

Camel REST was easier to work with but yes, you are right that less deps is better. I can try resteasy tomorrow.

> We may also think about using `camel-mock`

Not sure how? Injecting to the test class works only JVM mode.

ppalaga:
> No we don't have credentials.

May I provide mine then?

integration-tests/twitter/src/main/java/org/apache/camel/quarkus/component/twitter/CamelRoute.java
lburgazzoli: properties binding support should not be involved here, it is just properties resolution in an endpoint. you may need to prefix the property with ‘integration.’ as at the moment there’s no full support for mo config (I’m working on it). using camel as prefix works as it is one of the whitelist prefix

lburgazzoli: properties binding support should not be involved here, it is just properties resolution in an endpoint. you may need to prefix the property with ‘integration.’ as at the moment there’s no full support for mo config (I’m working on it). using camel as prefix works as it is one of the whitelist prefix

lburgazzoli:
> 
> > We may also think about using `camel-mock`
> 
> Not sure how? Injecting to the test class works only JVM mode.

It is not a mock framework but an endpoint you can use for testing like you put it at the end of the route and use it validate exchanges

ppalaga:
> > > We may also think about using `camel-mock`
> > 
> > 
> > Not sure how? Injecting to the test class works only JVM mode.
> 
> It is not a mock framework but an endpoint you can use for testing like you put it at the end of the route and use it validate exchanges

I know `camel-mock` but all usages I saw so far used an instance of the component directly from the test. That would work in the JVM mode (e.g. through injecting the context to the test), but not in the native mode. You have some other way of using `camel-mock` in mind?

ppalaga:
05d6f85 eliminates the routes altogether as agreed with @lburgazzoli.

Please do not merge yet, just review. I have exceeded my Twitter rate limit and I do not know whether the current version is actually passing.

ppalaga:
96196a9 is passing for me locally. It will fail on Jenkins unless https://github.com/apache/camel-quarkus/pull/140 is merged first.

ppalaga:
Removed the test scopes in 8d87b17 to make this PR independent of #140 .

It should be good to merge.

ppalaga:
Is there anything missing here, @lburgazzoli @oscerd ?

PR #146, opened 2019-08-19
chore(test): add test to validate camel registry hooks in ArC

lburgazzoli:


extensions/core/deployment/pom.xml
```diff
@@ -45,6 +45,21 @@
...
         </dependency>
+
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+
+
```
ppalaga: Isn't this too many empty lines?

lburgazzoli: feel free to fix it

PR #154, opened 2019-08-21
Remove unecessary conditional discovery

lburgazzoli:


extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -139,12 +139,15 @@ AdditionalBeanBuildItem createProducers(
     @Record(ExecutionTime.STATIC_INIT)
     @BuildStep
     void init(
+            // TODO: keep this field as we need to be sure ArC is initialized before starting events
+            //       We need to re-evaluate the need of fire events from context once doing
+            //       https://github.com/apache/camel-quarkus/issues/9
```
ppalaga: Thanks for the comment. I was wondering why the bean container is there.

PR #175, opened 2019-09-05
Fix checkstyle violations

jamesnetherton:
I think the extension generator stuff contributes a little bit to this, because it doesn't add license headers to all files and some of the generated code violates CS rules.

I could maybe follow up with another PR to address that.


oscerd:
I'm ok with any configuration that doens't make the build fail for code style problems

PR #178, opened 2019-09-05
Fix #97 Inspect routes to detect classes that may require reflection

### LINKED ISSUE #97: Inspect routes to detect classes that may require reflection
Assuming we have a route that contains something like:

    .bean(OrderService.class, "doSomething")

Then the `OrderService` should be listed among the classes that require reflection.

ppalaga:
This is currently just a PoC to study the problem.

The PoC works in many simple cases, but I assume it would fail in many more complex cases.

How it works: To get the `org.apache.camel.model.Model` where the `RouteDefinitions` can be inspected for reflective beans, the PoC instantiates a CamelContext and all known RoutesBuilders at build time. See https://github.com/apache/camel-quarkus/pull/178/files#diff-40a8979f4631139bb657ee652820dde7R52 
All these instances are thrown away after the inspection.

The main problem with the above approach is that calling `RoutesBuilder.configure()` may fail due to non-availability of various resources (config values, injected beans, ...) that we simply do not have at build time.

Now I plan to look at options to somehow access the `Model` without needing to instantiate the CamelContext and the RoutesBuilders.

### DIFF
--- a/extensions/bean/deployment/src/main/java/org/apache/camel/quarkus/component/bean/deployment/BeanProcessor.java
+++ b/extensions/bean/deployment/src/main/java/org/apache/camel/quarkus/component/bean/deployment/BeanProcessor.java
@@ -16,8 +16,24 @@
  */
 package org.apache.camel.quarkus.component.bean.deployment;
 
+import java.util.List;
+
+import org.apache.camel.RoutesBuilder;
+import org.apache.camel.model.BeanDefinition;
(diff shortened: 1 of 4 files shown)


gnodet:
I think that while the idea is interesting, it actually conflicts with the fact that I've just removed the build time initialization and moved everything to runtime to simplify things and limit the current scope.

ppalaga:
Out of date

PR #186, opened 2019-09-05
Attempt to fix master

gnodet:


integration-tests/pom.xml
```diff
@@ -41,6 +41,7 @@
          -->
         <native-image.docker-build>true</native-image.docker-build>
         <native-image.container-runtime>docker</native-image.container-runtime>
+        <native-image.docker-build>quay.io/quarkus/ubi-quarkus-native-image:19.2.0</native-image.docker-build>
```
lburgazzoli: looks like there are two `native-image.docker-build`

lburgazzoli: quarkus 0.21.x is still on 19.1.x, only master has switched to 19.2.0 afaik

gnodet: Duly noted, thx for checking !

PR #189, opened 2019-09-06
Make xml and jaxb disabled by default and opt-in when depending on camel

gnodet:


extensions/xml-common/deployment/src/main/java/org/apache/camel/quarkus/component/xml/deployment/ModelFactoryProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.camel.quarkus.core.deployment.CamelRegistryBuildItem;
+import org.apache.camel.spi.ModelJAXBContextFactory;
+
+public class ModelFactoryProcessor {
+
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep
+    CamelRegistryBuildItem jaxbContextFactory(SupportRecorder recorder) {
+            return new CamelRegistryBuildItem(
+                    "log",
```
lburgazzoli: looks like this is not the right name :)

extensions/xml-common/runtime/src/main/java/org/apache/camel/quarkus/component/xml/runtime/graal/Target_org_apache_camel_quarkus_core_runtime_support_FastModel.java
```diff
@@ -0,0 +1,45 @@
...
+@Substitute
+public class Target_org_apache_camel_quarkus_core_runtime_support_FastModel extends DefaultModel {
+
+    public Target_org_apache_camel_quarkus_core_runtime_support_FastModel(CamelContext camelContext) {
+        super(camelContext);
+    }
+
+    @Substitute
+    public void addRouteDefinitions(InputStream is) throws Exception {
+        RoutesDefinition def = ModelHelper.loadRoutesDefinition(getCamelContext(), is);
```
lburgazzoli: For my understanding:
- are we are replacing `FastMode` with `DefaultModel` ? 
- why this method is implemented in the substitution and not in `FastMode` ?

gnodet: We're not replacing `FastModel` with `DefaultModel`.  This is just a class containing a substitution for the `addRouteDefinitions` method. The one in `FastModel` now simply throws an `UnsupportedOperationException`.
The goal is to have the default `FastModel` not being tied to xml/jaxb (thus unable to load xml routes) and have this feature automatically added when adding the dependency to the `xml-common` extension.

extensions/xml-common/deployment/src/main/java/org/apache/camel/quarkus/component/xml/deployment/ModelFactoryProcessor.java
gnodet: Good catch !

extensions/xml-common/runtime/src/main/java/org/apache/camel/quarkus/component/xml/runtime/graal/Target_org_apache_camel_quarkus_core_runtime_support_FastModel.java
lburgazzoli: But that means that the JVM mode and the native mode are behaving differently right ?

lburgazzoli: FWIW, what I recall is that a class level `@Substitute` means that the whole class need to be replaced that's why I see this as replacement for `FastModel`

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -70,10 +73,17 @@
...
             List<CamelRegistryBuildItem> registryItems,
+            List<CamelContextBuildItem> contexts,
             BuildProducer<RuntimeBeanBuildItem> runtimeBeans) {
 
+        Supplier<CamelContext> contextSupplier;
+        if (contexts.size() == 0) {
+            contextSupplier = new FastCamelContextSupplier();
+        } else {
+            contextSupplier = contexts.get(0).getContext();
+        }
```
lburgazzoli: We may need to throw an exception in case of contexts > 1

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/FastModel.java
```diff
@@ -46,6 +47,16 @@ public FastModel(CamelContext camelContext) {
...
 
+    @Override
+    public void addRouteDefinitions(InputStream is) throws Exception {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void addRestDefinitions(InputStream is, boolean addToRoutes) throws Exception {
+        throw new UnsupportedOperationException();
+    }
```
lburgazzoli: Maybe we can remove those two methods from Model's interface in RC2 and re-factorize them as helpers ?

davsclaus:
I assume this can be closed as there is a new PR #249

PR #191, opened 2019-09-06
Add an option to control context auto startup

lburgazzoli:


lburgazzoli:
@gnodet 

This PR is only for discussion and needed to be revised after we get #9 implemented. 

The goal here is to let the application control the camel context life-cycle as it seems there are some odd behavior when looking up beans from the ArC container if quarkus is not fully started.

What I have observed is that

    RuntimeRegistry::findByType(Vertx.class) 

Which under the hoods leverage the `BeanManager` from ArC, returns `null` when the camel context starts before Quarkus is fully started but returns a proper `Vertx` instance when the same method is invoked after Quarkus is fully started.

This maybe need to be investigated more with @mkouba

Issue #171, opened 2019-09-03
Create a Camel MQTT extension


zhfeng:

davsclaus:
The camel-paho component is possible more active maintained than mqtt client that is using an old fusesource library.

zhfeng:
Thanks @davsclaus , I will take a look at the camel-paho. and the fusesource libray looks like some issues with the native mode. Does camel-paho support in the Syndesis ?

PR #197, opened 2019-09-11
Add the camel-paho extension to support the MQTT

zhfeng:
This fixes: https://github.com/apache/camel-quarkus/issues/171

### DIFF
--- a/integration-tests/paho/src/test/java/org/apache/camel/quarkus/component/paho/it/PahoTest.java
+++ b/integration-tests/paho/src/test/java/org/apache/camel/quarkus/component/paho/it/PahoTest.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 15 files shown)


integration-tests/paho/src/test/java/org/apache/camel/quarkus/component/paho/it/PahoTest.java
```diff
@@ -0,0 +1,41 @@
...
+@QuarkusTest
+@QuarkusTestResource(ActiveMQTestResource.class)
+class PahoTest {
+
+    @Test
+    public void test() {
+        final String msg = UUID.randomUUID().toString().replace("-", "");
+        RestAssured.given() //
+            .contentType(ContentType.TEXT).body(msg).post("/paho/publish") //
+            .then().statusCode(200);
```
ppalaga: I do not find this test sufficient. I think there should be an assertion that the message really passed the queue. You could e.g. add another REST endpoint that would forward the messages from the queue using a ConsumerTemplate, something like we have here: https://github.com/apache/camel-quarkus/blob/master/integration-tests/twitter/src/main/java/org/apache/camel/quarkus/component/twitter/CamelResource.java#L58

zhfeng: OK, I get your idea and will improve the test. Thanks @ppalaga

PR #200, opened 2019-09-12
improve the paho integration test

zhfeng:


zhfeng:
@ppalaga I make some changes to improve the paho test to check the message after publishing to the queue. Can you take a look ?

ppalaga:
@zhfeng perfect, thanks!

PR #201, opened 2019-09-13
Fix #184 Leverage platform http service

### LINKED ISSUE #184: Leverage platform http service
As today to leverage the http service provided by quarkus we do rely on `servlet` but we can do much better and bypass the servlet engine.

To do so we need to create a sort of "platform http component" that leverage what the platform camel runs on has to offer i.e.

- quarkus --> vertx-web
- spring --> ???
- widfly --> undertow/servlet
 
I would start by creating such component and quarkus binding here and once we have a stable enough API we can move the common part to main camel repository.

ppalaga:


extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/spi/PlatformHttpEngine.java
```diff
@@ -0,0 +1,31 @@
...
+ * An abstraction of an HTTP Server engine on which HTTP endpoints can be deployed.
+ */
+public interface PlatformHttpEngine {
+
+    /**
+     * Deploys the given {@link HttpEndpointSpec} on this {@link PlatformHttpEngine}
+     *
+     * @param spec the {@link HttpEndpointSpec} to deploy
+     */
+    void deploy(HttpEndpointSpec spec);
```
lburgazzoli: This can be simplified to something like:

    Consumer createConsumer(PlatformHttpEndpoint, Processor)

Remember than other than deploying you need also to take care of the "undeploy" i.e. when the component is stopped/suspended. 

As side effect this would made `HttpEndpointSpec` not needed.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpEndpoint.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.camel.Consumer;
+import org.apache.camel.Processor;
+import org.apache.camel.Producer;
+import org.apache.camel.component.http.server.spi.Method;
+import org.apache.camel.spi.HeaderFilterStrategy;
+import org.apache.camel.spi.HeaderFilterStrategyAware;
+import org.apache.camel.spi.UriParam;
+import org.apache.camel.support.DefaultEndpoint;
+
+public class PlatformHttpEndpoint extends DefaultEndpoint implements AsyncEndpoint, HeaderFilterStrategyAware {
```
lburgazzoli: It lacks all the annotations camel requires for an endpoint

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpComponent.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.camel.CamelContext;
+import org.apache.camel.Endpoint;
+import org.apache.camel.component.http.server.spi.Method;
+import org.apache.camel.support.DefaultComponent;
+
+public class PlatformHttpComponent extends DefaultComponent {
```
lburgazzoli: The component needs to be discoverable through standard camel discovery so you need to add the `@Component` annotation.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpComponent.java
```diff
@@ -0,0 +1,48 @@
...
+    protected Endpoint createEndpoint(String uri, String remaining, Map<String, Object> parameters) throws Exception {
+        final String[] remainingParts = remaining.split(":");
+        if (remainingParts.length != 2) {
+            throw new IllegalArgumentException("Expected two segments delimited by ':'; found " + remaining);
+        }
+        final Set<Method> methods = Method.parseList(remainingParts[0]);
+        final String path = remainingParts[1];
+        return new PlatformHttpEndpoint(uri, this, methods, path);
+    }
+
```
lburgazzoli: There should be a getter/setter to set the engine. Discovering it through the registry is one of the options but usually camel component have much more flexibility.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpConsumer.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.camel.Exchange;
+import org.apache.camel.ExchangePattern;
+import org.apache.camel.Processor;
+import org.apache.camel.component.http.server.spi.HttpEndpointSpec;
+import org.apache.camel.component.http.server.spi.Method;
+import org.apache.camel.component.http.server.spi.PlatformHttpEngine;
+import org.apache.camel.spi.HeaderFilterStrategy;
+import org.apache.camel.support.DefaultConsumer;
+
+public class PlatformHttpConsumer extends DefaultConsumer implements HttpEndpointSpec {
```
lburgazzoli: This class may not bee needed if the creation of the consumer is delegated to the engine bits.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpConsumer.java
```diff
@@ -0,0 +1,81 @@
...
+    public PlatformHttpConsumer(Endpoint endpoint, Processor processor) {
+        super(endpoint, processor);
+    }
+
+    @Override
+    protected void doStart() throws Exception {
+        super.doStart();
+        final PlatformHttpEngine platformHttpEngine = getEndpoint().getCamelContext().getRegistry()
+                .lookupByNameAndType(PlatformHttpConstants.PLATFORM_HTTP_ENGINE_NAME, PlatformHttpEngine.class);
+        platformHttpEngine.deploy(this);
```
lburgazzoli: We should "undeploy" when stopping/halting the component

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/spi/Method.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Locale;
+import java.util.Set;
+import java.util.TreeSet;
+
+/**
+ * An HTTP method.
+ */
+public enum Method {
```
lburgazzoli: camel has a `camel-http-common` artefact, we could think about improving/reusing  what we have there.

extensions/platform-http/component/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-support</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>spi-annotations</artifactId>
+        </dependency>
+    </dependencies>
```
lburgazzoli: looks like it lacks camel plugins to generate components metadata such has catalog entry and other bits

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/http/server/QuarkusPlatformHttpEngine.java
```diff
@@ -0,0 +1,317 @@
...
+import org.apache.camel.spi.HeaderFilterStrategy;
+import org.apache.camel.support.DefaultMessage;
+import org.apache.camel.support.ExchangeHelper;
+import org.apache.camel.support.MessageHelper;
+import org.apache.camel.support.ObjectHelper;
+import org.jboss.logging.Logger;
+
+@ApplicationScoped
+@Named(PlatformHttpConstants.PLATFORM_HTTP_ENGINE_NAME)
+public class QuarkusPlatformHttpEngine implements PlatformHttpEngine {
```
lburgazzoli: If possible I would like to do wiring with build steps

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpEndpoint.java
```diff
@@ -0,0 +1,74 @@
...
+        this.headerFilterStrategy = headerFilterStrategy;
+    }
+
+    public String getPath() {
+        return path;
+    }
+
+    public Set<Method> getMethods() {
+        return methods;
+    }
```
lburgazzoli: There should be a getter/setter to set the engine. Discovering it through the registry is one of the options but usually camel component have much more flexibility.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpComponent.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    protected Endpoint createEndpoint(String uri, String remaining, Map<String, Object> parameters) throws Exception {
+        final String[] remainingParts = remaining.split(":");
+        if (remainingParts.length != 2) {
+            throw new IllegalArgumentException("Expected two segments delimited by ':'; found " + remaining);
+        }
+        final Set<Method> methods = Method.parseList(remainingParts[0]);
+        final String path = remainingParts[1];
+        return new PlatformHttpEndpoint(uri, this, methods, path);
```
lburgazzoli: I would prefer the URI syntax to be similar to other components, like 

    platform-http://relative_path[?options]

Also follow the conventions other http components have to filter methods (see `httpMethodRestrict`)

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/spi/Method.java
ppalaga: Yeah, `camel-http-common` currently depends on servlet, which we definitely do not want here.   Splitting it and depending on the servlet-less part might be an option.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/http/server/QuarkusPlatformHttpEngine.java
ppalaga: Why is wiring with build steps better?

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpComponent.java
ppalaga: You propose more typing but less learning for those who know the older Camel components.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/spi/Method.java
ppalaga: Actually, depending on `camel-http-common` and excluding the servlet could be an option.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/http/server/QuarkusPlatformHttpEngine.java
lburgazzoli: it's an optimization thing as you can record the wiring and at runtime you do not need to involve ArC.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpComponent.java
lburgazzoli: It is about consistency

ppalaga: Let's allow methods before the path and also as an uriParam to have both good ergonomy and compatibilty with the older components. So all of the following variants would yeld the same result: 

```
platform-http:/path?httpMethodRestrict=GET,POST
platform-http:GET:/path?httpMethodRestrict=POST
platform-http:GET,POST:/path
```

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/spi/PlatformHttpEngine.java
ppalaga: I tried to provide a simple and hard to break interface for the implementors because I tend to mistrust camel model as fragile due to its mutability and overused inheritance. You are right it can be done like you say and it might feel more natural to camelists so I followed your wish.

`QuarkusPlatformHttpConsumer` now also takes care for the stop, suspend and resume.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpEndpoint.java
ppalaga: Thanks, I added some, plz re-check.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpComponent.java
ppalaga: Done, thanks.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpComponent.java
ppalaga: Done, thanks.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpConsumer.java
ppalaga: `HttpEndpointSpec` is gone, but `QuarkusPlatformHttpConsumer` is still needed IMO.

extensions/platform-http/component/src/main/java/org/apache/camel/component/http/server/PlatformHttpConsumer.java
ppalaga: Done in QuarkusPlatformHttpConsumer

extensions/platform-http/component/pom.xml
ppalaga: I looked into Camel main repo and I cannot say I understand which mojos are necessary. I vote to follow up on this later.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/http/server/QuarkusPlatformHttpEngine.java
ppalaga: I did not think there is any difference, but after some testing, I now see there is one. Thanks for the hint!

ppalaga:
39fcf89715a847d28f4e53e6f0f20734608710f2 : Addressed most of the points by @lburgazzoli .

The native mode test is timing out saying `Unable to start native image in 60s` which I saw in the past and solved by adding a dependency on `quarkus-undertow`. Not sure what should be the solution here.

ppalaga:
bdaeb18 adds the mojos to the component module.

I still have to figure out why the native test is timing out.

ppalaga:
A simple `vertx-web` itest hangs also on plain Quarkus, but only in 0.22.0 branch. It works on the master. I filed https://github.com/quarkusio/quarkus/issues/4047

ppalaga:
f99d4f6 solves the native image test problems.

@lburgazzoli are there still some open questions from your PoV?

ppalaga:
c13d462540edded6dda55cfa745937f525bb01cc removes a couple of `System.out.println()` leftovers.

I actually wanted to check whether throwing a RuntimeException here is the best what can be done: https://github.com/apache/camel-quarkus/pull/201/files#diff-8012f112cafc5e8d5658a2bbf6812475R95

extensions/platform-http/component/src/main/java/org/apache/camel/component/platform/http/PlatformHttpEndpoint.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.camel.support.DefaultEndpoint;
+
+@UriEndpoint(/* firstVersion = "3.?.0", */ scheme = "platform-http", title = "Platform HTTP", syntax = "platform-http:[methods:]path", label = "http")
+public class PlatformHttpEndpoint extends DefaultEndpoint implements AsyncEndpoint, HeaderFilterStrategyAware {
+
+    private final String path;
+    private final Set<Method> methods;
+
+    @UriParam(label = "consumer", description = "A comma separated list of HTTP methods to serve. This list will be merged with the methods specified in the URI path. E.g. platform-http:GET,POST:/path?httpMethodRestrict=PUT,DELETE will effectivelly result in GET,POST,PUT,DELETE. If no methods are specified, all methods will be served.")
+    private String httpMethodRestrict;
```
lburgazzoli: why not use `httpMethodRestrict` only as the other `http` components in camel ?

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/PlatformHttpRecorder.java
```diff
@@ -0,0 +1,24 @@
...
+import io.quarkus.arc.runtime.BeanContainer;
+import io.quarkus.runtime.RuntimeValue;
+import io.quarkus.runtime.annotations.Recorder;
+
+@Recorder
+public class PlatformHttpRecorder {
+
+    public void registerBeans(RuntimeValue<CamelRuntime> runtime, BeanContainer beanContainer) {
+        final Registry registry = runtime.getValue().getRegistry();
+        final QuarkusPlatformHttpEngine engine = beanContainer.instance(QuarkusPlatformHttpEngine.class);
```
lburgazzoli: Can't the `QuarkusPlatformHttpEngine` be a parameter of the recorder so that the bean container is not involved ?

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -0,0 +1,359 @@
...
+    @Override
+    public PlatformHttpEndpoint getEndpoint() {
+        return (PlatformHttpEndpoint) super.getEndpoint();
+    }
+
+    @Override
+    protected void doStart() throws Exception {
+        super.doStart();
+
+        synchronized (routeLock) {
```
lburgazzoli: don't think the lock is needed as `doStart` is atomic

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -0,0 +1,359 @@
...
+                } catch (Exception e1) {
+                    throw new RuntimeException(e1);
+                }
+            });
+        }
+    }
+
+    @Override
+    protected void doStop() throws Exception {
+        synchronized (routeLock) {
```
lburgazzoli: don't think the lock is needed as `doStop` is atomic

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpEngine.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.camel.component.platform.http.PlatformHttpConstants;
+import org.apache.camel.component.platform.http.PlatformHttpEndpoint;
+import org.apache.camel.component.platform.http.spi.PlatformHttpEngine;
+import org.jboss.logging.Logger;
+
+import io.vertx.ext.web.Router;
+
+@ApplicationScoped
+@Named(PlatformHttpConstants.PLATFORM_HTTP_ENGINE_NAME)
+public class QuarkusPlatformHttpEngine implements PlatformHttpEngine {
```
lburgazzoli: can't the `QuarkusPlatformHttpEngine` be created and configured by a build step ?

extensions/platform-http/component/src/main/java/org/apache/camel/component/platform/http/PlatformHttpEndpoint.java
ppalaga: `platform-http:GET:/platform-http/hello` is shorter, faster to type and faster to read than `platform-http:/platform-http/hello?httpMethodRestrict=GET`

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/PlatformHttpRecorder.java
ppalaga: `QuarkusPlatformHttpEngine` needs to `@Inject` the `Router`. How could that work without the bean container?

extensions/platform-http/component/src/main/java/org/apache/camel/component/platform/http/PlatformHttpEndpoint.java
lburgazzoli: it is super confusing as we have two competing way to set the same thing.
We do already have lot of discrepancies in components, please try to stick with what we have and if you think there is something better, please address it camel itself.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/PlatformHttpRecorder.java
lburgazzoli: There a router build item

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: Thanks, now I see there is a similar lock in ServiceSupport.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: +1

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/PlatformHttpRecorder.java
ppalaga: Indeed, but only since Quarkus 0.23. I'd prefer improving this later when we are on 0.23

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpEngine.java
ppalaga: Same as above - this is possible since Quarkus 0.23. I vote for improving this later.

extensions/platform-http/component/src/main/java/org/apache/camel/component/platform/http/PlatformHttpEndpoint.java
ppalaga: The component is incubating here and the API/SPI will eventually move to Camel when it is mature enough. So IMO this is a way how to address it in Camel. 

I agree that consistency with the old stuff is nice when migrating and to keep the old knowledge usable but the developer experience is the king for green field projects where the old stuff does not matter that much.

It is amazing to see how excited people get when they see Panache. Applying your argument there, Panache should not be introduced because it is reaching the same goal as the well established Hibernate? I do not think that would be the right approach.

I see `platform-http` as the new generation of the HTTP components where we SHOULD be compatible with the old generation as far as possible but at the same time it should not block us from improving things. 

Having to type 18 chars less to define a route is IMO an improvement that justifies having two ways of configuring the same thing.

ppalaga:
7ccc04d removes the synchromization in the QuarkusPlatformHttpConsumer lifecycle methods.

extensions/platform-http/component/src/main/java/org/apache/camel/component/platform/http/PlatformHttpEndpoint.java
lburgazzoli: Sorry but there's no old vs new.

ppalaga: I am not sure I follow. Could you please explain it in more detail? We do change substantial parts of the URI (no protocol, no host, no port) so there apparently is something new and we are not going to be 100% compatible with the older generation of the HTTP components.

ppalaga:
3dcfa9a should fix the test failure.

extensions/platform-http/deployment/src/main/java/org/apache/camel/quarkus/component/platform/http/deployment/PlatformHttpProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    @Record(ExecutionTime.RUNTIME_INIT)
+    @BuildStep
+    void platformHttpComponent(PlatformHttpRecorder recorder, CamelRuntimeBuildItem runtime, VertxWebRouterBuildItem router) {
+        recorder.registerPlatformHttpComponent(runtime.getRuntime(), router.getRouter());
```
lburgazzoli: you need to be sure this binding happens before the routes are started

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/PlatformHttpRecorder.java
ppalaga: In bb474bb, the bean container is not involved. The `QuarkusPlatformHttpEngine` is created in the recorder though, which is in line with the requirement of Quarkus that the runtime objects should not be created inside BuildSteps.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpEngine.java
ppalaga: Improved in bb474bb as far as possible.

extensions/platform-http/deployment/src/main/java/org/apache/camel/quarkus/component/platform/http/deployment/PlatformHttpProcessor.java
ppalaga: Good point. Let me figure out how to ensure that.

ppalaga:
> BTW, this component should be usable from the rest dsl so it may need some further development

Yes, there are many things missing, incl. more test coverage (I have a long list of test cases to add). Plz. file an issue for this one or other missing things that come to your mind.

extensions/platform-http/component/src/main/java/org/apache/camel/component/platform/http/PlatformHttpHeaderFilterStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.platform.http;
+
+import org.apache.camel.support.DefaultHeaderFilterStrategy;
+
+public class PlatformHttpHeaderFilterStrategy extends DefaultHeaderFilterStrategy {
```
davsclaus: Is this similar to a http header filter strategy, if so then I suggest to add a comment about this so in the future we know and can make sure to align these classes.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+import org.apache.camel.support.ObjectHelper;
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.jboss.logging.Logger;
+
+/**
+ * A Quarkus specific {@link Consumer} for the {@link PlatformHttpComponent}.
+ */
+public class QuarkusPlatformHttpConsumer extends DefaultConsumer {
+    private static final Logger LOG = Logger.getLogger("org.apache.camel.quarkus.component.platform.http");
```
davsclaus: We use the logger name for the FQN classname in Apache Camel so users can find the class

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+        final String path = getEndpoint().getPath();
+        final Route r = router.route(path);
+        final Set<Method> methods = getEndpoint().getEffectiveMethods();
+        if (!methods.equals(Method.getAll())) {
+            methods.stream().forEach(m -> r.method(HttpMethod.valueOf(m.name())));
+        }
+
+        Config cfg = ConfigProvider.getConfig();
+        final BodyHandler bodyHandler = BodyHandler.create();
+        /* Keep in sync with how the BodyHandler is configured in io.quarkus.vertx.web.runtime.VertxWebRecorder
```
davsclaus: Good code comment

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+    public PlatformHttpEndpoint getEndpoint() {
+        return (PlatformHttpEndpoint) super.getEndpoint();
+    }
+
+    @Override
+    protected void doStart() throws Exception {
+        super.doStart();
+
+        final String path = getEndpoint().getPath();
+        final Route r = router.route(path);
```
davsclaus: Name this router instead of r

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+                if (values != null) {
+                    response.putHeader(key, values);
+                } else if (firstValue != null) {
+                    response.putHeader(key, firstValue);
+                }
+            }
+        }
+
+        Object body = message.getBody();
+        final Exception exception = exchange.getException();
```
davsclaus: I think in some of the other http components we have options to turn this on|off as we should not leak server side exceptions by default to clients

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    void writeResponse(RoutingContext ctx, Exchange camelExchange, HeaderFilterStrategy headerFilterStrategy) {
+        final Object body = toHttpResponse(ctx.response(), camelExchange.getMessage(), headerFilterStrategy);
+
+        final HttpServerResponse response = ctx.response();
+        if (body == null) {
+            QuarkusPlatformHttpEngine.LOG.tracef("No payload to send as reply for exchange: %s", camelExchange);
+            response.putHeader("Content-Type", "text/plain; charset=utf-8");
+            response.end("No response available");
```
davsclaus: Isnt there a HTTP STATUS 204 (or something like that for -- OK but no content)

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+            } catch (TypeConversionException | NoTypeConversionAvailableException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+    }
+
+    Exchange toExchange(RoutingContext ctx, Exchange exchange, HeaderFilterStrategy headerFilterStrategy) {
+        Message in = toCamelMessage(ctx, exchange, headerFilterStrategy);
+        final String charset = ctx.parsedHeaders().contentType().parameter("charset");
```
davsclaus: What if this is null, then its likely better to not set these headers

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -0,0 +1,376 @@
...
+    //        if (predicateContextParams != null) {
+    //            // Remove this as it's an unwanted artifact of our Undertow predicate chain
+    //            predicateContextParams.remove("remaining");
+    //
+    //            for (String paramName : predicateContextParams.keySet()) {
+    //                headersMap.put(paramName, predicateContextParams.get(paramName));
+    //            }
+    //        }
+
+            // NOTE: these headers is applied using the same logic as camel-http/camel-jetty to be consistent
```
davsclaus: Thanks this is good to know

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: Are you sure? `router` would clash with the `router` field.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: Would you plz. file a followup issue?

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
davsclaus: Okay cant you get my point about using a single letter is bad (only e or i is usually common)

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: I think I copied this from the Undertow component. Let me check what the other HTTP components do.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
davsclaus: Ah netty-http component dont have this feature

davsclaus: Ah it was the transfer exception option. I created a new ticket for not dumping stacktraces so end users can turn this off for production usage etc: https://issues.apache.org/jira/browse/CAMEL-14025

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: So yes, this behavior is from Undertow. `org.apache.camel.http.common.DefaultHttpBinding.doWriteResponse(Message, HttpServletResponse, Exchange)` does not write anything if the body is null. 

The status code at this location can be both success or failure, so it would not always be correct to change it to 204.

extensions/platform-http/component/src/main/java/org/apache/camel/component/platform/http/PlatformHttpEndpoint.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public Producer createProducer() throws Exception {
+        throw new UnsupportedOperationException("Producer is not supported");
+    }
+
+    @Override
+    public Consumer createConsumer(Processor processor) throws Exception {
+        if (platformHttpEngine == null) {
+            platformHttpEngine = getCamelContext().getRegistry()
```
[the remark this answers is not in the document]
ppalaga: Done in 7c292ebc4ee38f0bda8ed182c29d3cdc3957a0dd 

The bottom line is that the omnipresent mutable state makes it very easy do something wrong. I wonder whether there ever was some discussion to aim at making Camel more functional.

extensions/platform-http/component/src/main/java/org/apache/camel/component/platform/http/PlatformHttpEndpoint.java
```diff
@@ -0,0 +1,132 @@
...
+                    .lookupByNameAndType(PlatformHttpConstants.PLATFORM_HTTP_ENGINE_NAME, PlatformHttpEngine.class);
+        }
+        return platformHttpEngine.createConsumer(this, processor);
+    }
+
+    public Set<Method> getEffectiveMethods() {
```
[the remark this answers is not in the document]
ppalaga: Removed in 7c292ebc4ee38f0bda8ed182c29d3cdc3957a0dd

extensions/platform-http/component/pom.xml
ppalaga: Added apt in 7c292ebc4ee38f0bda8ed182c29d3cdc3957a0dd

extensions/platform-http/component/src/main/java/org/apache/camel/component/platform/http/PlatformHttpHeaderFilterStrategy.java
ppalaga: Done in 7c292ebc4ee38f0bda8ed182c29d3cdc3957a0dd

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: Frankly have not got that point, sorry :) And never saw that as an issue for a variable ranging over just a couple of lines. No problem with changing it though.

ppalaga: How many letters is enough, BTW? Going to use 3 and wonder whether 2 would pass your filter too?

ppalaga: `rte` would not look self explaining. Took `newRoute` FWIW.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: So if there anything to do here, please file a followup issue.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: I personally find the current code good enough, but adding a note to my TODO list that this case should be covered by a test. BTW, which of the existing HTTP components is the most up to date? - i.e. where to look for the desired behavior?

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: Solved in 7c292ebc4ee38f0bda8ed182c29d3cdc3957a0dd

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: Solved in e66f2326c7bd4662dff548e7ba6272e2f1cd3668

ppalaga:
I removed the methods from the URI body after @lburgazzoli mentioned that we could eventually want to have this as a producer, where we indeed need the place for protocol, host, port, etc.

extensions/platform-http/component/src/main/java/org/apache/camel/component/platform/http/PlatformHttpEndpoint.java
ppalaga: Resolved in e66f2326c7bd4662dff548e7ba6272e2f1cd3668

extensions/platform-http/deployment/src/main/java/org/apache/camel/quarkus/component/platform/http/deployment/PlatformHttpProcessor.java
ppalaga: Will do this in a followup https://github.com/apache/camel-quarkus/issues/218

ppalaga:
In e66f2326c7bd4662dff548e7ba6272e2f1cd3668 all requests should be either resolved or should have followup issue.

ppalaga:
Is this good to merge now?

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -0,0 +1,380 @@
...
+        }
+        return body;
+    }
+
+    void writeResponse(RoutingContext ctx, Exchange camelExchange, HeaderFilterStrategy headerFilterStrategy) {
+        final Object body = toHttpResponse(ctx.response(), camelExchange.getMessage(), headerFilterStrategy);
+
+        final HttpServerResponse response = ctx.response();
+        if (body == null) {
+            QuarkusPlatformHttpEngine.LOG.tracef("No payload to send as reply for exchange: %s", camelExchange);
```
davsclaus: Use LOG from this class, and not some other class

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
davsclaus: Yeah ideally we should get the http components aligned. They should IMHO not return a hardcoded text, eg this is in english, then chinese users get english text etc. But lets log a ticket upstream at Apache Camel

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: +1. Must be a result of copying the code from somewhere else.

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
ppalaga: Filed https://github.com/apache/camel-quarkus/issues/220

ppalaga:
4495866 does just `s/QuarkusPlatformHttpEngine.LOG/LOG/`

PR #206, opened 2019-09-20
Improve the maven deployment of the synthetic test jars

ppalaga:
@oscerd yet another attempt to fix the release 0.2.0


oscerd:
I release on monday

PR #212, opened 2019-09-30
cleanup aws and paho extensions

lburgazzoli:


extensions/aws-eks/deployment/src/main/java/org/apache/camel/quarkus/component/aws/eks/deployment/AwsEKSProcessor.java
```diff
@@ -87,8 +77,8 @@ void process(CombinedIndexBuildItem combinedIndexBuildItem,
                 Region.class.getCanonicalName(),
                 Service.class.getCanonicalName(),
                 CredentialScope.class.getCanonicalName(),
-                LogFactoryImpl.class.getCanonicalName(),
-                Jdk14Logger.class.getCanonicalName(),
+                "org.apache.commons.logging.impl.LogFactoryImpl",
+                "org.apache.commons.logging.impl.Jdk14Logger",
```
ppalaga: Why is a string literal better in these two cases?

lburgazzoli: It should not make much difference in this case but having them expressed as literal string does not require to have such classes in the class-path so there is a little bit more freedom about how commons-logging deps are added to the mix.

If you do not like it, I can use classes

ppalaga: Thanks, the justification sounds good.

lburgazzoli:
> The changes look good and it would be nice if somebody could run the itests against AWS using some real creds before merging.

usually the one that submits the PR is likely also testing it :)

PR #222, opened 2019-10-03
chore: rename CamelRegistryBuildItem to a more meaningful CamelBeanBuildItem

lburgazzoli:


lburgazzoli:
@ppalaga renamed as I need it too for #9

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelBeanBuildItem.java
```diff
@@ -20,18 +20,18 @@
 
 import io.quarkus.builder.item.MultiBuildItem;
 
-public final class CamelRegistryBuildItem extends MultiBuildItem {
+public final class CamelBeanBuildItem extends MultiBuildItem {
```
ppalaga: A short class level javadoc stating the purpose and limitations would be nice. Something like

> A MultiBuildItem holding beans to add to CamelRegistry immediately after the registry is created. Can be produced only by methods that do not depend on CamelRuntime (because otherwise there is a circular dependency).

lburgazzoli:
> `RuntimeCamelBeanBuildItem`

I only need the build time one for the moment so I didn't added trhe runtime one because I do not yet know how it should look like.

About the naing I didn't add a prefix as I think adding it to the runtime one is enough and BuildTimeCamelBeanBuildItem looks  a little bit bloated (IMHO) but I can add it if you think it is better.

ppalaga:
> About the naing I didn't add a prefix as I think adding it to the runtime one is enough and BuildTimeCamelBeanBuildItem looks a little bit bloated (IMHO) but I can add it if you think it is better.

No problem with sticking to this convention, but plz add some javadoc to the build item so that the users (incl. future me) do not need to dive deep into the source code to figure these things out.

lburgazzoli:
> > About the naing I didn't add a prefix as I think adding it to the runtime one is enough and BuildTimeCamelBeanBuildItem looks a little bit bloated (IMHO) but I can add it if you think it is better.
> 
> No problem with sticking to this convention, but plz add some javadoc to the build item so that the users (incl. future me) do not need to dive deep into the source code to figure these things out.

done

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelBeanBuildItem.java
```diff
@@ -20,18 +20,22 @@
 
 import io.quarkus.builder.item.MultiBuildItem;
 
-public final class CamelRegistryBuildItem extends MultiBuildItem {
+/**
+ * A MultiBuildItem holding beans to add to CamelRegistry during static initialization phase.
+ * Can be produced only by methods that do not depend on CamelRuntime (because otherwise there is a circular dependency).
+ */
```
ppalaga: The text is perfect, but having MultiBuildItem CamelRegistry and CamelRuntime as links (or at least FQ names) would be nice.

lburgazzoli: done

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelBeanBuildItem.java
```diff
@@ -20,18 +20,24 @@
 
 import io.quarkus.builder.item.MultiBuildItem;
 
-public final class CamelRegistryBuildItem extends MultiBuildItem {
+/**
+ * A {@link MultiBuildItem} holding beans to add to {@link org.apache.camel.spi.Registry} during
+ * static initialization phase.
+ * Can be produced only by methods that do not depend on {@linke CamelRuntime} because otherwise
```
aldettinger: Small typo there if you have time.

PR #225, opened 2019-10-03
Rename substitutions class names

lburgazzoli:


extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/graal/SubstituteIntrospectionSupport.java
```diff
@@ -21,14 +21,17 @@
...
 @TargetClass(className = "org.apache.camel.support.IntrospectionSupport")
-final class Target_org_apache_camel_support_IntrospectionSupport {
-
+final class SubstituteIntrospectionSupport {
     @Alias
+    @TargetElement(name = "CACHE")
     @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.FromAlias)
-    private static Map<Class<?>, BeanIntrospection.ClassInfo> CACHE = LRUCacheFactory.newLRUWeakCache(256);
-
+    private static Map<Class<?>, BeanIntrospection.ClassInfo> cache = LRUCacheFactory.newLRUWeakCache(256);
```
ppalaga: Why `@TargetElement(name = "CACHE")` and `cache` field? Is our checkstyle protesting against an uppercase field name given it is not final?

lburgazzoli: Yes, one of the reason for the renaming was to be able to run checkstyle against substitutions but then `final` does not make much sense for substrate and uppercase makes checkstyle unhappy so this is a compromise.

As alternative I can explicit ignore that class.

oscerd:
LGTM, let me merge.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/graal/SubstituteIntrospectionSupport.java
ppalaga: Thanks for the explanation. BTW, there is also `// CHECKSTYLE:OFF` and `// CHECKSTYLE:ON` that can be used to skip specific lines of code.

lburgazzoli: Yes, everything adds a little bit of noise, if you feel the check-style comment is better, feel free to change it

ppalaga: No, your solution is one line less :-)

Issue #198, opened 2019-09-11
Remove custom registry implementation


lburgazzoli:
It does not seem we do store instances wrapped by a `RuntimeValue` in the registry any more so it may be possible to thing to simplify the registry and leverage the "bean repository" concept we have in Camel, something like:

    repository = new ArCBeanRepository();
    registry = new DefaultRegistry(repository);

This would reduce the maintenance costs as i.e. https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-13948 would not need to be implemented in bot Camel and Camel Quarkus.

//cc @gnodet

lburgazzoli:
Fixed by https://github.com/apache/camel-quarkus/pull/226

Issue #9, opened 2019-06-27
quarkus-camel - Use MainSupport as base for running Camel


davsclaus:
We should look at using MainSupport (parts of Camel Main) some more for bootstrapping camel-quarkus.

Then its similar across standalone Camel, Camel K, and Camel Quarkus. And also Camel Main has parts that are similar to Camel Spring Boot.

ppalaga:
This should be easier to do now that we are on Camel 3.0.0.M4. Any hints or examples @davsclaus how to do that?

lburgazzoli:
As today `core` is responsible to act like a `main` but for better flexibility (i.e. for other to build on top of `core`) I think we need to split responsibilities like:

1. `core` to set-up low level things such as:
    - camel context
    - registry
    - integration with quarkus event loop #131
2. `main` to set up application related things, like:
    - initialize, start/stop the camel context created by `core`
    - discover routes
    - discover and register life-cycle interceptors (to allow i.e. camel-k to hook into the static init build processors)
    - integration with quarkus health checks #141 
    - integration with quarkus metrics #128
    - fire CDI events

/cc @gnodet @davsclaus

lburgazzoli:
Fixed by https://github.com/apache/camel-quarkus/pull/226

PR #226, opened 2019-10-03
Use MainSupport as base for running Camel

lburgazzoli:
This is an initial draft implementation of #9 and #198 and it does some of the task listed [here](https://github.com/apache/camel-quarkus/issues/9#issuecomment-522317439).

Relevant changes to camel-quarkus-core:
- is now responsible to create the registry
- is now responsible to create the camel context but not to start it
- it does not discover routes any more
- it does not fire events any more

It is based on **Camel 3.0.0-SNAPSHOT**


lburgazzoli:
Relates to https://issues.apache.org/jira/browse/CAMEL-14031

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSupport.java
```diff
@@ -56,4 +69,67 @@ public static boolean isPublic(ClassInfo ci) {
             .flatMap(CamelSupport::safeWalk)
             .filter(Files::isRegularFile);
     }
+
+    public static Stream<String> getRouteBuilderClasses(IndexView view) {
+        Set<ClassInfo> allKnownImplementors = new HashSet<>();
```
davsclaus: I wonder if we may want to use a LinkedHashSet to have ordering of discovered classes

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSupport.java
```diff
@@ -56,4 +69,67 @@ public static boolean isPublic(ClassInfo ci) {
...
+            .map(ClassInfo::toString);
+    }
+
+    public static Stream<ServiceInfo> services(ApplicationArchivesBuildItem applicationArchivesBuildItem) {
+        return CamelSupport.resources(applicationArchivesBuildItem, CamelSupport.CAMEL_SERVICE_BASE_PATH)
+            .map(CamelSupport::services)
+            .flatMap(Collection::stream);
+    }
+
+    public static List<ServiceInfo> services(Path p) {
```
davsclaus: Should this be public method

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/FastCamelContext.java
```diff
@@ -365,11 +388,12 @@ public AsyncProcessor createMulticast(Collection<Processor> processors, Executor
     @SuppressWarnings("unchecked")
     protected <T> T resolve(Class<T> clazz, String type, String name, CamelContext context) {
```
davsclaus: What is this method used for? Can we add some javadoc, and the code that is disabled is it to be removed or not? Maybe add a TODO if this code needs revisit

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSupport.java
lburgazzoli: The problem is that the order also depends on how Jandex indexes entries so I'm not sure if it matter, maybe better is to have the RouteBuilder extends the Ordered interface and order them when routes are processed ?

davsclaus: Ah yeah this allows end users to control this if they have use-cases where this may matter. I dont recall if we in eg spring-boot discover of routes have any ordering too. But the Ordered is a good idea, we use it elsewhere in Camel too.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/FastCamelContext.java
lburgazzoli: it was originally used to auto configure component, languages and data formats but it is not more needed now as MainSupport does it, let me revisit it

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSupport.java
lburgazzoli: As far as I remember you can order things in spring-boot but we can assume that a routes that have an explicit order should take the precedence over any other ordering mechanics.

Needs to be tested but we can also document that the preferred way of ordering routes is through the Camel's Ordered interface.

lburgazzoli: https://issues.apache.org/jira/browse/CAMEL-14031

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSupport.java
lburgazzoli: done

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/FastCamelContext.java
lburgazzoli: done

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelProducers.java
```diff
@@ -16,51 +16,41 @@
...
 import javax.enterprise.inject.Produces;
+import javax.inject.Singleton;
 
 import org.apache.camel.CamelContext;
 import org.apache.camel.ConsumerTemplate;
 import org.apache.camel.ProducerTemplate;
 import org.apache.camel.spi.Registry;
 
-@ApplicationScoped
+@Singleton
```
ppalaga:
```suggestion
/**
 * Producers of beans that are injectable via CDI.
 */
@Singleton
```

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelRecorder.java
```diff
@@ -16,88 +16,98 @@
...
+    public RuntimeValue<Registry> createRegistry() {
+        return new RuntimeValue<>(new RuntimeRegistry());
     }
 
-    public void start(
-            ShutdownContext shutdown,
-            RuntimeValue<CamelRuntime> runtime,
-            CamelConfig.Runtime runtimeConfig) throws Exception {
+    @SuppressWarnings("unchecked")
+    public RuntimeValue<CamelContext> createContext(RuntimeValue<Registry> registry, BeanContainer beanContainer, CamelConfig.BuildTime buildTimeConfig) {
```
ppalaga: A JavaDoc stating whether this is a invoked from a build time or a runtime BuildStep would be nice.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelRecorder.java
```diff
@@ -16,88 +16,98 @@
...
-            runtime.getValue().getRegistry().bind(name, type, type.newInstance());
+            runtime.getValue().bind(name, type, type.newInstance());
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
     }
 
-    public BeanContainerListener initRuntimeInjection(RuntimeValue<CamelRuntime> runtime) {
-        return container -> container.instance(CamelProducers.class).setCamelRuntime(runtime.getValue());
+    public Supplier<CamelConfig.BuildTime> buildTimeSupplier(CamelConfig.BuildTime config) {
```
ppalaga: I vote for renaming to `buildTimeConfigSupplier`

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelRecorder.java
```diff
@@ -16,88 +16,98 @@
...
+        return new Supplier<CamelConfig.BuildTime>() {
+            @Override
+            public CamelConfig.BuildTime get() {
+                return config;
+            }
+        };
+    }
+
+    public Supplier<CamelConfig.Runtime> runtimeSupplier(CamelConfig.Runtime config) {
```
ppalaga: I vote for renaming to `runtimeConfigSupplier`

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/BeanManagerHelper.java
```diff
@@ -27,6 +27,7 @@
 import javax.enterprise.inject.spi.BeanManager;
 
 import io.quarkus.arc.Arc;
+import io.quarkus.arc.ArcContainer;
 
 @Vetoed
```
ppalaga:
```suggestion
/**
 * Helper methods to map {@link Registry} calls to {@link Arc}.
 */
@Vetoed
```

extensions/main/deployment/src/main/java/org/apache/camel/quarkus/main/deployment/BuildProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep
+    CamelMainBuildItem create(
+            CombinedIndexBuildItem combinedIndexBuildItem,
+            CamelMainRecorder recorder,
+            CamelContextBuildItem context,
+            List<CamelMainListenerBuildItem> listeners,
+            // TODO: keep is as it adds a dependency on Arc
```
ppalaga: It would be nice to document why is the dependency on Arc important, e.g. to be able to decide in the future that the param is not needed anymore.

extensions/main/deployment/src/main/java/org/apache/camel/quarkus/main/deployment/SubstrateProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+package org.apache.camel.quarkus.main.deployment;
+
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
+import io.quarkus.deployment.builditem.substrate.ReflectiveClassBuildItem;
+import org.apache.camel.quarkus.core.deployment.CamelSupport;
+import org.jboss.jandex.IndexView;
+
+public class SubstrateProcessor {
```
ppalaga: Why do we need two Processors? It would be nice to figure out from JavaDoc which tasks belong where.

extensions/main/deployment/src/main/java/org/apache/camel/quarkus/main/deployment/support/CamelMainBuildItem.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.main.deployment.support;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+import io.quarkus.runtime.RuntimeValue;
+import org.apache.camel.quarkus.main.CamelMain;
+
+public final class CamelMainBuildItem extends SimpleBuildItem {
```
ppalaga:
```suggestion
/**
 * Holds the {@link CamelMain} {@link RuntimeValue}.
 */
public final class CamelMainBuildItem extends SimpleBuildItem {
```

ppalaga: Minor: Do these two BuildItem classes deserve a separate package (with a meaningless `support` last segment name?)

extensions/main/runtime/src/main/java/org/apache/camel/quarkus/main/CamelMainEventDispatcher.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.main;
+
+import io.quarkus.arc.Arc;
+import org.apache.camel.CamelContext;
+import org.apache.camel.main.MainSupport;
+
+public class CamelMainEventDispatcher implements org.apache.camel.main.MainListener {
```
ppalaga: Please a single sentence summary JavaDoc what is this class doing.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelProducers.java
```diff
@@ -16,51 +16,41 @@
...
 public class CamelProducers {
+    private CamelContext context;
```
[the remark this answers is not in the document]
lburgazzoli: this is set by a recorder, don't know if make any difference

extensions/main/deployment/src/main/java/org/apache/camel/quarkus/main/deployment/SubstrateProcessor.java
lburgazzoli: what do you mean by two processors ?

extensions/main/deployment/src/main/java/org/apache/camel/quarkus/main/deployment/support/CamelMainBuildItem.java
lburgazzoli: they don't need to be separate, probably a leftover of something I was working on

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelRecorder.java
lburgazzoli: I don't think we need to explicit the phase here as for the recorder POV, it does not matter.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelRecorder.java
lburgazzoli: not used any more, will remove it

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelRecorder.java
lburgazzoli: not used any more, will remove it

lburgazzoli:
@ppalaga fixed your findings, at least, I think :)

integration-tests/pom.xml
```diff
@@ -80,7 +80,8 @@
         <!-- build those first -->
         <module>core</module>
         <module>core-impl</module>
-        <module>core-cdi</module>
+        <!--<module>core-cdi</module>-->
```
ppalaga: This commented line can perhaps be removed?

extensions/main/deployment/src/main/java/org/apache/camel/quarkus/main/deployment/SubstrateProcessor.java
ppalaga: There is `org.apache.camel.quarkus.main.deployment.BuildProcessor` and `org.apache.camel.quarkus.main.deployment.SubstrateProcessor`. Both contain build steps and it is not clear to me what would be the criteria to decide for the current and future build steps to decide into which one they belong?

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelRecorder.java
ppalaga: Sorry I meant static init vs. runtime init. It matters for the kinds of resources available inside the method and it gives a hint about ordering. Of course, the info can be looked up in the IDE, but when in a comment, it helps to faster understand the context.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/support/BeanManagerHelper.java
```diff
@@ -27,25 +27,35 @@
 import javax.enterprise.inject.spi.BeanManager;
 
 import io.quarkus.arc.Arc;
+import io.quarkus.arc.ArcContainer;
 
+/**
+ * Helper methods retrieve beans from the {@link Arc} container.
```
ppalaga: ... methods _to_ retrieve ... maybe?

extensions/main/deployment/src/main/java/org/apache/camel/quarkus/main/deployment/SubstrateProcessor.java
lburgazzoli: `SubstrateProcessor` is meant to contains build steps related to substrate like reflection, substrate properties and so on. The distinction is juts logical so one know where to look depending of the context.

extensions/servlet/deployment/src/test/java/org/apache/camel/quarkus/component/servlet/test/CustomDefaultServletClassTest.java
```diff
@@ -30,25 +37,45 @@
...
 
     @Test
-    public void customDefaultServletClass() {
-        RestAssured.when().get("/custom").then()
-            .body(IsEqual.equalTo("GET: /custom"))
-            .and().header("x-servlet-class-name", CustomServlet.class.getName());
+    public void customDefaultServletClass() throws Exception {
+        DefaultCamelContext context = new DefaultCamelContext();
+        context.addRoutes(new Routes());
+        context.start();
```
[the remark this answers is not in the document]
lburgazzoli: camel-core does not loads routes anymore

integration-tests/aws/pom.xml
```diff
@@ -30,6 +30,10 @@
     <description>The camel integration tests</description>
 
     <dependencies>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-main</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
lburgazzoli: camel-quarkus-core is supposed to be used by people who want to create their own "main" whereas camel-quarkus-main does provide a ready to use "main" so no, component extensions should not depends on main otherwise it won't be possible for people to create their own main

lburgazzoli:
@ppalaga findings fixed (I hope)

extensions/main/deployment/src/main/java/org/apache/camel/quarkus/main/deployment/SubstrateProcessor.java
ppalaga: Thanks, makes sense. How about adding a class level JavaDoc something like `Produces BuildItems from the io.quarkus.deployment.builditem.substrate package` ?

integration-tests/aws/pom.xml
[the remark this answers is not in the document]
ppalaga: What are the potential use cases for creating one's own "mains"?

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelRecorder.java
lburgazzoli: I can add them but it does not give much info and guarantees as in fact the behaviour depends to the target phase of the build steps that invokes the recorder.

integration-tests/aws/pom.xml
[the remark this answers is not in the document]
lburgazzoli: You want to have a better control about the context initialization, set-up and life cycle as example.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/runtime/CamelRecorder.java
ppalaga: True, the set of all current callers is the ultimate source of the truth for each recorder method. But isn't this particular method based in the assumption that it will be called exactly once in a very specific phase? I guess these assumptions are supposed to stay rather stable over time and any future change will require a thorough review. Therefore documenting the assumptions might be a good idea. Maybe I am just projecting something into the code that really is not there? I do not want to nitpick, I want to understand myself and eventually make the understanding easier to the people who will read the code after us.

integration-tests/aws/pom.xml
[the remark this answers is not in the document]
ppalaga: Do you need this in Camel K? Do you think 80+% of Camel Quarkus users will need this?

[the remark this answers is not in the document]
ppalaga: I am trying to understand the significance of this. In case this is meant to be a speciality for <20% of users, then I think we should figure out how not confuse the rest 80% of the users.

lburgazzoli:
Move camel-main bits to camel-core, user can opt-out - thus have the full resposability of setting up and starting camel contest - by setting `quarkus.camel.disable-main = false`

lburgazzoli:
@ppalaga @davsclaus do you mind a doing a final review ?

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.core;
+
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.camel.spi.BeanRepository;
+
+public final class RuntimeBeanRepository implements BeanRepository {
```
ppalaga: RuntimeBeanRepository being the only caller of BeanManagerHelper, is there a reason to keep them separate? I mean the invocations of BeanManagerHelper could be inlined here.

integration-tests/bean/pom.xml
```diff
@@ -30,6 +30,10 @@
     <description>Integration tests for Camel Bean extension</description>
 
     <dependencies>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-core</artifactId>
```
ppalaga: Core is redundant here, isn't it? And similarly in other itests.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
```diff
@@ -55,6 +47,21 @@
...
         public boolean disableXml;
+
+        /**
+         * Disable camel-main.
+         * When main is disabled, routes won't be automatically be loaded and
+         * started and the entire lifecycle of the Camel Context is under user
+         * control.
+         */
+        @ConfigItem(defaultValue = "false")
+        public boolean disableMain;
```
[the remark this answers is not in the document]
lburgazzoli: It uses negation to be consistent with other properties

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/Flags.java
```diff
@@ -38,4 +38,10 @@ public boolean getAsBoolean() {
         }
     }
 
+    public static final class MainDisabled implements BooleanSupplier {
```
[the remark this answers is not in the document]
lburgazzoli: Same as above

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
lburgazzoli: I've inlined it but can we please be pragmatic an try to solve one problem at time ? 

BeanManagerHelper was there long before this PR and was used in very same way so to me it is unrelated to this PR.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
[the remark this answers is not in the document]
ppalaga: I'd personaly vote for discontinuing the bad practice now and eventually rename the other props later.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
ppalaga: Yes, sorry, of course, that would be a valid reason to reject my proposal.

PR #228, opened 2019-10-05
Cleanup

lburgazzoli:


extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -58,29 +60,25 @@ void beans(BuildProducer<AdditionalBeanBuildItem> beanProducer) {
...
-                si -> {
-                    LOGGER.debug("Binding camel service {} with type {}", si.name, si.type);
+            CamelSupport.services(applicationArchives).forEach(si -> {
+                    LOGGER.debug("Binding bean with name: {}, type {}", si.name, si.type);
 
                     recorder.bind(
                         registry,
                         si.name,
-                        si.type
+                        recorderContext.classProxy(si.type)
```
ppalaga: As far as I could understand, `RecorderContext.classProxy(String)` postpones the class loading to runtime. Do some things break for us without using the classProxy? In other words, I wonder why is this better than passing the class name and do Class.forName() in the recorder that we had before? 

Maybe we need this because it postpones not only the class loading but also the class initialization?

PR #231, opened 2019-10-06
Add support for quarkus provided event loop

### LINKED ISSUE #131: Add support for quarkus provided event loop
We should back camel-core with event loop from quarkus, as example:

    https://github.com/quarkusio/quarkus/blob/master/extensions/netty/deployment/src/main/java/io/quarkus/netty/deployment/NettyProcessor.java#L57-L72

Or vert.x

lburgazzoli:
Fixes #131

### DIFF
--- a/extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
+++ b/extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
@@ -66,7 +66,22 @@ class BuildProcessor {
 
             RuntimeValue<Registry> registry = recorder.createRegistry();
 
-            CamelSupport.services(applicationArchives).forEach(si -> {
+            CamelSupport.services(applicationArchives)
+                .filter(si -> {
+                    //
+                    // by default all the service found in META-INF/service/org/apache/camel are
+                    // bound to the registry but some of the services are then replaced or set
+                    // to the camel context directly by extension so it does not make sense to
+                    // instantiate them in this phase.
+                    //
+                    boolean blacklisted = si.path.endsWith("reactive-executor") || si.path.endsWith("platform-http");
+                    if (blacklisted) {
+                        LOGGER.debug("Ignore service: {}", si);
+                    }
+
+                    return !blacklisted;
+                })
+                .forEach(si -> {
                     LOGGER.debug("Binding bean with name: {}, type {}", si.name, si.type);
 
                     recorder.bind(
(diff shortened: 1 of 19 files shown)


extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeRegistry.java
```diff
@@ -30,4 +34,26 @@ public Object unwrap(Object value) {
             ? ((RuntimeValue)value).getValue()
             : value;
     }
+
+    // TODO: fix upstream
```
davsclaus: Whats the issue upstream?

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -66,7 +66,26 @@ CamelRegistryBuildItem registry(
...
+            CamelSupport.services(applicationArchives)
+                .filter(si -> {
+                    //
+                    // we need to filter out some services for being registered as some
+                    // extension configure them programmatically and if not properly
+                    // configured, they may cause troubles.
+                    //
+                    // As example, a reactive-executor can be provided by adding a dep on
+                    // camel-reactive-executor-vertx but if that service get detected and
+                    // initialized, it will create an instance of Vert.x during static init
```
davsclaus: Its created in doStart lifecycle phase. So I wonder if quarkus-camel is a bit "wrong" if it starts Camel services during build phase? We should only use doInit phase IMHO for that.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeRegistry.java
lburgazzoli: Sorry was late and forgot to open an issue. The problem is that the default implementations does not merge the results from the repositories with the fallback one so if you bind something to the registry and one of the repositories return a non null set, then you' won't get the local instance but an empty answer.

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
lburgazzoli: Yeah, I need to digg into the camel-reactive-executor-vertx a little bit more as the camel context is not supposed to be started at build time but only at runtime init by camel main.

lburgazzoli: ok, found out that I was wrong so I probably had some bad code during the development that leaded to the reactive executor being started in the wrong phase.

I've updated the comment.

davsclaus:
Okay fixing the registry upstream
https://issues.apache.org/jira/browse/CAMEL-14040

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSupport.java
```diff
@@ -117,18 +117,21 @@ public static boolean isPublic(ClassInfo ci) {
      * services from resources belonging to META-INF/services/org/apache/camel.
      */
     public static class ServiceInfo {
+        public final Path path;
```
ppalaga: A piece of JavaDoc stating what Path is this would be nice. Something like `The path from which this {@link ServiceInfo} was read. E.g {@code META-INF/services/org/apache/camel/my-service}`

extensions/reactive-executor/deployment/src/main/java/org/apache/camel/quarkus/reactive/executor/deployment/BuildProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.annotations.ExecutionTime;
+import io.quarkus.deployment.annotations.Record;
+import io.quarkus.vertx.deployment.VertxBuildItem;
+import org.apache.camel.quarkus.core.Flags;
+import org.apache.camel.quarkus.core.deployment.CamelReactiveExecutorBuildItem;
+import org.apache.camel.quarkus.reactive.executor.ReactiveExecutorRecorder;
+
+public class BuildProcessor {
```
ppalaga: I must say I find using the same names like `BuildProcessor` in different extensions rather impractical. Having two `BuildProcessor` s open in an IDE makes it hard to figure out which editor tab belongs to which class.

Issue #234, opened 2019-10-07
ActiveMQTestResource should not eat the exception


WillemJiang:
We'd better to log the exceptions when ActiveMQ Server start or stop, in the [ActiveMQTestResource](https://github.com/apache/camel-quarkus/commit/bf856616cf18c2ddb81286e13f9118d037f91021#diff-6a006df9400353640e04e51565bc48ba).

lburgazzoli:
Fixed by https://github.com/apache/camel-quarkus/pull/235

PR #235, opened 2019-10-07
#234 Added log when starting or stoping ActiveMQ Broker failed

WillemJiang:


integration-tests/paho/src/test/java/org/apache/camel/quarkus/component/paho/it/ActiveMQTestResource.java
```diff
@@ -38,7 +42,7 @@
             broker.setDataDirectory("target");
             broker.start();
         } catch (Exception e) {
-
+            LOGGER.error("Starting the ActiveMQ broker with exception.", e);
```
ppalaga: Does it make any sense to continue with the execution of the tests if this fails? I wonder if throwing a RuntimeException would not be better, like they do e.g. here in Quarkus? https://github.com/quarkusio/quarkus/blob/master/integration-tests/artemis-core/src/test/java/io/quarkus/it/artemis/ArtemisTestResource.java#L23

WillemJiang:
> Thanks, please squash the two commits into one for the sake of a cleaner history.

Done, the two  commits were squashed.

PR #242, opened 2019-10-08
quarkus-camel-catalog

### LINKED ISSUE #75: Include quarkus info in camel-catalog
We should add information about what components are quarkus ready to the camel catalog

/cc @davsclaus @oscerd

davsclaus:
Fixes #75: Adding camel-quarkus-catalog

This implementation is similar to what we do at apache camel using maven plugins as tools to generate catalog metadata that gets included in the camel-quarkus-catalog JAR.

### DIFF
--- a/tooling/maven/package-maven-plugin/src/main/java/org/apache/camel/quarkus/maven/PackageHelper.java
+++ b/tooling/maven/package-maven-plugin/src/main/java/org/apache/camel/quarkus/maven/PackageHelper.java
@@ -67,6 +67,35 @@ public final class PackageHelper {
         }
     }
 
+    /**
+     * Converts the value to use title style instead of dash cased
+     */
+    public static String camelDashToTitle(String value) {
+        StringBuilder sb = new StringBuilder(value.length());
+        boolean dash = false;
+
+        for (char c : value.toCharArray()) {
+            if ('-' == c) {
+                dash = true;
+                continue;
+            }
+
+            if (dash) {
+                sb.append(' ');
+                sb.append(Character.toUpperCase(c));
+            } else {
+                // upper case first
+                if (sb.length() == 0) {
+                    sb.append(Character.toUpperCase(c));
+                } else {
+                    sb.append(c);
+                }
+            }
+            dash = false;
+        }
+        return sb.toString();
+    }
+
     public static void writeText(File file, String text) throws IOException {
         FileOutputStream fos = new FileOutputStream(file, false);
         try {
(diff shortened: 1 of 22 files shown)


tooling/maven/package-maven-plugin/src/main/resources/META-INF/LICENSE.txt
```diff
@@ -0,0 +1,203 @@
+
```
ppalaga: A license and notice files duplicated once again.

tooling/maven/package-maven-plugin/src/main/resources/camel-quarkus-package-maven-plugin.properties
```diff
@@ -0,0 +1,16 @@
+## ---------------------------------------------------------------------------
```
ppalaga: Is this empty properties file really needed?

tooling/maven/package-maven-plugin/src/main/java/org/apache/camel/quarkus/maven/PackageHelper.java
```diff
@@ -0,0 +1,99 @@
...
+     * Loads the entire stream into memory as a String and returns it.
+     * <p/>
+     * <b>Notice:</b> This implementation appends a <tt>\n</tt> as line
+     * terminator at the of the text.
+     * <p/>
+     * Warning, don't use for crazy big streams :)
+     */
+    public static String loadText(InputStream in) throws IOException {
+        StringBuilder builder = new StringBuilder();
+        InputStreamReader isr = new InputStreamReader(in);
```
ppalaga: I am not sure what possible resources is this going to read but using `new InputStreamReader(in)` that uses the given platform's default encoding does not seem to be right? I mean `new InputStreamReader(in, 'utf-8') or similar is perhaps a better bet?

ppalaga: If `new InputStreamReader(in)` throws an exception, then `in` is not closed. Could you please enclose all closeables in a try-with?

tooling/maven/package-maven-plugin/src/main/java/org/apache/camel/quarkus/maven/PackageHelper.java
```diff
@@ -0,0 +1,99 @@
...
+                    sb.append(c);
+                }
+            }
+            dash = false;
+        }
+        return sb.toString();
+    }
+
+    public static void writeText(File file, String text) throws IOException {
+        FileOutputStream fos = new FileOutputStream(file, false);
```
ppalaga: Could you please use try-with to have less lines of code?

tooling/maven/package-maven-plugin/src/main/java/org/apache/camel/quarkus/maven/PackageHelper.java
```diff
@@ -0,0 +1,99 @@
...
+        try {
+            fos.write(text.getBytes());
```
ppalaga: String.getBytes() uses the given platform's default encoding. `text.getBytes('utf-8')` or similar would perhaps be a better choice?

PR #243, opened 2019-10-09
Generate extension list readme file via tooling like we do at Apache Camel

davsclaus:


extensions/readme.adoc
```diff
@@ -0,0 +1,142 @@
+= Components
+
```
ppalaga: Not sure if this readme is produced to be included somewhere else where the context is clear, but here when the file is read as is, it is unclear what components are listed here. If they are supposed to be the components supported on Quarkus, then it should say so.

Further, the versions in the Available From column seem to be Apache Camel versions and the document should say so, so that the reader does not end up with a false impression that those are Camel Quarkus versions.

I like the format and we should consider re-using this data in https://github.com/apache/camel-quarkus/blob/master/docs/modules/ROOT/pages/list-of-camel-quarkus-extensions.adoc

davsclaus: Yeah lets use the quarkus release version

davsclaus: Yeah lets use camel-quarkus version here

Issue #245, opened 2019-10-09
Update doc with better extension list


davsclaus:
Use the maven plugin to update the website doc like we do at Apache Camel, instead of the groovy build script hack.

PR #250, opened 2019-10-10
#245 update the extensions list in the website docs also

davsclaus:


docs/modules/ROOT/pages/list-of-camel-quarkus-extensions.adoc
```diff
@@ -1,10 +1,143 @@
...
+== Camel Components
 
-In addition to the above extensions there are the following ancillary extensions:
+// components: START
+Number of Camel components: 22 in 18 JAR artifacts (0 deprecated)
 
-include::{partialsdir}/common-extensions.adoc[]
+[width="100%",cols="4,1,5",options="header"]
+|===
+| Component | Available From | Description
```
ppalaga: A non-native speaker's wording notice: Would maybe "Since" be a better alternative for "Available from"? I mean most Java devs should know the `@since` JavaDoc tag that means exactly what we want to express here.

PR #266, opened 2019-10-13
Fixes #265 - Add tarfile extension

### LINKED ISSUE #265: Add tarfile extension
Similar to zipfile so it ought not to be to difficult if the commons-compress works with graal.

davsclaus:


integration-tests/tarfile/src/test/java/org/apache/camel/quarkus/component/tarfile/it/TarfileTest.java
```diff
@@ -0,0 +1,60 @@
...
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        ByteArrayInputStream bis = new ByteArrayInputStream(body);
+        TarArchiveInputStream tis = (TarArchiveInputStream) new ArchiveStreamFactory().createArchiveInputStream(ArchiveStreamFactory.TAR, bis);
+
+        TarArchiveEntry entry = tis.getNextTarEntry();
+        if (entry != null) {
+            IOHelper.copy(tis, bos);
+        }
+
+        String str = bos.toString();
```
ppalaga: bos.toString() is using given platforms default encoding. This might be fine if RestAssured would post also using given platforms default encoding. I do not have enough time to investigate which exact encoding RestAssured uses, but quick and dirty googling seems to suggest that has some hard coded platform independent defaults (like latin 1 or utf-8). If that's really the case, this test can fail on platforms that use some incompatible encoding, say UTF-16.

ppalaga: Perhaps the easiest way to avoid the problem would be to send something like `.contentType("application/octet-stream").body("Hello World".getBytes())`

davsclaus: The TarfileResource is set to return octet-stream, the input is plain text

PR #280, opened 2019-10-17
Update to quarkus 0.25.0

lburgazzoli:


integration-tests/platform-http-engine/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+            <artifactId>camel-quarkus-platform-http</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jsonb</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jsonb</artifactId>
+        </dependency>
```
ppalaga: What is the purpose of this test? Check that platform-http works when resteasy is in the classpath?

PR #283, opened 2019-10-18
Document health, metrics & OpenTracing extensions

jamesnetherton:
I wrote some docs for extensions that I added. 

I tweaked the doc extensions list mojo so that If you drop an adoc file into the docs extensions directory, it'll be linked to in the extensions summary readme. 

It'd be nice to eventually template & auto generate this stuff. For now, I just want to get an initial baseline added.

### DIFF
--- a/docs/modules/ROOT/pages/extensions/microprofile-health.adoc
+++ b/docs/modules/ROOT/pages/extensions/microprofile-health.adoc
@@ -1,0 +1,57 @@
+[[microprofile-health]]
+= MicroProfile Health Extension
+
+*Available as of Camel Quarkus version 0.21*
+
+The MicroProfile Health extension bridges the Camel health check API with MicroProfile Metrics & SmallRye Health.
+
+Maven users will need to add the following dependency to their `pom.xml` for this extension.
+
+[source,xml]
+------------------------------------------------------------
+<dependency>
+    <groupId>org.apache.camel.quarkus</groupId>
+    <artifactId>camel-quarkus-microprofile-health</artifactId>
+</dependency>
+------------------------------------------------------------
+
+== Usage
+
+You can register health checks for your applications with the https://camel.apache.org/manual/latest/health-check.html[Camel health check API].
+
+By default, classes extending `AbstractHealthCheck` are registered as both liveness and readiness checks. For finer control over whether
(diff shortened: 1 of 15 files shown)


docs/modules/ROOT/pages/extensions/microprofile-health.adoc
```diff
@@ -0,0 +1,58 @@
...
+The MicroProfile Health extension bridges the Camel health check API with MicroProfile Metrics & SmallRye Health.
+
+Maven users will need to add the following dependency to their `pom.xml` for this extension.
+
+[source,xml]
+------------------------------------------------------------
+<dependency>
+    <groupId>org.apache.camel.quarkus</groupId>
+    <artifactId>camel-quarkus-microprofile-health</artifactId>
+    <version>${camel-quarkus.version}</version>
```
ppalaga: Aren't we (or shouldn't we if we don't) generally encouraging the users to import our BOM? So maybe we could omit the version here.

docs/modules/ROOT/pages/extensions/microprofile-health.adoc
```diff
@@ -0,0 +1,58 @@
...
+</dependency>
+------------------------------------------------------------
+
+== Usage
+
+You can register health checks for your applications with the https://camel.apache.org/manual/latest/health-check.html[Camel health check API].
```
ppalaga: @zregvart Is this the preferred way to link documents from other Camel guides? I know there is `xref:` - is that one not better for some reason?

docs/modules/ROOT/pages/extensions/microprofile-health.adoc
jamesnetherton: Yeah, I did ponder whether to bother including the version. I'm happy to remove it.

docs/modules/ROOT/pages/extensions/microprofile-health.adoc
zregvart: `xref` would be preferred, with `xref` there's the option of pointing to a specific version of the document and it's not tied to the URL structure of the website.

ppalaga: Would the `antora/xref-validator` we run in this repo pass even if the health-check document is in another git repository?

Issue #278, opened 2019-10-16
Add minimal extension metadata


lburgazzoli:
For an example, have a look at:

    https://github.com/quarkusio/quarkus/blob/master/extensions/arc/runtime/src/main/resources/META-INF/quarkus-extension.json

PR #284, opened 2019-10-18
Followup #278 Assert that quarkus-extension.json exists for each exte…

ppalaga:
…nsion

### DIFF
--- a/build/scripts/validate-extension-metadata.groovy
+++ b/build/scripts/validate-extension-metadata.groovy
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import groovy.io.FileType
+import groovy.json.JsonSlurper
+
+final String quarkusExtensionJsonRelPath = 'runtime/src/main/resources/META-INF/quarkus-extension.json'
+final List<String> messages = []
+project.basedir.eachFile FileType.DIRECTORIES, { extensionParentDir ->
+    if (new File(extensionParentDir, 'runtime/pom.xml').exists()) {
+        final File extensionJsonFile = new File(extensionParentDir, quarkusExtensionJsonRelPath)
+        final String shortName = extensionParentDir.getName()
+        final String shortPath = shortName + '/' + quarkusExtensionJsonRelPath
+        final boolean internal = (shortName.startsWith('core') || shortName.endsWith('-common'))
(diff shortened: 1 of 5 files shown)


lburgazzoli:
FWIW the file is auto generated by the quarkus plugin in any case

ppalaga:
> FWIW the file is auto generated by the quarkus plugin in any case

My understanding is that it is generated in `${project.build.outputDirectory}` unless it exists in `src/main/resources`. But we definitely want a custom file in `src/main/resources` because of the labels and name. So we should check that we have it, should we not?

lburgazzoli:
IMHO, the validation is meaningful if it checks that the content has some minimal info, checking only the existence of the file does not add real value

johnpoth:
Also all these build checks are a bit painful when developing... Any chance to move them to a Profile ? Similar to what is done in Camel with the sourcecheck profile?

ppalaga:
> IMHO, the validation is meaningful if it checks that the content has some minimal info, checking only the existence of the file does not add real value

Added some checks in 7203021

> Also all these build checks are a bit painful when developing... Any chance to move them to a Profile ? Similar to what is done in Camel with the sourcecheck profile?

I moved it to validate phase so that these issues pop up at the very end. I do not think this should be optional because failing to deliver a valid metadata is a real functional problem.

lburgazzoli:
@ppalaga can you rebase?

ppalaga:
2ce6dc7 rebased

build/scripts/validate-extension-metadata.groovy
```diff
@@ -0,0 +1,49 @@
...
+final List<String> messages = []
+project.basedir.eachFile FileType.DIRECTORIES, { extensionParentDir ->
+    if (new File(extensionParentDir, 'runtime/pom.xml').exists()) {
+        final File extensionJsonFile = new File(extensionParentDir, quarkusExtensionJsonRelPath)
+        final String shortName = extensionParentDir.getName()
+        final String shortPath = shortName + '/' + quarkusExtensionJsonRelPath
+        final boolean ancillary = (shortName.startsWith('core') || shortName.endsWith('-common'))
+        if (!extensionJsonFile.exists()) {
+            messages.add(shortPath + ' is missing')
+        } else if (!ancillary) {
```
ppalaga: Actually not sure we want to ignore all these checks in the ancillary extensions. WDYT @lburgazzoli ? Maybe they should have a name, empty labels (or maybe integration, internal ?) and the guide does not matter?

lburgazzoli: core-* do already have the extension metadata file (well core-cloud has an empty one but that's a mistake) and probably xml-common should now be renamed core-xml so yes, every non support extension should have proper  metdata

ppalaga: Thanks. One more question: do we still consider the extensions satisfying `shortName.startsWith('core') || shortName.endsWith('-common')` to be "not end user facing"? - I mean should those maybe always have some sort of `internal` label that the script should enforce?

lburgazzoli: don't think so, it' perfectly acceptable for people using core alone and internal is a little misleading i.e. right now if you want to have xml support you have to explicitly add xml-common

ppalaga:
0557eee updated based on the discussion with @lburgazzoli

extensions/http-common/runtime/src/main/resources/META-INF/quarkus-extension.json
```diff
@@ -1 +1,8 @@
-{}
+{
+  "name": "Camel Quarkus HTTP Common",
+  "labels": [
+    "integration",
+    "camel"
```
ppalaga: Let me add HTTP here

ppalaga:
0557eee Added labels to XML and HTML common

PR #295, opened 2019-10-21
Enable PlatformHttpTest.invalidMethod()

ppalaga:
https://github.com/quarkusio/quarkus/issues/4408 seems to be fixed as of Quarkus 0.25


lburgazzoli:
I think this does not happen anymore because there’s no resteasy, we should check what the platform-http-reasyeasy tests do

PR #298, opened 2019-10-21
Improve XML support

### LINKED ISSUE #292: Rename xml-common extension to core-xml
As now users have to explicit add xml-common to get XML support

### LINKED ISSUE #294: Add a support/xml artifact for common substrate support
AWS extension require some xml support but now xml-common brings much more, we should create a support/xml artifact with common bits

lburgazzoli:
Fixes #294
Fixes #292


tooling/maven/package-maven-plugin/src/main/java/org/apache/camel/quarkus/maven/PrepareCatalogQuarkusMojo.java
```diff
@@ -63,7 +63,7 @@
     private static final String DEFAULT_FIRST_VERSION = "0.2.0";
 
     private static final String[] EXCLUDE_EXTENSIONS = {
-            "http-common", "jetty-common", "support", "xml-common", "xstream-common"
+        "core-xml", "http-common", "jetty-common", "support", "xstream-common"
```
ppalaga: Should `core-xml` really be here, given we said it is end-user facing? 

Besides, we currently seem to exclude all `-common` extensions, so the list could perhaps be simplified to `["-common", "support"]`

PR #312, opened 2019-10-22
Timer example

davsclaus:


davsclaus:
@lburgazzoli the are some options in application.properties - such as the dump route. Not sure if they are working or have changed name? I dont see any dumped route (I assume its disabled as it may require JAXB/XML). If so we can remove the cruft from this file to make it minimal.

examples/timer-log/README.adoc
```diff
@@ -0,0 +1,27 @@
...
+in Quarkus Dev Mode, you can edit the source code and have live updates.
+For example try to change the logging output to be `Bye World`.
+
+== Native build
+
+To build as native:
+
+[source,text]
+----
+$ mvn package -Pnative-image
```
ppalaga: `-Pnative-image` is correct, but in the user guide we say `-Dnative` which is equivalent and a bit shorter.

I would perhaps also be worthwhile to note that the above requires Docker.

The result is actually a Linux executable that won't run on Windows and Mac. Not sure we should go in such details here?

examples/timer-log/README.adoc
```diff
@@ -0,0 +1,27 @@
...
+This is a basic hello world example that uses a Camel timer that
+triggers every second and prints to the log.
+
+To run it:
+
+[source,text]
+----
+$ mvn clean compile quarkus:dev -DnoDeps
+----
+
```
ppalaga:
```suggestion
This compiles the project and starts the Quarkus tooling in the https://quarkus.io/guides/maven-tooling#development-mode[develpment mode].
```

ppalaga:
> dump route. Not sure if they are working or have changed name? I dont see any dumped route (I assume its disabled as it may require JAXB/XML). If so we can remove the cruft from this file to make it minimal.

Good catch. `org.apache.camel.quarkus.core.CamelConfig.Runtime.dumpRoutes` still exists, but is unused. I think it can be removed. There is enough route status messages logged even without that.

examples/timer-log/README.adoc
davsclaus: Updated and the other example is using native-image so we can update those instead

PR #323, opened 2019-10-23
Fix #322 Make camel-attachments an optional dependency of platform-http

### LINKED ISSUE #322: Make camel-attachments an optional dependency of platform-http
Followup of #308 

We should try to provide some no-op handling of file uploads by default that does not depend on camel-attachments and javax.activation and handle the uploads properly via camel-attachments (wrapped in an extension) and javax.activation only if the user adds them as deps.

The benefit for the users is that if they do not wish to handle uploads, they can save 560 kB on the native image size.

ppalaga:


extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -272,4 +274,27 @@ void start(
...
+         * <p>
+         * Note that this {@link BuildStep} is effective only if {@code camel-quarkus-attachments} extension is not in
+         * the class path.
+         *
+         * @param recorder the {@link CoreAttachmentsRecorder}
+         * @return a new {@link UploadAttacherBuildItem}
+         */
+        @Overridable
+        @Record(value = ExecutionTime.STATIC_INIT, optional = true)
+        @BuildStep(onlyIfNot = Flags.AttachmentsAvailable.class)
```
lburgazzoli: Don't think this flag is needed as the build item is marked as @Overridable, thus is is overridden by the build item from the attachment extension

ppalaga: I tried without that and it complained that it was produced twice. Let me try again.

lburgazzoli: You need to mark the recorder optional = true on both build items afair

ppalaga: Hm... It works without the flag, but `optional = true` must be there. Otherwise it throws `Cannot provide/consume multiple values for class org.apache.camel.quarkus.core.deployment.UploadAttacherBuildItem` That's a bit strange. I would not spontaneously consider UploadAttacherBuildItem to be optional. One or the other provider is always there.

Anyway, thanks for the catch and let me remove the flag.

lburgazzoli: Yes, there's an issue I think for that

ppalaga:
7dce10b Added the since versions fix

ppalaga:
d0e1398 a dummy change to restart the CI

oscerd:
The PR https://builds.apache.org/job/camel-quarkus-pr/313/

passed

ppalaga:
@lburgazzoli is this good to merge now?

extensions/attachments/runtime/src/main/java/org/apache/camel/quarkus/component/attachments/AttachmentsRecorder.java
```diff
@@ -0,0 +1,28 @@
+package org.apache.camel.quarkus.component.attachments;
```
jamesnetherton: Missing license header

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/UploadAttacher.java
```diff
@@ -0,0 +1,22 @@
+package org.apache.camel.quarkus.core;
```
jamesnetherton: Missing license header

ppalaga:
> Spotted a couple of missing license headers.

My understanding is that as long as the PR CI does not enforce the license headers they are optional for common commits. They are only required for releases.

oscerd:
> > Spotted a couple of missing license headers.
> 
> My understanding is that as long as the PR CI does not enforce the license headers they are optional for common commits. They are only required for releases.

Please add the ASF license headers. They are optional if nobody spotted them, someone spotted them (thanks @jamesnetherton ) and it's better to add them, instead of having to add a bunch of license headers during the release process.

oscerd:
So, when we find codestyle or missing headers, it always better to fix the situation at the moment, otherwise we could forget about it

ppalaga:
Fixed the missing license headers and added more platform-http tests in d391f4f

ppalaga:
I cannot stand the feeling that checking the style manually is wasting of time of both reviewers and PR authors. Yes, I can alias my maven to do the checks locally, but due to that, I'll just end up fixing someone else's style violations, because non-compliant code is commonly committed. I vote for everybody fixing these little things before submitting and the default Maven build and CI enforcing them. That would be less work for both PR authors and reviewers.

ppalaga:
0f30536 fixed the failing encoding test in the native mode.

ppalaga:
The Jenkins nodes without Maven feel like Russian roulette :-/

ppalaga:
The CI passed finally. Is this good to merge?

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CoreAttachmentsRecorder.java
```diff
@@ -0,0 +1,37 @@
...
+import io.quarkus.runtime.annotations.Recorder;
+
+@Recorder
+public class CoreAttachmentsRecorder {
+
+    private static final Logger LOG = Logger.getLogger(CoreAttachmentsRecorder.class);
+
+    public RuntimeValue<UploadAttacher> createNoOpUploadAttacher() {
+        return new RuntimeValue<>((File localFile, String fileName, Message message) -> {
+            LOG.trace("File %s not attached to message %s because camel-quarkus-attachments is not in the class path");
```
lburgazzoli: Wonder if we should thrown an exception here or at least log a warning as the user won't be really aware why something is not behaving as expected

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -330,27 +330,28 @@ static void populateCamelHeaders(RoutingContext ctx, Map<String, Object> headers
...
     }
 
-    static Message toCamelMessage(RoutingContext ctx, Exchange exchange, HeaderFilterStrategy headerFilterStrategy, String fileNameExtWhitelist) {
-        Message result = new DefaultMessage(exchange);
+    Message toCamelMessage(RoutingContext ctx, Exchange exchange) {
+        final Message result = new DefaultMessage(exchange);
 
-        populateCamelHeaders(ctx, result.getHeaders(), exchange, headerFilterStrategy);
+        final HeaderFilterStrategy headerFilterStrategy = getEndpoint().getHeaderFilterStrategy();
+        System.out.println("headerFilterStrategy = "+  headerFilterStrategy);
```
lburgazzoli: This should be removed

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CoreAttachmentsRecorder.java
ppalaga: @lburgazzoli  made me agree in the end. I was originally thinking only about the malevolent clients and though that the app ops do not want to see a stack trace for every such attempt. I also assumed that such cases might be desired, but now I think they are not. It is OK to fail for both malevolent and broken clients.

ppalaga:
5a40b6f addressed the comments of @lburgazzoli

lburgazzoli:
ok to test

PR #335, opened 2019-10-25
Fix #326 platform-http should return 415 for an unaccepted content type

### LINKED ISSUE #326: platform-http should return 415 for an unaccepted content type
Found this when adding some tests to platform-http. Will fix later. 

Plz. assign to me.

ppalaga:
This is a rather straightforward fix. I found that the older HTTP components suffer from the same problem. That might also mean that a solution exists, and I just have not found it?


ppalaga:
69dc5c3 added an analogous fix and tests for the situations where 406 is appropriate

ppalaga:
> I found that the older HTTP components suffer from the same problem.

REST with servlet definitely does. Jetty has a test case (`RestJettyContentTypeTest`) so it probably does not. I am yet to figure out how the support for 415 and 406 is implemented in Jetty to be able to make sure that the current proposal does not go too far away.

ppalaga:
Anyway, any feeback here would be welcome.

PR #340, opened 2019-10-28
Improve camel service discovery and filtering

lburgazzoli:


extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelServiceFilter.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.core;
+
+import java.util.function.Predicate;
+
+@FunctionalInterface
+public interface CamelServiceFilter extends Predicate<CamelServiceInfo> {
```
ppalaga: I fail to see what value is CamelServiceFilter interface adding. I'd personally prefer using `Predicate<CamelServiceInfo>` directly because that way I'd immediately see that it is just a Predicate and nothing else.

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelServiceFilterBuildItem.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.core.deployment;
+
+import io.quarkus.builder.item.MultiBuildItem;
+import org.apache.camel.quarkus.core.CamelServiceFilter;
+
+public final class CamelServiceFilterBuildItem extends MultiBuildItem {
```
ppalaga: I'd like us all to document BuildItems properly. The class level BuildItem JavaDoc should make the extension developers understand what is the purpose of the given BuildItem and when/how they should use it. If we provide no JavaDoc, the extension developers are doomed to click through our BuildProcessors and collect the information investing a lot more effort.

PR #344, opened 2019-10-28
WIP: Move to Quarkus 0.27+ yaml descriptor

maxandersen:
Why:

 * in 0.27 Quarkus are moving to a updated structure that is more flexible
   (use of open-ended `metadata` for tooling/UI annoations/labels)
 * in 0.27 we are also changing `quarkus-extension.json` to `quarkus-extension.yaml`
 * For now this only affect tools/plaform usage, thus there should be no runtime impact.

This change addreses the need by:

 * scripted conversion from old to new format. Verified same number of files changed.

Note: I've done what I could do to try maintain full backwards compatability but things would
be much simpler if we could drop that as soon as possible. Thus if you can please include this
as soon as https://github.com/quarkusio/quarkus/pull/4935 is merged and released.

I'll keep this in draft mode until I merged the dependent PR but feel free to ask questions if any :)
The sooner the better!


ppalaga:
Thanks for the PR, Max. 

To make this work, our matedata validation script would have to get adapted too https://github.com/apache/camel-quarkus/blob/master/build/scripts/validate-extension-metadata.groovy#L21

I wonder whether Quarkus now offers any means to auto-generate or validate the metadata files?

maxandersen:
> I wonder whether Quarkus now offers any means to auto-generate or validate the metadata files?

we can look into do that post v1. open an issue please. At the moment the source files are trivial listing name and keywords and guides so we haven't had an issue yet. Not sure what you want to validate besides syntax since all the info in there are optional ?

maxandersen:
Ah you are validating camel specific things - like attributes contains camel and integration.

Maybe better we look at having it so the extensionmojodescriptor can also take values from the pom properties and you can make it something you have shared in a parent pom. i.e. "quarkus.extension.metadata.keywords=camel,integration" we will then split up into the right thing.

that would be fairly trivial to do. We do it for name  already using the pom.xml <name> element.

maxandersen:
marking as ready for review as quarkus master now has all in there to work with the new format.

lburgazzoli:
cherry picking the commits here into https://github.com/apache/camel-quarkus/pull/350 so I can work also on upgrading the validation script

lburgazzoli:
done by #350

PR #348, opened 2019-10-29
Fix #220 platform-http component should return 204 for success and no…

### LINKED ISSUE #220: platform-http component should return 204 for success and no body
Yeah ideally we should get the http components aligned. They should IMHO not return a hardcoded text, eg this is in english, then chinese users get english text etc. But lets log a ticket upstream at Apache Camel

_Originally posted by @davsclaus in https://github.com/apache/camel-quarkus/timeline_

ppalaga:
… body

### DIFF
--- a/extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
+++ b/extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
@@ -152,11 +152,8 @@ public class QuarkusPlatformHttpConsumer extends DefaultConsumer {
 
     static Object toHttpResponse(HttpServerResponse response, Message message, HeaderFilterStrategy headerFilterStrategy) {
         final Exchange exchange = message.getExchange();
-        final boolean failed = exchange.isFailed();
-        final int defaultCode = failed ? 500 : 200;
 
-        final int code = message.getHeader(Exchange.HTTP_RESPONSE_CODE, defaultCode, int.class);
-
+        final int code = determineResponseCode(exchange, message.getBody());
         response.setStatusCode(code);
 
         final TypeConverter tc = exchange.getContext().getTypeConverter();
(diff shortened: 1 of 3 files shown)


extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -111,13 +111,18 @@ protected void doStart() throws Exception {
...
-                    getProcessor().process(e);
-                    writeResponse(ctx, e, getEndpoint().getHeaderFilterStrategy());
+                    createUoW(exchange);
+                    getAsyncProcessor().process(
+                        exchange,
+                        doneSync -> writeResponse(ctx, exchange, getEndpoint().getHeaderFilterStrategy())
+                    );
                 } catch (Exception e) {
                     LOG.debugf(e, "Could not handle '%s'", path);
                     ctx.fail(e);
```
lburgazzoli: You may need to do something like: `getExceptionHandler().handleException(e)` like https://github.com/apache/camel/blob/master/components/camel-undertow/src/main/java/org/apache/camel/component/undertow/UndertowConsumer.java#L171

ppalaga: Yes, I'd see `getExceptionHandler().handleException(e)` as a good replacement for the current `LOG.debugf(e, "Could not handle '%s'", path)` because the default `LoggingExceptionHandler` just does some logging and the message it logs is actually better than `Could not handle '%s'`.

But we still need to mark the Vert.x `RoutingContext` as failed somehow, because otherwise it will hang forever, won't it? We have no custom ExceptionHandler on the Camel side that would do this and I see no good reason to create one, because calling `ctx.fail(e)` is much less code. WDYT?

davsclaus: yeah ctx.fail is fine

ppalaga:
bf76e48 replaced `LOG.debugf(e, "Could not handle '%s'", path)` with `getExceptionHandler().handleException("Failed handling platform-http endpoint " + path, exchg, e)` and enclosed `toExchange(ctx)` in the exception handler scope.

PR #371, opened 2019-11-04
Issue 337

gnodet:


build/eclipse-formatter-config.xml
```diff
@@ -0,0 +1,264 @@
...
+        <setting id="org.eclipse.jdt.core.formatter.comment.clear_blank_lines" value="true"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.format_comments" value="true"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.format_header" value="false"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.format_html" value="true"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.format_source_code" value="true"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.indent_parameter_description" value="true"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.indent_root_tags" value="true"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.insert_new_line_before_root_tags" value="insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.insert_new_line_for_parameter" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.line_length" value="80"/>
```
ppalaga: Is there a good reason to limit the line length in comments at 80? The common practice in this repo is 120.

build/eclipse-formatter-config.xml
```diff
@@ -0,0 +1,264 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<profiles version="8">
+    <profile name="Camel Java Conventions" version="8" kind="CodeFormatterProfile">
+        <setting id="org.eclipse.jdt.core.formatter.align_type_members_on_columns" value="false"/>
+        <setting id="org.eclipse.jdt.core.formatter.alignment_for_arguments_in_allocation_expression" value="18"/>
```
ppalaga: All these `value="18"` occurrences feel strange. I'd expect them to be something divisible by 4, e.g. 16, so that it is possible to get them by pressing tab four times in an IDE/editor?

build/eclipse-formatter-config.xml
```diff
@@ -0,0 +1,264 @@
...
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_between_empty_parens_in_annotation_type_member_declaration" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_between_empty_parens_in_constructor_declaration" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_between_empty_parens_in_enum_constant" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_between_empty_parens_in_method_declaration" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_between_empty_parens_in_method_invocation" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.keep_else_statement_on_same_line" value="false"/>
+        <setting id="org.eclipse.jdt.core.formatter.keep_empty_array_initializer_on_one_line" value="false"/>
+        <setting id="org.eclipse.jdt.core.formatter.keep_imple_if_on_one_line" value="false"/>
+        <setting id="org.eclipse.jdt.core.formatter.keep_then_statement_on_same_line" value="false"/>
+        <setting id="org.eclipse.jdt.core.formatter.lineSplit" value="180"/>
```
ppalaga: Lines having more than 128 chars are hard to review on GitHub.

build/eclipse-formatter-config.xml
```diff
@@ -262,5 +262,6 @@
         <setting id="org.eclipse.jdt.core.formatter.tabulation.char" value="space"/>
         <setting id="org.eclipse.jdt.core.formatter.tabulation.size" value="4"/>
         <setting id="org.eclipse.jdt.core.formatter.use_tabs_only_for_leading_indentations" value="false"/>
+        <setting id="org.eclipse.jdt.core.formatter.use_on_off_tags" value="true"/>
```
ppalaga: Adding the start/end tags explicitly would be nice, plus they should be the same as we have in checkstyle
```
<setting id="org.eclipse.jdt.core.formatter.disabling_tag" value="CHECKSTYLE:OFF"/>
<setting id="org.eclipse.jdt.core.formatter.enabling_tag" value="CHECKSTYLE:ON"/>
```

build/eclipse-formatter-config.xml
gnodet: I agree that 80 is a bit too small, but the original file comes directly from the camel tree.  I'm happy to make it easier to use though.

examples/timer-log-xml/src/main/java/org/acme/timer/package-info.java
```diff
@@ -1,18 +1,14 @@
...
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements. See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
- * The ASF licenses this file to You under the Apache License, Version 2.0
- * (the "License"); you may not use this file except in compliance with
- * the License. You may obtain a copy of the License at
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements. See the NOTICE
+ * file distributed with this work for additional information regarding copyright ownership. The ASF licenses this file
+ * to You under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+ * License. You may obtain a copy of the License at
```
ppalaga: This license header change looks unwanted. It is perhaps caused by the fact that it is using a javadoc comment style `/**` instead of plain `/*`

examples/rest-json/src/main/java/org/acme/rest/json/Fruit.java
```diff
@@ -58,7 +58,7 @@ public boolean equals(Object obj) {
             return false;
         }
 
-        Fruit other = (Fruit) obj;
+        Fruit other = (Fruit)obj;
```
ppalaga: Having a space after the cast seems to be more common, even in the main Camel code base.

The following setting would do that:

```
<setting id="org.eclipse.jdt.core.formatter.insert_space_after_closing_paren_in_cast" value="insert"/>
```

pom.xml
```diff
@@ -253,6 +253,24 @@
...
+                <version>2.10.0</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>format</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <configFile>${maven.multiModuleProjectDirectory}/build/eclipse-formatter-config.xml</configFile>
```
gnodet: We should use `camel.quarkus.project.root` instead

ppalaga:
-1 for the last commit `examples should import the BOM instead of inheriting it as users would do` 

The previous setup was intentional to keep apart the plugin config we need for our build from the  config the end users need for their builds. With the original setup it was much easy to copy an example out of our source tree and use it as a base for the user's real world project. The user guide is broken now.

gnodet:
> -1 for the last commit `examples should import the BOM instead of inheriting it as users would do`
> 
> The previous setup was intentional to keep apart the plugin config we need for our build from the config the end users need for their builds. With the original setup it was much easy to copy an example out of our source tree and use it as a base for the user's real world project. The user guide is broken now.

I would challenge the fact that it's easier to copy from the previous setup.  The only difference is that we would now have to remove the parent instead of removing the relative path.  That's not much a difference.
The benefit is that the examples do now inherit from the "global" settings (autoformatting, legal files, ...), else we'd have to add a bunch of plugins definitions to each example, or simply have them not complying with the rest of the build.
Also, I really think that the correct way of using BOMs is to import them, not to create a child project.
I can edit the user guide to reflect the new setup if you agree.

ppalaga:
> I really think that the correct way of using BOMs is to import them

+1 to that, but I still think that with a parent where the necessary plugins are configured, the end user can start faster.

> I can edit the user guide to reflect the new setup if you agree.

+1, thanks

gnodet:
> > I really think that the correct way of using BOMs is to import them
> 
> +1 to that, but I still think that with a parent where the necessary plugins are configured, the end user can start faster.

Yeah, that's true, especially as BOMs can not be used to define plugin management.  I think I'll go for a middle ground and create a new pom so that our example can be fully integrated and modify the user guide to point to the pom as it was before.

> 
> > I can edit the user guide to reflect the new setup if you agree.
> 
> +1, thanks

gnodet:
@ppalaga I've pushed a fix, let me know if that suits you

ppalaga:
> @ppalaga I've pushed a fix, let me know if that suits you

I'd like to have a look, but I am not finding any new related commit anywhere, neither in master or in this issue-337 branch. Could you please point me to where you pushed it?

gnodet:
> > @ppalaga I've pushed a fix, let me know if that suits you
> 
> I'd like to have a look, but I am not finding any new related commit anywhere, neither in master or in this issue-337 branch. Could you please point me to where you pushed it?

https://github.com/apache/camel-quarkus/commit/b35d5fd545eaf914766e5d4e044346a316b75339

PR #394, opened 2019-11-11
Fix license headers

ppalaga:


pom.xml
```diff
@@ -304,10 +305,14 @@
                                 <exclude>**/META-INF/persistence*.xsd</exclude>
                                 <exclude>**/pom.xml.versionsBackup</exclude>
                                 <exclude>ide-config/**</exclude>
+                                <exclude>**/*.mvel</exclude>
+                                <exclude>**/quarkus-extension.yaml</exclude>
```
ppalaga: I think we do not want the license headers in the quarkus-extension.yaml files. WDYT?

lburgazzoli: If it does not harm the quarkus processing I would say we should add to be consistent with other similar files (i.e. properties)

ppalaga: OK, removed the line.

lburgazzoli:
ok to test

PR #400, opened 2019-11-12
Add an option to filter auto-discovered routes

### LINKED ISSUE #396: Add an option to filter auto-discovered routes
camel-main has its own mechanism to include/exclude routes but this happen at runtime, we should implement a similar mechanism for build time discovery

lburgazzoli:
Fixes #396

### DIFF
--- a/extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSupport.java
+++ b/extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSupport.java
@@ -36,6 +36,8 @@ import org.apache.camel.RoutesBuilder;
 import org.apache.camel.builder.AdviceWithRouteBuilder;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.quarkus.core.CamelServiceInfo;
+import org.apache.camel.util.AntPathMatcher;
+import org.apache.camel.util.ObjectHelper;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.IndexView;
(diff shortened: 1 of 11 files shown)


lburgazzoli:
ok to test

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSupport.java
```diff
@@ -63,6 +65,29 @@ public static boolean isPublic(ClassInfo ci) {
         }
     }
 
+    public static boolean isPathFiltered(String path, Collection<String> excludePatterns, Collection<String> includePatterns) {
```
ppalaga: I'd be for a cleanly non-negated name and semantics here, e.g. `isPathIncluded`

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
```diff
@@ -40,5 +42,48 @@
...
+        public RoutesDiscoveryConfig routesDiscovery;
+    }
+
+    @ConfigGroup
+    public static class RoutesDiscoveryConfig {
+        /**
+         * Enable automatic discovery of routes during static initialization.
+         */
+        @ConfigItem(defaultValue = "true")
+        public boolean enabled;
```
ppalaga: I assume this option is taken into account only if main is enabled. Therefore I'd expect to find it under `camel.main` rather than as a sibling of camel main.

lburgazzoli: it is already, see https://github.com/apache/camel-quarkus/pull/400/files/e7a680d139f1e4d050cba26cb99572454976023f#diff-cd72e3ae157d3249a7dfe5705d1ef89bR50

ppalaga: Thanks, and sorry for my misunderstanding.

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSupport.java
lburgazzoli: done

lburgazzoli:
ok to test

lburgazzoli:
Need to be rebases after https://github.com/apache/camel-quarkus/pull/358 get merged

lburgazzoli:
Superseded by https://github.com/apache/camel-quarkus/pull/402

PR #404, opened 2019-11-12
Scheduler

davsclaus:


integration-tests/scheduler/src/test/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+@QuarkusTest
+class SchedulerTest {
+
+    @Test
+    public void test() throws Exception {
+        // give time for scheduler to run a bit
+        Thread.sleep(50);
+
+        String body = RestAssured.get("/scheduler/get").then().statusCode(200).extract().body().asString();
```
ppalaga: CI can be endlessly slow. Would it not be safer to poll `/scheduler/get` in a loop while it is equal 0 within some reasonable timeout?

lburgazzoli:
ok to test

ppalaga:
> I don't think 50 ms could lead to an endlessy slow CI job.

No, sorry for the confusion. "Slow" was meant for the system running the test, so 50ms might not be enough for the tested code to complete the expected tasks. I like the new version.

PR #405, opened 2019-11-12
core: avoid referencing application classes in bean build items and loading them in core build processors

lburgazzoli:
Processors should avoid loading application classes during augmentation [1] so the reference to
beans class has been removed from `Camel*BeanBuildItem` and replaced by a simple string. I will 
follow up this PR with an additional one after reviewing how `Camel*BeanBuildItem` are used by
extensions.

[1] https://quarkus.io/guides/writing-extensions#bootstrap-three-phases

### DIFF
--- a/extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
+++ b/extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
@@ -17,40 +17,21 @@
 package org.apache.camel.quarkus.core.deployment;
 
 import java.io.IOException;
+import java.lang.reflect.Modifier;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.lang.reflect.Modifier;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
-import io.quarkus.arc.deployment.BeanContainerBuildItem;
-import io.quarkus.deployment.ApplicationArchive;
-import io.quarkus.arc.deployment.BeanRegistrationPhaseBuildItem;
-import io.quarkus.arc.deployment.UnremovableBeanBuildItem;
-import io.quarkus.arc.deployment.UnremovableBeanBuildItem.BeanClassNamesExclusion;
-import io.quarkus.arc.processor.BeanInfo;
-import io.quarkus.arc.processor.BuildExtension;
-import io.quarkus.deployment.Capabilities;
-import io.quarkus.deployment.annotations.BuildProducer;
-import io.quarkus.deployment.annotations.BuildStep;
-import io.quarkus.deployment.annotations.ExecutionTime;
-import io.quarkus.deployment.annotations.Overridable;
-import io.quarkus.deployment.annotations.Record;
-import io.quarkus.deployment.builditem.ApplicationArchivesBuildItem;
-import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
-import io.quarkus.deployment.builditem.ServiceStartBuildItem;
-import io.quarkus.deployment.builditem.ShutdownContextBuildItem;
-import io.quarkus.deployment.recording.RecorderContext;
-import io.quarkus.runtime.RuntimeValue;
+import io.quarkus.arc.processor.BeanRegistrar;
 import org.apache.camel.CamelContext;
 import org.apache.camel.RoutesBuilder;
-import org.apache.camel.impl.converter.BaseTypeConverterRegistry;
 import org.apache.camel.builder.AdviceWithRouteBuilder;
 import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.impl.converter.BaseTypeConverterRegistry;
 import org.apache.camel.quarkus.core.CamelConfig;
 import org.apache.camel.quarkus.core.CamelMain;
 import org.apache.camel.quarkus.core.CamelMainProducers;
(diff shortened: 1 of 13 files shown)


lburgazzoli:
ok to test

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -71,6 +52,27 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
```
ppalaga: Isn't the official ordering `io.*` before `org.*`? We should introduce the `impsort-maven-plugin` to get rid of these problems.

extensions/microprofile-metrics/deployment/src/main/java/org/apache/camel/quarkus/component/microprofile/metrics/deployment/MicroProfileMetricsProcessor.java
```diff
@@ -41,7 +42,7 @@ FeatureBuildItem feature() {
     CamelBeanBuildItem metricRegistry(CamelMicroProfileMetricsRecorder recorder) {
         return new CamelBeanBuildItem(
                 MicroProfileMetricsConstants.METRIC_REGISTRY_NAME,
-                MetricRegistry.class,
+                MetricRegistry.class.getName(),
```
ppalaga: This should also be a DotName.

extensions/opentracing/deployment/src/main/java/org/apache/camel/quarkus/component/opentracing/deployment/OpenTracingProcessor.java
```diff
@@ -47,7 +48,9 @@ void process(BuildProducer<NativeImageResourceBuildItem> resource) {
     CamelBeanBuildItem setupCamelOpenTracingTracer(CamelOpenTracingConfig config, CamelOpenTracingRecorder recorder,
             BeanContainerBuildItem beanContainer) {
         // Configure & bind OpenTracingTracer to the registry so that Camel can use it
-        return new CamelBeanBuildItem("tracer", Tracer.class,
+        return new CamelBeanBuildItem(
+                "tracer",
+                Tracer.class.getName(),
```
ppalaga: DotName

extensions/platform-http/deployment/src/main/java/org/apache/camel/quarkus/component/platform/http/deployment/PlatformHttpProcessor.java
```diff
@@ -88,7 +88,7 @@ PlatformHttpEngineBuildItem platformHttpEngine(
     CamelRuntimeBeanBuildItem platformHttpEngineBean(PlatformHttpRecorder recorder, PlatformHttpEngineBuildItem engine) {
         return new CamelRuntimeBeanBuildItem(
                 PlatformHttpConstants.PLATFORM_HTTP_ENGINE_NAME,
-                QuarkusPlatformHttpEngine.class,
+                QuarkusPlatformHttpEngine.class.getName(),
```
ppalaga: DotName

extensions/platform-http/deployment/src/main/java/org/apache/camel/quarkus/component/platform/http/deployment/PlatformHttpProcessor.java
```diff
@@ -97,7 +97,7 @@ CamelRuntimeBeanBuildItem platformHttpEngineBean(PlatformHttpRecorder recorder,
     CamelRuntimeBeanBuildItem platformHttpComponentBean(PlatformHttpRecorder recorder, PlatformHttpEngineBuildItem engine) {
         return new CamelRuntimeBeanBuildItem(
                 PlatformHttpConstants.PLATFORM_HTTP_COMPONENT_NAME,
-                PlatformHttpComponent.class,
+                PlatformHttpComponent.class.getName(),
```
ppalaga: DotName

integration-tests/core/deployment/src/main/java/org/apache/camel/quarkus/core/support/deployment/SupportBuildStep.java
```diff
@@ -29,7 +29,7 @@
     CamelBeanBuildItem logComponent(SupportRecorder recorder) {
         return new CamelBeanBuildItem(
                 "log",
-                LogComponent.class,
+                LogComponent.class.getName(),
```
ppalaga: DotName

extensions/microprofile-metrics/deployment/src/main/java/org/apache/camel/quarkus/component/microprofile/metrics/deployment/MicroProfileMetricsProcessor.java
lburgazzoli: If read the PR comment, you'd find that I will follow up this PR with a new one in which I'll review how the bean build items are used by extensions.

lburgazzoli:
> +1 for the goal, I also thought of doing it. Passing a class proxy instead of the class did not work for me on this place: https://github.com/apache/camel-quarkus/pull/402/files#diff-ec6205752018e84a80fcd9eaa92af189R320 so it probably won't work for you here: https://github.com/apache/camel-quarkus/pull/405/files#diff-ec6205752018e84a80fcd9eaa92af189R191 It might well be a bug in Quarkus.

It should be fine now as the way it is used matches the pattern for which the class proxy is needed. In your code I suspect the problem was the fact that the proxy was owned by a build item and maybe triggering a bug.

The failure on CI is caused by the validator extension, don't know why.

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
lburgazzoli: [here](https://github.com/apache/camel/blob/master/buildingtools/src/main/resources/camel-checkstyle.xml#L138-L141) the checkstyle configuration for imports so it looks good to me

lburgazzoli:
ok to test

extensions/microprofile-metrics/deployment/src/main/java/org/apache/camel/quarkus/component/microprofile/metrics/deployment/MicroProfileMetricsProcessor.java
ppalaga: Sorry, I was not sure you mean exactly this.

ppalaga:
> the proxy be owned by a build item

That sounds like a valid hypothesis. Many thanks for mentioning it!

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
ppalaga: Strange. I think I saw `io.*` before `org.*` in some classes and started respecting it since then.

lburgazzoli:
@davsclaus ass the pr is failing for unrelated failures I'd go for merging it.

About XSLT, I've raised an issue https://github.com/apache/camel-quarkus/issues/419 and I'm working on a substitution

davsclaus:
Increate the CI timeout for that test, as it seems it took a bit more than 2 sec for replying

PR #407, opened 2019-11-13
Xslt

davsclaus:


oscerd:
I will never understand why you have to polemize each time, Peter.

ppalaga:
I hope it did not sound self-purposeful. I admit the usefulness of the git history is a topic for the dev list.

PR #414, opened 2019-11-13
chore: add timer-log-cdi example

lburgazzoli:


examples/timer-log-cdi/README.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+The above command compiles the project, starts the application and lets the Quarkus tooling watch for changes in your
+workspace. Any modifications in your project will automatically take effect in the running application.
+
+TIP: Please refer to the Development mode section of
+https://camel.apache.org/camel-quarkus/latest/user-guide.html#_development_mode[Camel Quarkus User guide] for more details.
+
+Then look at the log output in the console. As we run the example
+in Quarkus Dev Mode, you can edit the source code and have live updates.
+For example try to change the logging output to be `Bye World` or the property `timer.timeou` in application.properties.
```
ppalaga: s/timer.timeou/timer.period/ perhaps?

examples/timer-log-cdi/src/main/resources/application.properties
```diff
@@ -0,0 +1,38 @@
...
+# Quarkus
+#
+quarkus.log.file.enable = false
+
+#
+# Quarkus - Camel
+#
+
+# disable build time route discovery
+quarkus.camel.main.routes-discovery.enabled = false
```
ppalaga: After #136 was fixed, this should not be necessary for the example to work. Is it really necessary? (Wondering)

examples/timer-log-cdi/README.adoc
lburgazzoli: don't know from where this comes as I everywhere it is `timer.period` :)

examples/timer-log-cdi/src/main/resources/application.properties
lburgazzoli: It is not necessary just something you can do to completely avoid discovery for routes.

ppalaga: Related thought: `quarkus.camel.main.routes-discovery.enabled = false` can actually be expressed as `quarkus.camel.main.routes-discovery.exclude = **/*` so maybe we could remove `quarkus.camel.main.routes-discovery.enabled` ?

ppalaga: > It is not necessary just something you can do to completely avoid discovery for routes.

Will it not be confusing for the readers?

Note to myself: We should document that beans managed by Arc are not instantiated from a static initializer - a consequence of #136.

lburgazzoli: `quarkus.camel.main.routes-discovery.enabled` remove the needs to perform the discovery step so it gives a little boost to the build processor and it is much easier to remember as it matches the enable/disable patter used for other features and across platforms

lburgazzoli: > > It is not necessary just something you can do to completely avoid discovery for routes.
> 
> Will it not be confusing for the readers?
> 
Added some clarification to the property

lburgazzoli:
ok to test

PR #418, opened 2019-11-13
Fix #382 Support path parameters in platform-http

### LINKED ISSUE #382: Support path parameters in platform-http
I wonder if its in the rest-platform that we dont support rest-dsl with path placeholders

```
@ApplicationScoped
public class MyCamel extends RouteBuilder {

    @Override
        rest("/camel/").produces("text/plain")
            .get("{id}").to("bean:myBean");
    }
}
```

You can then not do localhost:8080/camel/123

This works in regular Camel

ppalaga:


extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
```diff
@@ -336,19 +341,16 @@ static void populateCamelHeaders(
...
-        //
-        // for (String paramName : predicateContextParams.keySet()) {
-        // headersMap.put(paramName, predicateContextParams.get(paramName));
-        // }
-        // }
+            /* Path parameters */
+            for (Entry<String, String> en : ctx.pathParams().entrySet()) {
+                final String name = en.getKey();
+                final String value = en.getValue();
+                if (!headerFilterStrategy.applyFilterToExternalHeaders(name, value, exchange)) {
```
ppalaga: @davsclaus do we actually need to filter the path params? Maybe not, because the keys are defined by developer, not by the app user.

lburgazzoli:
ok to test

extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/QuarkusPlatformHttpConsumer.java
davsclaus: Yeah if they are http context-path parameters then I dont see a reason for letting the camel filter strategy have a chance to filter them out.

ppalaga: OK, let me remove the filtering. It is not there in the Undertow component too.

ppalaga:
da24096 removed the filtering for the path params.

lburgazzoli:
ok to test

PR #437, opened 2019-11-18
xslt extension not working on java 11

### LINKED ISSUE #427: xslt extension not working on java 11
It appears that Java 11 introduces some changes to the underlying xslt engine that breaks how attributes are handled, in particular is is not possible to properly set the package name or the class name of the generated code as they get messed up

lburgazzoli:
Fixes #427

### DIFF
--- a/extensions/support/xalan/runtime/src/main/java/org/apache/camel/quarkus/support/xalan/XalanSupport.java
+++ b/extensions/support/xalan/runtime/src/main/java/org/apache/camel/quarkus/support/xalan/XalanSupport.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.support.xalan;
+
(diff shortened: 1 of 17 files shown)


lburgazzoli:
ok to test

extensions/support/xalan/runtime/src/main/java/org/apache/camel/quarkus/support/xalan/XalanSupport.java
```diff
@@ -0,0 +1,44 @@
...
+public final class XalanSupport {
+    private static final Logger LOGGER = LoggerFactory.getLogger(XalanSupport.class);
+
+    private XalanSupport() {
+    }
+
+    public static TransformerFactory newTransformerFactoryInstance() {
+        TransformerFactory factory = TransformerFactory.newInstance(
+                "org.apache.xalan.xsltc.trax.TransformerFactoryImpl",
+                Thread.currentThread().getContextClassLoader());
```
ppalaga: This looks like the key part of the solution. If so, it would be nice to add a short comment explaining what problem is this snippet solving so that the future readers do not need to wonder.

lburgazzoli: Should I expand the description of the problem [here](https://github.com/apache/camel-quarkus/issues/427) ?

ppalaga: Question: we cannot pass `org.apache.xalan.xsltc.trax.TransformerFactoryImpl` as a standard service provider, right? We'd thus get the right implementation, but the class loader would still be wrong, right?

lburgazzoli: Probably we can use a standard service provider but that's the most "secure" way as you can't override it with a system property. If you know a better way please go ahead and send a PR

ppalaga: > Should I expand the description of the problem [here](https://github.com/apache/camel-quarkus/issues/427) ?

I meant a comment in the above newTransformerFactoryInstance() method. It might not be clear to the future readers why it is there. A single sentence would be enough, something like `This solves the translet class loading problems on Java 11+, see https://github.com/apache/camel-quarkus/issues/427`

PR #440, opened 2019-11-19
Fix #428 XSLT extension does not work with file: URIs

### LINKED ISSUE #428: XSLT extension does not work with file: URIs
I just tried to extend the XsltTest with a `file:` use case and it fails. Let me try to fix it.

ppalaga:


extensions/xslt/deployment/src/main/java/org/apache/camel/quarkus/component/xslt/deployment/BuildTimeUriResolver.java
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.camel.util.FileUtil;
+import org.apache.camel.util.StringHelper;
+
+public class BuildTimeUriResolver {
+
+    /** A directory to use as a base for resolving {@code file:} URIs */
+    private final Path baseDir;
+
+    public BuildTimeUriResolver(Path baseDir) {
+        super();
```
lburgazzoli: not needed

extensions/xslt/runtime/src/main/java/org/apache/camel/quarkus/component/xslt/CamelXsltRecorder.java
```diff
@@ -46,8 +56,7 @@ protected void configureEndpoint(
                 String remaining,
                 Map<String, Object> parameters) throws Exception {
 
-            final String fileName = FileUtil.stripExt(remaining, true);
-            final String className = StringHelper.capitalize(fileName, true);
+            final String className = uriResolver.getTransletClassName(remaining);
```
lburgazzoli: can you start working on the upstream component to support this feature natively ?

extensions/xslt/deployment/src/main/java/org/apache/camel/quarkus/component/xslt/deployment/BuildTimeUriResolver.java
ppalaga: Sorry, now found the setting in Eclipse to omit it next time.

extensions/xslt/runtime/src/main/java/org/apache/camel/quarkus/component/xslt/CamelXsltRecorder.java
ppalaga: Not sure how that would make sense there because the upstream has no notion of build time/ runtime distinction? I mean a getTransletClassName() method could possibly be added to the upstream's XsltUriResolver, but the RuntimeUriResolver proposed here is quite different from it and am rather skeptical about letting RuntimeUriResolver to extend  XsltUriResolver. Or you mean something else?

ppalaga:
9fb37b4 removes the redundant `super()` invocations

extensions/xslt/runtime/src/main/java/org/apache/camel/quarkus/component/xslt/CamelXsltRecorder.java
lburgazzoli: I mean it would be nice to check what we need to to upstream to avoid having to subclass the component here.

lburgazzoli:
ok to test

extensions/xslt/runtime/src/main/java/org/apache/camel/quarkus/component/xslt/CamelXsltRecorder.java
ppalaga: OK, created https://github.com/apache/camel-quarkus/issues/441

PR #449, opened 2019-11-20
Add Hystirx component extension

### LINKED ISSUE #429: camel-hystrix extension
Even though quarkus has microprofile fault tolerance, then its based on hystrix. So we should add camel-hystrix as extension which we hopefully can get to native compilation too.

jamesnetherton:
fixes #429


jamesnetherton:
Looks like there's a little catalog issue I need to fix up...

lburgazzoli:
ok to test

lburgazzoli:
ok to test

Issue #352, opened 2019-10-30
Re-enable the FHIR tests once the FHIR testing service is back again


ppalaga:
Set `quarkus.camel.fhir.enable-dstu3=true` in https://github.com/apache/camel-quarkus/blob/master/integration-tests/fhir/src/main/resources/application.properties#L29

and
```
<quarkus.camel.fhir.enable-dstu2>true</quarkus.camel.fhir.enable-dstu2>
<quarkus.camel.fhir.enable-r4>true</quarkus.camel.fhir.enable-r4>
```
in  https://github.com/apache/camel-quarkus/blob/master/integration-tests/fhir/pom.xml#L81-L82

ppalaga:
@johnpoth I wonder whether it is exceptional that the FHIR testing service is down? 

In case it is rather normal that it is unstable, we should perhaps try to rewrite the test so that it runs only if an initial ping to the service succeeds.

lburgazzoli:
May be better to use something like:

    https://hapifhir.io/doc_server_tester.html

johnpoth:
Hi @ppalaga @lburgazzoli 

As @lburgazzoli pointed out, the goal would be to use our own FHIR server when testing which has been on my TODO list for quite some time (needed in Camel too). 

Ideally we would be using the server here https://github.com/hapifhir/hapi-fhir-jpaserver-starter. It would be cool to automatically find an unused port rather than a hard coded one.

I think we can disable the tests for now, I'll try to work on this sometime next week if that's ok ? Thanks !

lburgazzoli:
@johnpoth had any time to work on this ?

johnpoth:
@lburgazzoli I barely started and got pulled on something else ... should be okay this week :+1:

lburgazzoli:
We do use testcontainers now so we can close this one

PR #461, opened 2019-11-22
Startup a FHIR server when running the FHIR integration tests

johnpoth:
cc @lburgazzoli addresses #352 

This will fire up a FHIR server when running tests and the  `fhir-it` Maven profile is activated.

Note that the FHIR server, based on hapi-fhir-jpaserver-starter, only supports running _one_ FHIR version at a time. I'm working on a PR to address this. Therefore to test all FHIR specification versions, multiple runs are necessary.

The jetty-maven-plugin's deploy-war goal re-runs the _validate_ Maven phase therefore some plugins are run twice namely:

- directory-maven-plugin:0.3.1:highest-basedir (directories)
- groovy-maven-plugin:2.1.1:execute (sanity-checks)
- maven-enforcer-plugin:1.4.1:enforce ((enforce-maven-version, camel-quarkus-enforcer-rules)

By default, only the `FhirDataformatTest` tests are run and the FHIR server isn't started.

Thanks !

### DIFF
--- a/extensions/fhir/deployment/src/main/java/org/apache/camel/quarkus/component/fhir/deployment/FhirProcessor.java
+++ b/extensions/fhir/deployment/src/main/java/org/apache/camel/quarkus/component/fhir/deployment/FhirProcessor.java
@@ -26,6 +26,7 @@ import ca.uhn.fhir.model.dstu2.FhirDstu2;
 import ca.uhn.fhir.model.dstu2.resource.BaseResource;
 import ca.uhn.fhir.model.dstu2.valueset.*;
 import ca.uhn.fhir.rest.client.apache.ApacheRestfulClientFactory;
+import ca.uhn.fhir.util.jar.DependencyLogImpl;
 import ca.uhn.fhir.validation.schematron.SchematronBaseValidator;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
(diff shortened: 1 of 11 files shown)


ppalaga:
> This will fire up a FHIR server when running tests and the `fhir-it` Maven profile is activated.

Why don't we activate it by default?

> to test all FHIR specification versions, multiple runs are necessary.

Can't we start three jetty instances on three distinct ports in parallel?

I must say I am quite lost in the complexity of the setup.

johnpoth:
Hi @ppalaga,

Thanks for the feedback !

> > This will fire up a FHIR server when running tests and the `fhir-it` Maven profile is activated.
> 
> Why don't we activate it by default?
> 

I don't mind if we do. I put in a separate profile because integration tests that fire up external systems are usually not activated by default (e.g camel-infinispan).  

> > to test all FHIR specification versions, multiple runs are necessary.
> 
> Can't we start three jetty instances on three distinct ports in parallel?

Sure, this would be quite heavy though. I think we can wait until the hapi-fhir FHIR server supports a list of version ? The tests are setup so that even though the `fhir-it` profile is not activated, the Native tests still validate the FHIR dataformats.

> 
> I must say I am quite lost in the complexity of the setup.

I think it comes from the fact we are managing two integration test profiles: Native and one that fires up a FHIR server (fhir-it).

Thanks!

ppalaga:
> > > This will fire up a FHIR server when running tests and the `fhir-it` Maven profile is activated.
> > 
> > 
> > Why don't we activate it by default?
> 
> I don't mind if we do. I put in a separate profile because integration tests that fire up external systems are usually not activated by default (e.g camel-infinispan).

My understanding is that the jetty FHIR application fully simulates a FHIR service in such a way that it is not necessary to access any remote systems. If booting the FHIR service is within say 10 seconds and if its RAM utilization won't break the CI, I am for making it always active. Even better, if that opens a possibility to remove some conditionals that check what is enabled.

davsclaus:
All is green, lets merge ?

PR #462, opened 2019-11-22
Add spring-di example

lburgazzoli:


examples/timer-log-spring/README.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+The above command compiles the project, starts the application and lets the Quarkus tooling watch for changes in your
+workspace. Any modifications in your project will automatically take effect in the running application.
+
+TIP: Please refer to the Development mode section of
+https://camel.apache.org/camel-quarkus/latest/first-steps.html#_development_mode[Camel Quarkus User guide] for more details.
+
+Then look at the log output in the console. As we run the example
+in Quarkus Dev Mode, you can edit the source code and have live updates.
+For example try to change the logging output to be `Bye World` or the property `timer.period` in application.properties.
```
ppalaga: The `Bye World` suggestion is a bit outdated. The line is now `.setBody(() -> "Incremented the counter: " + counter.increment())` so we could perhaps suggest 

```
For example try to change `Incremented the counter` to `Counter state` in the message body or change the property `timer.period` in application.properties.

The same holds for the CDI example.

lburgazzoli:
ok to test

PR #464, opened 2019-11-22
Add FTP extension

### LINKED ISSUE #463: Create a Camel FTP extension

jamesnetherton:
fixes #463


lburgazzoli:
ok to test

jamesnetherton:
Just realised I need to fix up the `<firstVersion>` property.

jamesnetherton:
Can anyone else replicate this test failure in native mode? It's working fine for me.

lburgazzoli:
I think you may need to declare that you need ssl, like https://github.com/apache/camel-quarkus/blob/master/extensions/aws-s3/deployment/src/main/java/org/apache/camel/quarkus/component/aws/s3/deployment/AwsS3Processor.java#L59

jamesnetherton:
ok to test

lburgazzoli:
mmh, does not fails for me neither but still failing on jenkins... maybe something caused by the oracle jdk vs openjdk ?

jamesnetherton:
Yes, it's weird. I tested with the same JDK version Jenkins uses and it still passes.

lburgazzoli:
There’s an open issue here: https://github.com/quarkusio/quarkus/issues/2125

jamesnetherton:
Wondering if the Jenkins build environment is configured correctly?

If I unset `GRAALVM_HOME` locally I hit the same issue.

Maybe the other extensions that require native SSL support don't trigger the relevant code paths in their integration tests (AWS, Twitter & Salesforce tests don't run at all). Hence why we perhaps haven't hit this before.

lburgazzoli:
I think the main problem is that the code is built with docker but executed outside it, do you mind adding some notes to the linked issue or starting a conversation on zulip ?

jamesnetherton:
ok to test

davsclaus:
All is green, lets merge ? We can maybe mark this extension as preview

jamesnetherton:
Lets merge as-is. 

IMO it works fine in native mode when the environment is configured correctly. I raised https://github.com/apache/camel-quarkus/issues/468 as a follow up issue.

PR #475, opened 2019-11-26
Support dependency injection of FluentProducerTemplate/FluentConsumerTemplate

### LINKED ISSUE #474: Support dependency injection of FluentProducerTemplate/FluentConsumerTemplate
If you have a quarks rest service then you may want to inject Camel as a producer or consumer template (also the fluent builder kind)

```
@Path("/hello")
public class ExampleResource {

    @Inject
    private CamelContext camel;

    @GET
    @Produces(MediaType.TEXT_PLAIN)
    public String hello() {
        return "hello";
    }

    @GET
    @Path("/bye")
    @Produces(MediaType.TEXT_PLAIN)
    public String bye() throws Exception {
        return camel.createFluentProducerTemplate().to("direct:bye").request(String.class);
    }
}
```

But so far its only inject of `CamelContext` that works.

It would be nice to inject

    @Inject
    private ProducerTemplate camel;

And if you add the camel annotations for `@Produce` and `@Consume` and also the older `@EndpointInject` name which works on both producer and consumer templates

    @Inject
    @Produce("kafka:blah")
    private FluentProducerTemplate camel;

lburgazzoli:
Fixes #474

### DIFF
--- a/extensions/core/deployment/src/test/java/org/apache/camel/quarkus/core/runtime/CamelProducersTest.java
+++ b/extensions/core/deployment/src/test/java/org/apache/camel/quarkus/core/runtime/CamelProducersTest.java
@@ -22,6 +22,8 @@ import javax.inject.Inject;
 import io.quarkus.test.QuarkusUnitTest;
 import org.apache.camel.CamelContext;
 import org.apache.camel.ConsumerTemplate;
+import org.apache.camel.FluentProducerTemplate;
+import org.apache.camel.Produce;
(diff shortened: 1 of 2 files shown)


lburgazzoli:
@davsclaus looks like we do not have any `FluentConsumerTemplate`

lburgazzoli:
ok to test

davsclaus:
Thx ah yeah fluent consumer is likely not so useable as its api already is much simpler than producer template etc.

PR #479, opened 2019-11-27
Simplify the Tooling hierarchy

### LINKED ISSUE #476: List itests in an XML file for the Quarkus platform
rpkgtests-maven-plugin 0.4.0 has a mojo to generate a list of itests. It will simplify setting up the test modules on the platform side.

ppalaga:
b247f86 moves the plugin under `build` directory and removes some unneeded maven modules. Benefits: less code faster builds.
96aa619 should be self explaining


lburgazzoli:
both camel and camel-k-runtime have related maven plugins  and other bits under tooling, not a huge issue to move to build but maybe it should have been the opposite, like moving what is on build that should be moved to tooling ?

build/test-list/src/main/resources/test-jars.xml
```diff
@@ -0,0 +1,198 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<testJars>
+    <testJar>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-integration-test-aws</artifactId>
+        <version>0.4.1-SNAPSHOT</version>
```
lburgazzoli: Wouldn't be better to avoid having those version here ?

ppalaga: Good point. Let me figure out a better way how to get the version on the platform side.

lburgazzoli: can't we just use maven builtins to substitute place-holders ?

ppalaga: This XML file will be consumed via a versioned jar on the platform side - so the version can be taken from the version of the jar. The rpkgtest-maven-plugin mojo for generating the test modules just needs to get adapted for that.

ppalaga: Removing the commit from this PR for now. Will submit it separately.

ppalaga:
92745ec simplifies the tooling hierarchy. Let me add another commit to move the build stuff to tooling.

ppalaga:
367fc9f moves the resources from `build` to `tooling`. Please re-review

lburgazzoli:
ok to test

Issue #413, opened 2019-11-13
Provide aggregated tests for quarkus-platform


lburgazzoli:
The quarkus-platform now [includes](https://github.com/quarkusio/quarkus-platform/tree/master/integration-tests) the same tests as we have here but it does not seems very sustainable in term of maintenance and resources required to validate the quarkus-platform.

We should create some aggregated test that validates multiple components in a single test

ppalaga:
I was also thinking of how the platform can maintain our tests in the long term. My idea was to produce some metadata on our side (basically a list of our itest project GAVs) that the platform could leverage to generate/validate the modules in their source tree. I was thinking of adding a mojo to rpkgtests-maven-plugin to implement that.

The rpkgtests:rpkgtests mojo could also use the metadata so that the [testJars](https://github.com/quarkusio/quarkus-platform/blob/master/integration-tests/rpkgtests/pom.xml#L156) do not need to get configured manually.

This would be easier for us (we'd keep working as before, we'd just add a maven module that would generate the metadata). Not sure this would be acceptable for the platform.

ppalaga:
> My idea was to produce some metadata on our side (basically a list of our itest project GAVs) that the platform could leverage to generate/validate the modules in their source tree.

This is implemented now. Updating the tests in the platform is not that a big issue with the new tooling. I also would not say the number of the tests is an issue for them. Their CI runs only in JVM mode. 

Having said that, I do not see an immediate need to improve anything in this area.

PR #482, opened 2019-11-28
Fix #476 List itests in an XML file for the Quarkus platform

### LINKED ISSUE #476: List itests in an XML file for the Quarkus platform

ppalaga:
There is no version in the XML list of tests anymore as proposed by @lburgazzoli.
The XML is now deployed directly instead of being deployed in a jar.


lburgazzoli:
wonder if as future evolution, the xml with the list of the integration-test could be auto generated like all the it with a specific property are included (or excluded)

ppalaga:
bd419b9 fixes the license header problem

ppalaga:
> be auto generated like all the it with a specific property are included (or excluded)

(Not sure you noticed) it is currently autogenerated and there is some file path based filtering there. Is that not enough? https://github.com/apache/camel-quarkus/pull/482/files#diff-e0dff2635520bef5f237973025e0a295R51-R63

lburgazzoli:
oh, damn me. didn't notice

lburgazzoli:
got confused because I saw the generated file also included in the repo, maybe as it is auto generated it should not be checked in

ppalaga:
> auto generated it should not be checked in

I prefer to have it in git to be able to check anytime quickly what we export and also to have the history thereof. WDYT?

Now fixing the JAXB issue on Java 11.

lburgazzoli:
in general I'd go for not including auto generated code as It may be another source of conflicts and confusion as people may try to change them manually but it is not a huge issue

ppalaga:
Yes, potential merge conflicts in the XML file would be nice to avoid. 

The main motivation for me to keep the file in Git for now is that we have some irregularities in the `integration-tests` directory and the selection done by the plugin may get broken in the future. Having a chance to spot malign changes during the PR review could prevent that. - That's actually a sign that we should try to find a mechanism for exporting the tests that would be more robust than the current path matching. I vote for merging this PR as is and continue the discussion withing https://github.com/apache/camel-quarkus/issues/413 . Either the selection could work based on properties in the test POMs (the irregularities in main an core are still there) as you proposed or maybe we could put the aggregated tests into a new directory where we'd keep a strictly flat layout (so that the path matching would suffice).

lburgazzoli:
ok to test

lburgazzoli:
@ppalaga I have no objection to keep it now but long term solution would be better to remove it.

We may think to move support test extension in a dedicated support folder so main and core and future it tests are just like other tests and we only need to skip indexing the support folder

ppalaga:
> move support test extension in a dedicated support folder so main and core and future it tests are just like other tests

What an elegant option!

PR #486, opened 2019-11-28
Fix #484 The validator test should use a classpath URI for the XSD

### LINKED ISSUE #484: The validator test should use a classpath URI for the XSD
... so that the test is executable by the Quarkus Platform.

ppalaga:


oscerd:
ok to test

lburgazzoli:
ok to test

PR #488, opened 2019-11-28
removing sjms-component

wbrycki:
sjms-component can not be loaded as a dependency. Is this ready?


lburgazzoli:
what do you mean ?

ppalaga:
> sjms-component can not be loaded as a dependency. Is this ready?

No, it was not released yet. But will be soon. You may want to compile it from sources or use the Apache Snapshots repo [1], if you want to use it in the mean time. 

You probably see it in the list of extensions on https://camel.apache.org/camel-quarkus/latest/list-of-camel-quarkus-extensions.html 
"latest" in that URL currently means the master branch in this repo. The since version 0.5 there is correct - that's the one to be released.

Anyway, we acknowledge that the above interpretation of "latest" may be seen as misleading and we plan to change it after version 1.0.

[1] http://repository.apache.org/content/groups/snapshots

PR #492, opened 2019-11-29
Scripts to sort pom.xml files

ppalaga:


docs/modules/ROOT/pages/contributor-guide.adoc
```diff
@@ -118,6 +118,7 @@ may consider omitting explicit `nameBase` in some cases.
 +
 * Run `mvn license:format -Plicense` to add license headers to the new files.
 * Run `mvn process-resources` to re-generate the list of extensions and the Camel Quarkus Catalog
+* Run `cd extensions && mvn groovy:execute@sort-poms` to sort elements in various POM files properly
```
lburgazzoli: maybe as follow up we should have a profile that executes all the steps so users are not affected if we add/reove something

lburgazzoli:
ok to test

Issue #490, opened 2019-11-28
Add support for tidymarkup and tagsoup


jsight:
https://camel.apache.org/components/latest/tidyMarkup-dataformat.html

ppalaga:
@jsight are you by any chance planning to work on this?

jsight:
@ppalaga - I wasn't, but maybe this would maybe be a good starting point for me? I will give it a shot.

ppalaga:
You may want to check https://camel.apache.org/camel-quarkus/latest/contributor-guide.html
Feel free to ask here if you have specific questions.

jsight:
PR sent -- I didn't know about the contributor guide when I started, so this was done a bit manually. Hopefully the steps were taken correctly... happy to adjust based on feedback, though.

lburgazzoli:
Done with https://github.com/apache/camel-quarkus/pull/493

PR #493, opened 2019-11-30
Issue 490: Add tagsoup support

jsight:
https://github.com/apache/camel-quarkus/issues/490

### DIFF
--- a/extensions/tagsoup/deployment/pom.xml
+++ b/extensions/tagsoup/deployment/pom.xml
@@ -1,0 +1,81 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one or more
+    contributor license agreements.  See the NOTICE file distributed with
+    this work for additional information regarding copyright ownership.
+    The ASF licenses this file to You under the Apache License, Version 2.0
+    (the "License"); you may not use this file except in compliance with
+    the License.  You may obtain a copy of the License at
+
+         http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-tagsoup-parent</artifactId>
+        <version>0.4.1-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-tagsoup-deployment</artifactId>
+    <name>Camel Quarkus :: TagSoup :: Deployment</name>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.apache.camel.quarkus</groupId>
+                <artifactId>camel-quarkus-bom-deployment</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-core-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-core-xml-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-support-xalan-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-tagsoup</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.quarkus</groupId>
+                            <artifactId>quarkus-extension-processor</artifactId>
+                            <version>${quarkus.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/extensions/tagsoup/pom.xml
+++ b/extensions/tagsoup/pom.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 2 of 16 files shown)


jsight:
I don't really understand the failures here. The parent paths appeared to be correct to me, unless I'm misreading something?

extensions/tagsoup/deployment/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-tagsoup-parent</artifactId>
+        <version>0.3.2-SNAPSHOT</version>
```
lburgazzoli: Should be 0.4.1-SNAPSHOT

extensions/tagsoup/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-build-parent</artifactId>
+        <version>0.3.2-SNAPSHOT</version>
```
lburgazzoli: Should be 0.4.1-SNAPSHOT

extensions/tagsoup/runtime/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-tagsoup-parent</artifactId>
+        <version>0.3.2-SNAPSHOT</version>
```
lburgazzoli: Should be 0.4.1-SNAPSHOT

integration-tests/tagsoup/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-integration-tests</artifactId>
+        <version>0.3.2-SNAPSHOT</version>
```
lburgazzoli: Should be 0.4.1-SNAPSHOT

extensions/tagsoup/deployment/pom.xml
jsight: Oh... thanks :)

jsight:
@lburgazzoli - I think that I've fixed them now.

davsclaus:
ok to test

jsight:
Is this because it needs to be tested in native mode as well?

davsclaus:
Yes when we write that comment about its okay to test, then it does a full test incl native mode too.

extensions/tagsoup/runtime/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+        <artifactId>camel-quarkus-tagsoup-parent</artifactId>
+        <version>0.4.1-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-tagsoup</artifactId>
+    <name>Camel Quarkus :: TagSoup :: Runtime</name>
+
+    <properties>
+        <firstVersion>0.3.0</firstVersion>
```
ppalaga:
```suggestion
        <firstVersion>0.5.0</firstVersion>
```

extensions/tagsoup/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+name: "Camel Quarkus TagSoup"
+description: "Camel TagSoup support"
+metadata:
+  keywords:
+  - "camel"
```
ppalaga:
```suggestion
  - "camel"
  - "tagsoup"
  - "tidymarkup"
```

integration-tests/tagsoup/src/test/resources/test.html
```diff
@@ -0,0 +1,26 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<html>
+<head>
+    <title>Title</title>
+</head>
+<body>
+    <p>Paragraph Contents</p>
```
ppalaga: I'd say this document is too sane to see TagSoup at doing something really useful. Let's break it a bit  - e.g. let's at least remove the closing tag `</p>` to make it an invalid XML doc.

poms/bom/pom.xml
```diff
@@ -48,10 +48,7 @@
                 <scope>import</scope>
             </dependency>
 
-            <!-- Sort using cd extensions && mvn groovy:execute@sort-poms -->
-            <!-- Dependencies a..z -->
```
ppalaga: Please do not remove the above two lines. The second one is important for the `sort-poms` script. 

Also please remove the once removed `<!-- Camel (in alphabetical order) -->` line and re-run `cd extensions && mvn groovy:execute@sort-poms`.

integration-tests/tagsoup/src/test/resources/test.html
jsight: TBH, I wasn't really worried about that. The main thing is that it runs and returns a document. I will add a syntax error, though.

jsight:
I have pushed a fix but this just yields another error. I really do not know how to debug this last one.

poms/bom/pom.xml
jsight: I hope that you rethink this pattern, but I will try.

ppalaga:
> I have pushed a fix but this just yields another error. I really do not know how to debug this last one.

XPath is untested yet, so seeing it to fail in native mode is rather expected. But it is a good news actually, that the flow fails that late, because it means that TagSoup produces a valid DOM. I think the extension in itself works well, but we need to change the way we test it somehow.

I tried to serialize the DOM using a dummy XSLT but it fails due to issues that need to get fixed in the XSLT extension.

Then I tried to use the DOM API directly to return the content of the text node under /html/body/p and it works. I have sent a PR against your branch: https://github.com/jsight/camel-quarkus/pull/1 Please check whether it makes sense.

jsight:
Quarkus is at 1.0 but doesn't support... xpath? That is an interesting finding. Anyway, I have merged your PR. Hopefully that is all that needs to be done here now.

poms/bom/pom.xml
ppalaga: Sorry for the inconvenience, but these POMs will gradually contain hundreds of dependencies so sorting them programmatically to be able to figure out what is in there is a must.

ppalaga:
> all that needs to be done here now

Could you please squash all your (and mine) commits in this PR?

jsight:
Done

ppalaga:
ok to test

lburgazzoli:
ok to test

PR #501, opened 2019-12-03
Filter out beans from CDI from build time discovery

### LINKED ISSUE #499: Filter out beans from CDI from build time discovery
If i.e. a component is defined using CDI, like:

```java
@Singleton
public class CamelComponents {
    @Singleton
    @Named("direct")
    @Produces
    DirectComponent directComponent() {
        DirectComponent component = new DirectComponent();
        component.setTimeout(1234);

        return component;
    }
}
```

we should exclude it from build time auto discovery

lburgazzoli:
Fixes #499

### DIFF
--- a/extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
+++ b/extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
@@ -26,38 +26,13 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import io.quarkus.arc.processor.BeanRegistrar;
-import org.apache.camel.CamelContext;
-import org.apache.camel.RoutesBuilder;
-import org.apache.camel.builder.AdviceWithRouteBuilder;
-import org.apache.camel.builder.RouteBuilder;
-import org.apache.camel.impl.converter.BaseTypeConverterRegistry;
-import org.apache.camel.quarkus.core.CamelConfig;
-import org.apache.camel.quarkus.core.CamelMain;
-import org.apache.camel.quarkus.core.CamelMainProducers;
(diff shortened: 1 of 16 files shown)


lburgazzoli:
ok to test

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -153,12 +151,15 @@ CamelRegistryBuildItem registry(
...
+                BeanRegistrationPhaseBuildItem beanRegistrationPhase,
                 List<CamelBeanBuildItem> registryItems,
                 List<CamelServiceFilterBuildItem> serviceFilters) {
 
-            RuntimeValue<Registry> registry = recorder.createRegistry();
+            final RuntimeValue<Registry> registry = recorder.createRegistry();
+            final List<BeanInfo> beans = beanRegistrationPhase.getContext().get(BuildExtension.Key.BEANS);
 
             CamelSupport.services(applicationArchives)
+                    .filter(si -> !CamelSupport.isContainerBean(beans, si))
```
ppalaga: Here and on other places where CamelSupport.isContainerBean is invoked, I have concerns about the algo complexity. Now it is `m*n` (container beans x camel beans) in the worst case. To improve it, we could perhaps transform the container beans into a set of `CamelBeanInfo`s and store the set inside a BuildItem so that we can re-use it on all three call sites without needing to iterate over the list of container beans for every camel bean. WDYT?

lburgazzoli: Yes, not super efficient. We can improve it in a follow up PR

oscerd:
ok to test

oscerd:
ok to test

PR #503, opened 2019-12-03
Improve registration of DozerTypeConverter

jamesnetherton:


davsclaus:
ok to test

docs/modules/ROOT/pages/extensions/dozer.adoc
```diff
@@ -29,3 +29,10 @@ Use the following configuration property to specify the location(s) of your Doze
 ----
 quarkus.camel.dozer.mapping-files = dozerMappings.xml,file:foo/bar/custom-mappings.xml,http://foo/bar/mapping.xml
 ----
+
+The Dozer extension provides a type converter which is disabled by default. It can be enabled with the following configuration.
```
ppalaga: I think it would be helpful for the readers to give them some sort of an explanation why they might want to enable the converter.

jamesnetherton: I'll add some additional detail when the build completes.

PR #506, opened 2019-12-04
feat: base64 extension

davsclaus:


davsclaus:
ok to test

integration-tests/base64/src/test/java/org/apache/camel/quarkus/component/base64/it/Base64Test.java
```diff
@@ -0,0 +1,43 @@
...
+
+@QuarkusTest
+class Base64Test {
+
+    @Test
+    public void test() {
+        byte[] body;
+
+        ExtractableResponse response = RestAssured.given() //
+                .contentType(ContentType.BINARY).body("Hello World".getBytes()).post("/base64/post") //
```
ppalaga: String.getBytes() relies on the default encoding of the given platform and therefore this test will fail on exotic platforms that have some ASCII incompatible default encoding set. 
```suggestion
                .contentType(ContentType.BINARY).body("Hello World".getBytes("UTF-8")).post("/base64/post") //
```

PR #509, opened 2019-12-04
Do not use hard coded ports in integration-tests

### LINKED ISSUE #508: Do not use hard coded ports in integration-tests
some tests have hard-coded ports in routes definitions, test resources and so on which should be avoided to prevent CI failures

lburgazzoli:
Fixes #508


lburgazzoli:
ok to test

integration-tests/support/test-support/src/main/java/org/apache/camel/quarkus/test/AvailablePortFinder.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Gets the next available port.
+     *
+     * @throws IllegalStateException if there are no ports available
+     * @return the available port
+     */
+    public static int getNextAvailable() {
+        try (ServerSocket ss = new ServerSocket()) {
+            ss.setReuseAddress(true);
+            ss.bind(new InetSocketAddress((InetAddress) null, 0), 1);
```
ppalaga: Is this a solution proven to work in practice? Asking because I assume (I have not looked into the impl) that ServerSocket.bind() does not do any fancy locking, and therefore two racing threads or processes may get the same port number as available. But that's perhaps rather unlikely for real world scenarios and there is enough evidence that this works in practice?

davsclaus:
ok to test

integration-tests/support/test-support/src/main/java/org/apache/camel/quarkus/test/AvailablePortFinder.java
lburgazzoli: This is a copy and paste of an utility class from camel that is in place since quite a few years (but you can find the same implementation on other projects such as mina, spring and what not). 

Under the hood binding is done via syscall and in the kernel will assign a pseudo random free port so yes there is the theoretical possibility that two process get the same port but that would happen only in case you have a **lot** of ephemeral tcp connections.

A better solution would be to use unix socket but that's not very common for the majority of the components we have.

ppalaga: Thanks for the explanation!

PR #517, opened 2019-12-05
introduce testcontainers

### LINKED ISSUE #516: [question] introduce testcontainers
I'm experimenting with testcontainers to replace the embedded kafka server we have now for running camel-quarkus-kafka integration tests and it simplifies a lot the set-up. 

The downside is that some tests depends on docker.

Any objection to introduce testcontainers ?

lburgazzoli:
Fixes #516

### DIFF
--- a/integration-tests/kafka/pom.xml
+++ b/integration-tests/kafka/pom.xml
@@ -57,43 +57,18 @@
 
         <!-- test dependencies -->
         <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-junit5</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>io.rest-assured</groupId>
             <artifactId>rest-assured</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.assertj</groupId>
-            <artifactId>assertj-core</artifactId>
-            <scope>test</scope>
-        </dependency>
-
-        <!-- test dependencies - camel-quarkus -->
-        <dependency>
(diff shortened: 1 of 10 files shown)


oscerd:
ok to test

integration-tests/support/testcontainers-support/src/main/java/org/apache/camel/quarkus/testcontainers/ContainerUtil.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class ContainerUtil {
+    /**
+     * Wait for the container to be in running state
+     *
+     * @param container the container to wait for
+     */
+    public static void waitForInitialization(GenericContainer container) {
+        for (int retries = 0; retries < 5; retries++) {
+            if (container.isRunning()) {
```
bsideup: FYI `container.start()` will throw an error if the container is not running, so I assume this check always pass and this helper isn't needed. Let us know otherwise, there must be a bug then ;)

lburgazzoli: will have a look after we get this merged, thx for the support

oscerd: > FYI `container.start()` will throw an error if the container is not running, so I assume this check always pass and this helper isn't needed. Let us know otherwise, there must be a bug then ;)

Really nice to have you here @bsideup , thanks for helping

lburgazzoli:
ok to test

PR #521, opened 2019-12-06
[WIP] Fix #74 Telegram extension

### LINKED ISSUE #74: Create a Camel Telegram extension

ppalaga:
As agreed earlier, I switched the rest client impl from CXF to MicroProfile. The adjusted telegram component is part of the current PR for now and I plan to move it partly or completely to Camel depending on the discussion in https://issues.apache.org/jira/browse/CAMEL-13903


oscerd:
If we'll make it pluggable it would be great to backport on the component, otherwise it could stay only here

lburgazzoli:
I see some major issues with this pr:

1. we should not have duplicated components as it is a source of confusion and it increases the maintenances cost
2. this PR brings RESTEasy and MP Rest Client and it would be better to just use RESTEasy as we had a long discussion about bringing full RESTEasy client support in Quarkus
3. I'd would have done in the opposite way like making the camel-telegram pluggable and only after that creating the extension here as it make much easier to understand and review the changes that are required

ppalaga:
> 1. we should not have duplicated components as it is a source of confusion and it increases the maintenances cost

I agree that not having duplicated code is a good thing and I agree that as much code as possible from this PR needs to get ported to Camel. Whether all or just part and which parts of the component code should go to Camel is an open question for me and I'd like to reach some agreement before I spend time with the porting.

At the same time, as many of you may know, having a Telegram extension is handy when presenting. So I vote for temporarily hosting the duplicated component here.

> 2. this PR brings RESTEasy and MP Rest Client and it would be better to just use RESTEasy as we had a long discussion about bringing full RESTEasy client support in Quarkus

No problem, let me try that.

> 3. I'd would have done in the opposite way like making the camel-telegram pluggable and only after that creating the extension here as it make much easier to understand and review the changes that are required

That would be more work for me, esp. when requests like 2. appear during the review. So I chose not to go that far in the first iteration.

### Pluggability: yes or no?

Yes, the component may be made pluggable to allow for running with both CXF and RestEasy REST client. But I'd like us all to think once again whether it is a good idea to support two equivalent rest clients. That's more code, more bugs and more maintenance than having just one client. Do we really want to do that and can't we come up with any solution without the pluggability?

I can see these options:

#### A. Replace the CXF client with RestEasy: 

Reasons: The rest client is just an implementation detail. End users should not notice the change. The CXF client does not work OotB on Quarkus and it is safer for us to use libs that are known to work there.

There is also @oscerd 's statement in [CAMEL-13903](https://issues.apache.org/jira/browse/CAMEL-13903?focusedCommentId=16914300&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16914300)

#### B. Replace the CXF client with bare Apache HTTP client and Jackson.

That's basically, what both the CXF client and RestEasy client are using under the hood. There is just a handful of endpoints to cover. It would be a bit more code in the component, but less dependencies.

WDYT?

ppalaga:
> > 2. this PR brings RESTEasy and MP Rest Client and it would be better to just use RESTEasy as we had a long discussion about bringing full RESTEasy client support in Quarkus
> 
> No problem, let me try that.

I tried that and my conclusion is that it is not worth of doing it. Here is why: Quarkus MP Rest Client (MPRC) extension provides support for MP Rest Client on top of the RestEasy REST client. To ditch the MP Rest Client spec from the game would actually result in repeating most of the things done in MPRC extension. I do not think we want to do that. The MP Rest Client brings minimal overhead and porting RestEasy REST client separately is too much work that would not pay back.

Maybe I should explain why the current proposal depends both on the MP REST client and some RestEasy bits: the RestEasy bits are there for the multipart stuff, that the MP REST client API does not handle in any way.

Now I'd try to implement Telegram without CXF or RestEasy, taking some HTTP client and Jackson. Does anybody want to veto it (and name reasons) before I start?

lburgazzoli:
Maybe better to continue the discussion on the camel jira and/or dev mailing list and remember the component should continue to work on spring, karaf and eap.

jamesnetherton:
> the component should continue to work on spring, karaf and eap

Right, and maybe that makes the case for using plain old HTTP Client etc. I kinda agree with Peter, what are we really gaining from supporting multiple JAX-RS implementations? Aside from this project, who else is going to make use of it?

lburgazzoli:
This could be a good motivation, the point is that we should be very carefully in bringing dependencies that are tailored to a specific platform as it may be a fast path at implementation time but could be a pain later on (we do have some component based of spring bits that do not make it easy for example port them to camel-quakkus)

oscerd:
The problem is just: are we doing this to support telegram on Quarkus? Or are we doing this because there is an improvement in the component? If the answer is the first, then we are doing things wrong. The original issue was opened as "this is not working on Quarkus, lets change from CXF to Resteasy" and, frankly, this is not a good reason for changing something that works. Also, it should be the opposite, camel-quarkus should make it work on Quarkus and the component in Camel should remain exactly the same. I think it's just a matter of priorities. Quarkus is just one of the platform we cover, it's not the only platform we need to cover.

ppalaga:
I have sent a PR to Camel https://github.com/apache/camel/pull/3414. I will update this PR once that one gets merged.

ppalaga:
The telegram commit is now available in https://github.com/apache/camel-quarkus/pull/698

Issue #457, opened 2019-11-21
Provide test facilities


lburgazzoli:
In camel, we have camel-test-junit5 to help writing tests, we should provide similar facilities to test quarkus based integration

ppalaga:
Could you please be more specific about the anticipated functionality?

PR #524, opened 2019-12-06
refactor FHIR integration tests

lburgazzoli:
While experimenting a little bit about other topics (https://github.com/apache/camel-quarkus/issues/457, https://github.com/apache/camel-quarkus/issues/516), I've used the FHIR extension as target for my experiments.

I've started about providing some junit5 extensions so that tests can be easily excluded using the same `flags` we use at build time to enable/disable some build steps, like:

```java
    @Test
    @EnabledIf(FhirFlags.Dstu2Enabled.class)
    public void jsonDstu2() {
        ....
    }
```

The second thing is that I did work a little bit to enhance the [api-fhir-jpaserver-starter](https://github.com/hapifhir/hapi-fhir-jpaserver-starter) by adding an option to override properties from environment variables https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/67 and to create a [docker image](https://quay.io/repository/lburgazzoli/hapi-fhir-jpaserver-starter?tab=tags) we can use to simplify the test set-up of the quarkus extension by leverage `tescontainers` (the same image can be used also to the the camel component in the future).

The downside of this experiment is that the image used for tests is hosted in my own repository but I've raised an [issue](https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/68) to publish images every time a new release of the hapi happen.

@johnpoth what do you think ? does it make sense to merge it on master ?


ppalaga:
I did not have time to go through the actual changes but I like all the goals you mentioned in the comment.

johnpoth:
@lburgazzoli the `EnabledIfCondition` rocks ! Much needed in the FHIR integration tests and I think it will be useful for other complex extensions.

I like the idea of using `testcontainers` as it is a lot simpler to start/stop the hapi server for surefire tests. It also uses an annotation which simplifies the pom and makes the test class more self contained. The small price to pay is that you need to start the HAPI server for all tests. So +1

I also like the way you're configuring the component through the bean rather than the long URI I was using :)

besides CAMEL-14268 this is a big + 1 , Thanks !

lburgazzoli:
disabled the configuration of fhir component through beans as it does not work (see CAMEL-14268)

lburgazzoli:
ok to test

lburgazzoli:
@johnpoth are you o to merge this even if the testocontainer images is not yet the official one ?

johnpoth:
@lburgazzoli fine by me

lburgazzoli:
let see how it goes :)

PR #528, opened 2019-12-08
Re-introduce RoutesBuilderBuildItem

### LINKED ISSUE #526: Re-introduce RoutesBuilderBuildItem
RoutesBuilderBuildItem were used to set routes builder to camel-main at build time and then we changed it to put those discovered routes to the camel registry.

It appear that it was a wrong decision as routes from the registry are not subject to Camel's DI which seems be correct as those routes are coming from i.e. a container such as ArC thus they should rely on container's DI. 

Of course mixed DI i.e. CDI + CamelDI won't work.

@davsclaus what should be the right way of doing this ? One option could be that we do always apply Camel's DI to any route and let the container do it's own job.

As workaround I've implemented https://github.com/apache/camel-quarkus/pull/525 and reported the issue to camel [CAMEL-14271](https://issues.apache.org/jira/browse/CAMEL-14271)

lburgazzoli:
Fixes #526


lburgazzoli:
ok to test

PR #533, opened 2019-12-10
Add SQL component extension

### LINKED ISSUE #505: camel-sql extension
its using spring-jdbc but that may work with the spring data compatibility in quarkus?

jamesnetherton:
fixes #505

### DIFF
--- a/docs/modules/ROOT/pages/extensions/sql.adoc
+++ b/docs/modules/ROOT/pages/extensions/sql.adoc
@@ -1,0 +1,29 @@
+[[sql]]
+= SQL Extension
+
+*Since Camel Quarkus 1.1.0*
+
+The SQL component allows you to work with databases using JDBC queries.
+
+Maven users will need to add the following dependency to their `pom.xml` for this extension.
+
+[source,xml]
+------------------------------------------------------------
+<dependency>
+    <groupId>org.apache.camel.quarkus</groupId>
+    <artifactId>camel-quarkus-sql</artifactId>
+</dependency>
+------------------------------------------------------------
+
+== Usage
+
+The extension provides support for the Camel https://camel.apache.org/components/latest/sql-component.html[SQL Component].
+
+=== Configuration
+
+When configuring `sql` or `sql-stored` endpoints to reference script files from the classpath, set the following configuration property to ensure that they are available in native mode.
+
+[source,properties]
+----
+quarkus.camel.sql.script-files = queries.sql, classpath:sql/insert.sql
+----
--- a/extensions/sql/deployment/src/main/java/org/apache/camel/quarkus/component/sql/deployment/SqlProcessor.java
+++ b/extensions/sql/deployment/src/main/java/org/apache/camel/quarkus/component/sql/deployment/SqlProcessor.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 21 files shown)


extensions/sql/runtime/src/main/java/org/apache/camel/quarkus/component/sql/CamelSqlConfig.java
```diff
@@ -0,0 +1,33 @@
...
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+import java.util.List;
+
+@ConfigRoot(name = "camel.sql", phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
+public class CamelSqlConfig {
+
+    /**
+     * A comma separated list of paths to script files referenced by SQL endpoints.
```
ppalaga: This awakes the impression that the paths are meant to be local FS paths, which I think is not true. Also an example and a note that non-classpath resources do not need to get listed here would be nice.

docs/modules/ROOT/pages/extensions/sql.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+The extension provides support for the Camel https://camel.apache.org/components/latest/sql-component.html[SQL Component].
+
+=== Configuration
+
+When configuring `sql` or `sql-stored` endpoints to reference script files from the classpath, set the following configuration property to ensure that they are available in native mode.
+
+[source,properties]
+----
+quarkus.camel.sql.script-files = queries.sql, classpath:sql/insert.sql
```
ppalaga: If other URI schemes are supported, it should be noted here that they are supported and that they do not need to get listed in `quarkus.camel.sql.script-files`.

docs/modules/ROOT/pages/extensions/sql.adoc
```diff
@@ -0,0 +1,29 @@
+[[sql]]
+= SQL Extension
+
+*Since Camel Quarkus 1.1.0*
+
+The SQL component allows you to work with databases using JDBC queries.
```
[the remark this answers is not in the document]
jamesnetherton: It was a copy / paste from the camel [component docs](https://camel.apache.org/components/latest/sql-component.html).

I can change it in both places if we really want. I'm pretty relaxed about the wording.

extensions/sql/deployment/src/main/java/org/apache/camel/quarkus/component/sql/deployment/SqlProcessor.java
```diff
@@ -0,0 +1,68 @@
...
+    void sqlNativeImageResources(BuildProducer<NativeImageResourceBuildItem> nativeImage, CamelSqlConfig config) {
+        config.scriptFiles
+                .stream()
+                .map(scriptFile -> new NativeImageResourceBuildItem(scriptFile.replace("classpath:", "")))
+                .forEach(nativeImage::produce);
+    }
+
+    @BuildStep(onlyIf = NativeBuild.class)
+    void generateKParameterClass(BuildProducer<GeneratedClassBuildItem> generatedClass) {
+        // TODO: The native image build fails with a NoClassDefFoundError without this. Possibly similar to https://github.com/oracle/graal/issues/656.
```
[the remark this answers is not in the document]
jamesnetherton: It's a pretty harmless workaround IMO. I never got to the bottom of what was causing the issue so my TODO was just a marker for "we may want to investigate further and remove this in future".

I'll raise an issue as a reminder to follow up on it.

extensions/sql/runtime/src/main/java/org/apache/camel/quarkus/component/sql/CamelSqlConfig.java
jamesnetherton: I'll follow up with another PR to clarify.

docs/modules/ROOT/pages/extensions/sql.adoc
jamesnetherton: I'll follow up with another PR to clarify.

PR #535, opened 2019-12-10
chore: Clarify supported URI schemes in SQL script-files config property

jamesnetherton:
@ppalaga Follow up on your review comments in #533.


ppalaga:
This is just comments. Feel free to merge without CI.

PR #545, opened 2019-12-13
Fix commons-logging setup

lburgazzoli:


lburgazzoli:
> +1 for the cleanup.
> 
> So the rule you are hereby coining is that we should exclude in the BOM rather than in the runtime modules, right?
> 

I would say yes but no strong opinion on it, what's your ?

> We could actually ban commons-logging via an enforcer rule, WDYT?

Yes we should

lburgazzoli:
ok to test

ppalaga:
> > So the rule you are hereby coining is that we should exclude in the BOM rather than in the runtime modules, right?
> 
> I would say yes but no strong opinion on it, what's your ?

I do not mind such a rule, neither I have a strong opinion about where it should be done.

lburgazzoli:
> 
> I do not mind such a rule, neither I have a strong opinion about where it should be done.

So let's see how it goes

PR #548, opened 2019-12-13
Ensure custom services such as deataformats, languages and component are not removed from container

lburgazzoli:


lburgazzoli:
ok to test

PR #554, opened 2019-12-18
Modular RoutesCollector

lburgazzoli:
The aim here is to make the RoutesCollector more flexible so platform
such as camel-k can fine tune the behavior of the collector i.e. by
providing am ad-hoc way of discovery certain routes from the registry


lburgazzoli:
ok to test

lburgazzoli:
Failure is unrelated

PR #556, opened 2019-12-19
Restored the native profile for camel-quarkus-pdf integration tests

aldettinger:
This test is running on my machine, so the first aim of this PR is to check whether the issue is present on CI.


lburgazzoli:
ok to test

PR #569, opened 2019-12-21
Fix #568 Untrack Artemis test instance data

### LINKED ISSUE #568: Untrack Artemis test instance data
https://github.com/quarkusio/quarkus-platform/pull/19#issuecomment-568125279

ppalaga:


integration-tests/sjms/src/test/java/org/apache/camel/quarkus/component/sjms/it/CamelSjmsTestResource.java
```diff
@@ -41,14 +42,16 @@ public void inject(Object testInstance) {
     @Override
     public Map<String, String> start() {
         try {
-            FileUtils.deleteDirectory(Paths.get("./target/artemis").toFile());
+            final File dataDirectory = Paths.get("./target/artemis").toFile();
+            FileUtils.deleteDirectory(dataDirectory);
```
lburgazzoli: shouldn’t this be done also when the test ends?

ppalaga: A reason for keeping it after the test could be that sometimes it is handy to be able to inspect the state of a directory like this after the test.

Any reasons for removing it after the test?

PR #570, opened 2019-12-21
Fix #560 AHC extension

### LINKED ISSUE #560: AHC extension
Plz. assign to me.

ppalaga:


extensions/ahc/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+name: "Camel Quarkus HTTP Common"
+description: "Camel Quarkus HTTP Common"
```
lburgazzoli: looks like the name and description are wrong

extensions/pom.xml
```diff
@@ -100,6 +100,7 @@
         <module>vm</module>
         <module>xslt</module>
         <module>zipfile</module>
+        <module>ahc</module>
```
lburgazzoli: nor ordered

pom.xml
```diff
@@ -40,9 +40,11 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
+        <ahc.version>2.10.4</ahc.version>
```
lburgazzoli: do we need this version or can we inherit from camel-ahc ?

poms/bom-deployment/pom.xml
```diff
@@ -35,7 +35,6 @@
     <properties>
         <camel-quarkus.version>1.1.0-SNAPSHOT</camel-quarkus.version><!-- kept in sync with project.version by the release plugin -->
 
-        <jandex-maven-plugin.version>1.0.6</jandex-maven-plugin.version>
```
lburgazzoli: this jandex thing seems unrelated to the pr, can we make this change in a separate commit ?

extensions/ahc/runtime/src/main/resources/META-INF/quarkus-extension.yaml
ppalaga: +1

extensions/pom.xml
ppalaga: +1

pom.xml
ppalaga: Actually not for this PR directly, but it will be needed for the Telegram extension that needs AHC but has nothing to do with camel-ahc.

poms/bom-deployment/pom.xml
ppalaga: Uneeded indeed. Let me remove it.

pom.xml
lburgazzoli: yeah but this add some maintenance cost as we need to be sure this is in sync with the version used by camel

ppalaga:
Luca's requests addressed in 58ca689

pom.xml
ppalaga: Yes, that could be mitigated by 

(a) Importing the Camel BoM, or 
(b) Enabling the Dependency Convergence Rule on the Maven Enforcer plugin or
(c) A groovy script :)

davsclaus:
ok to test

pom.xml
lburgazzoli: I think we really need to consider option 1, eventually we may need to clean up a bit the camel's bom but I think that as long term we'll end up supporting most of the component camel supports and the chance to hit this issue again is in my opinion high

lburgazzoli:
ok to test

ppalaga:
88998b2 rebased and fixed conflicts

PR #580, opened 2020-01-02
Dozer extension should use JaxbFileRootBuildItem

### LINKED ISSUE #539: Dozer extension should use JaxbFileRootBuildItem
We should leverage the JAXB extension and `JaxbFileRootBuildItem` so that the Dozer `jaxb.index`  can be automatically read and the classes defined in it registered for reflection.

https://github.com/quarkusio/quarkus/issues/6070 currently blocks this.

jamesnetherton:
fixes #539


jamesnetherton:
ok to test

lburgazzoli:
ok to test

Issue #543, opened 2019-12-13
Unmarshal json in spring xml


jstakun:
I'm trying to migrate spring xml route to quarkus camel. Everything works except for json unmarshaling.

In original version I've configured data format in camel context:

```
<camelContext id="camel" xmlns="http://camel.apache.org/schema/spring">
        <dataFormats>
            <json id="transform-game" library="Jackson" unmarshalTypeName="com.openshift.wildwest.models.Game"/>
            <json id="transform-platform-object" library="Jackson" unmarshalTypeName="com.openshift.wildwest.models.PlatformObject"/>
        </dataFormats>
```
...
and the I'm calling it in the route:

 `<unmarshal ref="transform-game"/>`

In quarkus version I don't know how to declare dataFormats beacuse in the example camel context is declared in application.properties and not in the xml hence I'm declaring it in the route:

```
<unmarshal>
       <json library="Jackson" unmarshalTypeName="com.openshift.wildwest.models.Game" id="transform-game"/>
</unmarshal>
```
However this doesn't seem to unmarshal json body to pojo. 
Can you advice proper syntax?

lburgazzoli:
We don't support defining routes through spring xml at least at this stage so I'd recommend to use "routes fragments" (see the example https://github.com/apache/camel-quarkus/tree/master/examples/timer-log-xml).

jstakun:
Yes. That's exactly what I did. Unfortunetaly I can't create data formats in route fragment and hence I used

```
<unmarshal>
        <json library="Jackson" unmarshalTypeName="com.openshift.wildwest.models.Game" id="transform-game"/>
</unmarshal>
```
But this doesn't work correctly. Is that the right way to unmarshal json to pojo in spring xml, or should I try something else?

lburgazzoli:
You can create data formats programmatically using java i.e. CDI/Spring beans but what you wrote should work, what is the error ?

jstakun:
Object is created but all fields are not populated with data from json.
If this is correct I'll try to debug it further.

lburgazzoli:
does this happen in JVM or native mode ?

jstakun:
This happens in JVM

Here is the route:

```
<route id="start-game">
           <from uri="direct:start-game"/>
           <log id="init-game-log" message="Starting new game"/>
           <to id="_toD1" uri="netty-http:http://{{backend.service}}/createGame"/>
           <log id="_log22" message="Received ${body}"/>
           <unmarshal>
[...]

PR #581, opened 2020-01-02
Add tests for #543

lburgazzoli:


lburgazzoli:
ok to test

lburgazzoli:
We can but then we need to include the routes in the native image which is not yet done automatically, in general we should not assume that all the test have to run on the quarkus-platform

ppalaga:
> We can but then we need to include the routes in the native image which is not yet done automatically

I'd vote for including the XML file in the native image via application.properties' for now:

```
quarkus.native.additional-build-args = -H:IncludeResources=my-routes.xml
```

Not sure this will work OOTB on the platform, but it is a bit more flexible than `-H:ResourceConfigurationFiles` recommended in the Quarkus docs. 

> in general we should not assume that all the test have to run on the quarkus-platform

I think we should assume that all our tests should run there unless we define a clear filtering strategy.

lburgazzoli:
Peter, we already agreed that we need to create some consolidated test for the quarkus platform. T

lburgazzoli:
xml is now included in the native image

lburgazzoli:
ok to test

PR #583, opened 2020-01-02
Add basic tests for dev mode

lburgazzoli:


extensions/core-xml/deployment/src/test/java/org/apache/camel/quarkus/component/xml/deployment/CamelSupportServlet.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import org.apache.camel.CamelContext;
+
+@Path("/test")
+@ApplicationScoped
+public class CamelSupportServlet {
```
ppalaga: This is not a servlet. I'd vote for `s/Servlet/Resource/` on the next occasion. There are more classed like this in our code base.

extensions/core-xml/deployment/src/test/resources/routes.1
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ppalaga: Why not with an `.xml` suffix? - That would be more practical for editing and also for the license plugin.

PR #588, opened 2020-01-03
Some cleanup and improvements

lburgazzoli:


lburgazzoli:
ok to test

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelBeanBuildItem.java
```diff
@@ -18,6 +18,8 @@
 
 import java.util.Objects;
 
+import javax.annotation.Nullable;
```
ppalaga: I never used jsr305 and my naive expectation is that findbugs/spotbugs is requited to enforce it. Or it works differently? 

jsr305 artifact currently being a transitive dep of guava suggests to be cautious. Can't this potentially become a maintenance issue?

tooling/eclipse-formatter-config.xml
```diff
@@ -267,5 +267,6 @@
         <setting id="org.eclipse.jdt.core.formatter.enabling_tag" value="CHECKSTYLE:ON"/>
         <setting id="org.eclipse.jdt.core.formatter.comment.format_block_comments" value="false"/>
         <setting id="org.eclipse.jdt.core.formatter.comment.format_line_comments" value="false"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.format_javadoc_comments" value="false"/>
```
ppalaga: What problem is this solving?

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelBeanBuildItem.java
lburgazzoli: it is primary for documentation purpose here

tooling/eclipse-formatter-config.xml
lburgazzoli: The formatter plugin does not keep aliment with parameter which hurts readability

ppalaga: Could you please try if the following combination does what you want?

```
<setting id="org.eclipse.jdt.core.formatter.comment.format_javadoc_comments" value="true"/>
<setting id="org.eclipse.jdt.core.formatter.comment.align_tags_names_descriptions" value="true"/>
<setting id="org.eclipse.jdt.core.formatter.comment.indent_root_tags" value="false"/>
<setting id="org.eclipse.jdt.core.formatter.comment.align_tags_descriptions_grouped" value="false"/>
```

lburgazzoli: would you mind sending a follow up pr ? I don't want to waste another build cycle and I'm getting a little tired of this auto formatter thing

PR #618, opened 2020-01-14
Fix #617 Registerable and discoverable Camel services

### LINKED ISSUE #617: Registerable and discoverable Camel services
I stumbled upon this when I was rebasing my telegram branch on top of the current Camel master. It did not work because the endpoint and component configurers were not found. Those are currently looked up at runtime by inspecting the files in `META-INF/services/org/apache/camel/configurer` and we do not add them to the native image. 

Clearly, the problem could be solved by adding `META-INF/services/org/apache/camel/configurer` & co. to the native image, but that kind of solution would not utilize the full potential of Quarkus build time augmentation.

I am currently having a PoC that provides a camel-quarkus specific implementation of FactoryFinderResolver. It resolves the services based on a single map that is filled at build time.

I wonder whether anybody can think of a better strategy here?

I'll share my code when I get all tests passing.

ppalaga:


extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -128,9 +130,40 @@ void coreServiceFilter(BuildProducer<CamelServiceFilterBuildItem> filterBuildIte
...
+        void coreServices(
+                ApplicationArchivesBuildItem archives,
+                BuildProducer<CamelServiceBuildItem> services) {
+
+            /* Registered only */
+            CamelSupport.propertyFiles(
+                    archives,
+                    "META-INF/services/org/apache/camel/component",
+                    "META-INF/services/org/apache/camel/language",
+                    "META-INF/services/org/apache/camel/dataformat")
```
lburgazzoli: we need to fine tune language and data-format as we can store into the registry only singleton services (this requires also some changes to the FastCamelContext as it assumes languages and dataformats are stored in the registry)

lburgazzoli: Yes we do have this problem as today but this is a good change to fix it

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -128,9 +130,40 @@ void coreServiceFilter(BuildProducer<CamelServiceFilterBuildItem> filterBuildIte
...
+                    .filter(entry -> entry.getValue().getProperty("class") != null)
+                    .map(entry -> CamelServiceBuildItem.registeredOnly(entry.getKey(), entry.getValue().getProperty("class")))
+                    .forEach(services::produce);
+
+            /* Configurers need some name transformation */
+            CamelSupport.propertyFiles(
+                    archives,
+                    "META-INF/services/org/apache/camel/configurer")
+                    .filter(entry -> entry.getValue().getProperty("class") != null)
+                    .map(entry -> CamelServiceBuildItem.registeredOnly(
```
lburgazzoli: As future evolution, may be better not to expose such instances to the user through the registry, maybe better to try to bind them to the components at build time

ppalaga:
> It would be nice to have some build time properties to control factory files to be included/excluded like:
> 
> ```
> quarkus.camel.factories.discovery.include = **/*
> quarkus.camel.factories.discovery.exclude = org/apache/camel/something/*,org/apache/camel/xyz/**/*
> ```

I wonder how the set of services defined by the extensions should be composed with these user facing selectors?

A. (extension-includes + user-includes) - user-excludes

or 

B. extension-includes + (user-includes - user-excludes)

A. seems to make more sense. WDYT?

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
ppalaga: > we need to fine tune language and data-format 

Not sure how should we tune?

> as we can store into the registry only singleton services 

You seem to imply language and data-format are not singleton services, but could you plz. define "singleton services"?

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
ppalaga: I wonder how can we cover the case you mentioned on the chat - manually bind a second instance of a component under a different name?

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
lburgazzoli: The "tuning" here is about to carefully select which language/dataformat can be bound to the registry.

As example camel creates components only once so every time you use a schema, then you'll end up using the same component instance. 

For languages and dataformat that's not always true and this is because you can use the same dataformat multiple time in the same route with a different configuration.

lburgazzoli:
> A. (extension-includes + user-includes) - user-excludes

Yes something like that should work

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
lburgazzoli: I mean that we should not even bind them to the registry but scan the registry for components and bind related configurers through a recorder during STATIC_INIT

ppalaga: > the registry for components and bind related configurers through a recorder during STATIC_INIT

Yes, that would be easy to do. But when a user adds a second instance of a component under a different name at runtime, our configurer setting code would not catch it.

lburgazzoli: yep but as the second instance is probably configured at runtime, the component will fallback to the "factory finder" way which is reasonable

ppalaga:
d82e823 implements the requested build time config.

The includes/excludes need to be defined incl. the `META-INF/services/...` prefix. I hope that's not an issue.

I vote for creating a new issue for the fine-tuning of the languages and data-formats.

I'd also create a new issue for replacing the current usage of CamelServiceFilterBuildItem with the new CamelServicePatternBuildItem because those ServiceFilters are just excludes.

ppalaga:
4720897 rebased

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/util/PathFilter.java
```diff
@@ -0,0 +1,152 @@
...
+
+        public Builder combine(Builder other) {
+            includePatterns.addAll(other.includePatterns);
+            excludePatterns.addAll(other.excludePatterns);
+            return this;
+        }
+
+        public PathFilter build() {
+            final List<String> incl = includePatterns;
+            includePatterns = null; // avoid leaking the collection trough reuse of the builder
```
lburgazzoli: not a huge issue in our case but my understanding is that invoking `build()` multiple time on the same `builder` should produce the same result

lburgazzoli:
> I vote for creating a new issue for the fine-tuning of the languages and data-formats.

if you have time, it would be nice to at least remove them from the registry and use the default camel discovery for languages and dataformats

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/util/PathFilter.java
ppalaga: > invoking `build()` multiple time on the same `builder` should produce the same result

If the collection is not set to null here the following sequence would be possible:

```
PathFilter.Builder b = new PathFilter.Builder();
b.include("foo/bar");
PathFilter pf1 = b.build();
b.include("bar/baz");
PathFilter pf2 = b.build();
```

where pf1 and pf2 would have set the same list instance in their includePatterns fields. It find that really unwanted. That's a true reference leak that harms the immutability of PathFilter.

I hold re-using builders for rather unusual and so I find breaking the re-use better than copying the list in the PathFilter constructor as a way to ensure immutability.

lburgazzoli: At least document it as I think no-one expect the builder to be re-set once you invoke `build()`

ppalaga:
> remove them from the registry and use the default camel discovery for languages and dataformats

Done in 73f59ef

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/util/PathFilter.java
ppalaga: Done at https://github.com/apache/camel-quarkus/pull/618/files#diff-7d0dfbc8ca22f0eb14e77322fcda0902R143

valdar: Well another, in my opinion, cleaner option is to copy the collection in each object built making it immutable. A variation of this is used in most of the framework that generates builders, like Lombok and similars one.

My 2c.

lburgazzoli:
ok to test

lburgazzoli:
conflicts

ppalaga:
7e11340 rebased

ppalaga:
ok to test

oscerd:
ok to test

lburgazzoli:
The build seems to be running

    https://builds.apache.org/view/C/view/Apache%20Camel/job/camel-quarkus-pr/538/


even if it does not show up here

ppalaga:
This should be fixed by the `Do not register languages and dataformats ...` commit of https://github.com/apache/camel-quarkus/pull/618

ppalaga:
ok to test

lburgazzoli:
ok to test

PR #632, opened 2020-01-17
Add impsort-maven-plugin

ppalaga:


poms/build-parent/pom.xml
```diff
@@ -111,6 +112,17 @@
                         <configFile>${camel.quarkus.project.root}/tooling/eclipse-formatter-config.xml</configFile>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>net.revelc.code</groupId>
+                    <artifactId>impsort-maven-plugin</artifactId>
+                    <version>${impsort-maven-plugin.version}</version>
+                    <configuration>
+                        <groups>java.,javax.</groups>
```
lburgazzoli: maybe match: https://github.com/apache/camel/blob/master/buildingtools/src/main/resources/camel-checkstyle.xml#L142

ppalaga: Pulling `org.w3c` and `org.xml` up sounds a bit strange, but as you wish.

PR #634, opened 2020-01-17
Run native tests with GitHub Actions

lburgazzoli:


.github/workflows/pr-build.yaml
```diff
@@ -0,0 +1,365 @@
...
+  #     # of memory this we nee to limit it to avoid the process fail for OOM.
+  #     - name: Integration Tests
+  #       run: |
+  #         ./mvnw -V -B \
+  #           -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn clean verify \
+  #           -Dnative \
+  #           -Ddocker \
+  #           -DskipTests \
+  #           -Dnative-image.xmx=5g \
+  #           -pl :camel-quarkus-integration-test-fhir
```
ppalaga: fhir does not pass at all on GH actions?

pom.xml
```diff
@@ -602,6 +602,38 @@
                 <native-image.container-runtime>docker</native-image.container-runtime>
             </properties>
         </profile>
+
+        <profile>
+            <id>no-apache-snapshots</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
```
ppalaga: Remote snapshots are evil because they make builds non-reproducible. I vote for having this the other way round: off by default, active only in a non-default profile.

.github/workflows/pr-build.yaml
lburgazzoli: The native image compilation fails for out of memory, I have not yet found any combination to make it working in a reasonable amount of time (with the default settings it fails after ~ 30 minutes)

pom.xml
lburgazzoli: The apache snapshots have been enabled by default since long time ago and I have no problem to change the default behavior but I didn't want to change it while doing some experimentation because I do not remember if there is a reason or a job that need this set-up.

lburgazzoli:
> I have concerns about the maintainability. Maybe we should have a check in Maven that would ensure that every itest is listed in the GH actions yaml file?

There are certainly some tread off to take into account so we can have a check in place or we can even generate part of the yaml based on some properties but before going so far I'd like to experiment if this is really a viable option i.e. it may be the case that we can't run all the test like `fhir`

It may turn out that we use actions to run smoke tests only so I'd wait a little to define the tooling strategy till we have all the relevant information to take a decision.

lburgazzoli:
Without `fhir` the whole build took half the time it takes with jenkins:

image

pom.xml
lburgazzoli: I've now moved the snapshot repo to a dedicate profile

.github/workflows/pr-build.yaml
```diff
@@ -0,0 +1,381 @@
...
+      - name: Build Project
+        run: |
+          ./mvnw -V -B \
+            -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn \
+            clean install
+      - name: Save Cache
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: maven-${{ github.sha }}
```
zregvart: This would make the cache single-use only, there's [an example](https://github.com/actions/cache/blob/master/examples.md#java---maven) on how to do it by hashing the POM files.

lburgazzoli: if we hash the poms and we change the sources then we end up using wrong artifacts, isn't it ?

lburgazzoli: GitHub actions provide a few ways to "solve" this:

- caching
- deploy artefacts

None of them is optimal but the caching one is the easier to implement as today and as further step, we could cache only what the project produces instead of the whole local maven repo as the cache has also some limitations in term of space.

.github/workflows/pr-validate.yml
```diff
@@ -31,4 +35,10 @@ jobs:
       with:
         java-version: 1.8
     - name: mvn package
-      run: export LANG=en_US && ./mvnw -fae -B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn clean package -Pcheck-format -DskipTests
+      run: |
+        ./mvnw -fae -B \
+          -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn \
+          -DskipTests \
+          -Pcheck-format \
+          clean package
```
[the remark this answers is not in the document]
lburgazzoli: no problem to change it but since this is something outside the scope of this pr and as this behavior has not changed by this pr, I would change it with a follow up pr

.github/workflows/pr-build.yaml
```diff
@@ -0,0 +1,381 @@
...
+
+name: Build PR
+
+on:
+  pull_request:
+    branches:
+      - master
+
+env:
+  LANG: en_US
```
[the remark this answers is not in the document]
lburgazzoli: this has been introduced as workaround for a quarkus issue fixed by https://github.com/quarkusio/quarkus/pull/6681on quarkus main, since it is a workaround and should be removed I would leave it as the quarkus guys defined it

PR #639, opened 2020-01-20
Removed the Apache repositories since now the situation should be ok …

oscerd:
…in central

Lets try to build


oscerd:
ok to test

PR #647, opened 2020-01-21
Fix #518 Rely on configurers for Configuration classes instead of using

### LINKED ISSUE #518: Rely on configurers for Configuration classes instead of using reflection
Once a fix for https://issues.apache.org/jira/browse/CAMEL-14263 reaches us, this once rejected PR  https://github.com/apache/camel-quarkus/pull/513 contains most of the required changes.

ppalaga:
reflection

Removal of NettyConfiguration, NettyServerBootstrapConfiguration, PdfConfiguration registration led to native failures so these need a separate investigation.

### DIFF
--- a/extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/NativeImageProcessor.java
+++ b/extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/NativeImageProcessor.java
@@ -92,7 +92,8 @@ class NativeImageProcessor {
                     .filter(ai -> {
                         AnnotationValue av = ai.value("loader");
                         boolean isLoader = av != null && av.asBoolean();
-                        // filter out camel-base converters which are automatically inlined in the CoreStaticTypeConverterLoader
+                        // filter out camel-base converters which are automatically inlined in the
+                        // CoreStaticTypeConverterLoader
                         // need to revisit with Camel 3.0.0-M3 which should improve this area
                         if (ai.target().asClass().name().toString().startsWith("org.apache.camel.converter.")) {
                             log.debug("Ignoring core " + ai + " " + ai.target().asClass().name());
--- a/extensions/support/policy/deployment/src/main/java/org/apache/camel/quarkus/component/support/policy/deployment/PolicyProcessor.java
+++ b/extensions/support/policy/deployment/src/main/java/org/apache/camel/quarkus/component/support/policy/deployment/PolicyProcessor.java
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.support.policy.deployment;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
+import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.deployment.builditem.GeneratedResourceBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+import org.apache.camel.quarkus.core.deployment.UnbannedReflectiveBuildItem;
+import org.jboss.jandex.AnnotationTarget.Kind;
+import org.jboss.jandex.DotName;
+
+class PolicyProcessor {
+
+    private static final String FEATURE = "camel-policy";
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    UnbannedReflectiveBuildItem unbanReflectives() {
+        /**
+         * A list of classes annotated with <code>@UriParams</code> which we accept to be registered for reflection
+         * mostly because there are errors when they are removed. TODO: solve the underlying problems and remove as
+         * many entries as possible from the list.
+         */
+        return new UnbannedReflectiveBuildItem(
+                "org.apache.camel.support.processor.DefaultExchangeFormatter",
+                "org.apache.camel.component.pdf.PdfConfiguration",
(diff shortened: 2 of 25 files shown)


lburgazzoli:
ok to test

ppalaga:
Strange, the Consul test is passing for me locally. Let me figure out what is different here.

ppalaga:
A race for the configurers set on the component is my first theory.

ppalaga:
> A race for the configurers set on the component is my first theory.

Much simpler: stale deployment modules in my local Maven repo

Jenkins was right and I can reproduce.

Looks like this change needs to wait for Camel 3.1.0 because https://issues.apache.org/jira/browse/CAMEL-14263 was not fixed for all components in 3.0.1.

ppalaga:
36102f2 rebased and tested properly

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/NativeImageProcessor.java
```diff
@@ -68,6 +71,66 @@
...
+                "org.apache.camel.component.fhir.FhirSearchEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirTransactionEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirCreateEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirValidateEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirReadEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirCapabilitiesEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirHistoryEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirMetaEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirPatchEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirDeleteEndpointConfiguration",
```
lburgazzoli: would' be better to have a build item generated by the respective extensions ?

ppalaga: No, I'd prefer a central location to see better what is whitelisted. I hope the list will gradually disappear.

lburgazzoli: But external extension won't have a way to make it working

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/NativeImageProcessor.java
```diff
@@ -68,6 +71,66 @@
...
+
+            Set<String> violations = reflectiveClass.stream()
+                    .map(ReflectiveClassBuildItem::getClassNames)
+                    .flatMap(Collection::stream)
+                    .filter(cl -> !URI_PARAMS_WHITELIST.contains(cl))
+                    .filter(bannedClassNames::contains)
+                    .collect(Collectors.toSet());
+
+            if (!violations.isEmpty()) {
+                throw new IllegalStateException(
```
lburgazzoli: all the camel components have a property that can control if the generated configurer have to be used or not so I feel we should log a warning but I'd rather avoid to fail the build.

extensions/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/NativeImageProcessor.java
ppalaga: True. I was thinking of creating a dedicated extension for this (let's call it "policy extension") that would only be pulled by our tests. This would allow us to enforce the policy for our extensions and let 3rd party extensions do what they want. WDYT?

lburgazzoli: I really do not like to have to touch core to make an extension working so if it were me to decide, I'd go for a simple build item also, I do not know it it really make sense to put such enforcement

ppalaga: > I really do not like to have to touch core to make an extension working so if it were me to decide, I'd go for a simple build item.

I still think policy definitions are better to keep centrally, but OK, let me try to implement this with a build item.

> also, I do not know it it really make sense to put such enforcement

Well, I think this is the most effective, scalable (and sustainable) way to reduce the use of reflection. I do not think it is possible to be able to keep control once we will have 100+ extensions.

ppalaga:
c4c1663 introduces a build item as requested by @lburgazzoli

extensions/support/policy/deployment/src/main/java/org/apache/camel/quarkus/component/support/policy/deployment/PolicyProcessor.java
```diff
@@ -0,0 +1,112 @@
...
+
+        Set<String> violations = reflectiveClasses.stream()
+                .map(ReflectiveClassBuildItem::getClassNames)
+                .flatMap(Collection::stream)
+                .filter(cl -> !unbannedClassNames.contains(cl))
+                .filter(bannedClassNames::contains)
+                .collect(Collectors.toSet());
+
+        if (!violations.isEmpty()) {
+            throw new IllegalStateException(
```
lburgazzoli: Should we really fail the build ?

extensions/support/policy/deployment/src/main/java/org/apache/camel/quarkus/component/support/policy/deployment/PolicyProcessor.java
```diff
@@ -0,0 +1,112 @@
...
+                "org.apache.camel.component.fhir.FhirHistoryEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirMetaEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirPatchEndpointConfiguration",
+                "org.apache.camel.component.fhir.FhirDeleteEndpointConfiguration",
+
+                /* org.apache.camel.component.consul.* can be removed after the upgrade to Camel 3.1 */
+                "org.apache.camel.component.consul.ConsulConfiguration",
+                "org.apache.camel.component.consul.ConsulClientConfiguration",
+                "org.apache.camel.component.consul.health.ConsulHealthCheckRepositoryConfiguration",
+                "org.apache.camel.component.consul.cloud.ConsulServiceRegistryConfiguration");
```
lburgazzoli: It does not change much from the previous implementation as when we'll remove i.e. `PdfConfiguration` from the pdf extension we have to remember to get to this artefact and change it.

I really don't get why this should not be side-by-side with the code declaring the need for the reflective classes.

extensions/support/policy/deployment/src/main/java/org/apache/camel/quarkus/component/support/policy/deployment/PolicyProcessor.java
ppalaga: Yes, this is only with our tests. It won't happen at runtime (unless the user explictly depends on the new policy extension)

extensions/support/policy/deployment/src/main/java/org/apache/camel/quarkus/component/support/policy/deployment/PolicyProcessor.java
ppalaga: These are the anomalies we need to tackle. They will hopefully disappear soon. If some of these turns out to be more stable, the item can be moved to the respective extension.

lburgazzoli: I give up :)

PR #655, opened 2020-01-22
fhir: reduce uri options by leveraging component configuration

lburgazzoli:


lburgazzoli:
native fhir test is ok on my pc

PR #674, opened 2020-01-30
fix(rest-json/pom.xml): use quarkus-based dependency

hanzo2001:
Just switch dependency from `camel-jackson` to `camel-quarkus-jackson` in the `rest-json` example


oscerd:
Thank you

Issue #654, opened 2020-01-22
fhir native compilation is eating up lot of memory


lburgazzoli:
The native compilation of the camel-fhir extension is quite demanding in term of resources and the native-process easily reach 14GB of RSS (which is the hard coded limit the native-image can reach) which make not possible to test the fhir extension as examplle with GitHub Actions.

We should investigate how to optimize the build of the fhir extension

lburgazzoli:
I did try to limit the memory the native-image can require but in that case either the build take too long or fails because of OutOfMemory

johnpoth:
@lburgazzoli are you running native compilation with all FHIR specifications enabled ? (DSTU2...R5) 

Also what is the acceptable memory consumption when doing native compilation for a quarkus extension? Certainly not 14GB but I do feel like it's higher than average.

When testing I noticed that the more Java reflection FHIR used, the more memory consumption occurred when doing native compilation.

lburgazzoli:
I'm running the test suite for the fhir extension so in native mode it should take into account only a single specification.

I don't have any specific number in mind about how much memory a native-compilation can take but the public CI we can use have limitations around 7/8 GB so we should take that into account.

lburgazzoli:
I do not know if it could make sense to split the extension in code and specifications so we do not include any specification or 3th party library that is not strictly needed. I know that we do already have flags that tweak what we do include but the dependencies are still on the classpath and I'm not sure substrate can easily find what code paths are truly not possible thus they can be excluded from the analysis.

johnpoth:
@lburgazzoli I thought we were already doing that with the said flags. It would be cool if there was a substrate "flag" to ignore the specification libraries not used on the classpath. This would avoid creating as you suggested camel-fhir-dstu3 etc... quarkus extensions.

Another options would be to add the specification Maven dependencies in Profiles.

lburgazzoli:
I do not even know if doing that would really lower the requirements, just the first thing that came into my mind :)

johnpoth:
ah ok :) I'll look into it

johnpoth:
Quick update. I've tried removing the dependencies mentioned above and also removing the quarkus BuildSteps in [FhirProcessor](https://github.com/apache/camel-quarkus/blob/master/extensions/fhir/deployment/src/main/java/org/apache/camel/quarkus/component/fhir/deployment/FhirProcessor.java#L77) but I still get 11 GB of memory use. Further investigation is needed to understand whats taking up all this memory.

[...]

PR #678, opened 2020-02-03
camel-fhir record FHIR context instead of creating it at runtime. Onl…

johnpoth:
…y one FHIR context per specification is permitted in native mode

cc @lburgazzoli addresses #654 

I've changed the way HAPI FHIR integrates with Quarkus. The approach used here is similar to what camel-quarkus-core does.

Instead of enabling reflection everywhere, I've opted to record the FHIR context (for each specification). Therefore only one context per specification can be used in native mode. This limitation is in line with what is [recommended](https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirContext.java#L57).

Thanks !

### DIFF
--- a/extensions/fhir/deployment/src/main/java/org/apache/camel/quarkus/component/fhir/deployment/dstu2/Dstu2PropertiesBuildItem.java
+++ b/extensions/fhir/deployment/src/main/java/org/apache/camel/quarkus/component/fhir/deployment/dstu2/Dstu2PropertiesBuildItem.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.fhir.deployment.dstu2;
+
+import java.util.Properties;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+
+public final class Dstu2PropertiesBuildItem extends SimpleBuildItem {
+
+    private final Properties properties;
+
+    public Dstu2PropertiesBuildItem(Properties properties) {
+        this.properties = properties;
+    }
+
+    public Properties getProperties() {
+        return properties;
+    }
+}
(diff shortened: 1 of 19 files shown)


johnpoth:
Thanks @oscerd! Done

oscerd:
Thanks

johnpoth:
Thanks @oscerd !

extensions/fhir/deployment/src/main/java/org/apache/camel/quarkus/component/fhir/deployment/dstu3/Dstu3PropertiesBuildItem.java
```diff
@@ -0,0 +1,34 @@
...
+public final class Dstu3PropertiesBuildItem extends SimpleBuildItem {
+
+    private final Properties properties;
+
+    public Dstu3PropertiesBuildItem(Properties properties) {
+        this.properties = properties;
+    }
+
+    public Properties getProperties() {
+        return properties;
```
ppalaga: BuildItems should not expose mutable collections because they are consumed from concurrent threads. Mutability does not seem to be required here. Making the properties immutable would simplify the reasoning about the code.

extensions/fhir/deployment/src/main/java/org/apache/camel/quarkus/component/fhir/deployment/r4/R4PropertiesBuildItem.java
```diff
@@ -0,0 +1,34 @@
...
+public final class R4PropertiesBuildItem extends SimpleBuildItem {
+
+    private final Properties properties;
+
+    public R4PropertiesBuildItem(Properties properties) {
+        this.properties = properties;
+    }
+
+    public Properties getProperties() {
+        return properties;
```
ppalaga: Same as above: BuildItems should not expose mutable collections because they are consumed from concurrent threads. Mutability does not seem to be required here. Making the properties immutable would simplify the reasoning about the code.

extensions/fhir/deployment/src/main/java/org/apache/camel/quarkus/component/fhir/deployment/r5/R5PropertiesBuildItem.java
```diff
@@ -0,0 +1,34 @@
...
+public final class R5PropertiesBuildItem extends SimpleBuildItem {
+
+    private final Properties properties;
+
+    public R5PropertiesBuildItem(Properties properties) {
+        this.properties = properties;
+    }
+
+    public Properties getProperties() {
+        return properties;
```
ppalaga: Same as above: BuildItems should not expose mutable collections because they are consumed from concurrent threads. Mutability does not seem to be required here. Making the properties immutable would simplify the reasoning about the code.

extensions/fhir/deployment/src/main/java/org/apache/camel/quarkus/component/fhir/deployment/dstu2/Dstu2PropertiesBuildItem.java
```diff
@@ -0,0 +1,34 @@
...
+public final class Dstu2PropertiesBuildItem extends SimpleBuildItem {
+
+    private final Properties properties;
+
+    public Dstu2PropertiesBuildItem(Properties properties) {
+        this.properties = properties;
+    }
+
+    public Properties getProperties() {
+        return properties;
```
ppalaga: Same as above: BuildItems should not expose mutable collections because they are consumed from concurrent threads. Mutability does not seem to be required here. Making the properties immutable would simplify the reasoning about the code.

lburgazzoli:
@ppalaga as the PR is already merged maybe better to open an issue as it would make it easier to be aware of the problem and fix it

ppalaga:
> maybe better to open an issue as it would make it easier to be aware of the problem and fix it

I wanted to give the general public the chance to prove that I am wrong :)

lburgazzoli:
its is still easier with an issue

ppalaga:
Filed a followup: https://github.com/apache/camel-quarkus/issues/690

ppalaga:
> its is still easier with an issue

Not for a lazy reviewer like me :P

PR #679, opened 2020-02-04
Create enforce profile that disables dependency checks

johnpoth:
Hello,

Often when developing I find myself commenting out these checks so I moved them into a profile? To disable checking one can set the `dev` property e.g `maven clean install -Ddev`. Thanks !

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -210,27 +210,6 @@
 
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maven-enforcer-plugin</artifactId>
-                    <executions>
-                        <execution>
-                            <id>camel-quarkus-enforcer-rules</id>
-                            <goals>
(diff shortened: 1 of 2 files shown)


johnpoth:
@ppalaga @oscerd naming is hard ! How about `rules`, `check`,`enforce`, `constraints` ? We would disable it by setting `-Dcheck=false` or `-Denforce=false` etc...

oscerd:
Sounds good for me.

ppalaga:
`-Denforce=false` or also `-DnoEnforce` sound acceptable

johnpoth:
@ppalaga @oscerd I've renamed the profile name to 'enforce'. I think it is clearer when names don't have negations in them (so here I've preferred 'enforce' as opposed to 'no-enforce'). Also the profile does enforce sanity checks so it makes sense IMO. Hope it's okay, thanks !

PR #693, opened 2020-02-11
Add olingo4 extension

### LINKED ISSUE #692: Create a camel olingo4 extension

jamesnetherton:
fixes #692

### DIFF
--- a/extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
+++ b/extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.olingo4.deployment;
+
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.AdditionalApplicationArchiveMarkerBuildItem;
+import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
+import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+import org.apache.camel.component.olingo4.Olingo4AppEndpointConfiguration;
+import org.apache.camel.quarkus.core.deployment.UnbannedReflectiveBuildItem;
+import org.apache.olingo.server.core.ODataImpl;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.IndexView;
+
+class Olingo4Processor {
+
+    private static final String FEATURE = "camel-olingo4";
+    private static final DotName JSON_DESERIALIZE_DOT_NAME = DotName
+            .createSimple("com.fasterxml.jackson.databind.annotation.JsonDeserialize");
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    UnbannedReflectiveBuildItem whitelistOlingo4AppEndpointConfiguration() {
+        // TODO: Remove this and the associated ReflectiveClassBuildItem for this class in Camel 3.1
+        return new UnbannedReflectiveBuildItem(Olingo4AppEndpointConfiguration.class.getName());
+    }
+
+    @BuildStep
+    AdditionalApplicationArchiveMarkerBuildItem olingoArchiveMarker() {
+        return new AdditionalApplicationArchiveMarkerBuildItem("org/apache/olingo");
+    }
+
+    @BuildStep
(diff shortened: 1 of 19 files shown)


extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
```diff
@@ -0,0 +1,66 @@
...
+            .createSimple("com.fasterxml.jackson.databind.annotation.JsonDeserialize");
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    UnbannedReflectiveBuildItem whitelistOlingo4AppEndpointConfiguration() {
+        return new UnbannedReflectiveBuildItem(Olingo4AppEndpointConfiguration.class.getName());
```
ppalaga: Do you happen to know if this is necessary also on Camel master? If not, a comment would be handy that it can be removed after Camel 3.1. Otherwise, a Camel issue would be nice requiring to fix what is necessary in Camel.

extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @BuildStep
+    void registerForReflection(BuildProducer<ReflectiveClassBuildItem> reflectiveClass, CombinedIndexBuildItem combinedIndex) {
+        reflectiveClass.produce(new ReflectiveClassBuildItem(true, false, ODataImpl.class));
+        reflectiveClass.produce(new ReflectiveClassBuildItem(true, true, Olingo4AppEndpointConfiguration.class));
+
+        // Register Olingo Deserializer classes for reflection
+        IndexView index = combinedIndex.getIndex();
+        index.getAnnotations(JSON_DESERIALIZE_DOT_NAME)
```
ppalaga: I think this done by Quarkus Jackson extension. If that's really the case, we should perhaps prefer depending on them.

extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
```diff
@@ -0,0 +1,66 @@
...
+                .stream()
+                .map(annotation -> annotation.value("using").asClass().name().toString())
+                .filter(className -> className.startsWith("org.apache.olingo"))
```
ppalaga: Hm... why is this filter actually required?

extensions/support/httpclient/deployment/src/main/java/org/apache/camel/quarkus/support/httpclient/deployment/HttpClientProcessor.java
```diff
@@ -45,4 +46,10 @@ void registerForReflection(
             reflectiveClasses.produce(new ReflectiveClassBuildItem(true, false, info.name().toString()));
         }
     }
+
+    @BuildStep
+    NativeImageResourceBuildItem suffixListResource() {
+        // Required by org.apache.http.conn.util.PublicSuffixMatcher
```
ppalaga: I appreciate the comment!

extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
jamesnetherton: I have not checked. I assume it is fixed, but I can add a note. We could eventually remove this  on the `camel-master` branch.

extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
jamesnetherton: I'll double check, but I didn't see this being done by the Jackson extension.

extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
jamesnetherton: Should the extension concern itself with configuring reflection for stuff that is not associated with it? E.g non-olingo stuff.

extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
ppalaga: Here is what they do https://github.com/quarkusio/quarkus/blob/master/extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java#L94

extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
ppalaga: OK, I see the problem. Hard to say, I am not acquainted well enough with olingo. Are the users likely to write their own deserializers to be able to use the olingo extension?

extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
jamesnetherton: The Quarkus Jackson extension seems to ignore cases like this:

```java
@JsonDeserialize(using = ClientCsdlAction.ActionDeserializer.class)
class ClientCsdlAction extends CsdlAction implements Serializable {
...
}
```

It only takes the `using` part into consideration where the annotation is applied to a field or method.

I can add a comment about that and maybe raise an issue on the Quarkus side.

extensions/olingo4/deployment/src/main/java/org/apache/camel/quarkus/component/olingo4/deployment/Olingo4Processor.java
jamesnetherton: I don't think folks are going to want to write their own deserializers when interacting with the olingo component extension. So for now, I'd say it's safe to include the filter.

It will be simple enough to remove if we get feedback that proves that we don't need it.

PR #698, opened 2020-02-11
Camel master update

ppalaga:


lburgazzoli:
@ppalaga there are some build failures

ppalaga:
> @ppalaga there are some build failures

Yeah, there is never enough fun with remote snapshots. The Java 12 job was perhaps unlucky enough to get a newer snapshot that added a method to AbstractCamelContext. Because otherwise Java 8 should have failed too. Once we use the same cache for java 8 11 and 12 this should not happen anymore.

ppalaga:
ConfigurerResolver adapts to the newest changes in Camel

ppalaga:
Failed due to transfer problems again. Let's rebase on top of master once https://github.com/apache/camel-quarkus/pull/705 is merged and see if the transfer problems go away.

jamesnetherton:
@ppalaga #705 is merged. Can we rebase this one and try again?

ppalaga:
One more fix in e974195

ppalaga:
There is never enough fun with snapshots. The CI apparently uses a different snapshot than my machine. There is no way to figure out which one is used by the CI and there is no way to force the two machines to use the same snapshot. What do you guys do in situations like this?

jamesnetherton:
The error is reproducible by slurping down the latest SNAPSHOTs. E.g build with `-U`.

But you're right, it is a bit tricky to manage when the target keeps on moving.

lburgazzoli:
> But you're right, it is a bit tricky to manage when the target keeps on moving.

it is the funny part

ppalaga:
Updated what I wanted, let's wait for the next random error.

PR #700, opened 2020-02-11
CAMEL-QUARKUS-681: Created a camel gson extension

aldettinger:
Let's see how pr-builds go.

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -235,6 +235,7 @@ jobs:
             -DskipTests \
             -pl :camel-quarkus-integration-test-base64 \
             -pl :camel-quarkus-integration-test-bindy \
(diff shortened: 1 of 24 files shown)


.github/workflows/pr-build.yaml
```diff
@@ -235,6 +235,7 @@ jobs:
             -DskipTests \
             -pl :camel-quarkus-integration-test-base64 \
             -pl :camel-quarkus-integration-test-bindy \
+            -pl :camel-quarkus-integration-test-gson \
```
lburgazzoli: wonder if we can merge gson and jackson tests in a single dataformat-json module (similar to what we do for http)

aldettinger:
ok, let's see how pr builds go then I'm gonna check that.

PR #704, opened 2020-02-12
Ensure that catalog files are added to the native image

### LINKED ISSUE #686: Ensure that catalog files are added to the native image
Camel artifacts may brings some json files i.e. the http component brings `org/apache/camel/component/http/http.json` that are used by the runtime camel catalog.

We should ensure those files are added to the native image and have an option to disable camel catalog at all.

lburgazzoli:
Fixes #686

### DIFF
--- a/extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
+++ b/extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
@@ -39,7 +39,7 @@ public class CamelConfig {
     public ServiceConfig service;
 
     /**
-     * Build time configuration options for CamelRuntimeCatalog.
+     * Build time configuration options for {@link org.apache.camel.runtimecatalog.RuntimeCamelCatalog}.
      */
     @ConfigItem
     public RuntimeCatalogConfig runtimeCatalog;
@@ -183,16 +183,11 @@ public class CamelConfig {
     @ConfigGroup
     public static class RuntimeCatalogConfig {
         /**
-         * Enable {@link CamelRuntimeCatalog} functionaries.
-         */
-        @ConfigItem(defaultValue = "true")
-        public boolean enabled;
-
-        /**
          * Used to control the resolution of components catalog info.
          * <p>
          * Note that when building native images, this flag determine if the json metadata files related to components
-         * discovered at build time have to be included in the final binary.
+         * discovered at build time have to be included in the final binary. In JVM mode there is no real benefit of
+         * setting this flag to {@code false} if not to make the behavior consistent with native mode.
          */
         @ConfigItem(defaultValue = "true")
         public boolean components;
@@ -201,7 +196,8 @@ public class CamelConfig {
          * Used to control the resolution of languages catalog info.
          * <p>
          * Note that when building native images, this flag determine if the json metadata files related to languages
-         * discovered at build time have to be included in the final binary.
+         * discovered at build time have to be included in the final binary. In JVM mode there is no real benefit of
+         * setting this flag to {@code false} if not to make the behavior consistent with native mode.
          */
         @ConfigItem(defaultValue = "true")
         public boolean languages;
@@ -210,7 +206,8 @@ public class CamelConfig {
          * Used to control the resolution of dataformats catalog info.
          * <p>
          * Note that when building native images, this flag determine if the json metadata files related to dataformats
-         * discovered at build time have to be included in the final binary.
+         * discovered at build time have to be included in the final binary. In JVM mode there is no real benefit of
+         * setting this flag to {@code false} if not to make the behavior consistent with native mode.
          */
         @ConfigItem(defaultValue = "true")
         public boolean dataformats;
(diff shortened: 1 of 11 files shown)


extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
```diff
@@ -174,4 +180,48 @@
         public Optional<List<String>> includePatterns;
     }
 
+    @ConfigGroup
+    public static class RuntimeCatalogConfig {
+        /**
+         * Enable {@link CamelRuntimeCatalog} functionaries.
```
ppalaga: I have just learned about the existence of `functionary/functionaries` as a valid work of the English language. Anyway: is that not a typo?

In the description of the option, I'd like to learn what exactly happens when the option is true and what happens otherwise. 

If I am a new user, I may have no idea what is CamelRuntimeCatalog. Yes there is a link to CamelRuntimeCatalog, but that class has no JavaDoc. Moreover, the link will most probably not work everywhere (IDEs, Web docs) where the text will be displayed. So either a working link to a resource explaining the concept of the catalog or a short inline definition would be nice.

It would also be nice to document the relationship (if there is any) to other options in this file - e.g. does enabled = false imply that all other options in this file are ignored?

Does this option only matter for the native mode? What happens in the JVM mode then?

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
```diff
@@ -174,4 +180,48 @@
...
+         */
+        @ConfigItem(defaultValue = "true")
+        public boolean enabled;
+
+        /**
+         * Used to control the resolution of components catalog info.
```
ppalaga: Is "resolution" really what this does? Would not "inclusion" suit better?

Again, it may be unclear to the reader, what is catalog. Maybe a link to a Camel docs page explaining the concept would be nice. The description should give me a hint why/when should I care, why would I ever need this to be true?

This also applies to all other options in this file.

lburgazzoli: > Is "resolution" really what this does? Would not "inclusion" suit better?
> 


It is resolution, if you go through the code you'd see that this flag controls if the runtime catalog should attempt or not to resolve the given name to its model


> Again, it may be unclear to the reader, what is catalog. Maybe a link to a Camel docs page explaining the concept would be nice. The description should give me a hint why/when should I care, why would I ever need this to be true?
> 

The impact on the native image is described a few lines below.

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
lburgazzoli: > I have just learned about the existence of `functionary/functionaries` as a valid work of the English language. Anyway: is that not a typo?
> 
> In the description of the option, I'd like to learn what exactly happens when the option is true and what happens otherwise.
> 

It is a typo

> If I am a new user, I may have no idea what is CamelRuntimeCatalog. Yes there is a link to CamelRuntimeCatalog, but that class has no JavaDoc. Moreover, the link will most probably not work everywhere (IDEs, Web docs) where the text will be displayed. So either a working link to a resource explaining the concept of the catalog or a short inline definition would be nice.
> 

I don't think we need to describe what the RuntimeCamelContext is in camel-quarkus so if there's lack of documentation on camel side, please raise an issue there. 

> It would also be nice to document the relationship (if there is any) to other options in this file - e.g. does enabled = false imply that all other options in this file are ignored?
> 
> Does this option only matter for the native mode? What happens in the JVM mode then?

Agree, this need to be improved

lburgazzoli: should be fixed now

extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
lburgazzoli: should be fixed now

ppalaga:
Thanks for the improvements. I still think the config options we expose should be documented so that they make sense to a newcomer. If there is no Camel page explaining the nature and purpose of the catalog, then I vote for adding a sentence here rather than waiting for it to happen elsewhere. I'd be ready to write that sentence but I am quite frankly missing the required knowledge.

lburgazzoli:
> Thanks for the improvements. I still think the config options we expose should be documented so that they make sense to a newcomer. If there is no Camel page explaining the nature and purpose of the catalog.

Please rise an issue on camel side, we should improve camel in general rather than this specific sub-project so we can do the work only once for everyone to benefit.

btw, there are already some javadocs - maybe limited - on the related classes on camel side:
- https://github.com/apache/camel/blob/master/catalog/camel-catalog/src/main/java/org/apache/camel/catalog/CamelCatalog.java
- https://github.com/apache/camel/blob/master/core/camel-api/src/main/java/org/apache/camel/catalog/RuntimeCamelCatalog.java

ppalaga:
Done https://github.com/apache/camel-quarkus/issues/706 and https://issues.apache.org/jira/browse/CAMEL-14545

PR #705, opened 2020-02-12
Update mvnd.builder.rules

ppalaga:


ppalaga:
The previous iteration failed on a dependency download again. Lets try with `-Dmaven.wagon.http.retryHandler.requestSentEnabled=true` in 748db16

lburgazzoli:
You probably need to add somewhere what mvnd is and what's the meaning of the rules

ppalaga:
Added comments in a8526df

lburgazzoli:
LGTM

@ppalaga Wonder if mvnd should have a way to extend through service loader mechanism so one could contribute a logic to compute the rules without polluting the poms

ppalaga:
Sounds interesting. Would you mind proposing it via https://github.com/gnodet/mvnd/issues/new %

lburgazzoli:
maven download seems very unstable

PR #720, opened 2020-02-14
Fix #598 Use quarkus.package.type=native instead of the native-image mojo

### LINKED ISSUE #598: Use quarkus.package.type=native instead of the native-image mojo
Some of these options look a bit suspicious. We should check at some point whether enableHttpsUrlHandler enableJni and enableJni are really required.

https://github.com/apache/camel-quarkus/pull/597#discussion_r363190299

ppalaga:


ppalaga:
OK, I see that the props for the docker build have changed.

ppalaga:
ce43991 is using
```
                <quarkus.native.container-build>true</quarkus.native.container-build>
                <quarkus.native.container-runtime>docker</quarkus.native.container-runtime>
```

instead of the old ones

```
                <native-image.docker-build>true</native-image.docker-build>
```

ppalaga:
f833a32 :

* Rebased
* Reverted the Jira itest, because it still does not work without the custom options. I hope @ffang will solve that in his upcoming PR. 
* There was also a failure in the base64 itest due to missing camel-quarkus-base64.jar which I find unexplicable because it should have been provided by the `build` job. Maybe the Github actions cache is buggy?

PR #732, opened 2020-02-18
Removed useless Jenkinsfile(s)

oscerd:


oscerd:
I don't think we need to run the gh actions for this :-)

PR #738, opened 2020-02-19
Created a camel gson extension #681

aldettinger:


integration-tests/dataformats-json/src/main/java/org/apache/camel/quarkus/component/jackson/JsonDataformatsRoute.java
```diff
@@ -16,47 +16,93 @@
...
 import org.apache.camel.quarkus.component.jackson.model.PojoB;
+import org.apache.camel.spi.DataFormat;
 
 public class JsonDataformatsRoute extends RouteBuilder {
 
     @Override
     public void configure() {
-        JacksonDataFormat format = new JacksonDataFormat(DummyObject.class);
-        format.useList();
+        JacksonDataFormat jacksonDummyObjectDataFormat = new JacksonDataFormat(DummyObject.class);
```
ppalaga: Why do these data format variables have the `Dummy` infix in their name? I do not see anything "dummy" in how they are used.

aldettinger: The idea is that it's the dataformat serializing dummy object. That said, I would not veto a renaming of DummyObject too.

ppalaga: Makes sense. Thanks for the explanation :)

PR #754, opened 2020-02-24
Fix #960 Do not expose mutable collections from FHIR BuildItems

ppalaga:


ppalaga:
@johnpoth could you please review?

johnpoth:
@ppalaga looks great! However...

The idea here is to allow the user to add FHIR resources to the FHIR context before recording. Otherwise you'll get a ClassNotFoundException. See https://wiki.hl7.org/FHIR_Custom_Resources

Sorry for the late reply

lburgazzoli:
@johnpoth mind opening an issue ?

ppalaga:
> allow the user to add FHIR resources to the FHIR context before recording.

Could you please explain how is my change preventing that? (Maybe I do not understand properly what you mean with add "FHIR resources")

johnpoth:
Hi @ppalaga, the map contains the resources (classes) to be loaded. Now that the map is immutable it will make it harder for the user to add classes to the map. HTH

ppalaga:
> will make it harder for the user to add classes to the map

Yes, my patch is consciously preventing that for thread safety reasons. 

I do not think we should let users write their own extensions to be able to ad their custom resources. Couldn't we enable this via `application.properties` or allow the users to provide their own `fhirversion.properties`?

lburgazzoli:
What about introducing a FHIRModelBuildItem ? (or whatever name make sense)

johnpoth:
@lburgazzoli yes that would be the idea

ppalaga:
> What about introducing a FHIRModelBuildItem ?

Do we really want to let users write extensions to provide their custom FHIR resources? I wonder why is that necessary?

PR #758, opened 2020-02-25
Jira integration test fails in native mode

jamesnetherton:


jamesnetherton:
I'm going to merge this, previous runs were successful. I don't understand why we get these random failures on cache restoration.

Issue #223, opened 2019-10-03
Freemarker native support


carlosthe19916:

ppalaga:
@carlosthe19916 are you working on this one?

carlosthe19916:
@ppalaga I'd like to contribute in some way and I'm really curious about using Quakus and Freemarker in personal projects. I'm not sure if I will achieve the goal of adding a new extension but I'd like to give it a try.

ppalaga:
@carlosthe19916 new extensions are welcome! You may find https://camel.apache.org/camel-quarkus/latest/contributor-guide.html useful.

ppalaga:
@carlosthe19916 are you still interested to look into this?

carlosthe19916:
@ppalaga My apologies for not updating this Issue before. I'm still interested in this for sure. Thb, I've created my very basic version of Native Freemarker extension here https://github.com/project-openubl/xml-builder/blob/master/extensions/freemarker/deployment/src/main/java/org/openublpe/quarkus/freemarker/deployment/FreemarkerProcessor.java. I'm not sure how much different could be using camel

carlosthe19916:
@ppalaga Here I've just created a very basic first draft for the camel-freemarker extension. But I still need to enhance the code https://github.com/carlosthe19916/camel-quarkus/tree/camel-freemarker

ppalaga:
@carlosthe19916 tanks for the effort! If you want an early feedback, please feel free to send a draft pull request https://github.blog/2019-02-14-introducing-draft-pull-requests/

ppalaga:
A JVM-only extension (without a proper test) is underway.

vsevel:
is there any plan to offer freemarker as a quarkus extension (independently from camel), and get the camel-quarkus to depend on it?
this would be useful for people interested in quarkus+freemarker, and not necessarily in camel.

ppalaga:
> is there any plan to offer freemarker as a quarkus extension (independently from camel), and get the camel-quarkus to depend on it?

This would indeed make sense and we'd be eventually ready to depend on such a generic Freemarker extension. I think the Freemarker project would be the best place for this effort. Any help to drive that effort is welcome.

vsevel:
> I think the Freemarker project would be the best place for this effort.

or the quarkus project itself. that is where almost all of the extensions are developed.

ppalaga:
> > I think the Freemarker project would be the best place for this effort.
> 
[...]

PR #835, opened 2020-03-02
Camel-freemarker component

carlosthe19916:
Issue: https://github.com/apache/camel-quarkus/issues/223

Added camel-freemarker component.
Changes:
- The same test used for apache-camel-freemarker
- We can add the list of locations where the templates can be found in the classpath using: `quarkus.camel.freemarker.locations=myFolder1,myFolder2`

Looking forward to have some feedback.


extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerNativeImageProcessor.java
```diff
@@ -0,0 +1,120 @@
+package org.apache.camel.quarkus.component.freemarker.deployment;
```
oscerd: Please add the ASF license header.

carlosthe19916: @oscerd Added, thanks

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerProcessor.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @BuildStep
+    RuntimeInitializedClassBuildItem jythonWrapper() {
+        return new RuntimeInitializedClassBuildItem(JythonWrapper.class.getCanonicalName());
+    }
+
+    @BuildStep
+    void registerForReflection(BuildProducer<ReflectiveClassBuildItem> reflectiveClass) {
+        reflectiveClass.produce(new ReflectiveClassBuildItem(true, true,
+                Exchange.class,
```
ppalaga: Is Exchange really required here?

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerProcessor.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @BuildStep
+    RuntimeInitializedClassBuildItem jythonWrapper() {
+        return new RuntimeInitializedClassBuildItem(JythonWrapper.class.getCanonicalName());
+    }
+
+    @BuildStep
+    void registerForReflection(BuildProducer<ReflectiveClassBuildItem> reflectiveClass) {
+        reflectiveClass.produce(new ReflectiveClassBuildItem(true, true,
```
ppalaga: Wouldn't `new ReflectiveClassBuildItem(true, false, ...` be enough?

extensions/freemarker/runtime/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+        <artifactId>camel-quarkus-freemarker-parent</artifactId>
+        <version>1.1.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-freemarker</artifactId>
+    <name>Camel Quarkus :: Freemarker :: Runtime</name>
+
+    <properties>
+        <firstVersion>0.4.0</firstVersion>
```
ppalaga:
```suggestion
        <firstVersion>1.0.0-M5</firstVersion>
```

and re-run `mvn process-resources -Pformat -Denforce=false`

integration-tests/freemarker/src/main/resources/application.properties
```diff
@@ -0,0 +1,20 @@
...
+## Unless required by applicable law or agreed to in writing, software
+## distributed under the License is distributed on an "AS IS" BASIS,
+## WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+## See the License for the specific language governing permissions and
+## limitations under the License.
+## ---------------------------------------------------------------------------
+#
+# Quarkus
+#
+quarkus.camel.freemarker.locations = org/apache/camel/component/freemarker
```
ppalaga: Given that `org/apache/camel/component/freemarker` will be scanned recursivelly, I wonder whether all the class files are going to get included as native image resources? I would be interesting to compare the size of the native image using this locations value and using a new path that does not contain any class files.

ppalaga:
Could you please rebase and squash your commits? It would make it easier to review locally.

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerProcessor.java
carlosthe19916: You're right, I'll update that, Thanks :)

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerProcessor.java
carlosthe19916: Yes, adding Exchange for reflection is needed. According to this documentation: [freemarker context](https://camel.apache.org/components/latest/freemarker-component.html#_freemarker_context), it should be possible to use `exchange.properties` in the `.ftl` templates just like this example: https://github.com/carlosthe19916/camel-quarkus/blob/7d45a783e5b27d341b6a07df49b34087f636dad3/integration-tests/freemarker/src/main/resources/org/apache/camel/component/freemarker/example.ftl#L19. Without using reflection for Exchange it won't work.

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerNativeImageProcessor.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final String JAR_APPLICATION_MIGRATIONS_PROTOCOL = "jar";
+    private static final String FILE_APPLICATION_MIGRATIONS_PROTOCOL = "file";
+
+    private static final Logger LOGGER = Logger.getLogger(FreemarkerNativeImageProcessor.class);
+
+    @BuildStep(loadsApplicationClasses = true)
+    void build(
+            BuildProducer<NativeImageResourceBuildItem> resourceProducer,
+            CamelFreemarkerConfig camelFreemarkerConfig) throws IOException, URISyntaxException {
+        registerNativeImageResources(resourceProducer, camelFreemarkerConfig);
```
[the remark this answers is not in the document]
carlosthe19916: Thanks for the good advice, I'll do it inline just like you suggested :)

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerNativeImageProcessor.java
```diff
@@ -0,0 +1,136 @@
...
+            List<String> locations = new ArrayList<>(camelFreemarkerConfig.locations);
+            if (locations.isEmpty()) {
+                locations.add("freemarker/templates");
+            }
+
+            // Locations can be a comma separated list
+            for (String location : locations) {
+                // Strip any 'classpath:' protocol prefixes because they are assumed
+                // but not recognized by ClassLoader.getResources()
+                if (location != null && location.startsWith(CLASSPATH_APPLICATION_MIGRATIONS_PROTOCOL + ':')) {
```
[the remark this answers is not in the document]
carlosthe19916: my mistake, it shouldn't be possible to have null values since it comes from `CamelFreemarkerConfig`

integration-tests/freemarker/src/main/resources/application.properties
carlosthe19916: Only files with the extension `.ftl` will be included. You can see there is a filter for that [here](https://github.com/carlosthe19916/camel-quarkus/blob/7d45a783e5b27d341b6a07df49b34087f636dad3/extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerNativeImageProcessor.java#L119)... I've also created 2 native images, the first one using `org/apache/camel/component/freemarker` and the second one using `myfreemarkerfolder` and the sizes are exactly the same (45.8 MB)

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerNativeImageProcessor.java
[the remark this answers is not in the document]
carlosthe19916: done

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerNativeImageProcessor.java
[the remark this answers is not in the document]
ppalaga: Hm..., I wonder how the MicroProfile config impl used by Quarkus would parse a double comma value like this

```
quarkus.camel.freemarker.locations = my/path1,,my/path2
```

integration-tests/freemarker/src/main/resources/application.properties
ppalaga: > Only files with the extension `.ftl` will be included.

Thanks for the info, I have not noticed that and I must say I do not think we should be selecting only `*.ftl` files. I commonly used other or no extensions for FreeMarker template files in the past. I think many people do that because because of syntax highlighting in editors. There is even a couple of examples in this repo: https://github.com/apache/camel-quarkus/tree/master/tooling/create-extension-templates

I think we should either include whole subtrees or we should have path includes and excludes like we have in https://github.com/apache/camel-quarkus/blob/master/extensions/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java#L74-L100 WDYT?

extensions/freemarker/runtime/pom.xml
carlosthe19916: done

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerNativeImageProcessor.java
```diff
@@ -0,0 +1,136 @@
...
+        resource.produce(new NativeImageResourceBuildItem(nativeResources.toArray(new String[0])));
+    }
+
+    private List<String> discoverFreemarkerTemplates(
+            CamelFreemarkerConfig camelFreemarkerConfig) throws IOException, URISyntaxException {
+        List<String> resources = new ArrayList<>();
+        try {
+            List<String> locations = new ArrayList<>(camelFreemarkerConfig.locations);
+            if (locations.isEmpty()) {
+                locations.add("freemarker/templates");
```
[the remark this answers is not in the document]
carlosthe19916: I've added the default location `freemarker/templates` to CamelFreemarkerConfig and also added some docs in `docs/modules/ROOT/pages/extensions/freemarker.adoc`

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerNativeImageProcessor.java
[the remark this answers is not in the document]
carlosthe19916: @ppalaga good point. I've verified that `my/path1,,my/path2` will cause just 2 locations but `my/path1, ,my/path2` (notice the blank space) will generate 3 locations and generate an exception... I've added a basic filter to validate it https://github.com/carlosthe19916/camel-quarkus/blob/7c027254d03024ec7d64aeffd040ae05da4677ff/extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerNativeImageProcessor.java#L66

I also added this case to the test here https://github.com/carlosthe19916/camel-quarkus/blob/7c027254d03024ec7d64aeffd040ae05da4677ff/integration-tests/freemarker/src/main/resources/application.properties#L20

integration-tests/freemarker/src/main/resources/application.properties
carlosthe19916: Good point @ppalaga . Files with different extensions should still work. I agree and we should include the whole subtree but just to avoid adding files we don't want to add we can use `includePatterns` and `excludePatterns` as you suggested. Let me add it

extensions/freemarker/deployment/src/main/java/org/apache/camel/quarkus/component/freemarker/deployment/FreemarkerNativeImageProcessor.java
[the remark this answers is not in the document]
ppalaga: Thanks for the research, I have learned something.

integration-tests/freemarker/src/main/resources/application.properties
carlosthe19916: @ppalaga sorry for being so slow in this. I've added `includePatterns` and `excludePatterns`

carlosthe19916:
@ppalaga I've also added a new Config property called `classModels` which should be used to pass all Java Classes that will be marked for reflection. Freemarker uses POJOs for processing the templates but if those POJOs were not marked for reflection using Quarkus, then an exception will happen... Because of this change, I've removed this line https://github.com/apache/camel-quarkus/pull/835/commits/0e2d854d776a9e656b2bfafc40826ee330710b1b#diff-d5f7db497629630202e093f916c76b9eL65 WDYT?

docs/modules/ROOT/pages/extensions/freemarker.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+=== Configuration
+
+The Camel Freemarker extension exposes the following configuration properties.
+
+[cols="80,.^10,.^10"]
+|===
+|Configuration property |Type |Default
+
+|`quarkus.camel.freemarker.locations`
```
ppalaga: Could we perhaps remove the locations option now that we have includes and excludes? The includes option seems to offer a superset of `locations` functionality. The default can be changed to `freemarker/templates/**` and moved to the includes option too. WDYT?

docs/modules/ROOT/pages/extensions/freemarker.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+ Comma-separated list of locations to scan recursively for templates.
+|List<String>
+|`freemarker/templates`
+
+|`quarkus.camel.freemarker.include-patterns`
+
+ Used for inclusive filtering scanning of Freemarker templates. By default all tree files inside 'locations' will be included.
```
ppalaga: Plase add:

* Where we scan for the templates (classpath, filesystem)
* Are `classpath:` and `filesystem:` prefixes supported? Which one is the default? (I do not insist on supporting `filesystem:`) We have other extensions documenting similar features properly, just grep for `classpath:` in the source tree.

ppalaga:
It would be cool if the extension could do most/all of the following tasks at build time:

* Create a `freemarker.template.Configuration`
* Parse/warmup the templates.

That would make it truly quarkyfied!

ppalaga:
The tests were adapted in https://github.com/apache/camel-quarkus/pull/2028 and I'd like to implement the native support using the Quarkiverse Freemarker extension. Hence closing this one.

PR #841, opened 2020-03-03
Add JMS extension

jamesnetherton:


extensions/paho/deployment/src/main/java/org/apache/camel/quarkus/component/paho/deployment/PahoProcessor.java
```diff
@@ -39,7 +41,8 @@
     private static final List<Class<?>> PAHO_REFLECTIVE_CLASSES = Arrays.asList(
             JSR47Logger.class,
             TCPNetworkModuleFactory.class,
-            SSLNetworkModuleFactory.class);
+            SSLNetworkModuleFactory.class,
+            PahoConfiguration.class);
```
lburgazzoli: Looks suspicious that we need to add `PahoConfiguration[ as it seems correct on [camel side](https://github.com/apache/camel/blob/release/3.1.0/components/camel-paho/src/main/java/org/apache/camel/component/paho/PahoComponent.java#L55-L58), wonder if this is a regression either on camel or camel-quarkus side

jamesnetherton:
@ppalaga Latest commit addresses your review comments.

PR #848, opened 2020-03-05
Fix #785 JacksonXML support

### LINKED ISSUE #785: JacksonXML support
This might work already, we are perhaps just missing a test.

ppalaga:
Fix #785 JacksonXML support

### DIFF
--- a/extensions/support/jackson-dataformat-xml/deployment/src/main/java/org/apache/camel/quarkus/support/jackson/datafromat/xml/deployment/JacksonDataformatXmlSupportProcessor.java
+++ b/extensions/support/jackson-dataformat-xml/deployment/src/main/java/org/apache/camel/quarkus/support/jackson/datafromat/xml/deployment/JacksonDataformatXmlSupportProcessor.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.support.jackson.datafromat.xml.deployment;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Set;
+import java.util.stream.Stream;
+
+import javax.xml.stream.XMLEventFactory;
+import javax.xml.stream.XMLInputFactory;
+import javax.xml.stream.XMLOutputFactory;
+
+import com.ctc.wstx.shaded.msv.org_isorelax.verifier.VerifierFactoryLoader;
+import com.ctc.wstx.shaded.msv.relaxng_datatype.DatatypeLibraryFactory;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.ObjectCodec;
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.DeploymentClassLoaderBuildItem;
+import io.quarkus.deployment.builditem.FeatureBuildItem;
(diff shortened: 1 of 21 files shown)


ppalaga:
f45e450 added the missing `<module>jackson-dataformat-xml</module>` in the support parent.

extensions/support/jackson-dataformat-xml/deployment/src/main/java/org/apache/camel/quarkus/support/jackson/datafromat/xml/deployment/JacksonDataformatXmlSupportProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+        Stream.concat(
+                Stream.of(
+                        JsonFactory.class,
+                        ObjectCodec.class,
+                        VerifierFactoryLoader.class,
+                        DatatypeLibraryFactory.class,
+                        XMLEventFactory.class,
+                        XMLInputFactory.class,
+                        XMLOutputFactory.class)
+                        .map(Class::getName),
```
lburgazzoli: maybe we should just use class names as strings

extensions/support/jackson-dataformat-xml/deployment/src/main/java/org/apache/camel/quarkus/support/jackson/datafromat/xml/deployment/JacksonDataformatXmlSupportProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+                Stream.of(
+                        XMLValidationSchemaFactory.INTERNAL_ID_SCHEMA_DTD,
+                        XMLValidationSchemaFactory.INTERNAL_ID_SCHEMA_RELAXNG,
+                        XMLValidationSchemaFactory.INTERNAL_ID_SCHEMA_W3C,
+                        XMLValidationSchemaFactory.INTERNAL_ID_SCHEMA_TREX)
+                        .map(schemaId -> XMLValidationSchemaFactory.class.getName() + "." + schemaId))
```
lburgazzoli: As there's no discovery but we deal with a fixed set of ites, maybe we can use strings so we can avoid the the two `Stream.of` and the `Stream.concat`

extensions/support/jackson-dataformat-xml/deployment/src/main/java/org/apache/camel/quarkus/support/jackson/datafromat/xml/deployment/JacksonDataformatXmlSupportProcessor.java
ppalaga: Yeah, we discussed this some time ago with the result that strings are faster and they do not have any disadvantages related to class loading. In the meantime, I asked David about this and he said he finds the perf drawbacks negligible and so I started preferring maintainability. WDYT?

extensions/support/jackson-dataformat-xml/deployment/src/main/java/org/apache/camel/quarkus/support/jackson/datafromat/xml/deployment/JacksonDataformatXmlSupportProcessor.java
ppalaga: Yes, your proposal is valid and I can see the advantages but classes are easier to type in an IDE and any typos will cause a compilation error so it feels safer. I am open to discuss this.

lburgazzoli: Yes it is not really related to perf of loading classes but more about about the `complexity` of two nested streams with mapping like looking at the code is seems much more complex that what it actually does but there's no problem to keep it as it is

extensions/support/jackson-dataformat-xml/deployment/src/main/java/org/apache/camel/quarkus/support/jackson/datafromat/xml/deployment/JacksonDataformatXmlSupportProcessor.java
lburgazzoli: I'm always +1 for maintainability you know :)

extensions/support/jackson-dataformat-xml/deployment/src/main/java/org/apache/camel/quarkus/support/jackson/datafromat/xml/deployment/JacksonDataformatXmlSupportProcessor.java
ppalaga: I am at least hanging this quote from Plato's Timaeus above my table to remind me not to sin against the beauty in the future ;)

> The good, of course, is always beautiful, and the beautiful never lacks proportion.

PR #849, opened 2020-03-05
Support for camel-xml-io

### LINKED ISSUE #847: Support for camel-xml-io
camel-quarkus XML support requires camel-xml-jaxb but in camel 3.1.0 there is a new lightweight implementation camel-xml-io

We probably need to have two extensions:

- camel-quarkus-xml-jaxb
- camel-quarkus-xml-jaxp
- camel-quarkus-xml-io

To match camel's layout

lburgazzoli:
Fixes #847


lburgazzoli:
With this PR, `camel-quarkus-support-xml` has no dependencies and the only task is to add reflective classes.

I'm not sure if this extension should be kept as it is (i.e. the name is misleading) but I think we need to have a better view about the requirements of other extensions in relation to XML so we may re-factor it when we have a better idea.

ppalaga:
A general question: are all these new `camel-xml-*` extensions supposed to be used by the app devs directly? If not, they should rather be support extensions, should they not?

lburgazzoli:
the user need to decide what xml implementation is needed, we can certainly move them to support but there is the same artifact on camel-side

ppalaga:
I am struggling to understand the purpose of camel-xml-jaxp and camel-xml-io (in Camel). camel-xml-jaxp is just an alias for camel-support and and camel-xml-io is just an alias for camel-core-engine. What is the actual reason for their existence?

oscerd:
If this is your idea probably you didn't look too deep in the camel code. They are not just alias.

gnodet:
@ppalaga we're trying to make camel a bit more modular.  In addition to that, we want to have a camel which is lightweight enough in Quarkus.  Previsouly, xml routes were supported through jaxb only.  Now, we do support loading xml routes either using JAXB (using `camel-xml-jaxb` support) or using a generated lightweight parser (using `camel-xml-io`).
In the same idea, all the xml related support (mostly things that do use the JDK XML support) has been moved to `camel-xml-jaxp`. 
The overall goal is that camel quarkus native binaries will be as light as possible, not including the xml parsers (xerces / xalan embedded in the JDK) or jaxb if not needed.

ppalaga:
Ah, sorry, indeed there is actually some code in those modules. I am fully by Camel Quarkus where we just wrap things.

ppalaga:
OK, I hope I am starting to grasp this. 

Here a couple of note about this PR (that still may reside on my false assumptions; plz correct me in such cases):

* The naming of the modules that follows the naming in Camel is a good thing.
* If camel-xml-jaxb support and camel-xml-io are there for parsing routes from XML, their pom.xml descriptions (and yaml metadata) should say so. I do not understand them as a general purpose XML parser and a general purpose JAXB package. They are there for end user apps that require parsing routes from XML. They are not there for other Camel Quarkus extensions to get a general XML support (we may need to add new/adapt existing extensions for that). Right?
* camel-xml-jaxp looks like a general purpose JAXP package to used by other camel components on the camel side and I assume it should be similar here in Camel Quarkus: camel-quarkus-xml-jaxp should be used by other extensions that need JAXP. End user apps should not depend on it directly. Right? If so, the current camel-quarkus-support-xml should be merged with camel-xml-jaxp. WDYT?

extensions/xml-io/pom.xml
```diff
@@ -26,8 +26,8 @@
         <relativePath>../../poms/build-parent/pom.xml</relativePath>
     </parent>
 
-    <artifactId>camel-quarkus-core-xml-parent</artifactId>
-    <name>Camel Quarkus :: Core XML</name>
+    <artifactId>camel-quarkus-xml-io-parent</artifactId>
+    <name>Camel Quarkus :: XML :: Io</name>
```
ppalaga: `IO` maybe better? Here and elsewhere.

extensions/xml-jaxb/deployment/pom.xml
```diff
@@ -21,13 +21,13 @@
...
-        <artifactId>camel-quarkus-core-xml-parent</artifactId>
+        <artifactId>camel-quarkus-xml-jaxb-parent</artifactId>
         <version>1.1.0-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
 
-    <artifactId>camel-quarkus-core-xml-deployment</artifactId>
-    <name>Camel Quarkus :: Core XML :: Deployment</name>
+    <artifactId>camel-quarkus-xml-jaxb-deployment</artifactId>
+    <name>Camel Quarkus :: XML :: Jaxb :: Deployment</name>
```
ppalaga: `JAXB` here and elsewhere

extensions/xml-jaxp/deployment/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-xml-jaxp-parent</artifactId>
+        <version>1.1.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-xml-jaxp-deployment</artifactId>
+    <name>Camel Quarkus :: XML :: Jaxp :: Deployment</name>
```
ppalaga: `JAXP` here and elsewhere

lburgazzoli:
>     * If camel-xml-jaxb support and camel-xml-io are there for parsing routes from XML, their pom.xml descriptions (and yaml metadata) should say so. I do not understand them as a general purpose XML parser and a general purpose JAXB package. They are there for end user apps that require parsing routes from XML. They are not there for other Camel Quarkus extensions to get a general XML support (we may need to add new/adapt existing extensions for that). Right?
> 

Not really, `camel-xml-io` contains an xml parser implementation as far as i know

>     * camel-xml-jaxp looks like a general purpose JAXP package to used by other camel components on the camel side and I assume it should be similar here in Camel Quarkus: camel-quarkus-xml-jaxp should be used by other extensions that need JAXP. End user apps should not depend on it directly. Right? If so, the current camel-quarkus-support-xml should be merged with camel-xml-jaxp. WDYT?

The artefact contains as example the [XMLTokenizeLanguage](https://github.com/apache/camel/blob/master/core/camel-xml-jaxp/src/main/java/org/apache/camel/language/xtokenizer/XMLTokenizeLanguage.java) so I think it is safe to let the user if/when to include it.

ppalaga:
> > ```
> > * If camel-xml-jaxb support and camel-xml-io are there for parsing routes from XML, their pom.xml descriptions (and yaml metadata) should say so. I do not understand them as a general purpose XML parser and a general purpose JAXB package. They are there for end user apps that require parsing routes from XML. They are not there for other Camel Quarkus extensions to get a general XML support (we may need to add new/adapt existing extensions for that). Right?
> > ```
> 
> Not really, `camel-xml-io` contains an xml parser implementation as far as i know

Yes, there is a parser implementing Camels own interface `org.apache.camel.xml.io.XmlPullParser` and no other Camel component depends on it except for the core part that parses the XML routes. So, for me this does not look like a general purpose parser ATM. Is it supposed to become one in the future?

> > ```
> > * camel-xml-jaxp looks like a general purpose JAXP package to used by other camel components on the camel side and I assume it should be similar here in Camel Quarkus: camel-quarkus-xml-jaxp should be used by other extensions that need JAXP. End user apps should not depend on it directly. Right? If so, the current camel-quarkus-support-xml should be merged with camel-xml-jaxp. WDYT?
> > ```
> 
> The artefact contains as example the [XMLTokenizeLanguage](https://github.com/apache/camel/blob/master/core/camel-xml-jaxp/src/main/java/org/apache/camel/language/xtokenizer/XMLTokenizeLanguage.java) so I think it is safe to let the user if/when to include it.

OK, so we need a separate support JAXP extension for the native quirks - so `camel-quarkus-support-xml` could be renamed to `camel-quarkus-support-jaxp`. WDYT?

lburgazzoli:
> > > ```
> > > * If camel-xml-jaxb support and camel-xml-io are there for parsing routes from XML, their pom.xml descriptions (and yaml metadata) should say so. I do not understand them as a general purpose XML parser and a general purpose JAXB package. They are there for end user apps that require parsing routes from XML. They are not there for other Camel Quarkus extensions to get a general XML support (we may need to add new/adapt existing extensions for that). Right?
> > > ```
> > 
> > 
> > Not really, `camel-xml-io` contains an xml parser implementation as far as i know
> 
> Yes, there is a parser implementing Camels own interface `org.apache.camel.xml.io.XmlPullParser` and no other Camel component depends on it except for the core part that parses the XML routes. So, for me this does not look like a general purpose parser ATM. Is it supposed to become one in the future?
> 

That is a question for @gnodet 

> > > ```
> > > * camel-xml-jaxp looks like a general purpose JAXP package to used by other camel components on the camel side and I assume it should be similar here in Camel Quarkus: camel-quarkus-xml-jaxp should be used by other extensions that need JAXP. End user apps should not depend on it directly. Right? If so, the current camel-quarkus-support-xml should be merged with camel-xml-jaxp. WDYT?
> > > ```
> > 
> > 
> > The artefact contains as example the [XMLTokenizeLanguage](https://github.com/apache/camel/blob/master/core/camel-xml-jaxp/src/main/java/org/apache/camel/language/xtokenizer/XMLTokenizeLanguage.java) so I think it is safe to let the user if/when to include it.
> 
> OK, so we need a separate support JAXP extension for the native quirks - so `camel-quarkus-support-xml` could be renamed to `camel-quarkus-support-jaxp`. WDYT?

Yes that is something I was [thinking about ](https://github.com/apache/camel-quarkus/pull/849#issuecomment-595236629)

ppalaga:
> > we need a separate support JAXP extension for the native quirks - so `camel-quarkus-support-xml` could be renamed to `camel-quarkus-support-jaxp`. WDYT?
> 
> Yes that is something I was [thinking about](https://github.com/apache/camel-quarkus/pull/849#issuecomment-595236629)

Perfect. Please file a followup unless you want to amend now.

lburgazzoli:
> > > we need a separate support JAXP extension for the native quirks - so `camel-quarkus-support-xml` could be renamed to `camel-quarkus-support-jaxp`. WDYT?
> > 
> > 
> > Yes that is something I was [thinking about](https://github.com/apache/camel-quarkus/pull/849#issuecomment-595236629)
> 
> Perfect. Please file a followup unless you want to amend now.

Will open a follow up issue as soon this get merged so I have the right pointers

ppalaga:
The last open issue from my PoV is to document the purpose of the new extensions in their runtime/pom.xml `<description>`s and yaml metadata. Users should get an idea in which situations they need the dependency.

lburgazzoli:
should we fix it at camel side ? those extensions here are mainly wrapper

ppalaga:
> should we fix it at camel side ?

Ideally both.

lburgazzoli:
I think we should avoid to duplicate the effort and document how the xml thing wotk on camel side, we can link the page here but I’d avoid duplication

extensions/xml-io/runtime/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-xml-io-parent</artifactId>
+        <version>1.1.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-xml-io</artifactId>
+    <name>Camel Quarkus :: XML :: IO :: Runtime</name>
+    <description>Includes implementations of Camel's fast and lightweight XML parse</description>
```
ppalaga:
```suggestion
    <description>An XML stack for parsing XML route definitions. A fast an light weight alternative to camel-quarkus-xml-jaxp</description>
```

extensions/xml-jaxb/runtime/pom.xml
```diff
@@ -21,17 +21,17 @@
...
         <version>1.1.0-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
 
-    <artifactId>camel-quarkus-core-xml</artifactId>
-    <name>Camel Quarkus :: Core XML :: Runtime</name>
-    <description>Includes implementations of Java Architecture for XML Binding (JAXB) and Java API for XML Processing (JAXP)</description>
+    <artifactId>camel-quarkus-xml-jaxb</artifactId>
+    <name>Camel Quarkus :: XML :: JAXB :: Runtime</name>
+    <description>Includes implementations of Java Architecture for XML Binding (JAXB)</description>
```
ppalaga:
```suggestion
        <description>An XML stack for parsing XML route definitions. A legacy alternative to the fast an light weight camel-quarkus-xml-io</description>

```

extensions/xml-io/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -16,8 +16,8 @@
 #
 
 ---
-name: "Camel Quarkus XML Common"
-description: "The Camel Quarkus code XML module"
+name: "Camel Quarkus XML IO"
+description: "The Camel Quarkus XML IO module"
```
ppalaga:
```suggestion
description: "An XML stack for parsing XML route definitions. A fast an light weight alternative to camel-quarkus-xml-jaxp"
```

extensions/xml-jaxb/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+name: "Camel Quarkus XML Jaxb"
```
ppalaga:
```suggestion
name: "Camel Quarkus XML JAXB"
```

extensions/xml-jaxb/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,28 @@
...
+description: "The Camel Quarkus XML Jaxb module"
```
ppalaga:
```suggestion
description: "An XML stack for parsing XML route definitions. A legacy alternative to the fast an light weight camel-quarkus-xml-io"
```

extensions/xml-jaxp/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+name: "Camel Quarkus XML Jaxp"
```
ppalaga:
```suggestion
name: "Camel Quarkus XML JAXP"
```

gnodet:
> > > ```
> > > * If camel-xml-jaxb support and camel-xml-io are there for parsing routes from XML, their pom.xml descriptions (and yaml metadata) should say so. I do not understand them as a general purpose XML parser and a general purpose JAXB package. They are there for end user apps that require parsing routes from XML. They are not there for other Camel Quarkus extensions to get a general XML support (we may need to add new/adapt existing extensions for that). Right?
> > > ```
> > 
> > 
> > Not really, `camel-xml-io` contains an xml parser implementation as far as i know
> 
> Yes, there is a parser implementing Camels own interface `org.apache.camel.xml.io.XmlPullParser` and no other Camel component depends on it except for the core part that parses the XML routes. So, for me this does not look like a general purpose parser ATM. Is it supposed to become one in the future?
> 

The goal is not to make the parser public. If other parts/components need a lightweight parser, why not, but it should be part of a public API for users.
The only goal was to avoid having the dependency on the whole JAXP stack, but if users actually need to process XML data, it's safer to get the usual public API/ implementations.

extensions/xml-io/runtime/pom.xml
gnodet: I would not mention "XML stack", as this should not be made public.  (and the parser is very limited, so _stack_ is a bit too much for it).

PR #860, opened 2020-03-09
Add ActiveMQ extension

### LINKED ISSUE #424: Add support for camel-activemq
These are heavily used in integrations.

jamesnetherton:
Fixes #424

### DIFF
--- a/extensions/activemq/runtime/src/main/resources/META-INF/quarkus-extension.yaml
+++ b/extensions/activemq/runtime/src/main/resources/META-INF/quarkus-extension.yaml
@@ -1,0 +1,28 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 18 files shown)


lburgazzoli:
Wonder if it could make sense to put this among the messaging it

jamesnetherton:
Messaging has JMS 2 on the classpath so it potentially leads to false positives when testing. Hence I chose to separate into its own itest module.

extensions/activemq/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+#
+
+---
+name: "Camel Quarkus ActiveMQ"
+description: "Camel Quarkus ActiveMQ Support"
+metadata:
+  keywords:
+    - "camel"
+    - "ActiveMQ"
```
ppalaga: I think used lower case quite consequently so far.

jamesnetherton: I'll change it in a follow up commit. AFAIK, the keyword search on code.quarkus.io is case insensitive, so it'll still work fine with a mixture of upper & lower case.

PR #863, opened 2020-03-10
Fix #795 File Watch support

### LINKED ISSUE #795: File Watch support

ppalaga:


docs/modules/ROOT/pages/extensions/file-watch.adoc
```diff
@@ -0,0 +1,30 @@
+[[netty]]
```
jamesnetherton: Not netty :slightly_smiling_face:

ppalaga: +1, thanks

ppalaga:
a00ca02 fixes the issue found by @jamesnetherton

jamesnetherton:
One question before approving - what's the difference between the `mac` and `non-mac` profile? I didn't see anything obvious aside from the comment about exclusions.

ppalaga:
> what's the difference between the `mac` and `non-mac` profile?

Good catch, it is a leftover from my earlier experiments where I tried to exclude JNA for non-mac but it did not work well in the native mode.

Removed the profiles in 929889d.

ppalaga:
f83e7b8 removed the profiles correctly.

PR #864, opened 2020-03-10
Add GraphQL extension

### LINKED ISSUE #801: GraphQL support

jamesnetherton:
Fixes #801

### DIFF
--- a/extensions/graphql/deployment/src/main/java/org/apache/camel/quarkus/component/graphql/deployment/GraphQLProcessor.java
+++ b/extensions/graphql/deployment/src/main/java/org/apache/camel/quarkus/component/graphql/deployment/GraphQLProcessor.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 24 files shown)


extensions/graphql/deployment/src/main/java/org/apache/camel/quarkus/component/graphql/deployment/GraphQLProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+    private static final String FEATURE = "camel-graphql";
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    ExtensionSslNativeSupportBuildItem activateSslNativeSupport() {
+        return new ExtensionSslNativeSupportBuildItem(FEATURE);
```
ppalaga: Is SSL really required for everybody? I think we agreed with @lburgazzoli that we should mention this in the extension doc page and let the users decide whether they need it or not. It is similar with encodings and time zones.

extensions/graphql/runtime/src/main/java/org/apache/camel/quarkus/component/graphql/CamelGraphQLConfig.java
```diff
@@ -0,0 +1,39 @@
...
+@ConfigRoot(name = "camel.graphql", phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
+public class CamelGraphQLConfig {
+
+    /**
+     * A comma separated list of paths to files containing GraphQL queries for use by GraphQL endpoints.
+     *
+     * Query files that only need to be accessible from the classpath should be specified on this property.
+     *
+     * Paths can either be schemeless (E.g graphql/my-query.graphql) or be prefixed with the classpath: URI scheme (E.g
+     * classpath:graphql/my-query.graphql). Other URI schemes are not supported.
```
ppalaga: I appreciate the well written the doc text!

extensions/graphql/deployment/src/main/java/org/apache/camel/quarkus/component/graphql/deployment/GraphQLProcessor.java
ppalaga: I mean users should configure enable those things via their application.properties

jamesnetherton: In this case, there's stuff in `HttpClientBuilder` that wants to configure an `SSLContext`. Not sure if I should be looking to suppress that stuff (probably not).

PR #876, opened 2020-03-11
Fixed kafka itests as @Inject is not supported in native tests

aldettinger:


integration-tests/kafka/src/test/java/org/apache/camel/quarkus/component/kafka/it/CamelKafkaTestResource.java
```diff
@@ -32,14 +31,6 @@
     private static final String CONFLUENT_PLATFORM_VERSION = "5.3.1";
 
     private KafkaContainer container;
-    private CamelMain main;
-
-    @Override
-    public void inject(Object testInstance) {
-        if (testInstance instanceof CamelKafkaTest) {
-            this.main = ((CamelKafkaTest) testInstance).main;
```
aldettinger: Note that this line would have been invoked in native mode only due to distinct class loaders. But then @Inject is not authorized.

lburgazzoli: is that something documented on quarkus side ?

aldettinger: Honestly, I don't know but I could take a look.

lburgazzoli: yes please as this was working before and it has been used to avoid having warnings at the end of the tests

jamesnetherton: It was a change introduced in 1.3.0.CR1. See https://github.com/quarkusio/quarkus/commit/2cea862dbab1f9133ff5e0d7d9cdbef11e2d1e27.

I had to make a similar change for the Infinispan tests.

lburgazzoli: it looks strange as the pr is about the @Inject annotation

aldettinger: There must be 2 subjects:

\@Inject not supported in native mode is documented [there](https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/getting-started-testing.adoc#9-native-executable-testing), so it could have behaved differently previously but it's no more the case.

`testInstance instanceof CamelKafkaTest == false` because of distinct class loaders, I found no mention, so I would better open a question on quarkus side so that we know what behavior we can rely on.

lburgazzoli: oh now I realized what the issue, sorry for taking so much time. so this change is ok, there may ne something similar elsewhere

aldettinger: For the record, question on [quarkus side](https://github.com/quarkusio/quarkus/issues/7784). ok, I'm gonna check if I can find similar thing elsewhere in the context of another ticket.

PR #898, opened 2020-03-17
Fix #897 List the JVM-only extensions in the docs

### LINKED ISSUE #897: List the JVM-only extensions in the docs
A followup of https://github.com/apache/camel-quarkus/pull/895

ppalaga:
Fix #897


ppalaga:
d51d679 fixed the failing test

ppalaga:
e4543eb fixed the license header issues

ppalaga:
> We should not move the functionalities of the package-maven-plugin outside of the camel-quarkus repository as it will introduces yet another external dependency not under direct control of Apache Camel commiters.

It is true that not every Apache Camel committer has access, but OTOH, there are more than one Apache Camel committer that do have write access there.

The motivation was threefold: 

* The packaging mojos share some code with the create-extension and code formatting mojos.
* Having a plugin hosted in the same source tree where it is applied, is not ideal from the built performance PoV.
* When the mojos are hosted in a separate repo, they can be run without building them first, which makes the process easier esp. for the newcomers.

Moving the mojos to a separate repo under the ASF umbrella would mean loosing the flexibility of releasing on demand. My understanding is that the voting process cannot be relaxed even for tools like these, right?

So if you insist on your veto, I am going to move the changes back to camel-quarkus.

> In addition it creates a dependency issue as to update to a new version of Apache Camel would require to update the plugin first.

This is not an issue, as long as camel-catalog stays backwards compatible, because it is managed on the Camel Quarkus side https://github.com/apache/camel-quarkus/pull/898/files#diff-600376dffeb79835ede4a0b285078036R239-R243

lburgazzoli:
I think it is fine to have the extension generator outside the camel-quarkus repo as it is an optional thing but for the catalog generator and other functionalities provided by the package maven plugin it would be bad move so IMHO we should be keep it as part of the camel-quarkus repo.

@davsclaus @gnodet @oscerd what don you think ?

davsclaus:
Yes the quarkus camel catalog should be here, this is what we do in other projects too.

ppalaga:
00e3c63 brings the mojos back to the Camel Quarkus source tree

ppalaga:
d93f2be should fix the test failure.

ppalaga:
0f9eef7 fixed the create mojo config

PR #903, opened 2020-03-18
POM enhancements

ppalaga:


pom.xml
```diff
@@ -84,8 +84,7 @@
         <maven-source-plugin.version>3.1.0</maven-source-plugin.version>
         <maven-assembly-plugin.version>3.1.1</maven-assembly-plugin.version>
         <maven-dependency-plugin.version>3.1.1</maven-dependency-plugin.version>
-        <!-- NOTE: We pin to this version due to https://github.com/apache/camel-quarkus/issues/723 -->
-        <maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
+        <maven-surefire-plugin.version>3.0.0-M4</maven-surefire-plugin.version>
```
jamesnetherton: I'm not convinced all of the issues are fixed.

Run `NoDefaultServletTest` (You'll need to remove `@Disabled`) and the test passes with 3.x. Run it with 2.x and it'll fail.

ppalaga: That would mean that https://github.com/quarkusio/quarkus/issues/7203 is not fully fixed? Let me remove the commit then.

ppalaga:
149d0da removed the revert to maven-surefire-plugin.version 3.0.0-M4

Do we want to wait for the CI?

ppalaga:
> Do we want to wait for the CI?

Actually, we should, because I also rebased.

ppalaga:
Approved by @jamesnetherton

PR #905, opened 2020-03-18
Add Google Mail extension

### LINKED ISSUE #807: Google Mail support

jamesnetherton:
Fixes #807

### DIFF
--- a/integration-tests/google/src/main/resources/application.properties
+++ b/integration-tests/google/src/main/resources/application.properties
@@ -1,0 +1,22 @@
+## ---------------------------------------------------------------------------
+## Licensed to the Apache Software Foundation (ASF) under one or more
+## contributor license agreements.  See the NOTICE file distributed with
+## this work for additional information regarding copyright ownership.
+## The ASF licenses this file to You under the Apache License, Version 2.0
+## (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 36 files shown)


integration-tests/google/src/main/resources/application.properties
```diff
@@ -0,0 +1,25 @@
...
+##
+## Unless required by applicable law or agreed to in writing, software
+## distributed under the License is distributed on an "AS IS" BASIS,
+## WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+## See the License for the specific language governing permissions and
+## limitations under the License.
+## ---------------------------------------------------------------------------
+
+# Quarkus
+quarkus.native.enable-https-url-handler=true
```
ppalaga: If this is required for the regular case to work, then I vote for doing it from within the extension via `new NativeImageSystemPropertyBuildItem("quarkus.ssl.native", "true)` so that the users do not need to bother.

jamesnetherton: Yep - let me do that :+1: 

Actually, I think we can do likewise on a couple of other extensions. I'll check once we get this one merged.

jamesnetherton:
Latest changes remove the need for `quarkus.native.enable-https-url-handler` to be set.

PR #930, opened 2020-03-20
OpenStack support (JVM only)

jamesnetherton:


jamesnetherton:
> This is acceptable.
> 
> I wonder how much we should take care for the following when adding JVM only extensions:
> 
> 1. Creating support extensions for stuff like okhttp (I recently added two extensions who depended on it).
> 2. Adding dependencies on existing support or quarkus extensions (like jackson in this particular case).

Yes, perhaps we should do that. You could make a similar case for mongo-gridfs, where it should really reuse the Quarkus Mongo extension.

PR #957, opened 2020-03-23
Fix #936 Warn if JVM only extensions are used in native mode

### LINKED ISSUE #936: Warn if JVM only extensions are used in native mode

ppalaga:
Fix #936


ppalaga:
Looks like the Ci failed on a broken cache. Let me run the JDBC, MongoDB and SQL locally and merge in case they pass.

PR #962, opened 2020-03-23
AWS SWF and SDB

### LINKED ISSUE #958: AWS Simple Workflow support (JVM only)

ppalaga:
Fix #958 #959


ppalaga:
The failure is caused by `[warning]connect ETIMEDOUT 13.107.42.16:443` when loading the cache. All new extensions are JVM only, and the failed job does not cover them at all. Thus merging.

ppalaga:
Reported about the cache problems: https://github.com/actions/cache/issues/141#issuecomment-603083649

PR #968, opened 2020-03-24
Add AMQP extension

### LINKED ISSUE #143: Create a Camel AMQP extension

jamesnetherton:
Fixes #143

### DIFF
--- a/extensions/amqp/pom.xml
+++ b/extensions/amqp/pom.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one or more
+    contributor license agreements.  See the NOTICE file distributed with
+    this work for additional information regarding copyright ownership.
+    The ASF licenses this file to You under the Apache License, Version 2.0
+    (the "License"); you may not use this file except in compliance with
+    the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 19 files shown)


extensions/amqp/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+                <scope>import</scope>
+            </dependency>
+            <!-- quarkus-qpid-jms-deployment is currently not part quarkus-qpid-jms-deployment-->
+            <dependency>
+                <groupId>org.amqphub.quarkus</groupId>
+                <artifactId>quarkus-qpid-jms-deployment</artifactId>
+                <version>${quarkus.qpid.jms.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
ppalaga: I'd expect these to be in the deployment module. Is there any reason for them to be here in the parent?

ppalaga: By importing the deployment BOMs here in the parent, we awake an impression that it is legal to use the deployment artifacts in all descendants, incl. the runtime module, which I think really should not happen.

jamesnetherton: I'll fix this.

I also noticed some other extensions where this happens. I'll follow up with another PR to fix tomorrow.

ppalaga: I have some dep. management cleanup in progress. I'd fix all these things there, if you do not mind.

PR #972, opened 2020-03-25
Improve the documentations for camel-quarkus under user guide

AemieJ:
* Within the camel-quarkus documentation, there exists an **Examples** page within the user guide. This page only states to go through the source examples after reading the **First Steps** chapter. I think that the use of a separate page for **Examples** is unnecessary and can be appended to the **First Steps** page at the end of the documentation itself.

### DIFF
--- a/docs/modules/ROOT/pages/examples.adoc
+++ b/docs/modules/ROOT/pages/examples.adoc
@@ -1,4 +1,0 @@
-= Camel Quarkus Examples
-
-We offer several examples in our https://github.com/apache/camel-quarkus/tree/master/examples[source tree]. To learn
-how to use them, please follow the xref:first-steps.adoc[First steps] chapter of the User guide.
--- a/docs/modules/ROOT/pages/first-steps.adoc
+++ b/docs/modules/ROOT/pages/first-steps.adoc
@@ -198,3 +198,8 @@ That's under 35 MB of RAM!
 TIP: https://quarkus.io/guides/building-native-image-guide.html[Quarkus Native executable quide] contains more details
 including
 https://quarkus.io/guides/building-native-image-guide.html#creating-a-container[steps for creating a container image].
+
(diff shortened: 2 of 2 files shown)


ppalaga:
@AemieJ thanks for reaching out!

I think we wanted to have an autogenerated list of examples on that page. Not sure we want to remove it. What do others think?

AemieJ:
If we do want to provide the list on the examples, then we could keep the examples page. But if not, then the separate page for examples is unnecessary.

AemieJ:
> No need for the split, we would like to have a list of examples

So within the examples page, you would like a list of examples which could give the basic gist on the particular example and a direct link to the example?

oscerd:
> > No need for the split, we would like to have a list of examples
> 
> So within the examples page, you would like a list of examples which could give the basic gist on the particular example and a direct link to the example?

Something like what we have in Camel examples: https://github.com/apache/camel-examples/blob/master/examples/README.adoc

It's autogenerated.

AemieJ:
> > > No need for the split, we would like to have a list of examples
> > 
> > 
> > So within the examples page, you would like a list of examples that could give the basic gist on the particular example and a direct link to the example?
> 
> Something like what we have in Camel examples: https://github.com/apache/camel-examples/blob/master/examples/README.adoc
> 
> It's autogenerated.

The entire list is autogenerated? How is the category decided than if it isn't manually typed?

AemieJ:
> > > No need for the split, we would like to have a list of examples
> > 
> > 
> > So within the examples page, you would like a list of examples which could give the basic gist on the particular example and a direct link to the example?
> 
> Something like what we have in Camel examples: https://github.com/apache/camel-examples/blob/master/examples/README.adoc
> 
> It's autogenerated.

Besides the category, the list can be created similar to the apache camel examples.

PR #973, opened 2020-03-25
Clean up poms: introduce test BOM, manage only where necessary

ppalaga:


integration-tests/aws/pom.xml
```diff
@@ -39,6 +39,18 @@
...
+        <dependencies>
+            <dependency>
+                <groupId>org.apache.camel.quarkus</groupId>
+                <artifactId>camel-quarkus-bom-test</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
lburgazzoli: can't this be moved one level up to camel-quarkus-integration-tests ?

ppalaga: I thought of that option and I think that this way is better: We now have itests also under `extensions-jvm/*/integration-test`. Those will be easier to move to `integration-tests` when all tests are self-contained.

ppalaga:
197f65e rebased

PR #974, opened 2020-03-25
Protobuf and gRPC

### LINKED ISSUE #964: Protobuf dataformat support (JVM only)

ppalaga:
Fix #964 #965


ppalaga:
The style is correct, I checked locally. All JVM tasks are passing, thus merging.

PR #975, opened 2020-03-25
Speedup the Validate PR Style GH workflow by invoking only mvn validate

ppalaga:


ppalaga:
dea4ce4 more tweaks to make it work

PR #991, opened 2020-03-27
Make a tabular list of examples for the examples page

AemieJ:
* As the example page doesn't include the list of examples in the tabular form, I have made changes to the examples page to include the tabular form.

* However, the link to each example is to the `README.adoc` within GitHub itself, as there is no separate page for each example. Also, there doesn't exist a `README.adoc` for **timer-log-kotlin**, it direct to the GitHub repo for it.


ppalaga:
Thanks @AemieJ !

PR #998, opened 2020-03-30
Tika support

JiriOndrusek:
Fix: #799

PR contains also  change in xalan support - adds SAX capabilities (by extension of SAXTransformerFactory)


JiriOndrusek:
@ppalaga Current states is, that tika parser "works" in jvm and native, BUT

- Tike parser obviously has some issues in quarkus. Not every parser is working. There is a list of not native ready parsers hardcoded in git - https://github.com/quarkusio/quarkus/blob/master/extensions/tika/deployment/src/main/java/io/quarkus/tika/deployment/TikaProcessor.java#L39
- Even some of the parsers which should work, are not working. For example image parser - I've reported an issue about it https://github.com/quarkusio/quarkus/issues/8375 
- following change has to be present in camel https://github.com/apache/camel/pull/3705 (already merged) I'll prepare PR for integration branch once the branch is working with current camel

Do you see any direction what to do with this issue until integration branch is working?

.github/workflows/pr-build.yaml
```diff
@@ -154,11 +154,12 @@ jobs:
               braintree
               compression
               graphql
-              mustache
+              mustache
```
ppalaga, line 157: Plz. remove the tailing whitespace

extensions/tika/deployment/src/main/java/org/apache/camel/quarkus/component/tika/deployment/TikaProcessor.java
```diff
@@ -0,0 +1,59 @@
...
+    private static final Logger LOG = Logger.getLogger(TikaProcessor.class);
+    private static final String FEATURE = "camel-tika";
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    /*
+     * The bean-validator component is programmatically configured by the extension thus
```
ppalaga, line 43:
```suggestion
     * The tika component is programmatically configured by the extension thus
```

extensions/tika/deployment/src/main/java/org/apache/camel/quarkus/component/tika/deployment/TikaProcessor.java
```diff
@@ -0,0 +1,59 @@
...
+    CamelServiceFilterBuildItem serviceFilter() {
+        return new CamelServiceFilterBuildItem(CamelServiceFilter.forComponent("tika"));
+    }
+
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep
+    CamelRuntimeBeanBuildItem tikaComponent(BeanContainerBuildItem beanContainer, TikaRecorder recorder) {
+        return new CamelRuntimeBeanBuildItem(
+                "tika",
+                TikaRecorder.class.getName(),
```
ppalaga, line 56: `TikaRecorder` looks strange. `TikaComponent` maybe?

integration-tests/tika/src/main/java/org/apache/camel/quarkus/component/tika/it/TikaResource.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+// import org.apache.camel.ProducerTemplate;
```
ppalaga, line 30: Plz remove the commented code.

integration-tests/tika/src/test/java/org/apache/camel/quarkus/component/tika/it/TikaTest.java
```diff
@@ -0,0 +1,119 @@
...
+            UniversalEncodingDetector encodingDetector = new UniversalEncodingDetector();
+            detectedCharset = encodingDetector.detect(bodyIs, new Metadata());
+        } catch (IOException e1) {
+            Assertions.fail();
+        }
+
+        Assertions.assertTrue(detectedCharset.name().startsWith(Charset.defaultCharset().name()));
+    }
+
+    //    @Test
```
ppalaga, line 78: Maybe better this?
```suggestion
    @Test
    @Disabled("https://github.com/quarkusio/quarkus/issues/8375")
```

poms/bom-deployment/pom.xml
```diff
@@ -34,6 +34,7 @@
 
     <properties>
         <camel-quarkus.version>1.1.0-SNAPSHOT</camel-quarkus.version><!-- kept in sync with project.version by the release plugin -->
+        <quarkus-version>1.3.0.Final</quarkus-version>
```
ppalaga, line 37: `quarkus.version` is defined in the top pom so I think this one is not needed?

poms/bom-deployment/pom.xml
```diff
@@ -822,6 +823,16 @@
...
+            <dependency>
+                <groupId>org.apache.camel.quarkus</groupId>
+                <artifactId>camel-quarkus-tika-deployment</artifactId>
+                <version>${camel-quarkus.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-tika-deployment</artifactId>
+                <version>${quarkus.version}</version>
+            </dependency>
```
ppalaga, line 835: Plz move this one to the top where the mysql driver is and add a link to the quarkus PR where it is fixed in their BOM.

ppalaga:
And please rebase and squash your commits.

ppalaga:
You can add the missing license headers by running `mvn process-resources -Pformat`
The order of imports can be fixed by re-compiling the failing modules.

JiriOndrusek:
@Ppalaga all suggestions are applied, doc created, rebased to integration branch, squashed

ppalaga:
@JiriOndrusek Could you please check the failing test?

JiriOndrusek:
@ppalaga I'll look into it

extensions-support/xalan/runtime/src/main/java/org/apache/camel/quarkus/support/xalan/XalanTransformerFactory.java
```diff
@@ -110,4 +115,42 @@ public void setErrorListener(ErrorListener listener) {
     public ErrorListener getErrorListener() {
         return delegate.getErrorListener();
     }
+
+    private SAXTransformerFactory delegateAsSAXTransformerFactory() {
+        if (delegate instanceof SAXTransformerFactory) {
+            return (SAXTransformerFactory) delegate;
+        }
+        throw new IllegalArgumentException("Unsupported TransformerFactory feature " + SAXTransformerFactory.FEATURE);
```
ppalaga, line 123: I think it would be more effective to move this check to the constructor and change the type of the delegate field to SAXTransformerFactory thus eliminating the `delegateAsSAXTransformerFactory()` method. Given that we call `TransformerFactory.newInstance( "org.apache.xalan.xsltc.trax.TransformerFactoryImpl", ...)` we are quite safe to always get a SAXTransformerFactory

extensions/tika/runtime/src/main/doc/limitations.adoc
```diff
@@ -0,0 +1,20 @@
+Parameters `tikaConfig` and `tikaConfigUri` are not available in quarkus camel tika extension. Configuration
+can be changed only via `application.properties`.
+
+While you can use any of the available Tika parsers in JVM mode (https://tika.apache.org/[Apache Tika]),
+only several Tika parses are supported in native mode. See https://quarkus.io/guides/tika[QUARKUS - USING APACHE TIKA].
```
ppalaga, line 5:
```suggestion
While you can use any of the available https://tika.apache.org/1.24.1/formats.html[Tika parsers] in JVM mode,
only some of those are supported in native mode - see the https://quarkus.io/guides/tika[Quarkus Tika guide].
```

extensions/tika/runtime/src/main/doc/limitations.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+* `quarkus.tika.parsers`
+* optionally `quarkus.tika.parser.*`
+
+Example of `application.properties` follows :
+[source,properties]
+----
+quarkus.tika.parsers = pdf,odf,office
+quarkus.tika.parser.office = org.apache.tika.parser.microsoft.OfficeParser
+----
```
ppalaga, line 17: My understanding is that using `quarkus.tika.parsers` is recommended (not required) to reduce the application memory and native executable sizes. So we should perhaps also formulate it like that.

It is not clear to me whether each abbreviation used in `quarkus.tika.parsers` needs to be defined in `quarkus.tika.parser.my-abbrev` ? Or are there some well known abbreviations?

extensions/tika/runtime/src/main/doc/limitations.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+For more information about selecting parsers see https://quarkus.io/guides/tika[QUARKUS - USING APACHE TIKA]
```
ppalaga, line 19:
```suggestion
For more information about selecting parsers see https://quarkus.io/guides/tika[Quarkus Tika guide]
```

extensions/tika/runtime/src/main/java/org/apache/camel/quarkus/component/tika/TikaRecorder.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.tika.parser.Parser;
+
+@Recorder
+public class TikaRecorder {
+
+    public RuntimeValue<TikaComponent> createTikaComponent(BeanContainer container) {
+        return new RuntimeValue<>(new QuarkusTikaComponent(container.instance(TikaParserProducer.class)));
+    }
+
+    @org.apache.camel.spi.annotations.Component("tika")
```
ppalaga, line 53: Is this annotation required given that we produce a named bean above? https://github.com/apache/camel-quarkus/pull/998/files#diff-6af9bcae1d2af7449582ad99e6bdac3cR54

extensions/tika/runtime/src/main/java/org/apache/camel/quarkus/component/tika/TikaRecorder.java
```diff
@@ -0,0 +1,102 @@
...
+    static class QuarkusTikaComponent extends TikaComponent {
+
+        private final TikaParserProducer tikaParserProducer;
+
+        public QuarkusTikaComponent(TikaParserProducer tikaParserProducer) {
```
ppalaga, line 58: This is acceptable for now, but as a next step, could we perhaps define some sort of TikaParserProducer interface in Camel and have it there in the Camel Tika component and producer, so that we do not have to subclass here?

extensions/tika/runtime/src/main/java/org/apache/camel/quarkus/component/tika/TikaRecorder.java
JiriOndrusek: @ppalaga It would make sense to get rid of this annotation. 

But  without it, execution fails for `odf` parser with:

>Caused by: java.lang.LinkageError: loader constraint violation: loader (instance of <bootloader>) previously initiated loading for a different type with name "org/w3c/dom/Node"

I'm not sure about the reason of this behavior. There is a one related comment: https://github.com/quarkusio/quarkus/issues/8375#issuecomment-610309015

JiriOndrusek:
@ppalaga I've fixed Xalan and documentation. 

- Removal of annotation doesn't work - I'll try to find an explanation. 

- The last suggested would need a change in camel-tika. I've created an issue for this - https://issues.apache.org/jira/browse/CAMEL-15207

JiriOndrusek:
closes #799

extensions/tika/runtime/src/main/java/org/apache/camel/quarkus/component/tika/TikaRecorder.java
ppalaga: Strange, that looks like a symptom of some class loading nastyness. Eithere `org/w3c/dom/Node` is coming from two jars or it is loaded by two different class loaders.

integration-tests/tika/src/test/java/org/apache/camel/quarkus/component/tika/it/TikaTest.java
```diff
@@ -0,0 +1,158 @@
...
+        Charset detectedCharset = null;
+        try {
+            InputStream bodyIs = new ByteArrayInputStream(body.getBytes(StandardCharsets.UTF_16));
+            UniversalEncodingDetector encodingDetector = new UniversalEncodingDetector();
+            detectedCharset = encodingDetector.detect(bodyIs, new Metadata());
+        } catch (IOException e1) {
+            Assertions.fail();
+        }
+
+        Assertions.assertTrue(detectedCharset.name().startsWith(StandardCharsets.UTF_16.name()));
```
ppalaga, line 63: I wonder why do we need to test UniversalEncodingDetector here? It does not seem to be testing any Camel Quarkus code.

integration-tests/tika/src/test/java/org/apache/camel/quarkus/component/tika/it/TikaTest.java
```diff
@@ -0,0 +1,158 @@
...
```
ppalaga, line 79: Same as above: do we need this?

JiriOndrusek: You are right. This is covered by camel itself (https://github.com/apache/camel/blob/master/components/camel-tika/src/test/java/org/apache/camel/component/tika/TikaParseTest.java#L71). I'll remove both parts.

JiriOndrusek:
Both unnecessary asserts are removed.

JiriOndrusek:
@ppalaga  I've  discovered the reason of mandatory annotation.
As you can see from https://github.com/apache/camel/blob/master/core/camel-support/src/main/java/org/apache/camel/support/DefaultComponent.java#L410, if no annotation is present, then configurer is not used therefore camel throws an error, that not all of the properties are used. As I've cited linkage error in previous comment, it was my fault, it was not connected to this issue.

PR #1009, opened 2020-04-01
Fix #706 Improve RuntimeCatalogConfig docs

### LINKED ISSUE #706: Improve RuntimeCatalogConfig docs
https://issues.apache.org/jira/browse/CAMEL-14545 asks for providing a page that explains what is Camel Catalog and what it is good for. Once that page is ready a link to it should be added to RuntimeCatalogConfig.

This is a followup of https://github.com/apache/camel-quarkus/pull/704

ppalaga:
Fix #706

### DIFF
--- a/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
+++ b/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
@@ -39,7 +39,7 @@ public class CamelConfig {
     public ServiceConfig service;
 
     /**
-     * Build time configuration options for {@link RuntimeCatalogConfig}.
+     * Build time configuration options for {@link org.apache.camel.catalog.RuntimeCamelCatalog}.
      */
     @ConfigItem
     public RuntimeCatalogConfig runtimeCatalog;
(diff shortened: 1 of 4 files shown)


extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
```diff
@@ -290,41 +290,45 @@
     @ConfigGroup
     public static class RuntimeCatalogConfig {
         /**
-         * Used to control the resolution of components catalog info.
+         * Whether the {@code json} resources under {@code META-INF/services/org/apache/camel/component} should be
```
lburgazzoli, line 293: the json is not located at  `META-INF/services/org/apache/camel/component` but the catalog **may** uses service files to determine the class name of the component as the json is usually located in the same package as the component class.

Same for the other descriptions

ppalaga: Thanks, I have changed the text in 9affaa1 . Could you please re-review?

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelRuntimeCatalog.java
```diff
@@ -28,7 +28,8 @@ public CamelRuntimeCatalog(CamelConfig.RuntimeCatalogConfig config) {
     @Override
     public String modelJSonSchema(String name) {
         if (!config.models) {
-            return null;
+            throw new RuntimeException(
```
lburgazzoli, line 31: We need to check if throwing an exception is the right thing, I do not remember if there is any assumption in the code path that invokes this method

ppalaga: I am a troll, I have put it there to see who and how is actually using it :) I have fixed the tests in fc0bdf6. Are you using it in Camel K?

PR #1012, opened 2020-04-01
Workaround for the platform http not be able to handle matchOnUriPrefix

lburgazzoli:

### DIFF
--- a/extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/PlatformHttpRecorder.java
+++ b/extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/PlatformHttpRecorder.java
@@ -17,12 +17,14 @@
 package org.apache.camel.quarkus.component.platform.http.runtime;
 
 import java.util.List;
+import java.util.Map;
 
 import io.quarkus.runtime.RuntimeValue;
 import io.quarkus.runtime.annotations.Recorder;
 import io.vertx.core.Handler;
 import io.vertx.ext.web.Router;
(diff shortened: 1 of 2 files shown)


extensions/platform-http/runtime/src/main/java/org/apache/camel/quarkus/component/platform/http/runtime/PlatformHttpRecorder.java
```diff
@@ -37,7 +39,20 @@
...
     public RuntimeValue<PlatformHttpComponent> createComponent(RuntimeValue<PlatformHttpEngine> engine) {
-        PlatformHttpComponent component = new PlatformHttpComponent();
+        PlatformHttpComponent component = new PlatformHttpComponent() {
+            @Override
+            protected Endpoint createEndpoint(String uri, String remaining, Map<String, Object> parameters) throws Exception {
+                // remove matchOnUriPrefix as it will be fixed by camel 3.2 but will cause the context
+                // to fail as the property cannot be bound to the enpoint.
+                //
+                // TODO: remove once migrating to camel 3.2
+                parameters.remove("matchOnUriPrefix");
```
ppalaga, line 49: Shouldn't we warn at least that matchOnUriPrefix is ignored?

lburgazzoli: we could but actually the property is only added by platform-http component itself (when generating rest endpoints) even if it is not a property it  is able to handles so basically it would warn itself

lburgazzoli:
failure on java 11 tests seems not related to the PR

Issue #989, opened 2020-03-26
Couchdb native support for Producer Mode


aldettinger:
I'm gonna try this one.

ppalaga:
Thanks @aldettinger! You are probably the first one porting a JVM extension to native. It would be nice if you could start listing the necessary steps on a new doc page.

aldettinger:
good idea, I will take some notes.

aldettinger:
Follow up issue to list necessary steps to promote a JVM only extension to native in #1021

aldettinger:
Follow up issue for consumer support in native mode in #1022

PR #1024, opened 2020-04-02
Added native support for Couchdb extension in producer mode #989

aldettinger:


aldettinger:
Failing builds seems unrelated, so merged.

PR #1018, opened 2020-04-02
Fix #1017 Do not hardcode the TransformerFactory implementation irrev…

ppalaga:
Fix #1017

Issue #1017, opened 2020-04-02
Do not hardcode the TransformerFactory implementation irreversibly


ppalaga:
After this change https://github.com/apache/camel-quarkus/pull/1011/commits/9af2546d1602788a3ba9981b1b1976e9c0f90ff0#diff-e4d998f8f2cd322095bf13aad3daff36R26-R37 there is no way for the user to change the effective TransformerFactory implementation if it does not work for him. We should figure out a less intrusive and better maintainable way of promoting our TransformerFactory.

A PR follows.

PR #1027, opened 2020-04-03
Revisit #1017 Be even less intrusive when registering our Transformer…

ppalaga:
@lburgazzoli made me think about #1017 again after merging the first attempt #1018. He asked which service files get embedded into the native image. The previous attempt vastly ignored the question and forced our factory via system property. 

As a matter of fact, when multiple service files are available for the same path, using ServiceProviderBuildItem or NativeImageResourceBuildItem will always embed only one of them. There is not only no control which one is picked, but also the winner is different across GraalVM 19.3.1-java8, 19.3.1-java11, 20.0.0-java8 and 20.0.0-java11. So I am coming with the current proposal that pushes our preferred service file using low level GraaVM means. Using this mechanism, there is no need to mess with the system property and thus there is also no need to make it configurable (the user can use the system property on the command line to override our service file).


lburgazzoli:
Is this also taking care of forcing our impl in jvm mode ?

ppalaga:
> Is this also taking care of forcing our impl in jvm mode ?

It makes its best to do so. Our service file is written to the app, where it should™ be seen as the first and thus be the only effective one: https://github.com/apache/camel-quarkus/pull/1027/files#diff-27a4882bb0ecf7b9d1f46376b12351e9R79-R82

lburgazzoli:
Ah ok was confused because that core in the `XalanNativeImageProcessor` whereas it should probably bee moved to the `XalanProcessor` as it is not native specific

ppalaga:
All tests from the failing group are passing for me locally.

PR #1063, opened 2020-04-09
Post Camel 3.2.0 upgrade fixes

jamesnetherton:


poms/bom/pom.xml
```diff
@@ -2069,16 +2069,16 @@
...
             </dependency>
-            <dependency>
-                <groupId>org.yaml</groupId>
-                <artifactId>snakeyaml</artifactId>
-                <version>${snakeyaml.version}</version>
-            </dependency>
             <dependency>
                 <groupId>org.threeten</groupId>
                 <artifactId>threetenbp</artifactId>
                 <version>1.4.0</version>
```
lburgazzoli, line 2075: we should probably replace versions with properties

jamesnetherton: Yes, I'll do that. Thought we had some Groovy hackery which checked that stuff...

lburgazzoli: IIRC, not at the pom level but only at extension/integration test

extensions-jvm/pom.xml
```diff
@@ -51,7 +51,7 @@
         <module>aws2-translate</module>
         <module>cassandraql</module>
         <module>couchbase</module>
-        <module>debezium-mongodb</module>
+        <!--module>debezium-mongodb</module-->
```
ppalaga, line 54: Would you please file a followup issue?

extensions/pom.xml
```diff
@@ -98,8 +98,7 @@
         <module>mail</module>
         <module>microprofile-health</module>
         <module>microprofile-metrics</module>
-<!--        Disable MongoDB until quarkus uses mongbd 4.x -->
-<!--        <module>mongodb</module>-->
+        <!--        <module>mongodb</module>-->
```
ppalaga, line 101: Would you please file a followup issue? I think one is enough for all mongo extensions.

poms/bom/pom.xml
ppalaga: +1 for the prop.

extensions/pom.xml
jamesnetherton: https://github.com/apache/camel-quarkus/issues/1064

extensions-jvm/pom.xml
lburgazzoli: https://github.com/apache/camel-quarkus/issues/1064

PR #1088, opened 2020-04-15
Lightweight context without init / start phases

gnodet:
For testing mainly.


extensions-core/reactive-executor/deployment/src/main/java/org/apache/camel/quarkus/reactive/executor/deployment/BuildProcessor.java
```diff
@@ -25,9 +25,16 @@
...
+    CamelReactiveExecutorBuildItem reactiveExecutor(ReactiveExecutorRecorder recorder) {
+        return new CamelReactiveExecutorBuildItem(recorder.createReactiveExecutor());
+    }
+
     @Record(value = ExecutionTime.RUNTIME_INIT, optional = true)
     @BuildStep(onlyIf = Flags.MainEnabled.class)
-    CamelReactiveExecutorBuildItem reactiveExecutor(ReactiveExecutorRecorder recorder, VertxBuildItem vertx) {
-        return new CamelReactiveExecutorBuildItem(recorder.createReactiveExecutor(vertx.getVertx()));
+    void initReactiveExecutor(ReactiveExecutorRecorder recorder, CamelReactiveExecutorBuildItem executor,
+            VertxBuildItem vertx) {
```
lburgazzoli, line 37: I think we need a sort of barrier to avoid this to be executed after main is started

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
```diff
@@ -62,6 +62,13 @@
         @ConfigItem(defaultValue = "true")
         public boolean enabled;
 
+        /**
+         * If {@code true}, the camel context will be started at initialization time
+         * to boot up faster.
+         */
+        @ConfigItem
+        public boolean lightweight;
```
lburgazzoli, line 70: Wonder if `static` would be better as it would better reflect when it si initialized as well as the fact that it cannot be changed once configured ?

extensions-core/reactive-executor/deployment/src/main/java/org/apache/camel/quarkus/reactive/executor/deployment/BuildProcessor.java
gnodet: Right, I'll add a BuildItem to ensure the start of the context is done after.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
gnodet: {{static}} is a reserved word and can't be used as a property name...

lburgazzoli: Ah yeah, let maybe name it `static_` (as we do for the native one) and name the config item `static` ?

in any case, I don't really care much :)

extensions-core/reactive-executor/deployment/src/main/java/org/apache/camel/quarkus/reactive/executor/deployment/BuildProcessor.java
lburgazzoli: We may eventually use the `Consume/Produce` annotations with an [empty build item ](https://quarkus.io/guides/writing-extensions#empty-build-items)

gnodet: Well, I've just added an `CamelInitializedReactiveExecutorBuildItem` to ensure a 2 phase setup for the reactive executor.

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -525,12 +533,20 @@ UnremovableBeanBuildItem unremovableRoutesBuilders() {
...
         @BuildStep(onlyIf = Flags.MainEnabled.class)
         CamelReactiveExecutorBuildItem reactiveExecutor(CamelMainRecorder recorder) {
             return new CamelReactiveExecutorBuildItem(recorder.createReactiveExecutor());
         }
 
+        @Overridable
+        @Record(value = ExecutionTime.RUNTIME_INIT, optional = true)
+        @BuildStep(onlyIf = Flags.MainEnabled.class)
+        CamelInitializedReactiveExecutorBuildItem initReactiveExecutor(
+                CamelMainRecorder recorder, CamelReactiveExecutorBuildItem executor) {
```
ppalaga, line 546: Are the `recorder` param and `@Record` annotation required? If so, could you please add a comment explaining why?

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
```diff
@@ -62,6 +62,13 @@
         @ConfigItem(defaultValue = "true")
         public boolean enabled;
 
+        /**
+         * If {@code true}, the camel context will be started at initialization time
+         * to boot up faster.
+         */
+        @ConfigItem
```
ppalaga, line 69: boot up faster is better than boot up slower. Shouldn't the default be `true`?

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
ppalaga, line 70: I think the flag name does not express well what it does. How about `staticInitStart` or `startInStaticInit`?

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
```diff
@@ -62,6 +62,13 @@
         @ConfigItem(defaultValue = "true")
         public boolean enabled;
 
+        /**
+         * If {@code true}, the camel context will be started at initialization time
+         * to boot up faster.
```
ppalaga, line 67: If there are any situations in which `true` would not work well the JavaDoc should warn the user.

As far as I can understand the option moves camel context start from RUNTIME_INIT to STATIC_INIT. If so, the JavaDoc should not hesitate to use those concepts and should refer to https://quarkus.io/guides/writing-extensions#bootstrap-three-phases for the definitions of the terms.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/FastCamelContext.java
```diff
@@ -53,19 +53,22 @@
 import org.apache.camel.util.ObjectHelper;
 
 public class FastCamelContext extends AbstractCamelContext implements CatalogCamelContext, ModelCamelContext {
+    private final CamelContext reference;
```
ppalaga, line 56: I'd appreciate a short field level JavaDoc explaining the purpose of this field.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
gnodet: It should not, but I intend to run a PR to check that it can be enabled and still works.  I just need to make all the tests pass before.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
gnodet: Yes, but see my comment above.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
gnodet: I'll add a flag for that later.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/FastCamelContext.java
gnodet: Yes, I can add that, but it may be better to document the `getCamelContextReference()` method instead.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
lburgazzoli: IIRC the lightweight context has some limitation also in term of EIPs so it should be a conscious decision (unless we remove such limitations of course)

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
ppalaga: Another comment I'd appreciate here is an explanation why is this optional/overridable, something like

```
This produces the default CamelInitializedReactiveExecutorBuildItem simply by repackaging the excutor produced by reactiveExecutor(). Both this method and reactiveExecutor() are overridable to allow plugging in a Vertx based executor by adding the camel-quarkus-reactive-executor extension to the class path.
```

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/BuildProcessor.java
```diff
@@ -525,12 +533,20 @@ UnremovableBeanBuildItem unremovableRoutesBuilders() {
         }
 
         @Overridable
-        @Record(value = ExecutionTime.RUNTIME_INIT, optional = true)
+        @Record(value = ExecutionTime.STATIC_INIT, optional = true)
         @BuildStep(onlyIf = Flags.MainEnabled.class)
         CamelReactiveExecutorBuildItem reactiveExecutor(CamelMainRecorder recorder) {
```
ppalaga, line 538: Like in initReactiveExecutor(), I'd appreciate a comment explaining why is this optional/overridable.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
ppalaga: > has some limitation also in term of EIPs

I'd very much like to see all such limitations documented.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/FastCamelContext.java
ppalaga: Yes, documenting on `getCamelContextReference()` would be fine too.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelRecorder.java
```diff
@@ -69,23 +69,42 @@ public void addTypeConverterLoader(RuntimeValue<TypeConverterRegistry> registry,
...
+            context.setTypeConverterRegistry(typeConverterRegistry.getValue());
+            context.setLoadTypeConverters(false);
+            context.setModelJAXBContextFactory(contextFactory.getValue());
+            context.build();
+
+            // register to the container
+            beanContainer.instance(CamelProducers.class).setContext(context);
+
+            return new RuntimeValue<>(context);
+        }
```
ppalaga, line 107: This would be easier to read an maintain:

```suggestion
        final ExtendedCamelContext context;
        if (config.main.lightweight) {
            context = new FastLightweightCamelContext(
                    factoryFinderResolver.getValue(),
                    version,
                    xmlLoader.getValue(),
                    xmlModelDumper.getValue());
        } else {
            context = new FastCamelContext(
                    null,
                    factoryFinderResolver.getValue(),
                    version,
                    xmlLoader.getValue(),
                    xmlModelDumper.getValue());
        }

        context.setRuntimeCamelCatalog(new CamelRuntimeCatalog(config.runtimeCatalog));
        context.setRegistry(registry.getValue());
        context.setTypeConverterRegistry(typeConverterRegistry.getValue());
        context.setLoadTypeConverters(false);
        context.setModelJAXBContextFactory(contextFactory.getValue());
        context.build();

        // register to the container
        beanContainer.instance(CamelProducers.class).setContext(context);

        return new RuntimeValue<>(context);
```

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelInitializedReactiveExecutorBuildItem.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.core.deployment;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+import io.quarkus.runtime.RuntimeValue;
+import org.apache.camel.spi.ReactiveExecutor;
+
+/**
+ * Holds the initialized {@link ReactiveExecutor} {@link RuntimeValue}.
```
ppalaga, line 24: I'd appreciate a bit more explanations here: What does "initialized" mean here? Esp. what does that mean for the potential consumers of this BuildItem? When should they consume this one rather then the plain CamelReactiveExecutorBuildItem?

Similarly the CamelReactiveExecutorBuildItem should give hints when to use that one vs. this one.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
ppalaga: > I think the flag name does not express well what it does. How about `staticInitStart` or `startInStaticInit`?

Taking back, now that I understand better what the PR does. The option name is fine. Sorry for the confusion.

ppalaga:
@gnodet I am missing some general context for this change (this might be my fault because I might have missed changes in other repos or discussions on the ML, sorry in that case). Could you please explain the following?

1. Is the new Lightweight context implementation intended to become default at some point? Would that be short/middle/long term?

2. Is the new Lightweight context implementation intended to become the only supported implementation at some point? Would that be short/middle/long term?

I have some concerns about the maintainability of supporting two quite different context implementations. I do not think we can test both properly.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
gnodet: This is really a first step in the PR.  I'll submit more commits so that we actually have the most interesting part which is to switch the `context.init()` call to the `STATIC_INIT` phase.

gnodet:
> @gnodet I am missing some general context for this change (this might be my fault because I might have missed changes in other repos or discussions on the ML, sorry in that case). Could you please explain the following?
> 
> 1. Is the new Lightweight context implementation intended to become default at some point? Would that be short/middle/long term?
> 2. Is the new Lightweight context implementation intended to become the only supported implementation at some point? Would that be short/middle/long term?
> 
> I have some concerns about the maintainability of supporting two quite different context implementations. I do not think we can test both properly.

The goal is to continue on top of this PR with the main change which will switch the context initialization from RUNTIME_INIT to STATIC_INIT.  Once that's done, we'll keep the choice to use the lightweight context + STATIC_INIT or default context + RUNTIME_INIT.
The latter will be needed when:
  * dynamic loading of routes is needed (most certainly in camel-k or other specific use cases)
  * the limitations of the lightweight context are problematic in the needed scenario.

The limitations are currently mainly related to EIPs that use some kind of dynamic routing: the lightweight context can not create endpoints at runtime.  That's expected, as the goal is not reduce the amount of code needed in native mode.

lburgazzoli:
@gnodet do you mind if I make this pr a draft pr ? so it cannot be merged unintentionally ?

gnodet:
> @gnodet do you mind if I make this pr a draft pr ? so it cannot be merged unintentionally ?

Not at all ? How is that done ?

lburgazzoli:
Therese is an option right under the reviewer list:

image

btw, I've just noticed it, before it was possible to create a draft pr but was not possible to convert a PR to a draft one

ppalaga:
Out of date

PR #1107, opened 2020-04-20
Remove unneccessary reflective class registration for camel configuration

### LINKED ISSUE #843: Property binding requires reflective access to configuration classes
As per https://github.com/apache/camel-quarkus/pull/841#discussion_r387304810, it seems property binding is not working well in native mode, as the component configuration class has to be registered for reflection in order for it to work properly.

jamesnetherton:
Fixes #843


jamesnetherton:
> Thanks for doing this, @jamesnetherton . Are you going to solve also the remaining occurrences of the same problem?

Not immediately, but I'll create an issue for it. The remaining cases all relate to components that are built using the `camel-api-component-maven-plugin` & unfortunately they all still require reflective access.

jamesnetherton:
I'm going to merge this as the two failures are not related to this change.

PR #1117, opened 2020-04-21
Dynamic endpoint cannot be resolved properly (toD)

### LINKED ISSUE #1116: Dynamic endpoint cannot be resolved properly (toD)
Using latest version of Camel Quarkus I cannot get a `toD()` with dynamic http endpoints to be resolved correctly.

` .toD("{{esma.download.url}}")`

which is resolved to this:

`https://registers.esma.europa.eu/solr/esma_registers_firds_files/select?q=*&fq=publication_date:%5B${date:now-72h:yyyy-MM-dd}T00:00:00Z%20TO%20${date:now-24h:yyyy-MM-dd}T23:59:59Z%5D&wt=xml&indent=false&start=0&rows=100`

But the dynamic values (a date calculation) is not resolved.

I have tried to debug it and I don't really grasp how and where but the findings so far is that in class `SendDynamicProcessor.java` there's a property called `dynamicAware` that doesn't get resolved (null). The reason here is that is searches for a path
`META-INF/services/org/apache/camel/send-dynamic/https` in a `classMap` which is not found (`FastFactoryResolver.java`).

lburgazzoli:
Fixes #1116

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -147,6 +147,7 @@ jobs:
               quartz
               scheduler
               seda
+              send-dynamic-http
               ref
           - category: Memoryhog Dataformat JSON
             test-modules: >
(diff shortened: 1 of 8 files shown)


lburgazzoli:
@hakuseki I've reworked the PR so now it does not depend on an external service but the behavior is the same, please let me know if there's anything I've missed

PR #1130, opened 2020-04-23
Add a doc page to describe the CI setup

jamesnetherton:


docs/modules/ROOT/nav.adoc
```diff
@@ -7,3 +7,4 @@
 ** xref:create-new-extension.adoc[Create new extension].
 ** xref:promote-jvm-to-native.adoc[Promote JVM extension to Native].
 * xref:list-of-camel-quarkus-extensions.adoc[List of Camel Quarkus extensions]
+* xref:ci.adoc[Camel Quarkus Continuous Integration Builds]
```
ppalaga, line 10: I think this should better be under the contributor guide one line higher with `**`. WDYT?

jamesnetherton: Yeah, probably better. I'll tweak it.

PR #1151, opened 2020-04-27
build: create build-parent-it pom to collect common integration tests set-up

lburgazzoli:

### DIFF
--- a/extensions-jvm/avro-rpc/pom.xml
+++ b/extensions-jvm/avro-rpc/pom.xml
@@ -23,9 +23,9 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.apache.camel.quarkus</groupId>
-        <artifactId>camel-quarkus-build-parent</artifactId>
+        <artifactId>camel-quarkus-build-parent-it</artifactId>
         <version>1.1.0-SNAPSHOT</version>
-        <relativePath>../../poms/build-parent/pom.xml</relativePath>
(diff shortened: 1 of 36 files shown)


tooling/create-extension-templates/parent-pom.xml
```diff
@@ -23,9 +23,9 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.apache.camel.quarkus</groupId>
-        <artifactId>camel-quarkus-build-parent</artifactId>
+        <artifactId>camel-quarkus-build-parent-it</artifactId>
         <version>[=version]</version>
-        <relativePath>../../poms/build-parent/pom.xml</relativePath>
+        <relativePath>../../poms/build-parent-it/pom.xml</relativePath>
```
ppalaga, line 28: I think this should stay as is. This the template for the parent of runtime and deployment modules.

The template that needs change is `tooling/create-extension-templates/integration-test-pom.xml`. I think we can hardcode the artifactId `camel-quarkus-build-parent-it` there.

lburgazzoli: oh sorry, I changed the wrong one, maybe having a structure similar to the the final directory layout could help a little

lburgazzoli: @ppalaga done

ppalaga: > having a structure similar to the the final directory layout could help a little

That's probably doable if the config of the `create-extension` mojo is adjusted too.

PR #1169, opened 2020-04-30
Replace hard coded native test categories with dynamic lookup

jamesnetherton:


.github/workflows/camel-master-cron.yaml
```diff
@@ -74,6 +76,13 @@ jobs:
         with:
           name: maven-repo
           path: maven-repo.tgz
+      - name: Setup Native Test Matrix
+        id: set-native-matrix
+        run: |
+          sudo wget -O /usr/local/bin/yq "https://github.com/mikefarah/yq/releases/download/3.2.1/yq_linux_amd64"
+          sudo chmod +x /usr/local/bin/yq
+          CATEGORIES=$(cat .github/test-categories.yaml | yq r - --printMode p "categories.*" | sed -e 's/categories.//g' -e "s/\(.*\)/'\1'/" | sed ':a;N;$!ba;s/\n/,/g')
```
ppalaga, line 84: I wonder if we could eliminate the `sed -e 's/categories.//g'` step by flattening the `.github/test-categories.yaml` file?

jamesnetherton: Yes, good point. I think I only kept the `categories` parent in case we ever wanted to define other stuff in there. But that's probably an unlikely scenario, so I can look at flattening it.

PR #1209, opened 2020-05-11
Stick to kubernetes-client version managed in Quarkus BOM

ppalaga:


ppalaga:
kube-client 4.9.0 in Quarkus 1.4.2, and kube-client 4.9.1 in Quarkus master.
This changes is actually motivated by the issues I see when upgrading to Camel 3.3.0 that uses kube client 4.10

ppalaga:
> This changes is actually motivated by the issues I see when upgrading to Camel 3.3.0 that uses kube client 4.10

After testing properly, I can say that the current change does not help. Let me come up with something different.

PR #1213, opened 2020-05-12
Do not use version literals in BOMs

ppalaga:


ppalaga:
The failed native tests are passing for me locally. Let's handle if we see oom exceptions in the future.

PR #1224, opened 2020-05-15
Upgrade to Camel 3.3.0

ppalaga:


ppalaga:
@jamesnetherton @lburgazzoli would you mind throwing a final look?

ppalaga:
I have manually cancelled the push workflow

poms/bom/pom.xml
```diff
@@ -2085,6 +2085,31 @@
                 <artifactId>javassist</artifactId>
                 <version>${javassist.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.mongodb</groupId>
+                <artifactId>bson</artifactId>
+                <version>${mongodb.version}</version>
```
jamesnetherton, line 2108: Just curious - how do these overrides work out in the Quarkus Universe BOM?

Do we end up overriding their Mongo version? Or does their version take precedence again?

lburgazzoli:
LGTM

poms/bom/pom.xml
ppalaga: Good point I should have documented my intention. Actually the main problem is that Quarkus does not manage `bson`. That causes conflicts on our side because camel-mongodb pulls a different version transitively. Best would be to let Quarkus manage `bson` too.

ppalaga: I'd improve it in a followup if you do not mind.

ppalaga:
> bump & align `spring.version`

Let me do it in a followup.

poms/bom/pom.xml
ppalaga: Filed https://github.com/quarkusio/quarkus/issues/9330

PR #1238, opened 2020-05-18
Remove duplicate kubernetes-client.version property

ppalaga:


ppalaga:
The failures are unrelated

Issue #1180, opened 2020-05-05
Avro dataformat native support


aldettinger:

aldettinger:
Closing as build issue is unrelated.

PR #1239, opened 2020-05-18
Added native support for avro dataformat #1180

aldettinger:
I've also put a `@BuildTimeAvroDataFormat` annotation that could parse the schema at build time.


docs/modules/ROOT/pages/extensions/avro.adoc
```diff
@@ -0,0 +1,38 @@
...
+== Usage
+
+The extension provides support for the Camel https://camel.apache.org/components/latest/dataformats/avro-dataformat.html[Avro Data Format].
+
+=== Configuration
+
+Beyond standard usages described above, Camel Quarkus adds the possibility to parse the Avro schema at build time both in JVM and Native mode via the `@BuildTimeAvroDataFormat` annotation.
+
+For instance below, in a first step the `user.avsc` schema resource is parsed at build time.
+In a second step, an AvroDataFormat instance using the previously parsed schema is injected in the `buildTimeAvroDataFormat` field at runtime. At the end of the day, the injected data format is used
```
ppalaga, line 27:
```suggestion
In the second step, an AvroDataFormat instance using the previously parsed schema is injected in the `buildTimeAvroDataFormat` field at runtime. At the end of the day, the injected data format is used
```

extensions/avro/runtime/src/main/java/org/apache/camel/quarkus/component/avro/AvroSchemaRegistry.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.avro;
+
+import java.util.HashMap;
+
+import org.apache.avro.Schema;
+
+public class AvroSchemaRegistry extends HashMap<String, Schema> {
```
ppalaga, line 23: I have some concerns about this. Let me write more tomorrow.

ppalaga: I was about to say that as long as we need just Map.get() and Map.put(), we should prefer aggregation to inheritance. But in the end, can't we fully avoid the AvroSchemaRegistry class in this specific case and use a stock HashMap instead?

extensions/avro/runtime/src/main/java/org/apache/camel/quarkus/component/avro/AvroSchemaRegistrySubstitution.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * {@code AvroSchemaRegistrySubstitution} offers a way to bypass Quarkus
+ * default serialization. This workaround has been introduced as Quarkus default
+ * serialization used to failed with {@code org.apache.avro.Schema}.
+ */
+public class AvroSchemaRegistrySubstitution implements ObjectSubstitution<AvroSchemaRegistry, byte[]> {
+
+    @Override
+    public byte[] serialize(AvroSchemaRegistry registry) {
+        return SerializationUtils.serialize(registry);
```
ppalaga, line 31: SerializationUtils.(de)serialize() seem to be the only methods used from commons-lang3. Inlining these methods (which should go quite easily) would allow us to eliminate the commons-lang3 dependency and thus reduce number of classes GraalVM has to analyze and eo ipso speed up the native image builds.

extensions/avro/deployment/src/main/java/org/apache/camel/quarkus/component/avro/deployment/AvroProcessor.java
```diff
@@ -0,0 +1,124 @@
...
+            try (InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream(schemaResourceName)) {
+                Schema avroSchema = new Schema.Parser().parse(is);
+                avroSchemaRegistry.put(injectedFieldId, avroSchema);
+                LOG.debug("Parsed the avro schema at build time from resource named " + schemaResourceName);
+            } catch (SchemaParseException | IOException ex) {
+                final String message = "An issue occured while parsing schema resource on field " + injectedFieldId;
+                throw new BuildTimeAvroSchemaParseException(message, ex);
+            }
+        }
+        avroRecorder.recordDataFormatProducerRegistryInitialization(beanContainer.getValue(), avroSchemaRegistry);
```
ppalaga, line 121: I wonder whether it would not be safer to not to rely on HashMap serialization and instead create the HashMap in the recorder and pass the key-schema pairs one by one there. We'd thus still depend on Schema serialization but the amount of possible breakage would be a bit smaller.

integration-tests/avro/pom.xml
```diff
@@ -21,13 +21,12 @@
...
-        <artifactId>camel-quarkus-build-parent-it</artifactId>
+        <artifactId>camel-quarkus-integration-tests</artifactId>
         <version>1.1.0-SNAPSHOT</version>
-        <relativePath>../../../poms/build-parent-it/pom.xml</relativePath>
     </parent>
 
-    <artifactId>camel-quarkus-avro-integration-test</artifactId>
-    <name>Camel Quarkus :: Avro :: Integration Test</name>
+    <artifactId>camel-quarkus-integration-test-avro</artifactId>
+    <name>Camel Quarkus :: Integration Test :: Avro</name>
```
jamesnetherton, line 29: Minor nit - for consistency with other itest modules, change to:
```suggestion
    <name>Camel Quarkus :: Integration Tests :: Avro</name>
```

extensions/avro/runtime/src/main/java/org/apache/camel/quarkus/component/avro/AvroSchemaRegistry.java
aldettinger: The AvroSchemaRegistry class can't be avoided due to [#9446](https://github.com/quarkusio/quarkus/issues/9446) however aggregation looks feasible.

extensions/avro/deployment/src/main/java/org/apache/camel/quarkus/component/avro/deployment/AvroProcessor.java
aldettinger: Schema serialization would fail due to Quarkus issue [#9446]

extensions/avro/runtime/src/main/java/org/apache/camel/quarkus/component/avro/AvroSchemaRegistry.java
ppalaga: #9446 can be worked around using a custom `ObjectSubstitution<Schema, byte[]>`. There is no need to pass a HashMap to the recorder, IMO.

ppalaga: Something like this: https://github.com/ppalaga/camel-quarkus/commit/5d03784f8907b045188f46ec9aaeebee20b9018a#diff-9b69190fd7df50ae0801559a77503de0R34

extensions/avro/deployment/src/main/java/org/apache/camel/quarkus/component/avro/deployment/AvroProcessor.java
ppalaga: I mean something like this: https://github.com/ppalaga/camel-quarkus/commit/5d03784f8907b045188f46ec9aaeebee20b9018a#diff-4a5870489e7ebf2120398aaf3099ee6aL106-R115

extensions/avro/deployment/src/main/java/org/apache/camel/quarkus/component/avro/deployment/AvroProcessor.java
```diff
@@ -0,0 +1,124 @@
...
+            String schemaResourceName = annotation.value().asString();
+            FieldInfo fieldInfo = annotation.target().asField();
+            String injectedFieldId = fieldInfo.declaringClass().name() + "." + fieldInfo.name();
+            try (InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream(schemaResourceName)) {
+                Schema avroSchema = new Schema.Parser().parse(is);
+                avroSchemaRegistry.put(injectedFieldId, avroSchema);
+                LOG.debug("Parsed the avro schema at build time from resource named " + schemaResourceName);
+            } catch (SchemaParseException | IOException ex) {
+                final String message = "An issue occured while parsing schema resource on field " + injectedFieldId;
+                throw new BuildTimeAvroSchemaParseException(message, ex);
```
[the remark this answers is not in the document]
aldettinger: Sure

extensions/avro/runtime/src/main/java/org/apache/camel/quarkus/component/avro/AvroSchemaRegistry.java
aldettinger: Ok I see, indeed we can remove the AvroSchemaRegistry level if we pass schemas one by one in the recorder. It's less code and I think startup time will not be significantly impacted. So I'm ok.

extensions/avro/deployment/src/main/java/org/apache/camel/quarkus/component/avro/deployment/AvroProcessor.java
aldettinger: ok, linked to discussion above.

extensions/avro/runtime/src/main/java/org/apache/camel/quarkus/component/avro/AvroSchemaRegistrySubstitution.java
aldettinger: Should be ok now.

aldettinger:
@ppalaga , @jamesnetherton thanks for review guys :+1:  Issues should be fixed now.

extensions/avro/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -24,8 +24,10 @@
...
 metadata:
-  unlisted: true
-  guide: "https://camel.apache.org/components/latest/avro-dataformat.html"
+  keywords:
+  - "camel"
+  - "messaging"
+  - "transformation"
+  guide: "https://camel.apache.org/components/latest/dataformats/avro-dataformat.html"
   categories:
   - "integration"
```
ppalaga, line 33: We recently switched to 100% generated quarkus-extension.yaml. The generator takes the input data from runtime POM and from Camel Catalog. Could you please run `mvn -N cq:update-quarkus-metadata` from the repo root directory and check if the result makes sense?

Based on the discussion in https://groups.google.com/forum/#!topic/quarkus-dev/uc3N_cQzG1g we stopped using keywords in most cases. They are mostly redundant because code.quarkus.io searches name and description by default anyway. So keywords make sense only in rather rare cases, if the given term is both important and occurs neither in extension name nor description. In such as case the runtime POM's property `quarkus.metadata.keywords` should be used, like in https://github.com/apache/camel-quarkus/blob/master/extensions/aws-ec2/runtime/pom.xml#L34

I should document these things ASAP and sorry that I have not done that yet.

aldettinger: Good to know. Thanks for sharing :+1:

aldettinger:
Ok, I'm gonna squash and rebase that in another PR to check the build.

PR #1240, opened 2020-05-18
Reduce the number of Qute component dependencies

ppalaga:
Esp. do not pull camel-test and camel-bom


ppalaga:
9d349fe fixed the license headers.

PR #1248, opened 2020-05-20
Add support for MicroProfile Fault Tolerance

### LINKED ISSUE #1234: Add support for MicroProfile Fault Tolerance
https://camel.apache.org/components/latest/others/microprofile-fault-tolerance.html

jamesnetherton:
Fixes #1234

### DIFF
--- a/docs/modules/ROOT/pages/list-of-camel-quarkus-extensions.adoc
+++ b/docs/modules/ROOT/pages/list-of-camel-quarkus-extensions.adoc
@@ -764,7 +764,7 @@ Level | Since | Description
 == Miscellaneous Extensions
 
 // others: START
-Number of miscellaneous extensions: 12 in 12 JAR artifacts (0 deprecated)
+Number of miscellaneous extensions: 13 in 13 JAR artifacts (0 deprecated)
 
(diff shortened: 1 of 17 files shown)


jamesnetherton:
Database native test failures seems unrelated to this change. I've seen `java.net.BindException: Address already in use` pop up a few times recently. Wonder if its related to the problem mentioned in https://github.com/apache/camel/pull/3842 given that we copied the original implementation of `AvailablePortFinder`?

lburgazzoli:
If we don't use the ranged method, it should not be the same thing, wonder if we should increase the range of ports the OS is allowed to allocate

jamesnetherton:
Or I wonder if it is because we do not randomise `quarkus.http.test-ssl-port`? In the logs I see the default value being passed` -Dquarkus.http.ssl-port=8444`.

lburgazzoli:
> Or I wonder if it is because we do not randomise `quarkus.http.test-ssl-port`? In the logs I see the default value being passed` -Dquarkus.http.ssl-port=8444`.

oh, yeah that could be

jamesnetherton:
Merging as Kudu test failure is not unrelated.

Issue #1232, opened 2020-05-16
Per-extension documentation pages


ppalaga:
Following the discussion in https://camel.465427.n5.nabble.com/Website-Camel-Quarkus-Per-extension-pages-td5865580.html

We currently have manually curated per-Camel-Quarkus-extension pages
rather rarely, esp. when there is some additional config option or
important difference between the given Camel Quarkus extension and the
underlying Camel component. Here are some examples:

* https://camel.apache.org/camel-quarkus/latest/extensions/ahc.html
* https://camel.apache.org/camel-quarkus/latest/extensions/couchdb.html
* https://camel.apache.org/camel-quarkus/latest/extensions/dozer.html
* https://camel.apache.org/camel-quarkus/latest/extensions/fhir.html
* https://camel.apache.org/camel-quarkus/latest/extensions/graphql.html

The scope of this is issue is to find a way how to semi-generate per-extension pages for all extensions we have. The generated parts include:

* Name
* The description if included Camel bits from Camel Catalog
* Links to those Camel bits' pages
* Maven/Gradle coordinates

Per-extension page URLs will be used as Guide URLs in Quarkus metadata.

ppalaga:
Related: @jamesnetherton pointed out in https://github.com/apache/camel-quarkus/pull/1230#discussion_r426121770 that

> https://camel.apache.org/components/latest/core-language.html Guide link returns a 404 - same for Camel Reactive Executor. There may be others.

PR #1254, opened 2020-05-21
#1232 Extension pages

ppalaga:


jamesnetherton:
Just so that I understand how the generated bits work....

The main doc file is generated and if we need to document custom camel-quarkus config then we add that to `src/main/doc/configuration.adoc` in the extension?

ppalaga:
> The main doc file is generated and if we need to document custom camel-quarkus config then we add that to `src/main/doc/configuration.adoc` in the extension?

Yes, there `src/main/doc/configuration.adoc` and a couple of others: `intro.adoc`, `usage.adoc` and `limitations.adoc`. All of them optional. I need to document that.

PR #1268, opened 2020-05-27
Add Peter Palaga's PGP key

ppalaga:


tooling/create-extension-templates/runtime-pom.xml
```diff
@@ -33,7 +33,7 @@
     <description>[=description]</description>
 
     <properties>
-        <firstVersion>1.0.0-M8</firstVersion>
+        <firstVersion>1.0.0-CR2</firstVersion>
```
davsclaus, line 36: Should it not be CR1 ?

ppalaga: No it  is correct. That's the template for newly created extensions, which will effectively appear in the n+1 release.

Issue #1269, opened 2020-05-28
Azure native test failure


galderz:
* OpenJDK 11u-dev
* Graal 20.1
* Quarkus `master` branch
* Apache Quarkus `quarkus-master` branch

```
Caused by: java.security.InvalidKeyException: No installed provider supports this key: javax.crypto.spec.SecretKeySpec
	at javax.crypto.Mac.chooseProvider(Mac.java:392)
	at javax.crypto.Mac.init(Mac.java:435)
	at com.microsoft.azure.storage.StorageCredentialsAccountAndKey.getHmac256(StorageCredentialsAccountAndKey.java:190)
	at com.microsoft.azure.storage.core.StorageCredentialsHelper.computeHmac256(StorageCredentialsHelper.java:76)
	at com.microsoft.azure.storage.core.StorageCredentialsHelper.signBlobQueueAndFileRequest(StorageCredentialsHelper.java:133)
	at com.microsoft.azure.storage.core.StorageRequest.signBlobQueueAndFileRequest(StorageRequest.java:310)
	at com.microsoft.azure.storage.queue.CloudQueue$7.signRequest(CloudQueue.java:907)
	at com.microsoft.azure.storage.queue.CloudQueue$7.signRequest(CloudQueue.java:889)
	at com.microsoft.azure.storage.core.ExecutionEngine.setupStorageRequest(ExecutionEngine.java:341)
	... 50 more
Caused by: java.security.NoSuchAlgorithmException: class configured for Mac (provider: SunJCE) cannot be found.
	at java.security.Provider$Service.getImplClass(Provider.java:1857)
	at java.security.Provider$Service.newInstance(Provider.java:1818)
	at javax.crypto.Mac.chooseProvider(Mac.java:365)
	... 58 more
Caused by: java.lang.ClassNotFoundException: com.sun.crypto.provider.HmacCore$HmacSHA256
	at com.oracle.svm.core.hub.ClassForNameSupport.forName(ClassForNameSupport.java:60)
	at java.lang.Class.forName(DynamicHub.java:1194)
	at java.security.Provider$Service.getImplClass(Provider.java:1842)
```

[Longer output](https://gist.github.com/galderz/5cce7afe616ca38255c5ea16d08d383e).¯

galderz:
Can someone assign this to me?

galderz:
Thx @jamesnetherton.

[...]

PR #1272, opened 2020-05-29
Add missing security configuration to azure test #1269

galderz:
Fix for https://github.com/apache/camel-quarkus/issues/1269.

* Security provider needs to enabled and security extension needs adding.

### DIFF
--- a/integration-tests/azure/pom.xml
+++ b/integration-tests/azure/pom.xml
@@ -48,6 +48,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-resteasy</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-security</artifactId>
+        </dependency>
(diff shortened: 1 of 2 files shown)


lburgazzoli:
@jamesnetherton do we need to add those bits to the ms extensions themselves ?

jamesnetherton:
@lburgazzoli Yeah I was just about to suggest that. It's probably better to do something like this in `AzureProcessor`:

```
    @BuildStep
    ExtensionSslNativeSupportBuildItem activateSslNativeSupport() {
        return new ExtensionSslNativeSupportBuildItem(FEATURE);
    }
```

Rather than adding the properties file and the additional dependency to the tests.

galderz:
@lburgazzoli @jamesnetherton Sounds good. I'll rework the PR.

galderz:
Tried it and it works. Created a new PR for the new fix https://github.com/apache/camel-quarkus/pull/1294.

PR #1294, opened 2020-06-02
Enable SSL for Azure extension #1269

galderz:
PR for https://github.com/apache/camel-quarkus/issues/1269. Replaces https://github.com/apache/camel-quarkus/pull/1272.

### DIFF
--- a/extensions/azure/deployment/src/main/java/org/apache/camel/quarkus/component/azure/deployment/AzureProcessor.java
+++ b/extensions/azure/deployment/src/main/java/org/apache/camel/quarkus/component/azure/deployment/AzureProcessor.java
@@ -17,6 +17,7 @@
 package org.apache.camel.quarkus.component.azure.deployment;
 
 import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.ExtensionSslNativeSupportBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 
 class AzureProcessor {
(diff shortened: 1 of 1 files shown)


galderz:
Shall I send a PR to quarkus in parallel to add Camel Azure to the list [here](https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/native-and-ssl.adoc)?

jamesnetherton:
> Shall I send a PR to quarkus in parallel to add Camel Azure to the list [here](https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/native-and-ssl.adoc)?

@ppalaga WDYT? There's actually a few other camel extensions that enable SSL support. Not sure if we should document that over on the Quarkus side or whether it'd be better to do it exclusively in our own documentation (probably the better option IMO).

ppalaga:
> > Shall I send a PR to quarkus in parallel to add Camel Azure to the list [here](https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/native-and-ssl.adoc)?
> 
> @ppalaga WDYT? There's actually a few other camel extensions that enable SSL support. Not sure if we should document that over on the Quarkus side or whether it'd be better to do it exclusively in our own documentation (probably the better option IMO).

Yeah, I think syncing to Quarkus docs is not sustainable. The question is how detailed we should document it on our side. I have created https://github.com/apache/camel-quarkus/issues/1300 Let's discuss there so that we do not block this PR.

galderz:
Thanks all!

PR #1296, opened 2020-06-02
1.0.0-CR2 post release polishing

ppalaga:


ppalaga:
The Container startup failure in the FHIR test on Java 8 is unrelated.

PR #1326, opened 2020-06-10
Avoid duplicate feature registration of camel-support-debezium

jamesnetherton:


ppalaga:
I checked locally, that it compiles and I think it is safe to merge without waiting for the CI.

PR #1339, opened 2020-06-11
The next is 1.0.0-CR3

ppalaga:


ppalaga:
Passing on JVM  - that's enough for this kind of change.

PR #1342, opened 2020-06-11
Make the local Antora build produce .htaccess file with redirects based

ppalaga:
on :page-aliases: attributes in the AsciiDoc pages the content of that
file can by copied to
https://github.com/apache/camel-website/blob/master/static/.htaccess

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -4,8 +4,13 @@
   "description": "Apache Camel extensions for Quarkus Documentation",
   "license": "Apache-2.0",
   "devDependencies": {
-    "@antora/cli": "~2.3.0",
-    "@antora/site-generator-default": "~2.3.0",
+    "@antora/asciidoc-loader": "2.3.3",
+    "@antora/cli": "2.3.3",
+    "@antora/content-aggregator": "2.3.3",
+    "@antora/content-classifier": "2.3.3",
+    "@antora/document-converter": "2.3.3",
+    "@antora/playbook-builder": "2.3.3",
+    "@antora/site-generator-default": "2.3.3",
     "@antora/xref-validator": "gitlab:antora/xref-validator",
     "lite-server": "^2.4.0"
   },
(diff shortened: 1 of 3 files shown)


docs/package.json
```diff
@@ -4,10 +4,25 @@
...
+    "escape-string-regexp": "~2.0",
+    "gulp": "~4.0",
+    "gulp-cheerio": "^1.0.0",
+    "gulp-htmlmin": "~5.0",
+    "html-validate": "~1.2",
+    "hugo-cli": "~0.11",
+    "link-checker": "~1.2",
+    "npm-run-all": "~4.1",
+    "opal-runtime": "1.0.11",
+    "toml": "~3.0"
```
ppalaga, line 25: @djencks here is what I did to overcome the `urls.redirect_facility: must be one of the possible values: ["disabled","netlify","nginx","static"]: value was "httpd"` error trying to follow your advice to get "configuration for the checks that assures that the xref-checker is using a 2.3+ Antora":

I copied all devDependencies from camel and ran `yarn install && yarn upgrade`. 

As a result `yarn checks` pass now. 

That could mean mission completed, but I suspect copying all devDependencies from camel is a bit too brute force, but I confess I have no clue which ones could be left out. Is it OK to leave it as is or you perhaps want to propose some changes?

djencks:
I think this is enough:

"devDependencies": {
  "@antora/asciidoc-loader": "2.3.3",
  "@antora/cli": "2.3.3",
  "@antora/content-aggregator": "2.3.3",
  "@antora/content-classifier": "2.3.3",
  "@antora/document-converter": "2.3.3",
  "@antora/playbook-builder": "2.3.3",
  "@antora/site-generator-default": "2.3.3",
  "@antora/xref-validator": "gitlab:antora/xref-validator",
  "lite-server": "^2.4.0"
},

Does this work for you?
It works for me with v. 2.3.0, but why not upgrade?

Thanks!
David Jencks

ppalaga:
@djencks many thanks, the deps proposed by you seem to work.

PR #1366, opened 2020-06-17
Fix #838 TimeZone-less DTSTART and DTEND not changed to GMT

### LINKED ISSUE #838: [Quarkus 1.6][iCal] TimeZone-less DTSTART and DTEND changed to GMT in native mode
Probably caused by https://github.com/quarkusio/quarkus/issues/5016 

As a result, only zoned DTSTART and DTEND work properly.

ppalaga:
in native mode with GraalVM 20.1.0+

Waiting for Quarkus 1.6 whose docker image brings GraalVM 20.1.0


ppalaga:
1ffce1c rebased. Should be good to merge now that we are on Quarkus 1.6.0.CR1

davsclaus:
Should we get this merged before we do a new release?

ppalaga:
I need to figure out why is it failing.

ppalaga:
OK, I can reproduce the failure locally by setting the system TZ to Etc/UTC via `sudo timedatectl set-timezone Etc/UTC` or any other TZ with an offset different from Europe/Paris used in the test. Let me figure out how to solve it.

ppalaga:
Java 8 failure is unrelated.

PR #1371, opened 2020-06-18
Upgrade to Camel 3.4.0

jamesnetherton:


poms/bom/pom.xml
```diff
@@ -2185,7 +2190,7 @@
                 <artifactId>debezium-embedded</artifactId>
                 <version>${debezium.version}</version>
             </dependency>
-            <dependency>
+	    <dependency>
```
ppalaga, line 2193:
```suggestion
            <dependency>
```

jamesnetherton: Yeah I saw that. I'll address that in a follow up PR as I plan to address some of the outstanding TODO items.

jamesnetherton:
Seems the native tests hit some issues pulling from quay.io :grimacing: 

I'm going to close this and create another PR.

PR #1372, opened 2020-06-18
Type safe component injection

lburgazzoli:
This allows to:

```java
@ApplicationScoped
public static class Configurer {
    @Inject
    LogComponent log;

    @PostConstruct
    void setUpLogComponent() {
        log.setBasicPropertyBinding(true);
        log.setExchangeFormatter(new MyExchangeFormatter());
    }

    public LogComponent getLog() {
        return log;
    }
}

@ApplicationScoped
public static class Holder {
    private LogComponent log;

    @Inject
    public Holder(LogComponent log) {
        this.log = log;
    }

    public LogComponent getLog() {
        return log;
    }
}
```

### DIFF
--- a/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/InjectionPointsProcessor.java
+++ b/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/InjectionPointsProcessor.java
@@ -1,0 +1,124 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


davsclaus:
Hmm would users want to write this much code for setting up a component?

You already have component dsl
https://camel.apache.org/manual/latest/component-dsl.html

Isnt there something we can do with this instead that makes this better?

davsclaus:
Or maybe have the @Produces annotatatio (or whatever its called in CDI) In Camel you can use @BindToRegistry for this - but in Spring Boot its @Bean or @Produces or what its named.

```
@Produces
public LogComponent createMyFancyLogComponent(LogComponent log) {
   log.setXXXX
   return log;
}

We can make it take in itself as argument (if that is possible) where you just get a basic new LogComponent using its no-arg ctr.

**Comment by astefanutti** [comment]
We could automatically declare a `ComponentsBuilderFactory` bean so that users can write:

```
@Produces
public LogComponent createMyFancyLogComponent(ComponentsBuilderFactory builder) {
   return builder.log()...build();
}
```

lburgazzoli:
The initial motivation behind this enhancement are:

- make it possible to use @Inject to get component as you would do with @BeanInject inside a route but more familiar for CDI developer
- make it possible to get a reference of a component created at build time as by using the component-dsl or by producing a custom component via @Produces one would override the component created and pre-configured at build time

lburgazzoli:
> We could automatically declare a `ComponentsBuilderFactory` bean so that users can write:
> 
> ```
> @Produces
> public LogComponent createMyFancyLogComponent(ComponentsBuilderFactory builder) {
>    return builder.log()...build();
> }
> ```

That requires a camel-componentdsl extension, will open an issue

lburgazzoli:
Created: https://github.com/apache/camel-quarkus/issues/1376

Issue #1365, opened 2020-06-17
[suggestion] log message for FileNotFound exception not clear enough


nonnenmacher:
when loading `xml-routes` from a `file:` reference, the log message when raising `FileNotFoundException` in `org.apache.camel.quarkus.core.CamelRoutesCollector#collectXmlRoutesFromDirectory`
is not *explicit* enough.

It log 

```
catch (FileNotFoundException e) {
                LOGGER.debug("No XML routes found in {}. Skipping XML routes detection.", part);
```

instead of more explicitely output that *it didn't find the file* (hence no routes were added)
 
so one will see something like 

```
2020-06-17 15:28:28,588 INFO  [org.apa.cam.qua.cor.CamelRoutesCollector] (Quarkus Main Thread) Loading additional Camel XML routes from: file:src/main/resources/routes/camel-routes.xml
2020-06-17 15:28:28,589 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (Quarkus Main Thread) Apache Camel 3.3.0 (CamelContext: camel-xml-route) is starting
2020-06-17 15:28:28,591 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (Quarkus Main Thread) StreamCaching is not in use. If using streams then its recommended to enable stream caching. See more details at http://camel.apache.org/stream-caching.html
2020-06-17 15:28:28,591 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (Quarkus Main Thread) *Total 0 routes, of which 0 are started*
```

And search for other reasons on the XML file itself instead of more simpler cause : *didn't find the file: or class path: resource specified*

    NOTE : it would also be a nice to have an addendum about *absolute* path, in the `Example:file-split-log`, as when running it without attention on the indicated `application.properties`

is like 

```
camel.context.name = camel-quarkus-xml
camel.main.xml-routes = file:src/main/resources/routes/camel-routes.xml
camel.file.route.folder = src/main/resources/file
```

Which lead to this specific misleading message.

JiriOndrusek:
Please, assign to me.

PR #1379, opened 2020-06-18
Log message for FileNotFound exception not clear enough #1365

JiriOndrusek:
Fixes: https://github.com/apache/camel-quarkus/issues/1365

Misleading log message was changed to avoid false understanding that file (defined as xml-routes) exist.
Log level was changed from debug to warn, which makes problem more visible to user. (if the resource can not be parsed because of different reason, exception is thrown and application fails to start.)

### DIFF
--- a/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelRoutesCollector.java
+++ b/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelRoutesCollector.java
@@ -76,7 +76,7 @@ public class CamelRoutesCollector implements RoutesCollector {
                     }
                 }
             } catch (FileNotFoundException e) {
-                LOGGER.debug("No XML routes found in {}. Skipping XML routes detection.", part);
+                LOGGER.warn("File {} can not be found. Skipping XML routes detection.", part);
             } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelRoutesCollector.java
```diff
@@ -76,7 +76,7 @@ public XMLRoutesDefinitionLoader getXmlRoutesLoader() {
                     }
                 }
             } catch (FileNotFoundException e) {
-                LOGGER.debug("No XML routes found in {}. Skipping XML routes detection.", part);
+                LOGGER.warn("Resource {} can not be parsed. Skipping XML routes detection.", part);
```
ppalaga, line 79: Isn't this still confusing? FileNotFoundException suggests that no parsing happened, because the file does not exist. So the message shout rather say that the given path does not exist.

JiriOndrusek: Yes I agree. I've overcombined it all. I'll change it to "File {} can not be found".
This will work in a case that the file doesn't exist or for example if target resource is a folder.

Issue #723, opened 2020-02-14
Surefire 3.0.0-M4 passing even if a build step throws an exception


ppalaga:
The XsltProcessor is throwing NoSuchMethodError when the app is augmented for testing (the same exception is mentioned in #721 for the dev mode). However with maven-surefire-plugin 3.0.0-M4, this exception is swallowed and the whole build is passing and 0 tests run are reported:

```
[INFO] Running org.apache.camel.quarkus.component.xslt.it.XsltTest
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.863 s - in org.apache.camel.quarkus.component.xslt.it.XsltTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
```

When the maven-surefire-plugin is downgraded to 2.22.1, it fails as expected:

```
[INFO] Running org.apache.camel.quarkus.component.xslt.it.XsltTest
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.871 s <<< FAILURE! - in org.apache.camel.quarkus.component.xslt.it.XsltTest
[ERROR] org.apache.camel.quarkus.component.xslt.it.XsltTest  Time elapsed: 1.871 s  <<< ERROR!
java.lang.RuntimeException: 
java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
        [error]: Build step org.apache.camel.quarkus.component.xslt.deployment.XsltProcessor#xsltResources threw an exception: java.lang.NoSuchMethodError: javax.xml.transform.TransformerFactory.newInstance(Ljava/lang/String;Ljava/lang/ClassLoader;)Ljavax/xml/transform/TransformerFactory;
        at org.apache.camel.quarkus.support.xalan.XalanTransformerFactory.<init>(XalanTransformerFactory.java:41)
...
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   XsltTest » Runtime java.lang.RuntimeException: io.quarkus.builder.BuildExcepti...
[INFO] 
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[...]

PR #1383, opened 2020-06-19
Upgrade to surefire/failsafe plugins 3.0.0-M8

JiriOndrusek:
Fixes: https://github.com/apache/camel-quarkus/issues/723

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -194,8 +194,7 @@
         <maven-resources-plugin.version>3.1.0</maven-resources-plugin.version>
         <maven-scm-plugin.version>1.11.3</maven-scm-plugin.version>
         <maven-shade-plugin.version>3.4.1</maven-shade-plugin.version>
-        <!-- NOTE: We pin to this version due to https://github.com/apache/camel-quarkus/issues/723 -->
-        <maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
+        <maven-surefire-plugin.version>3.0.0-M8</maven-surefire-plugin.version>
         <os-maven-plugin.version>${os-maven-plugin-version}</os-maven-plugin.version>
         <properties-maven-plugin.version>1.0.0</properties-maven-plugin.version>
         <protobuf-maven-plugin.version>${protobuf-maven-plugin-version}</protobuf-maven-plugin.version>
@@ -398,6 +397,7 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${maven-surefire-plugin.version}</version>
                     <configuration>
+                        <useModulePath>false</useModulePath>
                         <failIfNoTests>false</failIfNoTests>
                         <systemPropertyVariables>
                             <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
@@ -417,6 +417,9 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-failsafe-plugin</artifactId>
                     <version>${maven-surefire-plugin.version}</version>
+                    <configuration>
+                        <useModulePath>false</useModulePath>
+                    </configuration>
                 </plugin>
 
                 <plugin>


pom.xml
```diff
@@ -119,7 +119,7 @@
         <maven-jar-plugin.version>3.2.0</maven-jar-plugin.version>
         <maven-release-plugin.version>2.5.3</maven-release-plugin.version>
         <!-- NOTE: We pin to this version due to https://github.com/apache/camel-quarkus/issues/723 -->
```
jamesnetherton, line 121: I think we can safely remove the NOTE comment now

ppalaga:
@jamesnetherton do you have any idea how to ensure that all tests are executed (beyond checking known suspects manually)?

JiriOndrusek:
@ppalaga It could be possible to manually check reports from previous builds and compare it with the current one.

ppalaga:
Maybe we could:

* Downloading full CI logs before and after
* Grep the logs for test summaries
* Comparing the greps should show zero diffs

WDYT?

jamesnetherton:
> Grep the logs for test summaries

Yeah, I think that would work.

pom.xml
JiriOndrusek: I'll do it

JiriOndrusek:
There were some failing test on CI  after upgrade to surefire-plugin 3.0.0-M5. Problem was caused by using JPMS during execution of tests. **There is a different behavior in default test executions on Oracle JDK and AdoptOpenJDK.** Java modular system is by default turned on (since https://issues.apache.org/jira/browse/SUREFIRE-1570), but there is an option to turn it off (https://issues.apache.org/jira/browse/SUREFIRE-1531). This PR turns off modular system for both JVM and native tests

JiriOndrusek:
I've reported an issue on surefire about different behavior in JPMS with different Java vendors - https://issues.apache.org/jira/browse/SUREFIRE-1809

pom.xml
```diff
@@ -558,6 +561,7 @@
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-surefire-plugin</artifactId>
                         <configuration>
+                            <useModulePath>false</useModulePath>
```
ppalaga, line 564: I do not think this line is required given that there is `<useModulePath>false</useModulePath>` in the pluginManagement section.

ppalaga:
> I've reported an issue on surefire about different behavior in JPMS with different Java vendors - https://issues.apache.org/jira/browse/SUREFIRE-1809

Good work, thanks!

Is my understanding right that any resolution of SUREFIRE-1809 will not impact us, because useModulePath's default is true and we want false anyway? SUREFIRE-1809 is just about a broken autodetection that we do not want to rely on, right?

JiriOndrusek:
@ppalaga yes. I understand it in the sane way. There will be no impact by fix of SUREFIRE-1809, because we are overriding default behavior to different one. It is reported only for the sake of surefire plugin.

ppalaga:
I have approved, but actually have you checked whether the same set of tests is executed before and after this change?

> * Downloading full CI logs before and after
> * Grep the logs for test summaries
> * Comparing the greps should show zero diffs

JiriOndrusek:
I'll verify it. It makes sense because the original issue was about ignored tests.

JiriOndrusek:
Please do not merge this PR. 
There are some differences between execution with surefire 3.0.0-M5 and 2.22.2
I'm switching back to draft PR and I'll try to fix it.

JiriOndrusek:
There is another difference in test summaries between 2.22.2 and 3.0.0-M5 -> reported as https://issues.apache.org/jira/browse/SUREFIRE-1812

Tibor17:
Can you show me your logs?
Did the plugin crash or how did it behave with this error?

JiriOndrusek:
Restested and commented here: https://issues.apache.org/jira/browse/SUREFIRE-1812?focusedCommentId=17276293&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17276293

JiriOndrusek:
Ab issue in maven-surefire plugin is confirmed, fix won't make it into M6 - https://issues.apache.org/jira/browse/SUREFIRE-1812?focusedCommentId=17279194&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17279194

JiriOndrusek:
Problem reported  as https://issues.apache.org/jira/browse/SUREFIRE-1812 for version 3.0.0-M5 does not happen on 3.0.0-M6. I'm going to compare results on cq for current surefire plugin and for 3.0.0-M8.

JiriOndrusek:
There is another problem with 3.0.0-M8, where skipped counts are not correct when @ParametrizedTest is used  (worked correctly in 2.22.2 or 3.0.0-*M7). Reported as https://issues.apache.org/jira/browse/SUREFIRE-2143

Tibor17:
Hi Jiri,

I am aware of such issue SUREFIRE-2143
<https://issues.apache.org/jira/browse/SUREFIRE-2143>. It was never
working. We are reworking the code in many places which would fix several
issues at once. It will be a big change.

aldettinger:
It should be fixed with surefire 3.0.0-M10.

JiriOndrusek:
The newest version is [surefire-3.1.2](https://github.com/apache/maven-surefire/releases/tag/surefire-3.1.2). It's worth a try with that version.

jamesnetherton:
Superseded by #5053.

PR #1391, opened 2020-06-20
Make camel-quarkus-rest depend on camel-quarkus-platform-http

### LINKED ISSUE #1244: Make camel-quarkus-rest depend on camel-quarkus-platform-http
We could make the camel-quarkus-rest to depend on camel-quarkus-platform-http as that is the recommended set-up we do recommend.

lburgazzoli:
Fixes #1244

### DIFF
--- a/docs/modules/ROOT/pages/extensions/rest.adoc
+++ b/docs/modules/ROOT/pages/extensions/rest.adoc
@@ -27,3 +27,8 @@ Please refer to the above links for usage and configuration details.
 ----
 
 Check the xref:user-guide/index.adoc[User guide] for more information about writing Camel Quarkus applications.
+
+== Additional Camel Quarkus configuration
+
+By default this extension configures the xref:extensions/platform-http.adoc as the component that provides the REST transport.
+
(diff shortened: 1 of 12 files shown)


lburgazzoli:
Requires some more work on camel side

docs/modules/ROOT/pages/extensions/rest.adoc
```diff
@@ -27,3 +27,8 @@ Please refer to the above links for usage and configuration details.
 ----
 
 Check the xref:user-guide/index.adoc[User guide] for more information about writing Camel Quarkus applications.
+
+== Additional Camel Quarkus configuration
+
+By default this extension configures the xref:extensions/platform-http.adoc as the component that provides the REST transport.
```
ppalaga, line 33: "By default this extension configures" awakes the impression that the end user can change the default. I guess the user can change it by calling `context.getRestConfiguration().setComponent(...)` from his RoutesBuilder but there is no way to remove platform-http from the class path. I think it would be useful to inform the end user about that, e.g.:

---8<-----

You can override the default from your `RouteBuilder`:

```

public class MyRouteBuilder extends RouteBuilder {
    @Override
        getContext().getRestConfiguration().setComponent(...);
    }
}
```

Note that by changing the default, the `platform-http` extension is not removed from your application's class path.

---8<-----

PR #1407, opened 2020-06-22
Create a camel-componentdsl extension

### LINKED ISSUE #1376: Create a camel-componentdsl extension
So among other things, we can creates component using a fluent builder like:

```java
@Produces
public LogComponent createMyFancyLogComponent() {
   return ComponentsBuilderFactory.log()...build();
}
```

lburgazzoli:
Fixes #1376


docs/modules/ROOT/pages/extensions/componentdsl.adoc
```diff
@@ -0,0 +1,22 @@
+// Do not edit directly!
+// This file was generated by camel-quarkus-package-maven-plugin:update-extension-doc-page
+
+[[componentdsl]]
+= Component DSL
+
+[.badges]
+[.badge-key]##Since Camel Quarkus##[.badge-version]##1.0.0-M3## [.badge-key]##JVM##[.badge-supported]##supported## [.badge-key]##Native##[.badge-supported]##supported##
+
+Create Camel component with a fluent Java DSL
```
ppalaga, line 10: The description in the runtime module is different. Could you please regenerate?

ppalaga, line 10:
```suggestion
Create Camel components with a fluent Java DSL
```

Other locations look good, so adding `s` should be enough.

extensions/componentdsl/runtime/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+        <version>1.1.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-componentdsl</artifactId>
+    <name>Camel Quarkus :: Component DSL :: Runtime</name>
+    <description>Create Camel components with a fluent Java DSL</description>
+
+    <properties>
+        <firstVersion>1.0.0-M3</firstVersion>
```
ppalaga, line 34:
```suggestion
        <firstVersion>1.0.0-CR3</firstVersion>
```

lburgazzoli: I will update this but wonder if it does make any sense to set the first version to a pre 1.0 release, shouldn’t we update all the first version to 1.0 once we release it ?

davsclaus: Yes they should all be 1.0.0 when its GA. All older versions are EOL and dead. And those M2, RC2, RC3 versions dont add value but just more confusion.

ppalaga: It is still practical to be able to tell (for a tweet and/or blog post) what's new in 1.0.0 compared to CR2 for the users who know/use CR2.

ppalaga: So I'd vote for consistently using the CR3 label for the newest stuff until shortly before the release and eliminate all labels when releasing.

docs/modules/ROOT/pages/extensions/componentdsl.adoc
lburgazzoli: I think the PR job should fail if it ends up generating/amending files

extensions/componentdsl/runtime/pom.xml
lburgazzoli: I think no one is questioning about having CR3 here now but the question was about what we should do when we'll release 1.0

ppalaga: Yes, changing all firstVersions to 1.0.0 is a good idea. Sorry that I have not expressed that straight away.

PR #1409, opened 2020-06-22
Update dependencies

lburgazzoli:


pom.xml
```diff
@@ -323,6 +323,13 @@
                             <artifactId>groovy-all</artifactId>
```
davsclaus, line 323: What are we using groovy-all for? In camel 3 repo we just use groovy and groovy-xml groovy-ant etc.

lburgazzoli: it  is used for some validation and generation steps so groovy-all simplifies the set-up but we can think about having a fine grain list of dependencies instead

Issue #1439, opened 2020-07-01
JAX-RS endpoint accessible before Camel context is started


ppalaga:
Steps to reproduce:

Add `camel-quarkus-main` to the Box itest.

```
cd integration-tests/box

export BOX_USER_NAME=...
export BOX_USER_PASSWORD=...
export BOX_CLIENT_ID=...
export BOX_CLIENT_SECRET=...

mvn clean test

[INFO] Running org.apache.camel.quarkus.component.box.it.BoxTest
2020-07-01 15:38:29,194 INFO  [org.apa.cam.qua.cor.CamelBootstrapRecorder] (main) bootstrap runtime: org.apache.camel.quarkus.main.CamelMainRuntime
2020-07-01 15:38:29,197 INFO  [io.quarkus] (main) Quarkus 1.6.0.CR1 on JVM started in 1.569s. Listening on: http://0.0.0.0:42833
2020-07-01 15:38:29,197 INFO  [io.quarkus] (main) Profile test activated. 
2020-07-01 15:38:29,197 INFO  [io.quarkus] (main) Installed features: [camel-box, camel-core, camel-direct, camel-main, camel-policy, camel-support-common, cdi, resteasy]
2020-07-01 15:38:29,242 INFO  [org.apa.cam.mai.BaseMainSupport] (Thread-42) Auto-configuration summary:
2020-07-01 15:38:29,243 INFO  [org.apa.cam.mai.BaseMainSupport] (Thread-42)     camel.component.box.configuration.userPassword={{env:BOX_USER_PASSWORD}}
2020-07-01 15:38:29,243 INFO  [org.apa.cam.mai.BaseMainSupport] (Thread-42)     camel.component.box.configuration.userName={{env:BOX_USER_NAME}}
2020-07-01 15:38:29,243 INFO  [org.apa.cam.mai.BaseMainSupport] (Thread-42)     camel.component.box.configuration.clientSecret={{env:BOX_CLIENT_SECRET}}
2020-07-01 15:38:29,243 INFO  [org.apa.cam.mai.BaseMainSupport] (Thread-42)     camel.component.box.configuration=#class:org.apache.camel.component.box.BoxConfiguration
2020-07-01 15:38:29,244 INFO  [org.apa.cam.mai.BaseMainSupport] (Thread-42)     camel.component.box.configuration.authenticationType=STANDARD_AUTHENTICATION
[...]

PR #1443, opened 2020-07-01
Workaround #1439 JAX-RS endpoint accessible before Camel context is s…

ppalaga:
…tarted

This solves #1439 in quite a dirty and straightforward way. A proper fix would require an API change in Camel.


ppalaga:
Of course we need to discuss whether this the way to go. I am not sure myself. I am curious to see whether any tests are going to fail.

ppalaga:
#1444 is more complete. Let's prefer that one.

Issue #1430, opened 2020-06-29
Extension for camel-threadpoolfactory-vertx


davsclaus:
In Camel 3.5 we have a new module that would be good to make camel-quarkus use. We could consider having it by default in camel-quarkus or add an extension for it so users can add it or not.

This example was updated
https://github.com/apache/camel-examples/commit/014f9081122e32a1308621062f2099a7ec08c5ec

So basically it allows to use vertx worker threads for EIPs that are using concurrency (if they dont use a custom thread pool). 
https://github.com/apache/camel/blob/master/components/camel-threadpoolfactory-vertx/src/main/docs/reactive-threadpoolfactory-vertx.adoc

This requires Camel 3.5 onwards

JiriOndrusek:
Please, assign to me. 

I'll try to use it in camel-quarkus. I'll try to add this module as an extension, having it by default in camel-quarkus can be the next step.

PR #1462, opened 2020-07-07
Extension for camel-threadpoolfactory-vertx #1430

JiriOndrusek:
Fixes: https://github.com/apache/camel-quarkus/issues/1430


docs/modules/ROOT/pages/extensions/threadpoolfactory-vertx.adoc
```diff
@@ -0,0 +1,28 @@
...
+= ThreadPoolFactory Vert.x
+
+[.badges]
+[.badge-key]##Since Camel Quarkus##[.badge-version]##1.0.0-CR4## [.badge-key]##JVM##[.badge-supported]##supported## [.badge-key]##Native##[.badge-supported]##supported##
+
+ThreadPoolFactory for camel-core using Vert.x
+
+== What's inside
+
+* https://camel.apache.org/components/latest/others/threadpoolfactory-vertx.html[ThreadPoolFactory Vert.x]
```
aldettinger, line 14: threadpoolfactory-vertx.html: is it https://camel.apache.org/components/latest/others/reactive-threadpoolfactory-vertx.html ?

.github/test-categories.yaml
```diff
@@ -105,6 +105,7 @@ platform:
   - platform-http-engine
   - reactive-streams
   - rest-openapi
+  - threadpoolfactory-vertx
```
jamesnetherton, line 108: Because there is no dedicated integration test module for `threadpoolfactory-vertx`. You may need to add it to the list of exclusions in `scripts/validate-github-workflows.groovy` in order to prevent the validation checks from failing.

JiriOndrusek:
@aldettinger I've fixed doc link according your suggestion (from the content of the suggested page it is the correct page)
@jamesnetherton I've removed this module from `test-categories.yaml` because it doesn't have a separate testing module, is this state correct?

jamesnetherton:
> is this state correct

Yep - if the itest module does not exist, then there's no need to reference it in `test-categories.yaml`. Your new tests are in `main` which is already specified in that file.

JiriOndrusek:
Test failure is not connected to this change: 
>There are uncommitted changes
HEAD detached at pull/1462/merge
	modified:   docs/modules/ROOT/pages/extensions/aws-s3.adoc
	modified:   docs/modules/ROOT/pages/extensions/aws2-s3.adoc
	modified:   docs/modules/ROOT/pages/list-of-camel-quarkus-extensions.adoc

Restarting test (locally after build from test revision, there is nothing unstaged)

JiriOndrusek:
Check problem is caused by failing debezium test. Error is caused by an issue in debezium 1.2.0.Final - https://github.com/apache/camel-quarkus/issues/1490

davsclaus:
@JiriOndrusek can you fix the pom.xml conflict

davsclaus:
This PR also has a commit to update to 3.5-SNAPSHOT of camel, is this really needed?

JiriOndrusek:
@davsclaus I'll fix it. PR is meant for branch camel-master, so there should already be 3.5-SNAPSHOT version of camel in pom. I'll fix the conflict and look why it  looks differently.

JiriOndrusek:
Test fails because of https://github.com/apache/camel-quarkus/issues/1493

docs/modules/ROOT/pages/extensions/threadpoolfactory-vertx.adoc
ppalaga: Filed a followup https://github.com/apache/camel-quarkus/issues/1506

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/BaseModel.java
```diff
@@ -110,6 +112,91 @@ public synchronized RouteDefinition getRouteDefinition(String id) {
...
 
+    @Override
+    public List<RouteTemplateDefinition> getRouteTemplateDefinitions() {
+        return routeTemplateDefinitions;
+    }
+
+    @Override
+    public RouteTemplateDefinition getRouteTemplateDefinition(String id) {
+        for (RouteTemplateDefinition route : routeTemplateDefinitions) {
+            if (route.idOrCreate(camelContext.adapt(ExtendedCamelContext.class).getNodeIdFactory()).equals(id)) {
```
ppalaga, line 123:
```suggestion
        final ... nodeIdFactory = camelContext.adapt(ExtendedCamelContext.class).getNodeIdFactory();
        for (RouteTemplateDefinition route : routeTemplateDefinitions) {
            if (route.idOrCreate(nodeIdFactory).equals(id)) {
```

It would be more effective to get the factory once before starting the loop.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/BaseModel.java
```diff
@@ -110,6 +112,91 @@ public synchronized RouteDefinition getRouteDefinition(String id) {
...
+    public void addRouteTemplateDefinitions(Collection<RouteTemplateDefinition> routeTemplateDefinitions) throws Exception {
+        if (routeTemplateDefinitions == null || routeTemplateDefinitions.isEmpty()) {
+            return;
+        }
+        this.routeTemplateDefinitions.addAll(routeTemplateDefinitions);
+    }
+
+    @Override
+    public void addRouteTemplateDefinition(RouteTemplateDefinition routeTemplateDefinition) throws Exception {
+        this.routeTemplateDefinitions.addAll(routeTemplateDefinitions);
```
ppalaga, line 140: Does not look right. Perhaps this?
```suggestion
        this.routeTemplateDefinitions.addAll(routeTemplateDefinition);
```

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/BaseModel.java
```diff
@@ -110,6 +112,91 @@ public synchronized RouteDefinition getRouteDefinition(String id) {
...
+    }
+
+    @Override
+    public void removeRouteTemplateDefinitions(Collection<RouteTemplateDefinition> routeTemplateDefinitions) throws Exception {
+        routeTemplateDefinitions.removeAll(routeTemplateDefinitions);
```
ppalaga, line 145: Does not look right either. Maybe this?
```suggestion
        this.routeTemplateDefinitions.removeAll(routeTemplateDefinitions);
```

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/BaseModel.java
JiriOndrusek: No problem, I'll refactor it

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/BaseModel.java
JiriOndrusek: I agree that current state doesn't make sense. I'll fix it.

JiriOndrusek:
@ppalaga I'll rebase it. I think, that the last 2 change requests will disappear, as they were not the part of the original changes, but it seems that they were added during some  rebases. 
I'll fix the first change request (about performance)

JiriOndrusek:
@ppalaga I see that these 2 last issues are already in camel-master, I'll fix both of them.
https://github.com/apache/camel-quarkus/blob/camel-master/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/BaseModel.java#L140
https://github.com/apache/camel-quarkus/blob/camel-master/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/BaseModel.java#L145

JiriOndrusek:
@ppalaga I've embedded all 3 fixes into this PR (they are all from current camel-master branch) I think that creating a simple PR for all 3 of them could be better solution, what do you think?

ppalaga:
> I've embedded all 3 fixes into this PR (they are all from current camel-master branch) I think that creating a simple PR for all 3 of them could be better solution, what do you think?

No problem to have unrelated changes in one PR, as long as they are in separate commits.

JiriOndrusek:
@ppalaga I'll change it into separate commit. Now it is amended

JiriOndrusek:
Good to merge from my point of view

lburgazzoli:
@ppalaga @jamesnetherton ok to merge ?

ppalaga:
I will take care

Issue #1466, opened 2020-07-08
Add Grok Extension


aldettinger:

ppalaga:
Still not a fan of auto-closing ;), @aldettinger?

aldettinger:
You caught me, this time I had really forgotten to close it :)

PR #1504, opened 2020-07-29
Added grok support #1466

aldettinger:


extensions/grok/deployment/src/main/java/org/apache/camel/quarkus/component/grok/deployment/GrokProcessor.java
```diff
@@ -0,0 +1,36 @@
...
+    private static final String FEATURE = "camel-grok";
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    NativeImageResourceBuildItem registerDefaultPatternsNativeResources() {
+        return new NativeImageResourceBuildItem("patterns/patterns");
```
ppalaga, line 34: I heart of grok for the first time today, but anyway, isn't `patterns/patterns` a typo?

Regardless whether it is correct, wouldn't it be worthwhile to document path?

Shouldn't the itest cover it somehow (I cannot see `patterns/patterns` folder in the itest)?

aldettinger: Thanks for raising @ppalaga.

camel-grok comes with what the [camel doc](https://camel.apache.org/components/latest/dataformats/grok-dataformat.html#_predefined_patterns) calls `predefined patterns`, under the hood it's [java-grok default patterns](https://github.com/thekrakken/java-grok/blob/master/src/main/resources/patterns/patterns). Grok default patterns includes things like mac, ip, number and so on.

That said, `patterns/patterns` may be a bit vague but is not under our control. The test coverage must be correct as we are using mac, ip and so on in itests. Concerning the doc, I would let it in camel, maybe just clarify the predefined patterns section.

WDYT ?

ppalaga: Oh, I see, so, `patterns/patterns` is not intended to be used by end users, right? All is fine then. Thanks for explaining.

aldettinger: Not for end user indeed. Thanks for asking, this exchange will help me to improve the camel doc a bit and may help users in the future.

PR #1524, opened 2020-08-10
1.0.0 post release cleanup

ppalaga:


ppalaga:
I remember to have agreed with somebody (was it @lburgazzoli ?) to collapse all 1.0.0-M* and 1.0.0-CR* version labels to just 1.0.0 once we release 1.0.0.

PR #1526, opened 2020-08-12
CAMEL-14297: Introduce RouteBuilderConfigurer

davsclaus:


davsclaus:
The test in
org.apache.camel.quarkus.core.CoreTest#testLookupRoutes

has a TODO as the 2nd route from RouteBuilderConfigurer is not discovered.

It uses @Produces annotation from JEE but Camel cannot discover it. Not sure how we can make this possible.
The regular RouteBuilder classes are discovered via jandex index and added during recorder magic.

I would assume a @Produces annotation from CDI/JEE would also work. But since the bean is not injected somewhere then arc may not trigger it. So maybe we need some jandex magic to discover all methods that returns a RouteBuilderConfigurer and are annotated with @Produces should then record the method, or whatever needs to be done.

lburgazzoli:
you can mark the bean as `@Unremovable` to avoid ArC to eliminate it, we can probably also add the `RouteBuilderConfigurer` class to the list of unmoveable beans

davsclaus:
Thanks @lburgazzoli that fixes it

integration-tests/core/src/test/java/org/apache/camel/quarkus/core/CoreTest.java
```diff
@@ -33,12 +33,18 @@
...
+
     @Test
     public void testContainerLookupFromRegistry() {
         RestAssured.when().get("/test/registry/lookup-registry").then().body(is("true"));
         RestAssured.when().get("/test/registry/lookup-context").then().body(is("true"));
     }
 
+    @Test
+    public void testLookupRoutes() {
+        RestAssured.when().get("/test/routes/lookup-routes").then().body(is("true"));
```
ppalaga, line 45: Comparing the list of routeIds here would make it easier to troubleshoot if something goes wrong.
```suggestion
        RestAssured.when().get("/test/routes/lookup-routes").then().body(is("foo,bar"));
```

davsclaus: Yeah good idea

PR #1535, opened 2020-08-18
Upgrade camel to v3.4.3 (1.0.x)

lburgazzoli:
@jamesnetherton do we need to add 1.0.x to the list of branches for which we need to run the CI ?


jamesnetherton:
Yes it'll need to be added to the branch list. For the `push` and `pull_request` events.

We could maybe add it as a regex if we're going to end up with multiple 0.0.x branches in future.

PR #1543, opened 2020-08-19
Add missing netty dependencies to olingo4 extension

jamesnetherton:


oscerd (approved):

ppalaga (approved):
Out of curiosity: is there any issue with the deps not being there?

jamesnetherton:
> Out of curiosity: is there any issue with the deps not being there?

I'm glad you asked....

While working on something else, I noticed that `quarkus-resteasy` transitively pulls in `quarkus-netty`. And since almost every itest uses resteasy, I was curious whether any extensions were getting their `quarkus-netty` deps for 'free'. Turns out the only problem was with olingo4.

ppalaga:
> While working on something else, I noticed that `quarkus-resteasy` transitively pulls in `quarkus-netty`. And since almost every itest uses resteasy, I was curious whether any extensions were getting their `quarkus-netty` deps for 'free'. Turns out the only problem was with olingo4.

Wow, I admire your systematism!

BTW, I remember we had a discussion with @lburgazzoli about a year ago where I was holding the view that we should not test all our extensions trough resteasy because otherwise we can oversee issues like this. I am glad we have a real example now! Of course most of our transports pull netty and vert.x so my proposal to use other transports wouldn't be effective against missing those deps anyway.

The question is whether we can detect these situations somehow at all?

ppalaga:
> we can detect these situations somehow at all?

Maybe we should check whether the dep set of the camel jar contains anything from the dep set of resteasy and other quasi-test deps?

lburgazzoli:
The issue is that any extension you bring in can actually mask this kind of problems (see for example the fact that rest-assured has dependencies on jaxb and we got some test working only because of that) 

The only way would be to write a test framework that is based on the stock classes only but I'd rather prefer fixing some issues like this one than having to maintain my own test framework :)

jamesnetherton:
> Of course most of our transports pull netty and vert.x

Yeah, your're right. It's unlikely users would ever notice this problem, but IMO each extension should accurately declare the bits they depend on. 

> Maybe we should check whether the dep set of the camel jar contains anything from the dep set of resteasy and other quasi-test deps?

We could do something like that. It's a similar story when we group many extensions together in a single itest module. There's always a chance that one extension silently provides bits that are missing from another.

ppalaga:
Actually, inspecting the dependency set would be easier to do in the runtime module itself. We could compare it against the dependency sets of Quarkus extensions.

jamesnetherton:
Test failure is not related so merging this.

We can continue the discussion elsewhere if needed.

PR #1552, opened 2020-08-20
Compute the component counts using JavaScript to avoid merge conflicts

ppalaga:
The main idea is to compute the counts on the following pages in the browser:

* https://camel.apache.org/camel-quarkus/latest/reference/index.html
* https://camel.apache.org/camel-quarkus/latest/reference/components.html
* https://camel.apache.org/camel-quarkus/latest/reference/dataformats.html
* https://camel.apache.org/camel-quarkus/latest/reference/languages.html
* https://camel.apache.org/camel-quarkus/latest/reference/others.html

The implementation is not especially elegant, but works™. I used asciidoc pass-trough `++++` to smuggle the javascript to the html page. I wonder whether you see any issues with this @zregvart ?

### DIFF
--- a/docs/modules/ROOT/pages/reference/components.adoc
+++ b/docs/modules/ROOT/pages/reference/components.adoc
@@ -4,483 +4,511 @@
 [camel-quarkus-components]
 = Camel components supported on Quarkus
 
-158 components in 123 JAR artifacts (0 deprecated, 17 JVM only)
+[#cq-components-table-row-count]##?## components in [#cq-components-table-artifact-count]##?## JAR artifacts ([#cq-components-table-deprecated-count]##?## deprecated, [#cq-components-table-jvm-count]##?## JVM only)
 
-[width="100%",cols="4,1,1,1,5",options="header"]
+[#cq-components-table.counted-table,width="100%",cols="4,1,1,1,5",options="header"]
 |===
 | Component | Artifact | Support Level | Since | Description
 
-| xref:reference/extensions/activemq.adoc[ActiveMQ] | camel-quarkus-activemq | Native +
+| xref:reference/extensions/activemq.adoc[ActiveMQ] | [.camel-element-artifact]##camel-quarkus-activemq## | [.camel-element-Native]##Native## +
 Stable | 1.0.0 | Send messages to (or consume from) Apache ActiveMQ. This component extends the Camel JMS component.
 
-| xref:reference/extensions/amqp.adoc[AMQP] | camel-quarkus-amqp | Native +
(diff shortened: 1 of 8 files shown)


djencks:
What is the source of the data shown on these pages?  Are the tables and counts derived from querying the Antora content catalog? Or from querying a json file?  Or something else?  It looks to me as if these pages are similar to the index pages in "main" camel that I changed to be generated by an asciidoctor antora indexer extension, so I wonder if a similar approach is possible here.  I've since also written an extension that can query json files.

ppalaga:
> What is the source of the data shown on these pages?

The html table in the page itself. The javascript is counting elements using `document.getElementsBy*()` methods.

djencks:
I realize the javascript you are proposing adding looks at the html and counts rows.  What is the source of the data shown in the html table?  I could study the maven plugin code  but I'm hoping you'll just tell me where it gets the info it puts in the table :-)

ppalaga:
> What is the source of the data shown in the html table?

Ah sorry, I misunderstood your initial question. The data comes from the catalog that was just prepared by another Maven mojo (`camel-quarkus-maven-plugin:prepare-catalog-quarkus`) in `catalog/target/classes/org/apache/camel/catalog/quarkus`. The catalog is a mix of property files and json files. A catalog with the same structure is produced in Camel source tree. 

> asciidoctor antora indexer extension

I remember, you mentioned this in the past but I have no idea how those extensions work. Would you mind explaining a bit? 

* Which language are those extensions written in? 
* Where are they hosted? In the camel-website repo?
* When are those extensions run? I guess during the site build (`yarn build`)?
* Does the existing indexer extension read a camel catalog?

> where the merge conflicts come from?

The conflicts typically appear in the asciidoc line counting the components. If there are two PRs, both of which add a new extension, then they both change the line. When the first PR is merged, the conflict needs to be solved in the second PR. 
Here an example PR that adds an extension and changes the line: https://github.com/apache/camel-quarkus/pull/1548/files#diff-a5f808e798bf3612d1e44875e6bef126L7-R7

ppalaga:
Anyway, the current proposal, however dirty it is, solves the most burning problem we have and I'd like to get it merged as a temporary solution, so that we can start adding extensions a bit faster. I am all for working on a cleaner solution.

jamesnetherton:
Alternatively, consider removing the counts entirely.

ppalaga:
> Alternatively, consider removing the counts entirely.

What an idea! :D

I see their main value in being able to compare quickly where we stand compared to Camel. There are other ways to get the info of course.

If there is a general consent to remove them temporarily or forever, I am not against that.

PR #1555, opened 2020-08-21
Update mvnd rules and split their entries by newlines where merge con…

ppalaga:
…flicts may happen


lburgazzoli:
I reember there were some discussion to make mvnd able to use a groovy script or something like that instead of using a list in each pom.

Did you/Guillaume implement it ?

ppalaga:
mvnd has the option, but nobody has implemented the groovy script for camel quarkus yet

PR #1606, opened 2020-08-25
Fix #1602 Speed up the CI

### LINKED ISSUE #1602: Speed up the CI
Here an idea what we could do:

## Current state

* The initial build on Java 11 runs with tests `./mvnw -V -ntp ${BRANCH_OPTIONS} clean install` 
* As a consequence of that all the other jobs have to wait for the JVM tests (40+ mins IIRC) to finish

## Proposed state

* The initial build on Java 11 should skip the tests and Quarkus mojos:
`./mvnw -V -ntp ${BRANCH_OPTIONS} clean install -DskipTests -Dquarkus.build.skip` 
* The skipped tests should be run in three (groups of) separate jobs:

1. Functional unit tests (under `extensions-core` and `extensions` directories) 
2. JVM tests under `extensions-jvm`
3. JVM and native tests under `integration-tests`

The group 3. is covered by the current `native-tests` job. I think there is no need to change it.

Groups 1. and 2. are new.

### How would this play with alternative JVM Jobs? 

We currently both build and JVM-test on alternative JVMs (8 and 14). I think it would be enough to test on both, re-using the artifacts built by the initial job. Intead of `mvn clean verify` in the root dir, we could do something like `cd extensions-jvm && mvn test` and `cd integration-tests && mvn test`. I think there is no need to run the functional tests on Java 8 and 14.

### I hope the groups 1., 2., and 3. cover all tests we have?

WDYT?

ppalaga:
Fix #1602

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -90,10 +90,9 @@ jobs:
             && ./mvnw -B clean install -DskipTests -DskipITs
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Build Project
+      - name: mvn clean install -DskipTests
         run: |
-          ./mvnw -V -ntp ${BRANCH_OPTIONS} \
-            clean install
+          ./mvnw -V -ntp ${BRANCH_OPTIONS} clean install -DskipTests -Dquarkus.build.skip
       - name: Fail if there are uncommitted changes
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


oscerd (approved):

.github/workflows/ci-build.yaml
```diff
@@ -151,14 +150,101 @@ jobs:
...
           df -h /
-      - name: Build on ${{ matrix.java }}
+      - name: cd extensions-core && mvn test
         run: |
-          ./mvnw -V -ntp ${BRANCH_OPTIONS} \
-            clean verify
-      - name: Fail if there are uncommitted changes
+          cd extensions-core
+          ../mvnw -V -ntp ${BRANCH_OPTIONS} \
+            -Dformatter.skip -Dimpsort.skip -Denforce=false -Dcamel-quarkus.update-extension-doc-page.skip \
```
jamesnetherton, line 157: Maybe we could look at doing what Quarkus does and combine most of these under a `quickly` profile. E.g `mvn clean install -Dquickly`.

jamesnetherton (approved):

.github/workflows/ci-build.yaml
ppalaga: Hm... I am a bit undecided, what should be there in `-Dquickly`. I tend to think that most of times when I want to build quickly, `-Dformatter.skip -Dimpsort.skip -Denforce=false -Dcamel-quarkus.update-extension-doc-page.skip` should not be there, because in that way, I'd end up with unformated source tree with bad deps and stale docs.

ppalaga:
The initial build went from ~49min down to ~17min - that's great!

Compare with this recent job: https://github.com/apache/camel-quarkus/pull/1595/checks?check_run_id=1025749383

.github/workflows/ci-build.yaml
lburgazzoli: It depends what you want to achieve i.e. if you check out master and you want a quick way to install everything because you are working on a side project that depends on latest camel-quarkus bits, you don't care about formattings or docs

ppalaga: Yeah, that sounds like a valid use case. Would you mind creating a followup issue?

lburgazzoli: https://github.com/apache/camel-quarkus/issues/1607

PR #1619, opened 2020-08-26
Clarified application developer responsibility when using onException in native mode

aldettinger:

### DIFF
--- a/extensions/flatpack/deployment/src/main/java/org/apache/camel/quarkus/component/flatpack/deployment/FlatpackProcessor.java
+++ b/extensions/flatpack/deployment/src/main/java/org/apache/camel/quarkus/component/flatpack/deployment/FlatpackProcessor.java
@@ -19,8 +19,6 @@ package org.apache.camel.quarkus.component.flatpack.deployment;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.NativeImageResourceBuildItem;
-import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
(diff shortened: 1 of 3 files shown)


aldettinger:
I'm pushing 2 commits at least for review. I'm happy to discuss rebasing them if preferred.

extensions/flatpack/deployment/src/main/java/org/apache/camel/quarkus/component/flatpack/deployment/FlatpackProcessor.java
```diff
@@ -36,8 +34,4 @@ NativeImageResourceBuildItem registerNativeImageResources() {
         return new NativeImageResourceBuildItem("net/sf/flatpack/xml/flatpack.dtd", "fpconvert.properties");
     }
 
-    @BuildStep
-    ReflectiveClassBuildItem registerReflectiveClasses() {
-        return new ReflectiveClassBuildItem(false, false, FlatpackException.class);
```
jamesnetherton, line 41: Maybe we should keep this?

If it's a class provided by Camel, then I see no harm in the extension registering it for reflection. If the exception types are provided by the user application, then it seems reasonable that they need to register them for reflection manually.

oscerd (approved):

extensions/flatpack/deployment/src/main/java/org/apache/camel/quarkus/component/flatpack/deployment/FlatpackProcessor.java
aldettinger: There is just the cost of embedding this class while it may not be used. But, yeah there is not real harm and that could prevent camel-quarkus user to register this camel class. I'm gonna drop this commit and push locally.
Many thanks all for review :+1:

PR #1627, opened 2020-08-26
Add IPFS, IRC & JSch JVM only extensions

jamesnetherton:


oscerd (approved):

ppalaga (approved):

ppalaga:
nessus-ipfs-client is not available in Central. Similar with Weka. We'll have to add https://repository.jboss.org/

PR #1657, opened 2020-08-28
Hipchat, IEC 60870, Ignite, IOTA and JBPM JVM support

ppalaga:


oscerd (approved):

ppalaga:
Ups, forgot to set the JVM label again. The JVM-only tests passed, so this is good to merge.

PR #1681, opened 2020-09-03
Add vertx-websocket component extension

### LINKED ISSUE #1301: [Camel 3.5] Create a vert.x websocket extension
See: https://issues.apache.org/jira/browse/CAMEL-14956.

jamesnetherton:
Fixes #1301


oscerd (approved):

extensions/vertx-websocket/runtime/src/main/java/org/apache/camel/quarkus/component/vertx/websocket/VertxWebsocketRecorder.java
```diff
@@ -0,0 +1,59 @@
...
+@Recorder
+public class VertxWebsocketRecorder {
+
+    public RuntimeValue<VertxWebsocketComponent> createVertxWebsocketComponent(RuntimeValue<Vertx> vertx) {
+        QuarkusVertxWebsocketComponent component = new QuarkusVertxWebsocketComponent();
+        component.setVertx(vertx.getValue());
+        return new RuntimeValue<>(component);
+    }
+
+    @Component("vertx-websocket")
```
ppalaga, line 39: What is the purpose of this annotation?

extensions/vertx-websocket/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,31 @@
...
+#
+
+# This is a generated file. Do not edit directly!
+# To re-generate, run the following command from the top level directory:
+#
+#   mvn -N cq:update-quarkus-metadata
+#
+---
+name: "Camel Vert.x WebSocket"
+description: "Camel WebSocket support with Vert.x"
```
ppalaga, line 25: This description does not match the style we generally adhere to with extension and component descriptions. I see that it stems from Camel. Could you please re-word it there to match the "verb first" style informing the reader about what can be done with the given extension/component? It should also make clear whether this ws client, ws server or both. 

Inspiration: 
* For AHC client, we have `Exchange data with external Websocket servers using Async Http Client`
* For JSR356 server, we have `Expose websocket endpoints using JSR356`

ppalaga (approved):
Thanks, looks good + two comments inline.

extensions/vertx-websocket/runtime/src/main/java/org/apache/camel/quarkus/component/vertx/websocket/VertxWebsocketRecorder.java
jamesnetherton: Ah, good spot. That was a leftover from some early experiments. I have removed it.

extensions/vertx-websocket/runtime/src/main/resources/META-INF/quarkus-extension.yaml
jamesnetherton: > Could you please re-word it there to match the "verb first" style informing the reader about what can be done with the given extension/component?

Yes, no problem. I'll update the description to something better over at the main camel project.

jamesnetherton:
Seems there's an issue resolving the Vert.x router. It doesn't happen for me locally so I need to dig into what the problem is.

PR #1682, opened 2020-09-03
Align dependencies with latest camel & quarkus releases

jamesnetherton:


oscerd (approved):

jamesnetherton:
> Have you done this manually? We should perhaps try to setup some sort of a tool.

Yeah, I was thinking we should have some means of doing this automatically. In wildfly-camel there's a unit test which at least verifies the properties are in sync with Camel. We could do a similar thing as a bare minimum.

We should maybe review some of those properties because I've no idea what half of them were introduced for. Maybe some are no longer relevant.

Issue #1672, opened 2020-09-02
Check whether it is possible to reduce the amount of RAM spent by the AS2 itest


ppalaga:
As seen in this job, https://github.com/apache/camel-quarkus/pull/1670/checks?check_run_id=1061031500#step:9:1345
the native compilation of the AS2 itest takes too much RAM and too much time. I wonder whether there is any potential to strip down the amount of compiled code somehow so that less RAM and less time is needed. @JiriOndrusek  would you mind having a look?

JiriOndrusek:
@ppalaga Please, assign to me. I'll investigate it.

JiriOndrusek:
Hi @ppalaga,
I've looked into this problem and I see 2 items:

- Highlighted log line shows that it took 25 mins to finish as2 integration test

> [INFO] Camel Quarkus :: Integration Tests :: AS2 .......... SUCCESS [25:41 min]

On my local computer it takes usually about 5-6 mins or similar. So I'm not sure what could be wrong.

- From my local RAM consumption it seems, that about a half of the consumption is caused by presence of bouncy castle (extension support-bouncycastle). Here are some measurements:

#### Local execution with **current code**:
>[camel-quarkus-integration-test-as2-1.1.0-SNAPSHOT-runner:14157]    classlist:   9,643.39 ms,  1.68 GB
...
[camel-quarkus-integration-test-as2-1.1.0-SNAPSHOT-runner:14157]      [total]: 304,349.07 ms, 12.07 GB
[INFO] [io.quarkus.deployment.pkg.steps.NativeImageBuildStep] Execute [objcopy, --strip-debug, /home/jondruse/git/camel-quarkus/integration-tests/as2/target/camel-quarkus-integration-test-as2-1.1.0-SNAPSHOT-runner]
[INFO] [io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in **308998ms**
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  **05:34 min**

#### Local execution **without bouncycastle**:
>[camel-quarkus-integration-test-as2-1.1.0-SNAPSHOT-runner:16243]    classlist:   8,989.76 ms,  1.17 GB
...
[camel-quarkus-integration-test-as2-1.1.0-SNAPSHOT-runner:16243]      [total]: 154,163.27 ms,  5.72 GB
[...]

PR #1686, opened 2020-09-03
Check whether it is possible to reduce the amount of RAM spent by the AS2 itest #1672

JiriOndrusek:
Fixes https://github.com/apache/camel-quarkus/issues/1672

Adds configuration for bouncycastle extension. It now requires user to define all ciphers and digest at build time, which has impact on memory requirements.

This change can be merged. It is an isolated fix. 
I'll continue with the investigation  in case that there is another way of reducing requirements of as2 integration test module.


JiriOndrusek:
@ppalaga The current build time is 7 mins faster then before
> [INFO] Camel Quarkus :: Integration Tests :: AS2 .......... SUCCESS [18:39 min]

(unfortunately I've missed one `System.ot.println` in my code, so now it is building again wit the updated code)

I'll try to find another optimization, but IMHO result form the current build is much better then I hoped for.
(I'm attaching log from the previous execution)
[9_Native Tests - misc.txt](https://github.com/apache/camel-quarkus/files/5175141/9_Native.Tests.-.misc.txt)
.

ppalaga:
> The current build time is 7 mins faster then before

That's great!

I'd like to clear some questions on Monday, switching to draft till then.

PR #1688, opened 2020-09-04
Feat(mongodb): add support for named client

### LINKED ISSUE #1608: MongoClientName annotation class not available when using camel-quarkus-mongodb extension
**Describe the issue**
`@io.quarkus.mongodb.runtime.MongoClientName` -annotation class not available when using `camel-quarkus-mongodb` extension. Annotation works with `quarkus-mongodb-client` but not if both extensions are present.

**To Reproduce**
Add mongo extension and set `quarkus.mongodb.users.connection-string` in application.properties 

Try to inject MongoClient into any bean:
```Java
@Inject
@MongoClientName("users")
MongoClient mongoClient1;
```

Observer `java.lang.NullPointerException: Annotation class not available` -error

squakez:
As referenced in #1608 we were not yet supporting the possibility to inject named client connection. With a fix provided to quarkus extension and this work PR we will be able to inject a "named" client that can be configurable in `application.properties` and referenced in route with its name.

Closes #1608

### DIFF
--- a/extensions-support/mongodb/deployment/src/main/java/org/apache/camel/quarkus/support/mongodb/deployment/SupportMongoDBProcessor.java
+++ b/extensions-support/mongodb/deployment/src/main/java/org/apache/camel/quarkus/support/mongodb/deployment/SupportMongoDBProcessor.java
@@ -35,20 +35,23 @@ class SupportMongoDBProcessor {
     }
 
     @BuildStep
-    void registerCamelMongoClientProducer(
+    void registerCamelMongoClientProducers(
             List<MongoClientBuildItem> mongoClients,
             BuildProducer<CamelRuntimeBeanBuildItem> runtimeBeans) {
 
         for (MongoClientBuildItem mongoClient : mongoClients) {
-            // If there is a default mongo client instance, then bind it to the camel registry
-            // with the default mongo client name used by the camel-mongodb component
-            if (MongoClientBeanUtil.isDefault(mongoClient.getName())) {
-                runtimeBeans.produce(
-                        new CamelRuntimeBeanBuildItem(
-                                "camelMongoClient",
-                                "com.mongodb.client.MongoClient",
-                                mongoClient.getClient()));
-            }
+            String clientName = getMongoClientName(mongoClient.getName());
+            runtimeBeans.produce(
+                    new CamelRuntimeBeanBuildItem(
+                            clientName,
+                            "com.mongodb.client.MongoClient",
+                            mongoClient.getClient()));
         }
     }
+
+    private String getMongoClientName(String clientName) {
+        // Use the default mongo client instance name if it is the default connection
+        return MongoClientBeanUtil.isDefault(clientName) ? "camelMongoClient" : clientName;
+    }
+
 }
--- a/docs/modules/ROOT/pages/reference/extensions/mongodb.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/mongodb.adoc
@@ -39,3 +39,26 @@ The Camel Quarkus MongoDB extension automatically registers a MongoDB client bea
     from("direct:start")
     .to("mongodb:camelMongoClient?database=myDb&collection=myCollection&operation=findAll")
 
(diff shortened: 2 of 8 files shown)


extensions-support/mongodb/deployment/src/main/java/org/apache/camel/quarkus/support/mongodb/deployment/SupportMongoDBProcessor.java
```diff
@@ -35,20 +35,23 @@ FeatureBuildItem feature() {
...
-            // If there is a default mongo client instance, then bind it to the camel registry
-            // with the default mongo client name used by the camel-mongodb component
-            if (MongoClientBeanUtil.isDefault(mongoClient.getName())) {
-                runtimeBeans.produce(
-                        new CamelRuntimeBeanBuildItem(
-                                "camelMongoClient",
-                                "com.mongodb.client.MongoClient",
-                                mongoClient.getClient()));
-            }
+            String clientName = getMongoClientName(mongoClient.getName());
```
lburgazzoli, line 43: isn't quarkus already publishing the MongoClients to CDI ?

oscerd (changes requested):
I think quarkus publish MongoClients to cdi

squakez:
Thanks for the review. I am not sure of how to proceed further as I've just started working with quarkus extensions. I see [this method](https://github.com/quarkusio/quarkus/blob/master/extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java#L288-L306) in the `quarkus` extension that we are later using in this camel extension to bind a client to the registry.

I've played a bit removing the whole `@BuildStep` on the extension resulting in a `No bean could be found in the registry for: mongoClient1 of type: com.mongodb.client.MongoClient` exception. Please @oscerd, @lburgazzoli any hint or example on how to proceed would help me a lot :)

lburgazzoli:
we need to check on the quarkus side if such beans are also published and if they are why the don't get resolved.

squakez:
I've made a few tests and according to what I've managed to see there is no client bean into CDI, unless that `MongoClientBuildItem` is used (either for a default or named connection). I've opened a thread to figure out if that is the expected behavior and why: https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/MongoDB.20clients.20CDI

squakez:
As we're discussing with Quarkus community, it seems the extension is not putting the CDI names to the clients, reason why we don't automatically bind them to the registry. 

There is an [opened PR to fix that issue](https://github.com/quarkusio/quarkus/pull/11998), but it opened a [discussion](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/MongoDB.20clients.20CDI) that is still ongoing and probably won't be resolved in short term.

As the work I've done is not really adding any further dependency, I propose to follow up with this development for now as it adds the possibility to use a named client on the mongo component within a method that is already called on our side. I will open a follow up request on our side in order to track the progress on the Quarkus community and refine this component once they will introduce those required adjustments.

@oscerd, @lburgazzoli what do you think?

lburgazzoli:
@squakez fine with me

squakez:
@oscerd there is still pending a change requested on your side. As I wrote in previous comment this is actually the only way to go ahead at the moment, can you please have a second look and confirm if you have any objection on merging this? Thanks!

oscerd (approved):

oscerd:
Sorry, I forgot about this PR

squakez:
> Sorry, I forgot about this PR

No problem, thanks for reviewing it!

jamesnetherton:
This looks ok to me. We should add a test for the named client.

squakez:
> This looks ok to me. We should add a test for the named client.

I can have a look at that.

squakez:
@jamesnetherton I've added a suite of test to cover also the named clients.

jamesnetherton (approved):

jamesnetherton:
Thanks @squakez!

PR #1766, opened 2020-09-09
Quarkus 1.8.0.final

jamesnetherton:


ppalaga (approved):

oscerd (approved):

pom.xml
```diff
@@ -75,7 +75,7 @@
         <okhttp.version>3.14.6</okhttp.version><!-- keep in sync with okio -->
         <okio.version>1.17.2</okio.version><!-- keep in sync with okhttp -->
         <quarkus.version>1.8.0.Final</quarkus.version>
-        <quarkus-qpid-jms.version>0.17.0</quarkus-qpid-jms.version>
+        <quarkus-qpid-jms.version>0.18.0</quarkus-qpid-jms.version>
```
ppalaga, line 78: I wonder if we should not add this step to the release guide?

jamesnetherton: Yes, we could do...

It would mean that we're perpetually dependent on the qpid folks to get their release out on time. If we didn't bump `quarkus-qpid-jms`, it's probably not a huge issue, unless there were breaking API changes in Quarkus. Then we'd definitely have to wait for their release.

ppalaga: I'd be OK to consider it optional as long as their release is available. The main concern for me is not to forget to check.

Issue #1557, opened 2020-08-21
Use asciidoctor-antora-indexer to produce lists of extensions and components on the reference pages


ppalaga:
Followup of https://github.com/apache/camel-quarkus/pull/1552

The reference pages

* https://camel.apache.org/camel-quarkus/latest/reference/index.html
* https://camel.apache.org/camel-quarkus/latest/reference/components.html
* https://camel.apache.org/camel-quarkus/latest/reference/dataformats.html
* https://camel.apache.org/camel-quarkus/latest/reference/languages.html
* https://camel.apache.org/camel-quarkus/latest/reference/others.html

are currently produced by `camel-quarkus-maven-plugin` out of the camel catalog data in `catalog/target/classes/org/apache/camel/catalog/quarkus`

@djencks mentioned that we could use Antora extensions to produce these pages.

ppalaga:
Any progress on this, @djencks ? I'd give this a try because of https://github.com/apache/camel-quarkus/issues/1770 unless you are in the middle of it?

ppalaga:
I have a PoC that I'd like get feedback about shortly.

djencks:
I got thoroughly distracted from this.  I expect I can take a look at your PoC though!

ppalaga:
> I got thoroughly distracted from this. I expect I can take a look at your PoC though!

No problem. It would be great if you could please comment on https://github.com/apache/camel-quarkus/pull/1777

ppalaga:
FIxed via #1777

PR #1777, opened 2020-09-10
Use asciidoctor-antora-indexer to produce lists of extensions

ppalaga:
Rel #1557 

This is it, @djencks .

The first two commits expose the extension metadata on extension pages. This part is probably useful regardless whether we ditch the third commit.

The third commit brings an AsciiDoc extension to render the list of extensions on https://camel.apache.org/camel-quarkus/latest/reference/index.html
This part is not something I insist on. It's just a way how me, a JavaScript noob was able to solve the problem in the most straightforward way.
I guess this could be done using your `@djencks/asciidoctor-antora-indexer` but I frankly was not able to figure out how to call your macros. So your proposals how to leverage `asciidoctor-antora-indexer` here is highly welcome.


oscerd (approved):

djencks:
Nice!  I have most of a version using my extension working, but there are a couple glitches to solve.  I had to extend my extension to camel-case attribute names in expressions, and haven't published the update yet.  I hope to get it all working tomorrow.

I don't understand where the "deprecated" comes from that gets counted above the table.  If this is real, can it be added to the extension adoc page?

ppalaga:
> Nice! I have most of a version using my extension working, but there are a couple glitches to solve. I had to extend my extension to camel-case attribute names in expressions, and haven't published the update yet. I hope to get it all working tomorrow.

Great!

> I don't understand where the "deprecated" comes from that gets counted above the table. If this is real, can it be added to the extension adoc page?

Of course, let me do that.

ppalaga:
81b1de6 : 

* Added `cq-deprecated` attribute to the extension pages.
* Removed the third commit so that you can eventually push your own commit to this branch. My original version is here for the reference: https://github.com/ppalaga/camel-quarkus/commits/i1557.1

djencks:
I made a PR against your branch, https://github.com/ppalaga/camel-quarkus/pull/5

I added a capability to the indexer extension to repeat a “template” block for each selected page, and wondered if it would be useful here.  I added a second index page that uses this format, next to the one using a table.  I think it would be easy to get stuff like the badges into the template, but I didn’t try.  If we do this here, we should presumably do the same for the rest of the camel site.

We’ll need to update the main camel-website package.json to use the 0.0.5 version of the indexer extension.

I don’t understand how the styling for the status in the table is supposed to work…. I don’t see anything visibly different from it.  Is there a css update pending somewhere?

David Jencks

ppalaga:
Thanks @djencks , more inline

> I made a PR against your branch, [ppalaga#5](https://github.com/ppalaga/camel-quarkus/pull/5) I added a capability to the indexer extension to repeat a “template” block for each selected page, and wondered if it would be useful here. 

Here a screenshot for the others to get an idea what it looks like: 

image

It is an interesting experiment, but I must say, I like the existing table layout more. It takes less space and it is easier to scan visually. What do others think?

> I don’t understand how the styling for the status in the table is supposed to work…. I don’t see anything visibly different from it. Is there a css update pending somewhere? 

No, I do not think there was supposed to be any special styling of the status in the table, at least not from me.

Unless somebody wants to veto, I'd extract the changes related to the extensions table from @djencks PR https://github.com/ppalaga/camel-quarkus/pull/5 and add them to this PR. The [extensions table](https://camel.apache.org/camel-quarkus/latest/reference/index.html) will thus look much like today, but will be powered by `@djencks/asciidoctor-antora-indexer` AsciiDoc extension.
I'll possibly add my work to display separate since versions for JVM and native on top of that.

djencks:
The block arrangement if condensed would take 3 lines per extension rather than what shows up as 2 lines for me, and I think looks a little nicer and more flowing for one extension, and I like the on-this-page TOC.  However, I don't actually use this so my opinion is not all that valuable.

Since no one else has spoken in support of the block arrangement I'll take it out of my PR.  It's easy enough to bring back should we change our minds.

ppalaga:
08ad76b:

* Reference tables ganerated using asciidoctor-antora-indexer AsciiDoc extension
* JVM since and Native since shown separately.

Please review

ppalaga:
Follow ups:

* Remove `UpdateDocExtensionsListMojo`, its templates and config
* Discuss whether there is a better solution than having attribute-only `components/*.adoc` files that actually get transformed to HTML

catalog/pom.xml
```diff
@@ -4127,7 +4127,7 @@
                         <!-- prepare the catalog and update doc files, etc. -->
                         <goals>
                             <goal>prepare-catalog-quarkus</goal>
-                            <goal>update-doc-extensions-list</goal>
+                            <!-- <goal>update-doc-extensions-list</goal> -->
```
jamesnetherton, line 4130: Can this be removed?

jamesnetherton (approved):

catalog/pom.xml
ppalaga: Yes, I want to remove the whole mojo - see my commet above

> Remove UpdateDocExtensionsListMojo, its templates and config

Issue #1717, opened 2020-09-07
Be able to determine that the test was run against external service or not


llowinge:
There exist some tests which are run against mocked services (eg. SlackTest) by default. If we set ENV property, it should be run against external service but we cannot say that with 100%, becasue eg. ENV property can change the name and so the test would be run against mocked service only.

Suggested way would be adding some property eg. -Dtest.external.service=true, which would skip the test, if the needed ENV properties are not set.

ppalaga:
Fixed via https://github.com/apache/camel-quarkus/pull/1818

PR #1782, opened 2020-09-11
Make sure that Telegram tests are not run against mock API by setting…

ppalaga:
… CAMEL_QUARKUS_START_MOCK_BACKEND=false #1717

@llowinge this is an idea how we could do that. Done only for Telegram, but other tests could use the same pattern with the same env var `CAMEL_QUARKUS_START_MOCK_BACKEND`, so that you can set it once for all tests. WDYT?

### DIFF
--- a/integration-tests/telegram/src/main/java/org/apache/camel/quarkus/component/telegram/it/TelegramRoutes.java
+++ b/integration-tests/telegram/src/main/java/org/apache/camel/quarkus/component/telegram/it/TelegramRoutes.java
@@ -21,37 +21,86 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.stream.Stream;
 
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Produces;
+import javax.inject.Named;
+
+import io.quarkus.arc.Unremovable;
 import org.apache.camel.Exchange;
 import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.telegram.TelegramComponent;
+import org.apache.camel.quarkus.test.mock.backend.MockBackendUtils;
 import org.apache.camel.support.ResourceHelper;
 import org.apache.camel.util.IOHelper;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
 
+@ApplicationScoped
 public class TelegramRoutes extends RouteBuilder {
+
+    @ConfigProperty(name = "camel.quarkus.start-mock-backend", defaultValue = "true")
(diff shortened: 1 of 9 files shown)


oscerd (approved):

ppalaga:
c31201e fixed the formatting

llowinge:
@ppalaga I like the idea with possible switching between mock/external service. Would community accept the `CAMEL_QUARKUS_START_MOCK_BACKEND`  be set by default to `false` and in your tests would be switched to `true` ? In this manner we QE can be safe that the tests will really be run against external service even if property name `CAMEL_QUARKUS_START_MOCK_BACKEND` will change in future.

ppalaga:
> Would community accept the `CAMEL_QUARKUS_START_MOCK_BACKEND` be set by default to `false`

I discussed this with @jamesnetherton  and @oscerd and we prefer not to do that, because it would require all devs to export CAMEL_QUARKUS_START_MOCK_BACKEND=true

@jamesnetherton came with an alternative proposal that is implemented in 7042574

With this solution in place, you can do the following:

* Run the tests with CAMEL_QUARKUS_START_MOCK_BACKEND=false
* Count the occurrences of `<artifactId>camel-quarkus-integration-test-support-mock-backend</artifactId>` in `integration-tests/*/pom.xml`
* Count the occurrences of `Real backend will be used for` in the output of all tests
* The numbers should be non-null and equal (as long as you only run the JVM tests)

WDYT?

integration-tests/telegram/src/main/java/org/apache/camel/quarkus/component/telegram/it/TelegramRoutes.java
```diff
@@ -21,37 +21,86 @@
...
+import org.apache.camel.quarkus.test.mock.backend.MockBackendUtils;
 import org.apache.camel.support.ResourceHelper;
 import org.apache.camel.util.IOHelper;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
 
+@ApplicationScoped
 public class TelegramRoutes extends RouteBuilder {
 
+    @ConfigProperty(name = "camel.quarkus.start-mock-backend", defaultValue = "true")
+    boolean startMockBackend;
```
llowinge, line 41: I'm just thinking if it would be somehow possible to move this parameter into `MockBackendUtils` and there would be the logic with one generic method `logBackendUsed`. Wdyt @jamesnetherton ?

ppalaga:
Superceded by https://github.com/apache/camel-quarkus/pull/1818

Issue #1694, opened 2020-09-07
String template native support


ppalaga:

JiriOndrusek:
Please, assign to me.

Issue #1807, opened 2020-09-18
allowContextMapAll requires template engines to have reflective access to camel core classess 


jamesnetherton:
Thinking a bit more about [this comment](https://github.com/apache/camel-quarkus/pull/1804#discussion_r490314030).

We probably have a general issue with all of the templating component extensions (qute, mustache, mvel, velocity etc) where the template engine may need reflective access to complex objects like `Exchange`, `Message` or `CamelContext`. E.g as described [here](https://camel.apache.org/components/latest/velocity-component.html#_velocity_context) when `allowContextMapAll` is enabled.

I don't think we have much (or any) test coverage for this, so we haven't noticed any issues thus far.

We could:

* Add a general purpose support extension, that would contain a `BuildStep` to configure reflection for these classes. Then the various extensions could depend on it.

* Not offer any OOTB support for `allowContextMapAll` in native mode, and advise folks to keep things simple by only referencing the message body, headers & properties in templates

* Similar to the above but force users to manually do the reflective class configuration for the necessary bits.

davsclaus:
allowContextMapAll is a security vulnerabilty fix, so I am fine with no support for setting = true in camel-quarkus native mode or even jvm mode

PR #1828, opened 2020-09-24
String template native support #1694

JiriOndrusek:
Issue: https://github.com/apache/camel-quarkus/issues/1694


oscerd (approved):

extensions/stringtemplate/deployment/src/main/java/org/apache/camel/quarkus/component/stringtemplate/deployment/StringtemplateProcessor.java
```diff
@@ -17,30 +17,23 @@
...
-    @BuildStep(onlyIf = NativeBuild.class)
-    @Record(value = ExecutionTime.RUNTIME_INIT)
-    void warnJvmInNative(JvmOnlyRecorder recorder) {
-        JvmOnlyRecorder.warnJvmInNative(LOG, FEATURE); // warn at build time
-        recorder.warnJvmInNative(FEATURE); // warn at runtime
+    @BuildStep
+    ReflectiveClassBuildItem registerForReflectionWithMethods() {
+        return new ReflectiveClassBuildItem(true, false,
+                //required for values in properties
+                DefaultExchange.class.getName());
```
jamesnetherton, line 37: Maybe we should not do this. See my comments in https://github.com/apache/camel-quarkus/issues/1807.

JiriOndrusek: If it may be security vulnerability, I'll remove corresponding test, I'll remove this build item and add into limitations, that this feature is not supported in native mode. @jamesnetherton Do you agree?

jamesnetherton: Fine with me :+1:

JiriOndrusek: I've fixed it as I suggested. (also discovered a bug in camel component - https://issues.apache.org/jira/browse/CAMEL-15577)

integration-tests/stringtemplate/src/test/java/org/apache/camel/quarkus/component/stringtemplate/it/StringtemplateIT.java
```diff
@@ -0,0 +1,47 @@
...
+
+import static org.hamcrest.CoreMatchers.containsString;
+
+@NativeImageTest
+class StringtemplateIT extends StringtemplateTest {
+
+    @Test
+    public void testDisabledContextMapAll() {
+        //allowContextMapAll has to be disabled in native mode
+        RestAssured.given()
```
JiriOndrusek, line 34: I've added test for native to verify that parameter `allowContextMapAll` is disabled. 
But because there are usually no special tests for native mode, I'd like to ask. @jamesnetherton , @ppalaga   Should I keep it or remove it?

ppalaga: No problem with having a special test for native. We definitely had such in the past, not sure we still have them elsewhere.

ppalaga:
There are conflicts. Could you please rebase?

extensions/stringtemplate/deployment/src/main/java/org/apache/camel/quarkus/component/stringtemplate/deployment/StringtemplateProcessor.java
ppalaga: > Maybe we should not do this. See my comments in #1807.

I thought the same when I saw it, but I was not able to find the place where Claus mentioned that allowContextMapAll is a security vulnerabilty fix. Thanks for mentioning #1807 @jamesnetherton !

integration-tests/stringtemplate/src/test/java/org/apache/camel/quarkus/component/stringtemplate/it/StringtemplateIT.java
jamesnetherton: We should agree on whether [this](https://github.com/apache/camel-quarkus/blob/bd845d3959984d575441d3e79cd9a58095b6aa34/extensions/stringtemplate/runtime/src/main/java/org/apache/camel/quarkus/component/stringtemplate/StringtemplateEndpointSubstitute.java) is the right thing to do.

Originally, I saw the 'fix' to #1807 as purely a documentation exercise. Where we just state for each affected extension that `contextMapAll` will not work in native mode.

Potentially forcing `contextMapAll` to `false` is a bit confusing because you would run into problems at runtime. Where as without it, the native build would fail, so it's sort like a 'fail fast'.

JiriOndrusek: So, removing substitution (which forcescontextMapAll = true), rewriting limitations (that exchange.properties won't work) and removing test, should be the correct solution, yes?

jamesnetherton: IMO, yes. @ppalaga WDYT?

JiriOndrusek: I think about documentation like this one:
> If property `allowContextMapAll` is set to `true` there is a limitation in native mode. Reflective calls to methods
of several classes in camel are not allowed (for example `org.apache.camel.support.DefaultExchange`). Therefore 
some expression won't work (for example `<exchange.properties.*>).
>
>You can allow reflective calls to such classes (for example by application.properties), but keep in mind that it brings 
some security risks.

WDYT? @ppalaga , @jamesnetherton

ppalaga: +1 for removing the substitution and letting the native compilation fail

jamesnetherton: Sounds ok to me. Maybe tweak `some expression` to `some expressions`.

JiriOndrusek: done

ppalaga (approved):

ppalaga:
Ups, conflicts again. Sorry. Could you please rebase?

ppalaga:
Something went wrong with the rebase: https://github.com/apache/camel-quarkus/pull/1828/checks?check_run_id=1166121710#step:9:16

PR #1855, opened 2020-09-30
Document allowContextMapAll native mode limitations

### LINKED ISSUE #1807: allowContextMapAll requires template engines to have reflective access to camel core classess

jamesnetherton:
Fixes #1807


oscerd (approved):

tooling/maven-plugin/src/main/java/org/apache/camel/quarkus/maven/UpdateExtensionDocPageMojo.java
```diff
@@ -316,6 +320,29 @@ static boolean detectNativeSsl(Path deploymentBasePath) {
         }
     }
 
+    static boolean detectAllowContextMapAll(CqCatalog catalog, String artifactId) {
+        final String allowContextMapAll = "allowContextMapAll";
+        Optional<ArtifactModel<?>> optional = catalog.filterModels(artifactId).findFirst();
```
ppalaga, line 325: This may not work as expected if the first model is a LanguageModel or a Dataformat model. I think it would be safer to 
```suggestion
        return catalog.filterModels(artifactId)
                .filter(m -> m instanceof ComponentModel)
                .map(m -> (ComponentModel) m)
                .anyMatch(/* your options condition here */)
```

ppalaga (commented):
A suggestion inline

ppalaga:
I like general idea to auto-document `allowContextMapAll` based on the data from the Catalog

tooling/maven-plugin/src/main/java/org/apache/camel/quarkus/maven/UpdateExtensionDocPageMojo.java
jamesnetherton: Thanks. Latest commit implements this suggestion.

PR #1856, opened 2020-09-30
Geocoder native support fixes #1645

### LINKED ISSUE #1645: Geocoder native support

zbendhiba:


oscerd (approved):

ppalaga:
I have restarted the CI.

integration-tests/geocoder/src/test/java/org/apache/camel/quarkus/component/geocoder/it/GeocoderGoogleTest.java
```diff
@@ -0,0 +1,73 @@
...
+    public static void setup() {
+        // enables the integration tests if an API key exists
+        enabled = ConfigProvider.getConfig()
+                .getOptionalValue(GOOGLE_GEOCODER_API_KEY, String.class).isPresent();
+    }
+
+    @Test
+    public void loadCurrentLocation() {
+        // disable test if no API KEY
+        if (enabled) {
```
jamesnetherton, line 45: I think it'd be better to follow the pattern we use in other tests, where annotations determine whether the test should run. For example, see the Braintree test:

https://github.com/apache/camel-quarkus/blob/master/integration-tests/braintree/src/test/java/org/apache/camel/quarkus/component/braintree/it/BraintreeTest.java#L29-L31

https://github.com/apache/camel-quarkus/blob/master/integration-tests/braintree/src/test/java/org/apache/camel/quarkus/component/braintree/it/BraintreeIT.java#L23-L25

Can you also add a README to the itest module directory that explains which environment variable(s) need to be set and how to obtain API keys etc. E.g like this:

https://github.com/apache/camel-quarkus/blob/master/integration-tests/braintree/README.adoc

zbendhiba: thanks!! I'll add that annotation then !!

integration-tests/geocoder/src/test/java/org/apache/camel/quarkus/component/geocoder/it/GeocoderGoogleTest.java
```diff
@@ -0,0 +1,73 @@
...
+@TestHTTPEndpoint(GeocoderGoogleResource.class)
+class GeocoderGoogleTest {
+
+    private static boolean enabled;
+
+    @BeforeAll
+    public static void setup() {
+        // enables the integration tests if an API key exists
+        enabled = ConfigProvider.getConfig()
+                .getOptionalValue(GOOGLE_GEOCODER_API_KEY, String.class).isPresent();
```
ppalaga, line 39: You can use `@EnabledIfEnvironmentVariable` for disabling the test like we do in https://github.com/apache/camel-quarkus/blob/1175fdc6dd5e6ee4a7782afa56b404803ad2e73d/integration-tests/braintree/src/test/java/org/apache/camel/quarkus/component/braintree/it/BraintreeIT.java#L25

If you do that that way, the test app is not started at all and you actually do not need to check in the app whether property is set. 

I think in case the GOOGLE_GEOCODER_API_KEY is not provided, the test should run against a mock API, something like we do with Telegram and elsewhere. If a mock testcontainer is available we should use it. No need to do this now, I'll file a followup.

ppalaga (changes requested):
`application.properties` and the `.adoc` file requests are blocking from my PoV, the rest can be done in a followup.

ppalaga:
Here is the follow up: https://github.com/apache/camel-quarkus/issues/1860

integration-tests/geocoder/src/test/java/org/apache/camel/quarkus/component/geocoder/it/GeocoderGoogleTest.java
zbendhiba: OK

extensions/geocoder/runtime/src/main/doc/configuration.adoc
```diff
@@ -0,0 +1,3 @@
+The extension uses Locale.English to do the conversion of Double latitude and longitude values, in order to match the format required by the Google Maps API.
+
+An example of setting the Locale en_US for GraalVm: https://www.graalvm.org/reference-manual/ruby/UTF8Locale/
```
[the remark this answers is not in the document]
zbendhiba: LGTM!

ppalaga:
The foundation test group failure is unrelated.

zbendhiba:
Thanks @ppalaga

PR #1858, opened 2020-09-30
Tidy up pom.xml files

ppalaga:
Some sorting and moving properties around

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,9 +37,6 @@
     <inceptionYear>2019</inceptionYear>
 
     <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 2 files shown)


oscerd (approved):

pom.xml
```diff
@@ -277,6 +286,12 @@
                     </configuration>
                 </plugin>
 
+                <plugin>
+                  <groupId>org.apache.maven.plugins</groupId>
+                  <artifactId>maven-resources-plugin</artifactId>
+                  <version>3.1.0</version>
```
jamesnetherton, line 292: Shouldn't that reference the version property?

ppalaga: Good catch, thanks!

PR #1897, opened 2020-10-09
File specific issues for TODOs where necessary #1285

### LINKED ISSUE #1285: File specific issues for TODOs where necessary
We have a number of TODOs in the source, eg git grep TODO.
I can see some have been fixed by 3rd party etc so lets revisit them and see which ones are already done or we can fix now.

ppalaga:
Fix #1285


oscerd (approved):

ppalaga:
This is just comments, no need to wait for the full CI

Issue #1701, opened 2020-09-07
[Dependency hell] Spark native support


ppalaga:

ppalaga:
Giving up the initial attempt to produce a JVM-only extension due to too many dependency alignment problems.

ppalaga:
We now have an idiosyncratic JVM-only Spark extension https://github.com/apache/camel-quarkus/pull/1916
Let's use this issue for the native support from now on.

ppalaga:
We will remove the existing JVM-only Spark extension https://github.com/apache/camel-quarkus/issues/3466

PR #1916, opened 2020-10-14
Spark JVM support

ppalaga:
Related https://github.com/apache/camel-quarkus/issues/1701

This is the idea how we could handle this kind of dependency hell components. The artifacts are managed in the runtime module so that they do not to pollute the BOM.


oscerd (approved):

jamesnetherton (approved):

ppalaga:
Let me write a test

ppalaga:
a1b5695 :
* Added a proper test
* dependencyManagement in a separate camel-quarkus-spark-bom

Please re-review

jamesnetherton (approved):
Looks good.

I wonder if we should place the spark specific version properties within the spark bom and keep them out of the root pom. WDYT?

ppalaga:
> I wonder if we should place the spark specific version properties within the spark bom and keep them out of the root pom. WDYT?

I'd like to have them in the top pom so that everyone can see that those artifacts are used somewhere. I find this important for the case that they pop up through some other extension. In such a case we should make a conscious decision whether they should be managed in the main BOM or somewhere else. That could get forgotten easily if there is no trace of them in the top pom. WDYT?

ppalaga:
I should document the peculiarities.

jamesnetherton:
> That could get forgotten easily if there is no trace of them in the top pom. WDYT?

Ok - fine with me to keep them in the root pom.

Issue #1759, opened 2020-09-09
quarkus-spring-web breaks startup of camel-quarkus-activemq


lburgazzoli:
placeholder for https://github.com/quarkusio/quarkus/issues/11996

jamesnetherton:
The problem is that `quarkus-spring-di` has this plugin configuration:

```xml
<plugin>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-bootstrap-maven-plugin</artifactId>
    <configuration>
        <excludedArtifacts>
            <excludedArtifact>org.springframework:spring-core</excludedArtifact>
            <excludedArtifact>org.springframework:spring-beans</excludedArtifact>
            <excludedArtifact>org.springframework:spring-context</excludedArtifact>
        </excludedArtifacts>
    </configuration>
</plugin>
```

This effectively bans the core spring libraries from the application. The Quarkus Spring extensions repackage the Spring dependencies and only include a minimal set of packages to reduce the overall footprint. Hence we get a CFNE at runtime.

I don't think there's a simple way to work around this.

We could maybe use a similar approach to Quarkus, where we have our own custom extensions for `spring-jms` / `spring-jdbc` and repackage the Spring bits that we depend on. That would get around the artifact ban.

PR #1932, opened 2020-10-19
Increase test coverage for Spring dependent extensions

jamesnetherton:
Prelude to fixing https://github.com/apache/camel-quarkus/issues/1759.

### DIFF
--- a/integration-tests/jpa/src/main/java/org/apache/camel/quarkus/component/jpa/it/JpaResource.java
+++ b/integration-tests/jpa/src/main/java/org/apache/camel/quarkus/component/jpa/it/JpaResource.java
@@ -17,17 +17,21 @@
 package org.apache.camel.quarkus.component.jpa.it;
 
 import java.net.URI;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import javax.inject.Inject;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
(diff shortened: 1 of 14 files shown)


oscerd (approved):

integration-tests/jpa/src/main/java/org/apache/camel/quarkus/component/jpa/it/JpaResource.java
```diff
@@ -49,6 +56,51 @@
...
+    }
+
+    @Path("/get/query/named/{id}")
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @SuppressWarnings("unchecked")
+    public List<Fruit> getFruitByNamedQuery(@PathParam("id") int id) {
+        Map<String, Object> queryParams = new HashMap<>();
+        queryParams.put("fruitId", id);
+        context.getRegistry().bind("parameters", queryParams);
```
ppalaga, line 83: Shouldn't we add a comment here that this is not a good pattern to copy in a real application?

ppalaga (approved):
Impressive!

integration-tests/jpa/src/main/java/org/apache/camel/quarkus/component/jpa/it/JpaResource.java
jamesnetherton: I can rework it so that it gets bound to the registry via a producer method, which is a bit more 'real world'.

ppalaga: No, this OK for a test, esp. if we do it only once. I just wanted that ppl. see that this not good for areal app.

PR #1937, opened 2020-10-21
Enable Spring dependent extensions to work with Quarkus Spring

jamesnetherton:
Works around the issue I identified in https://github.com/apache/camel-quarkus/issues/1759.

I did some analysis of Spring code paths that get executed in the Camel unit tests and came up with a list of package includes & excludes for shading. 

The itests pass ok, but I can't check for every possible code path. Is it better / safer to shade all packages, instead of selectively including them? Obviously there's a chance that some required stuff may have been left out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ pom.xml.releaseBackup
 pom.xml.versionsBackup
 pom.xml.next
 release.properties
+dependency-reduced-pom.xml
 
 # Eclipse
 .project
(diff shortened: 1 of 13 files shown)


oscerd (approved):

ppalaga (approved):
Do we have a piece of docs somewhere explaining why we do this? If not, a README or `<description>` in the pom would come in handy.

jamesnetherton:
> Do we have a piece of docs somewhere explaining why we do this? If not, a README or `<description>` in the pom would come in handy.

Latest commit adds a README to the spring support extension.

PR #1939, opened 2020-10-21
Exclude glassfish dependencies from hbase-testing-util

jamesnetherton:
Seems there's a SNAPSHOT lurking in the `hbase-testing-util` dependency tree for `org.glassfish:javax.el` which caused some failures in #1937. Hence excluding `org.glassfish.web:*`.


ppalaga (approved):

jamesnetherton:
Merging since the JVM tests passed.

PR #1953, opened 2020-10-26
Replace ${camel.quarkus.project.root} set by directory-maven-plugin w…

ppalaga:
…ith ${maven.multiModuleProjectDirectory}

This should work as long as we have `.mvn` directory in the root.

### DIFF
--- a/catalog/pom.xml
+++ b/catalog/pom.xml
@@ -4249,9 +4249,9 @@
                         <phase>process-resources</phase>
                         <configuration>
                             <extensionDirectories>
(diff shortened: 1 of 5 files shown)


oscerd (approved):

jamesnetherton:
IIRC the maven wrapper binaries / scripts are excluded from the source distribution.

ppalaga:
> IIRC the maven wrapper binaries / scripts are excluded from the source distribution.

An empty `.mvn` would be enough. Let me check.

ppalaga:
I have changed the assembly descriptor to include `./mvn/maven.config`.

jamesnetherton (approved):

Issue #1961, opened 2020-10-27
json-validator native build fails with Camel 3.7.0-SNAPSHOT


jamesnetherton:
Camel has upgraded `json-schema-validator` to 1.0.44 which seems to contain some changes that break native support.

Looks like there's now support for an [optional library](https://github.com/networknt/json-schema-validator/blob/master/pom.xml#L91-L96) for [ECMA 262](https://github.com/networknt/json-schema-validator/blob/master/doc/ecma-262.md) validation. Which is probably the root cause of the CNFE.

```
Error: Class initialization of com.networknt.schema.PatternValidator$PatternValidatorEcma262 failed. Use the option --initialize-at-run-time=com.networknt.schema.PatternValidator$PatternValidatorEcma262 to explicitly request delayed initialization of this class.
Original exception that caused the problem: java.lang.NoClassDefFoundError: org/jcodings/Encoding
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized0(Native Method)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized(Unsafe.java:1042)
	at jdk.unsupported/sun.misc.Unsafe.ensureClassInitialized(Unsafe.java:698)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.ensureClassInitialized(ConfigurableClassInitialization.java:177)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.computeInitKindAndMaybeInitializeClass(ConfigurableClassInitialization.java:594)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.computeInitKindAndMaybeInitializeClass(ConfigurableClassInitialization.java:130)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.maybeInitializeHosted(ConfigurableClassInitialization.java:168)
	at com.oracle.svm.hosted.SVMHost.registerType(SVMHost.java:236)
	at com.oracle.graal.pointsto.meta.AnalysisUniverse.createType(AnalysisUniverse.java:271)
	at com.oracle.graal.pointsto.meta.AnalysisUniverse.lookupAllowUnresolved(AnalysisUniverse.java:209)
	at com.oracle.graal.pointsto.infrastructure.WrappedConstantPool.lookupType(WrappedConstantPool.java:164)
[...]

PR #1965, opened 2020-10-28
Added the joni optional dependency in order fix the native build #1961

aldettinger:


oscerd (approved):

aldettinger:
Proposal fix for issue https://github.com/apache/camel-quarkus/issues/1961.
The ECMA 262 feature is not exposed from camel-json-validator. However, I've found clean maintenance way to cut the problematic code path with native substitution.
So, I propose to take the option dependency.

@jamesnetherton What do you think ?

jamesnetherton:
> So, I propose to take the optional dependency to please native image static points to analysis.

I think that's fine for now. We can revisit in furture if we need to.

PR #1974, opened 2020-10-30
Automate the process of creating jvm only extensions

jamesnetherton:
This is a bit of an experiment which we can always remove if it ends up being not useful.

It lets you create JVM only extensions for a list of camel components. To run you'd do:

Actions -> Generate JVM Extensions -> Run Workflow

Then enter a list of artifact ids to add. It'll run the cq-maven-plugin, make sure the tests pass & create a PR.

### DIFF
--- a/.github/workflows/generate-jvm-extension.yaml
+++ b/.github/workflows/generate-jvm-extension.yaml
@@ -1,0 +1,82 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


oscerd (approved):

.github/workflows/generate-jvm-extension.yaml
```diff
@@ -0,0 +1,82 @@
...
+# limitations under the License.
+#
+
+name: Generate JVM Extensions
+
+on:
+  workflow_dispatch:
+    inputs:
+      artifactIds:
+        description: 'Comma separated list of Camel component artifact ids'
```
ppalaga, line 24: An example would be handy to see whether they should be with or without the `camel-` prefix.

ppalaga (approved):

.github/workflows/generate-jvm-extension.yaml
jamesnetherton: It can be with or without, as the `camel-` prefix is stripped out. But I can tweak the description so that it's obvious.

ppalaga: Nifty!

jamesnetherton:
Latest commit updates the input parameter description.

I'll merge and try it out on `aws2-eventbridge`.

Issue #1969, opened 2020-10-29
Configuration option (not) to start the runtime


ppalaga:
The RuntimeBuildItem has an option to configure if the runtime has to be started or not but we don't have any way to set it through a property.

jamesnetherton:
Is there anything else needed for this?

ppalaga:
Nope, thanks for checking.

PR #1977, opened 2020-10-30
Configuration option (not) to start the runtime #1969

ppalaga:


ppalaga:
@lburgazzoli would this be what you need?

oscerd (approved):

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
```diff
@@ -159,6 +159,7 @@ public CamelRuntimeBuildItem runtime(
         }
 
         return new CamelRuntimeBuildItem(
-                recorder.createRuntime(beanContainer.getValue(), context.getCamelContext()));
+                recorder.createRuntime(beanContainer.getValue(), context.getCamelContext()),
+                config.bootstrap.enabled);
```
lburgazzoli, line 163: should this option be renamed to auto-start ? 
it is perfectly fine for me to leave as it is but not sure if it is clear for evenryone

ppalaga: I think we do not need to rename as long as the option docs explains the option properly, which I think it does.

lburgazzoli (approved):

PR #1988, opened 2020-11-04
WireMockTestResourceLifecycleManager improvements

jamesnetherton:


oscerd (approved):

integration-tests-support/wiremock/src/main/java/org/apache/camel/quarkus/test/wiremock/WireMockTestResourceLifecycleManager.java
```diff
@@ -78,30 +82,35 @@
...
+                                }
                             }
                         }
                     }
                 }
+                server.stop();
             }
-            server.stop();
+        } finally {
+            servers.remove(getServerKey());
```
ppalaga, line 113: How is the caching supposed to work, when you remove the server on every stop()? I thought the cache was meant to keep the same server instance for multiple test classes referring to the same TestResourceLifecycleManager. Maybe you meant something different?

ppalaga (commented):
A question inline

integration-tests-support/wiremock/src/main/java/org/apache/camel/quarkus/test/wiremock/WireMockTestResourceLifecycleManager.java
jamesnetherton: If an integration test has multiple test classes annotated with `@TestResource(Foo.class)`. Then `start()` / `stop()` is only invoked once. I.e when the first test starts and the last test completes.

ppalaga: I see. Is the TestResourceLifecycleManager instance on which start() and stop() are called different?
A comment somewhere would be nice so that future readers do not need to wonder.

jamesnetherton: > the TestResourceLifecycleManager instance on which start() and stop() are called different?

No it's the same.

I have tricked myself into thinking that caching was needed. It was something odd in the Geocoder tests that made me go down this path.

Let me rework things a little. I can probably drop the first commit entirely...

jamesnetherton:
Thanks for the review, it helped to clear my head a bit :+1: 

I'll close this and make d37e453 part of another PR where I plan port some itests over to WireMock.

Issue #1726, opened 2020-09-07
Nagios native support


ppalaga:

aldettinger:
Follow up filed for possibly adding real tests in #1992.

PR #1991, opened 2020-11-04
Added nagios native support #1726

aldettinger:


oscerd (approved):

ppalaga:
Overall, this looks good, thanks @aldettinger !

I'd like to ask about some details:

* Since recently, we try to make tests runnable in two modes: real backend service vs. mocked backend. There is also some log message based auditing convention so that interested parties can make sure that the tests were actually run against real services. IIRC, Twillio is one of the recent examples. The current proposal seems to support only the mock mode. I wonder whether the real mode could be supported too? cc @llowinge 
* @jamesnetherton recently started to use WireMock for mocking backend services. I wonder whether it would also make sense to use it also here? I have no idea whether WireMock supports raw TCP.

jamesnetherton:
> * I have no idea whether WireMock supports raw TCP.

Its for HTTP only IIRC.

aldettinger:
Thanks for paying attention to this one guys.
My current understanding is that twilio itests uses wiremock which supports only HTTP based mock server.

In the case at hand, I've explored different options to run a test against a real NSCA server to no avail. So, I've ended up simulating the protocol with camel-netty. On top of that, do we see any improvement/alignment for this PR ?

ppalaga:
> Thanks for paying attention to this one guys.
> My current understanding is that twilio itests uses wiremock which supports only HTTP based mock server.

OK, let's forget about WireMock for raw tC

> In the case at hand, I've explored different options to run a test against a real NSCA server to no avail.

The test does not need to manage the real service. It may well be the responsibility of the testing person to set it up, start it and pass some env vars or similar local config to the test to avoid running against the mock. I guess something like that would be possible with NSCA server?

aldettinger:
It should be possible, but I was not able to set it up in reasonable time. I wonder if we should file a follow up issue ? maybe even in camel ?

ppalaga:
+1, let's file a followup and let @llowinge express his opinion.

aldettinger:
I'm filing #1992 as follow up

PR #1993, opened 2020-11-04
Migrate ServiceNow, Slack, Geocoder & Telegram tests to WireMock

jamesnetherton:


extensions/geocoder/deployment/src/main/java/org/apache/camel/quarkus/component/geocoder/deployment/GeocoderProcessor.java
```diff
@@ -50,6 +50,7 @@ ExtensionSslNativeSupportBuildItem activateSslNativeSupport() {
         items.add(new ReflectiveClassBuildItem(false, true, "com.google.maps.model.Bounds"));
         items.add(new ReflectiveClassBuildItem(false, true, "com.google.maps.model.LatLng"));
         items.add(new ReflectiveClassBuildItem(false, true, "com.google.maps.model.LocationType"));
+        items.add(new ReflectiveClassBuildItem(false, true, "com.google.maps.model.GeolocationPayload"));
```
ppalaga, line 53: Your effort pays back!

integration-tests/telegram/pom.xml
```diff
@@ -56,17 +56,18 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-attachments</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.camel.quarkus</groupId>
-            <artifactId>camel-quarkus-integration-test-support-mock-backend</artifactId>
-        </dependency>
```
ppalaga, line 62: The auditing idea was based on the assumption that the number of <artifactId>camel-quarkus-integration-test-support-mock-backend</artifactId> occurrences in poms must be equal to the number of the magic log tokens. Now we are stopping to use <artifactId>camel-quarkus-integration-test-support-mock-backend</artifactId> . 

Yes, <artifactId>camel-quarkus-integration-wiremock-support</artifactId> is a replacement, but probably not all mock will be HTTP, so the interested parties (cc @llowinge ) will have to add the number of <artifactId>camel-quarkus-integration-wiremock-support</artifactId> occurrences to the number of <artifactId>camel-quarkus-integration-test-support-mock-backend</artifactId>.

ppalaga (approved):

integration-tests/telegram/pom.xml
jamesnetherton: Let me think about this some more because there are probably other (better) ways of achieving this.

oscerd (approved):

Issue #1973, opened 2020-10-30
[Camel 3.7] JSON-B data format support


ppalaga:
JSON-B is new in Camel 3.7 https://github.com/apache/camel/pull/4452/files

JiriOndrusek:
Please, assign to me.

ppalaga:
Fixed in camel-master

PR #2001, opened 2020-11-09
[Camel 3.7] JSON-B data format support

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/1973

Camel dataformat for json-b doesn't contain implementation for jsonb. As you can see in `configuration.adoc` and in a test module, I've used johnzon implementation of jsonb.

@ppalaga I can imagine also creation of jsonb extension containing johnzon implementation - which would be easier to use (without reflection configuration). What do you think?


oscerd (approved):

jamesnetherton:
> I can imagine also creation of jsonb extension containing johnzon implementation - which would be easier to use (without reflection configuration). What do you think?

Can we not use `quarkus-jsonb`?

JiriOndrusek:
@jamesnetherton Unfortunatelly using `quarkus-jsonb` is not a solution. `Camel-jsonb` is based on geronimo_spec (which is by the way using Johnzon as default implementation, see https://github.com/apache/geronimo-specs/blob/trunk/geronimo-jsonb_1.0_spec/src/main/java/javax/json/bind/spi/JsonbProvider.java#L30).

 `Quarkus-jsonb` is using `yasson` (see https://github.com/quarkusio/quarkus/blob/master/extensions/jsonb/runtime/pom.xml#L19). Yasson implements `javax.json.bind:javax.json.bind-api` not `org.apache.geronimo.specs:geronimo-jsonb_1.0_spec`.

jamesnetherton:
But aren't they both implementations of JSR-367? So plugging either library in should work so long as they both conform to the spec(s).

Couldn't we exclude the `geronimo` bits and use `quarkus-jsonb`? Or am I missing something?

JiriOndrusek:
@jamesnetherton  I agree with you. `camel-jsonb` which this extension is for is using geronimo_spec, so excluding geronimo didn't come to my mind. 

But from my POV geronimo should be removed from `camel-jsonb`, which will allow `yasson`. Then similarly as here (https://www.codeflow.site/de/article/java-json-binding-api) yasson could be used (without need of any exclusion) or johnzon (but with addition of geronimo spec)

I've tested it locally and without geronimo we can use yasson. So there is only a question, whether we want

1. exlude geronimo with yasson
2. or include geronimo with johnzon

(option 1 seems to be more usual from my POV)

JiriOndrusek:
Do not merge this PR please, several changes will be done.

JiriOndrusek:
Dependency `io.quarkus:quarkus-jsonb` is added to the extension, so no configuration is required from users.

PR could be merged.

integration-tests/jsonb/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+            <type>pom</type>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <!-- The following dependencies guarantee that this module is built after them. You can update them by running `mvn process-resources -Pformat -N` from the source tree root directory -->
```
jamesnetherton, line 92: Just one little nit - can you remove the duplicated comment.

JiriOndrusek: Thanks, I missed that - it is fixed now.

Issue #1941, opened 2020-10-21
Camel Avro RPC component native support


ppalaga:
This one is for camel-avro-rpc component. We already have the avro dataformat in native https://github.com/apache/camel-quarkus/issues/782

JiriOndrusek:
Please, assign to me.

PR #2025, opened 2020-11-24
Camel Avro RPC component native support #1941

JiriOndrusek:
fixes: https://github.com/apache/camel-quarkus/issues/1941

Change overrides version of avro dependencies from 1.9.1 (camel) to 1.10.0 (used by quarkus). Once issue https://issues.apache.org/jira/browse/CAMEL-15879 is solved, restriction could be removed. But version has to be still aligned manually, because camel uses dependencies like `avro-ipc*`, which are not used in quarkus.

There is a "questionable" substitution - `GenericDatumReaderSubstitute`.
As you can see, there is a method `getStringClass` in avro (see https://github.com/apache/avro/blob/branch-1.10/lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java#L503), which uses local cache created right during instantiation (see https://github.com/apache/avro/blob/branch-1.10/lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java#L500).
This cache is **null** (in some tests) during execution in native mode. I don't see how this is possible. But I've created substitution, which creates this cache if it is null. With this substitution tests are successful. (but I still think that it should work without this substitution, which obviously does not)
From my POV it could be nice, to create a separate issue to investigate this behavior. WDYT? 
  


integration-tests/avro-rpc/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+                </plugins>
+            </build>
+        </profile>
+
+        <!-- Profile for generation of protocol files into package `org.apache.camel.quarkus.component.avro.rpc.it.specific.generated` -->
+        <profile>
+            <id>avro</id>
+            <activation>
+                <property>
+                    <name>avro</name>
```
ppalaga, line 123: This should either be active by default (and inactive in `quickly`) or we should at least document here under which circumstances should the classes be regenerated.

integration-tests/avro-rpc/src/main/java/org/apache/camel/quarkus/component/avro/rpc/it/AvroRpcResource.java
```diff
@@ -0,0 +1,139 @@
...
+    @Inject
+    ProducerTemplate producerTemplate;
+
+    @Path("/reflectionProducerSet")
+    @POST
+    @Consumes(MediaType.TEXT_PLAIN)
+    public void reflectionProducerSet(@QueryParam("protocol") ProtocolType protocol, String name) throws Exception {
+        Object[] request = { name };
+        producerTemplate.requestBody(String.format(
+                "avro:%s:localhost:{{%s}}/setName?protocolClassName=%s&singleParameter=true",
```
ppalaga, line 64: I'd vote for configuring the endpoint in the same way here and in `AvroRpcRouteBuilder`. I personally find the `@ConfigProperty` fields in `AvroRpcRouteBuilder` a bit better readable than Camel property expansion.

ppalaga (commented):
Nice test coverage, @JiriOndrusek ! thanks! Some comments inline.

The matter with the substitution is indeed strange. +1 for filing a followup for that.
A field being null may happen eg. when a superclass is calling a method implemented in the subclass from the superclass'es constructor. Something like

```
abstract class Parent {

  Parent() {
    String s = getString(); // s is null, when this is called via new Child()
  }
  abstract String getString();
}

class Child extends Parent {

  private final String string = "foo";
  String getString() {
     return this.string;
  }
}
```

I am not sure there are other situations where this can happen.

pom.xml
```diff
@@ -41,6 +41,8 @@
         <ahc.version>2.12.1</ahc.version>
         <animal-sniffer.version>1.18</animal-sniffer.version>
         <antlr3.version>3.5.2</antlr3.version><!-- Spark, Stringtemplate and probably others -->
+	    <avro-ipc-jetty-version>1.10.0</avro-ipc-jetty-version>
+        <avro-ipc-netty-version>1.10.0</avro-ipc-netty-version>
```
jamesnetherton, line 45: If these ipc deps are always going to have the same version, lets have a single property for them.

integration-tests/avro-rpc/src/main/java/org/apache/camel/quarkus/component/avro/rpc/it/AvroRpcResource.java
JiriOndrusek: @ppalaga I'm not sure, what do you mean.
Here is endpoint form `AvroRpcResource.java`
>String.format(
                "avro:%s:localhost:{{%s}}/setName?protocolClassName=%s&singleParameter=true",
                protocol,
                protocol == ProtocolType.http ? REFLECTIVE_HTTP_SERVER_PORT_PARAM : REFLECTIVE_NETTY_SERVER_PORT_PARAM,
                TestReflection.class.getCanonicalName())

Here is a route definition from 'AvroRpcRouteBuilder'
>from(String.format("avro:http:localhost:%d/setTestPojo?protocolClassName=%s&singleParameter=true", httpPort,
                TestReflection.class.getCanonicalName()))
                        ....

I don't see much difference here. May I ask you for more information?

integration-tests/avro-rpc/pom.xml
JiriOndrusek: I've made it active by default (and inactive in `quickly)`

pom.xml
JiriOndrusek: I changed it into single property

integration-tests/avro-rpc/src/main/java/org/apache/camel/quarkus/component/avro/rpc/it/AvroRpcResource.java
ppalaga: My point is that there's perhaps much more future readers of this code who know MicroProfile config that those who know Camel property expansion `{{my.property}}`. So you could do here what you do in `AvroRpcRouteBuilder`:

1. Add at the ConfigProperty fields

```
@ConfigProperty(name = AvroRpcResource.REFLECTIVE_HTTP_TRANSCEIVER_PORT_PARAM)
    Integer httpPort;

@ConfigProperty(name = AvroRpcResource.REFLECTIVE_NETTY_TRANSCEIVER_PORT_PARAM)
    Integer nettyPort;
```

2. Change the handler

```
producerTemplate.requestBody(String.format(
                "avro:%s:localhost:%d/setName?protocolClassName=%s&singleParameter=true",
                protocol,
                protocol == ProtocolType.http ? httpPort: nettyPort,
                TestReflection.class.getCanonicalName()), request);
```

WDYT?

JiriOndrusek: ok, now I understand and agree, thanks

JiriOndrusek: done

ppalaga (approved):
Thanks @JiriOndrusek!

JiriOndrusek:
Current build problem does not reate wit avro-rpc - https://github.com/apache/camel-quarkus/pull/2025/checks?check_run_id=1453121702#step:10:1528

PR #2048, opened 2020-12-01
CSimple language support

### LINKED ISSUE #2036: CSimple language support
https://camel.apache.org/components/latest/languages/csimple-language.html
https://github.com/apache/camel/blob/81f8d1dd4ce6371285af3310a7574e53677ff638/core/camel-core/src/test/java/org/apache/camel/language/csimple/CSimpleTest.java#L23
https://github.com/apache/camel-examples/tree/master/examples/camel-example-csimple

ppalaga:
Fix #2036

A draft to discuss.

Depends on https://github.com/apache/camel/pull/4708


davsclaus (commented):
Whoa this is really fantastic Peter - great to see that csimple part of the quarkus build process and with discovery and native compilation. 

Keep on the great work. I would like to see a quarkus-example use csimple which we can have at quarkus-examples git repo. And see this from "user point of view" with both jvm and native compiled working.

ppalaga:
I am improving some parts after the feedback from @jamesnetherton

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CSimpleLanguageProcessor.java
```diff
@@ -0,0 +1,455 @@
...
+                final Class<?> cl = loader.loadClass(className);
+
+                if (RouteBuilder.class.isAssignableFrom(cl)) {
+                    try {
+                        final RouteBuilder rb = (RouteBuilder) cl.newInstance();
+                        rb.configure();
+                        final Map<String, Boolean> expressions = collector.collect(
+                                "csimple",
+                                rb.getRouteCollection(),
+                                rb.getRestCollection());
```
ppalaga, line 114: @jamesnetherton this is new since yesterday. Let's call it pure man's build time routes introspection. So far it works only for `org.apache.camel.builder.RouteBuilder`. We simply instantiate the classes, call `.configure()` and introspect its RouteCollection and RestCollection. Under the hoot, it is using JAXB, so the underlying code can be used also for XML.

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CSimpleLanguageProcessor.java
```diff
@@ -0,0 +1,455 @@
...
+            final ExpressionCollector collector = new ExpressionCollector(loader);
+
+            for (CamelRoutesBuilderClassBuildItem routesBuilderClass : routesBuilderClasses) {
+                final String className = routesBuilderClass.getDotName().toString();
+                final Class<?> cl = loader.loadClass(className);
+
+                if (RouteBuilder.class.isAssignableFrom(cl)) {
+                    try {
+                        final RouteBuilder rb = (RouteBuilder) cl.newInstance();
+                        rb.configure();
```
lburgazzoli, line 110: I'm not very sure if this is always enough or if you need to set a camel context, as example if the route invokes [this method](https://github.com/apache/camel/blob/master/core/camel-core-model/src/main/java/org/apache/camel/builder/RouteBuilder.java#L286), then you'll get an NPE. In general the builder assumes that the camel context is set when `configure()`.

ppalaga: Yeah, thanks for the feedback. I thought something like this could be the case. I think we can set a dummy context to support a couple of more cases. For the rest, I'd say, at this experimental stage, we simply won't compile their expressions and either make them fail at runtime (recommending to use simple language instead) or have some runtime "interpreted" fallback. In the long term we'd probably need to redesign the existing route builder or add a new kind of route builder such that it would allow introspecting the route defs without requiring any runtime info.

lburgazzoli: The problem is what the user would do in the `configure()` as you can for example, bind things to the registry

ppalaga:
Already cherry-picked in camel-master

PR #2056, opened 2020-12-04
Polish how we pass -Djavax.net.ssl.trustStore to the Solr test #2029

### LINKED ISSUE #2029: Polish how we pass -Djavax.net.ssl.trustStore to the Solr test
Followup https://github.com/apache/camel-quarkus/pull/2026#discussion_r529877052

ppalaga:
Fix #2029

Could you please review, @zbendhiba ?


oscerd (approved):

zbendhiba:
@ppalaga For JVM, mode it looks good to me. I can't tell for native mode.

ppalaga:
Ah, the native test would work if the JVM would be run in advance, but apparently on the CI this is not the case. Let me fix it.

ppalaga:
In 08e7dc6 I abandoned the idea with JUnit extension. All is done via Maven pom.xml. I think it is easier to follow like that. The downside is that some special pom setup will be needed on the Platform. Could you please re-review, @zbendhiba ?

ppalaga:
I have filed https://github.com/quarkusio/quarkus/issues/13752 and added [this](https://github.com/apache/camel-quarkus/pull/2056/commits/9b6accb4343d44ab38f21fd6ea9fac39b30d0973#diff-9a5b2e1a1cad09bff38e48afe7731592b775f58b95cb353f1d15e970ddf6c20eR102-R107) and [this](https://github.com/apache/camel-quarkus/pull/2056/commits/9b6accb4343d44ab38f21fd6ea9fac39b30d0973#diff-bfe8c44974870081fc957f86c599dedf9ef2116a3d8a2d8c6fd0e0bbf7f190dcR21) as a workaround.

PR #2088, opened 2020-12-21
Added JSONata extension

aldettinger:


integration-tests/jsonata/src/main/resources/spec/expressions.json
```diff
@@ -0,0 +1,5 @@
+{
+    "name": FirstName & " " & Surname,
+    "mobile": Phone[type = "mobile"].number
```
ppalaga, line 3: I see that this is supposed to be some sort of JSONata transformation recipe, but I still wonder whether it is valid JSON?

oscerd (approved):

pom.xml
```diff
@@ -60,7 +60,7 @@
         <google-http-client.version>1.34.0</google-http-client.version>
         <google-auth-library-oauth2-http.version>0.21.1</google-auth-library-oauth2-http.version> <!-- Mess in google-cloud-bom -->
         <guava.version>29.0-jre</guava.version>
-        <gson.version>2.8.5</gson.version>
+        <gson.version>2.8.6</gson.version><!-- JSONata -->
```
ppalaga, line 63: This is OK for now, but in the long term, we should aim at inheriting `${gson-version}` from camel-dependencies here. Could you plz. perform the necessary changes in Camel so that all their components dependent on Gson use their `${gson-version}` that should be upgraded to 2.8.6?

ppalaga (approved):

integration-tests/jsonata/src/main/resources/spec/expressions.json
aldettinger: It's not valid JSON. The best term I could find is "spec" from the camel-jsontata documentation, so I would propose to rename it to expressions.spec then. What do you think ?

pom.xml
aldettinger: That could be discussed in [CAMEL-15981](https://issues.apache.org/jira/browse/CAMEL-15981).

integration-tests/jsonata/src/main/resources/spec/expressions.json
ppalaga: Yes, I think using somethig else than .json would be less misleading for both users and IDEs/editors :)

pom.xml
ppalaga: Could you plz change the comment to something like 

```
<!-- 2.8.6 is required by JSONata; replace with ${gson-version} from camel-dependencies once https://issues.apache.org/jira/browse/CAMEL-15981 is fixed -->
```

oscerd: 2.8.6 doesn't work in OSGi. This is why is not updated.

ppalaga: Thanks for the info. Is there a Gson issue so that we can track the progress?

oscerd: https://github.com/google/gson/issues/1677

This one is the latest one. It seems they will solve with 2.8.7, but there is no ETA

Issue #2040, opened 2020-11-27
Minio native support


jamesnetherton:
https://camel.apache.org/components/latest/minio-component.html

There's a Minio extension in the quarkiverse that could help with this:

https://github.com/quarkiverse/quarkiverse-minio

JiriOndrusek:
Please, assign to me.

PR #2102, opened 2021-01-04
Minio native support #2040

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/2040


oscerd (approved):

jamesnetherton (commented):
I've not done a full review yet, but I have some comments / questions.

* Seems `mysql-connector-java-8.0.20.jar` was committed by mistake.

* Any reason why we are not leveraging `quarkiverse-minio`? In theory it has already solved the Minio native support, so we'd not have to duplicate it here.

JiriOndrusek:
@jamesnetherton thanks for the reviewing

- `mysql-connector-java-8.0` was commited by mistake, it should be now reverted

- about `quarkiverse-minio` - this is my mistake. I was preparing this extension before the Christmas, when camel was using older version of minio (7.1.4), quarkiverse is using 8.x. Today the version from camel is 8.0.3, which is the same as in quarkiverse. Today I've fixed my previous work and using quarkiverse doesn't occur to me.
- I try to look into a possibility to use existing extension and not to duplicate some code.

JiriOndrusek:
@jamesnetherton I'm facing a problem with quarkiverse-minio. Extension is meant to create minio client (seems to be the main purpose of extension). It requires mandatory configuration for the client - [source](https://github.com/quarkiverse/quarkiverse-minio/blob/master/runtime/src/main/java/io/quarkiverse/minio/client/MinioProducer.java).

It should be possible to use quakiverse extension from cq if this client was not present (camel component is able to create its own client for each endpoint if necessary [source](https://github.com/apache/camel/blob/master/components/camel-minio/src/main/java/org/apache/camel/component/minio/MinioEndpoint.java#L87)

I can possibly imagine solution with quarkiverse, when the client created by quarkiverse could be used by camel component. But this approach would require to fill connection parameters even if user then decides to use some kind of dynamic values in endpoint parameters.

WDYT?

jamesnetherton:
> even if user then decides to use some kind of dynamic values in endpoint parameters

I actually encountered a similar thing when I implemented Google BigQuery native support, which leverages a quarkiverse extension. There's a `projectId` config param that has to be specified:

https://github.com/apache/camel-quarkus/blob/master/integration-tests/google-bigquery/src/main/resources/application.properties#L19

For the initial implementation we could state in the docs that it's mandatory to configure the Minio client using the Quarkus way and Camel will autowire it into the Mino component. Then later we revisit this and try to improve things.

Alternatively, we could forget about leveraging the quarkiverse extension altogether and just do our own thing (not my preference).

@ppalaga @aldettinger @zbendhiba WDYT?

aldettinger:
I would stick to quarkiverse as far as possible. Chances are that we'll find at trick to handle those kind of situations later on.

ppalaga:
Yes, I'd also prefer reusing the quarkiverse extension as far as possible, possibly proposing changes there to get a better support for our use cases. Doing our own thing should be reserved for situations where there are serious technical/design obstacles between us and them.

JiriOndrusek:
Thanks for the suggestions. I'll use quarkiverse extension. 

I'm closing this PR because current solution is not the preferred one.

Issue #2098, opened 2020-12-30
CSimple NPE even if CSimple language is not used


ennishol:
The route below runs into NPE while packaging a project:
```java
@ApplicationScoped
public class MyRoute extends EndpointRouteBuilder {
    @Inject
    MyProcessor myProcessor;

    @Override
    public void configure() {
        from("direct:start").process(myProcessor::process);  }

    @ApplicationScoped
    public static class MyProcessor {
        public String process(Exchange el) {    return "";   }
    }
}
```
As confirmed by @davsclaus the route does not use csimple at all, so that quarkus builder
shouldn't cause that NPE bug. 

Using lambda works however .process(e -> myProcessor.process(e))

```sh
[WARN] [org.apache.camel.quarkus.core.deployment.CSimpleRouteDefinitionProcessor]
Could not extract CSimple language expressions from the route
definition Routes: [Route[From[direct:start] -> []]] in class class
com.simple.temp.MyRoute.
java.lang.NullPointerException: null
        at java.base/java.util.Objects.requireNonNull(Objects.java:222)
        at com.simple.temp.MyRoute.configure(MyRoute.java:17)
        at org.apache.camel.quarkus.core.deployment.CSimpleRouteDefinitionProcessor.collectCSimpleExpresions(CSimpleRouteDefinitionProcessor.java:115)
        at jdk.internal.reflect.GeneratedMethodAccessor305.invoke(Unknown
Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:564)
        at io.quarkus.deployment.ExtensionLoader$2.execute(ExtensionLoader.java:972)
        at io.quarkus.builder.BuildContext.run(BuildContext.java:277)
        at org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
        at org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:2046)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1578)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1452)
        at java.base/java.lang.Thread.run(Thread.java:832)
[...]

PR #2103, opened 2021-01-05
CSimple NPE even if CSimple language is not used

ppalaga:
Related: #2098

This is a PoC how we could better support the build time analysis of Camel route definitions. The main idea is to have a new kind of model builder (called `CamelDefinitionsBuilder` in this proposal) that would make the build time inspection explicit for the end users. 
Instead of plain `configure()` in the existing Camel `RouteBuilder`, there is `configure(DefinitionsBuilderContext)` in the new `CamelDefinitionsBuilder`. `DefinitionsBuilderContext.getPhase()` can be used by end users to figure out whether the builder is called at build time or at runtime. They can thus add some logic (if necessary) to avoid accessing runtime resources at build time, including CamelContext available via `CamelDefinitionsBuilder.DefinitionsBuilderContext.getContext()`.

Pros:

* The way of writing routes stays very close to the existing `RouteBuilder` 's Java DSL.
* The new `CamelDefinitionsBuilder` are opt-in. Who decides for them, explicitly accepts both the perf. benefits and any eventual development overhead.

Cons:

* This solution makes the end users responsible for making their routes inspectable at build time. This may result in unclean code full of ifs.
* The API does not prevent end users from writing non-working code. That would require a redesign of the Java DSL with properly modeled runtime resources. Hard to say without trying whether something like that is even possible.

Open questions/issues:

* Figure out how would this or any other alternative solution play with CDI injection of fields in `CamelDefinitionsBuilder`s.

### DIFF
--- a/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CSimpleRouteDefinitionProcessor.java
+++ b/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CSimpleRouteDefinitionProcessor.java
@@ -26,6 +26,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


jamesnetherton:
In general my preference for fixing #2098 would be to put the CSimple code into a separate extension and make it a conscious choice for the user to opt-in. That'll buy us some time to figure out the best way to implement build time introspection.

ppalaga:
> In general my preference for fixing #2098 would be to put the CSimple code into a separate extension and make it a conscious choice for the user to opt-in. That'll buy us some time to figure out the best way to implement build time introspection.

+1, let me do that.

ppalaga:
Let me keep this open as a study material for now.

lburgazzoli:
@ppalaga 

I've been thinking a little bit about this and one of the options could be that we create a lightweight route builder that has some limitations, like it does not provide a camel context:

| Type               |  Kind                                                        |
|--------------------|--------------------------------------------------------------|
| RoutesBuilder      |  interface                                                   |
| SimpleRouteBuilder |  provide limited functionalities, no access to camel context |
| RouteBuilder       |  same as today                                               |

Then camel-quarkus can safely inspect `SimpleRouteBuilder` at build time so it is up to the user to consciously pick the builder that fits better. This may also help cleaning up the `RouteBuilder` a bit.

I'm not convinced that giving to the user an option to know in which phase the `configure()` method is invoke would be a good options as, IMHO, this detail should no leak to the end user.

Another options would be to have some sort of proxy for some of the bits such the camel context (similar to what @gnodet did for the "camel context lite" to deal with build vs runtime properties) but I don't know how complex to implement and maintain it would be.

PR #2108, opened 2021-01-06
CSimple NPE even if CSimple language is not used 

ppalaga:
Fix #2103


jamesnetherton (approved):

docs/modules/ROOT/pages/reference/extensions/csimple.adoc
```diff
@@ -0,0 +1,56 @@
...
+:cq-status: Stable
+:cq-description: Aggregation repository using csimple as datastore
+:cq-deprecated: false
+:cq-jvm-since: 1.5.0
+:cq-native-since: 1.5.0
+
+[.badges]
+[.badge-key]##JVM since##[.badge-supported]##1.5.0## [.badge-key]##Native since##[.badge-supported]##1.5.0##
+
+Aggregation repository using csimple as datastore
```
davsclaus, line 15: I think this description is not correct

ppalaga: Good catch, thanks!

ppalaga: Fixed in 5d7dac1

PR #2117, opened 2021-01-08
Remove docker.io prefix for images

llowinge:
With docker.io prefix the auth from `docker login` is not working in testcontainers because this line (https://github.com/docker-java/docker-java/blob/master/docker-java-core/src/main/java/com/github/dockerjava/core/DockerConfigFile.java#L51) compare with `https://index.docker.io/v1/` which fails and no auth is used. And from documentation (https://www.testcontainers.org/quickstart/junit_5_quickstart#2-get-testcontainers-to-run-a-redis-container-during-our-tests) i can see there is no docker.io prefix used either.

wdyt @ppalaga @jamesnetherton should i raise issue on testcontainers/docker-java as well ?


ppalaga (approved):
> wdyt @ppalaga @jamesnetherton should i raise issue on testcontainers/docker-java as well ?

I don't think it is necessary.

Issue #2134, opened 2021-01-14
Reimplement creation of minio client to avoid minio extension limitations


JiriOndrusek:
With this issue (https://github.com/quarkiverse/quarkiverse-minio/issues/22) solved, it should be possible to change obtaining mechanism of miio client and therefore remove limitations caused by current implementation.

JiriOndrusek:
Please, assign to me.

JiriOndrusek:
Issue is fixed by https://github.com/apache/camel-quarkus/pull/2140/

PR #2140, opened 2021-01-15
Avoid automatic minio client autowiring #2134

JiriOndrusek:
fixes https://issues.apache.org/jira/browse/CAMEL-16037

### DIFF
--- a/docs/modules/ROOT/pages/reference/extensions/minio.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/minio.adoc
@@ -32,13 +32,22 @@ Please refer to the above link for usage and configuration details.
 
 Check the xref:user-guide/index.adoc[User guide] for more information about writing Camel Quarkus applications.
 
-== Camel Quarkus limitations
+== Additional Camel Quarkus configuration
 
-This extension leverages the Quarkiverse Minio. Standard application.properties mechanism is used to define connection (see http://github.com/quarkiverse/quarkiverse-minio#configuration-reference[documentation]).
+Depending on Minio configuration, this extension may require SSL encryption on its connections. In such cases, you will need
+to add `quarkus.ssl.native=true` to your `application.properties`.
+See also https://quarkus.io/guides/native-and-ssl[Quarkus native SSL guide] and xref:user-guide/native-mode.adoc[Native mode]
+section of Camel Quarkus user guide.
 
-It is mandatory to configure the Minio client in properties and Camel will autowire client into the Minio component.
-Configuration via application.properties allows definition of only one minio client, therefore it isn't possible to define several different minio endpoints, which run together.
+There are two different configuration approaches:
 
-Please upvote https://github.com/quarkiverse/quarkiverse-minio/issues/22[this issue]
-if you do not like the present state.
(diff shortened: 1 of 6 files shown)


oscerd (approved):

jamesnetherton (commented):
Thinking more about what this PR is trying to solve, maybe there is a simpler solution....

If multiple Minio clients are required, then we mandate that they must be configured and added to the camel registry (via a CDI producer / bean etc). Then they can be referenced on the endpoint URI like `minio:foo?minioClient=#clientA` & `minio:foo?minioClient=#clientB` etc. Autowiring does not occur when multiple instances are found (Camel checks this).

What this solution does is disable all autowiring (I.e for every component) when `quarkus.camel.minio.autowired-enabled=true`. I don't think we want that.

JiriOndrusek:
@jamesnetherton I agree. I see only one disadvantage - minio component is capable of creating clients, but this approach will ignore it and force user to create clients manually. From my POV it is a small "fee", but I 'd like to mention it.

jamesnetherton:
> minio component is capable of creating clients

Right - and for that, Camel already provides `camel.component.minio.autowired-enabled`.

JiriOndrusek:
I was reffering this code: https://github.com/apache/camel/blob/master/components/camel-minio/src/main/java/org/apache/camel/component/minio/MinioEndpoint.java#L173 (fallback when there is no autowired client)

extensions/minio/runtime/src/main/java/org/apache/camel/quarkus/component/minio/MinioRecorder.java
```diff
@@ -0,0 +1,45 @@
...
+        private final CamelMinioConfig config;
+
+        public MinioContextCustomizer(CamelMinioConfig config) {
+            this.config = config;
+        }
+
+        @Override
+        public void customize(CamelContext camelContext) {
+            if (config.autowiredEnabled.isPresent())
+                camelContext.setAutowiredEnabled(config.autowiredEnabled.get());
```
lburgazzoli, line 41: this would disable autowiring for any component, isn't it ?

JiriOndrusek: yes, now I understand what is wrong with it. I'll verify option to provide several clients into registry and change documentation accordingly (and remove this change, as it is not necessary)

JiriOndrusek:
@jamesnetherton @lburgazzoli 
I reworked fix. Now it contains only doc change + change in tests. Implementation is not affected.

JiriOndrusek:
Test failure is not caused by minio extension.

>org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-test) on project camel-quarkus-integration-test-hazelcast: There are test failures.

extensions/minio/runtime/src/main/java/org/apache/camel/quarkus/component/minio/MinioRecorder.java
JiriOndrusek: refactored so disabled autowiring is not suggested

PR #2178, opened 2021-01-29
Azure Event Hubs support

### LINKED ISSUE #2079: Azure Event Hubs support
https://camel.apache.org/components/latest/azure-eventhubs-component.html

ppalaga:
Fix #2079


oscerd (approved):

ppalaga:
@jamesnetherton the tests are currently run only against real Azure API. 

As for mocking I'd like to point out the following: 

* Wiremock is not enough because there is some qpid communication going on
* Azurite does not support eventhubs either (I guess it is for the same reason)

Actually the eventhubs client can be configured to use websockets. Maybe it would not be that hard to mock.

Given the above, do you think @jamesnetherton we need to aim at having mock tests at some point? I could file a followup with the above details. My standpoint is that we do not need to care as long as we know that @llowinge 's team tests against the real API.

@llowinge you are welcome to review/test too.

jamesnetherton:
> do you think we need to aim at having mock tests at some point?

Possibly... but perhaps it's not something that we should put any significant effort into. I.e maybe we could see if Azurite are willing to implement something on their side that we can eventually use.

Ultimately, it's probably not realistic to expect that we can mock & stub everything. There will always be some things that prevent this, like stuff that are non-free, paid for services etc.

ppalaga:
Filed https://github.com/Azure/Azurite/issues/690

ppalaga:
Filed https://github.com/apache/camel-quarkus/issues/2180

jamesnetherton (approved):

extensions/azure-eventhubs/deployment/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.amqphub.quarkus</groupId>
+                <artifactId>quarkus-qpid-jms-deployment</artifactId>
+                <version>${quarkus-qpid-jms.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <!-- quarkus-qpid-jms-deployment is currently not part quarkus-qpid-jms-deployment-->
```
aldettinger, line 44: Sorry, I don't get this one

integration-tests/azure-eventhubs/README.adoc
```diff
@@ -0,0 +1,18 @@
+== Azure Azure Event Hubs Service integration tests
```
aldettinger, line 1: Maybe an extra Azure above. The prerequisites section is really nice.

integration-tests/azure-eventhubs/src/test/java/org/apache/camel/quarkus/component/azure/eventhubs/it/AzureEventhubsTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+        Awaitility.await().pollInterval(1, TimeUnit.SECONDS).atMost(120, TimeUnit.SECONDS).until(() -> {
+            final String body = RestAssured.given()
+                    .get("/azure-eventhubs/receive-events")
+                    .then()
+                    .extract().body().asString();
+            LOG.infof("Expected message body: '%s'; actual: '%s'", messageBody, body);
+            return body != null && body.contains(messageBody);
+        });
+
```
aldettinger, line 58: Would the test fail when body is still null after 120s ?

aldettinger (changes requested):
Well done Peter, there is quite a lot of substitutions in there. My current understanding is that the test always passing may need change.

extensions/azure-eventhubs/deployment/pom.xml
ppalaga: It is copied from `camel-quarkus-qpid` (by @jamesnetherton IIRC) and I think it originally intended to say that quarkus-qpid-jms-deployment itself was not managed in quarkus-qpid-jms-deployment. It would be worth checking whether it is still the case.

ppalaga: Actually, importing quarkus-qpid-jms-deployment it is perhaps not necessary anymore. In some phase I experimented with depending on quarkus-qpid-jms to solve some problem that turned out to be solvable differently. Let me check if it can be removed.

integration-tests/azure-eventhubs/README.adoc
ppalaga: Yes, thanks!

extensions/azure-eventhubs/deployment/pom.xml
ppalaga: Yes, it can be removed.

integration-tests/azure-eventhubs/src/test/java/org/apache/camel/quarkus/component/azure/eventhubs/it/AzureEventhubsTest.java
ppalaga: `body != null && body.contains(messageBody)` is false as long as the first conjunct is false. So yes, the test would fail if body is still null after 120s.

extensions/azure-eventhubs/deployment/pom.xml
ppalaga: Thanks for noticing

ppalaga (commented):
ab5281d: Addressed the comments of @aldettinger (thanks!) except for the until() expression in the test which I believe was correct.

extensions/azure-eventhubs/deployment/pom.xml
aldettinger: Thanks to you Peter, I had no idea it should be removed but just find it strange :)

aldettinger (approved):
Yes it's correct.

PR #2206, opened 2021-02-05
Enable test back again

llowinge:
It works locally. Should be investigated why it fails on CI.


jamesnetherton:
I think @zbendhiba already looked into this. IIRC there are some fixes in Camel 3.8. The test is already enabled on our 'camel next' branch.

zbendhiba:
@jamesnetherton  yes, I confirm! It's related to a fix in camel 3.8

jamesnetherton:
We'll pick up this change when we upgrade to Camel 3.8.

Issue #1355, opened 2020-06-16
CassandraQL native support


aldettinger:

aldettinger:
Some hurdles in native mode but a quarkus cassandra extension is [on the way](https://github.com/quarkusio/quarkus/issues/7876) and may fix most issues.

JiriOndrusek:
Please, assign to me.

JiriOndrusek:
Fix is blocked by several issues:

- https://datastax-oss.atlassian.net/jira/software/c/projects/JAVA/issues/JAVA-2867
- https://github.com/datastax/cassandra-quarkus/issues/124
- https://github.com/datastax/cassandra-quarkus/issues/123

If mentioned issues are solved, it should be possible to use following changes to add support - https://github.com/JiriOndrusek/camel-quarkus/tree/QUARKUS1355_cassandra-native
(depending on driver version in cassandra-quarkus) it could be necessary to modify runtime/pom to align versions correctly.

JiriOndrusek:
Current status of issues:

- https://datastax-oss.atlassian.net/jira/software/c/projects/JAVA/issues/JAVA-2867 - fixed in driver version 4.9.0. (Camel componet is already updated)
- datastax/cassandra-quarkus#124 - problem could be solved by using parameter `eager-session-init`
- datastax/cassandra-quarkus#123 - next release would contain datastax driver 4.9. 
It should be possible to continue on the solutrion ( by forcing driver version 4.9 even if cassandra-quarkus depends on 4.7.2) 

I'll try to continue with this issue.

JiriOndrusek:
I've asked for an improvement on cassandra-quarkus (https://github.com/datastax/cassandra-quarkus/issues/135), which could make use of `eager-session-init` attribute much more friendly.

JiriOndrusek:
Issue is still blocked. Waiting for release  `1.0.0-alpha4` of `cassandra-quarkus`

ppalaga:
@JiriOndrusek waiting for 1.0.0-alpha4 of cassandra-quarkus may take long. I assume you have some tests in your branch that you could add to the JVM module straight away? That would increase our confidence that the extension works well in the JVM mode at least.

jamesnetherton:
cassandra-quarkus 1.0.0 is out now.

JiriOndrusek:
I'll look into it and try o finish my work

JiriOndrusek:
PR is ready to be merged.

[...]

PR #2219, opened 2021-02-09
CassandraQL native support #1355

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/1355


oscerd (approved):

integration-tests/cassandraql/src/main/java/org/apache/camel/quarkus/component/cassandraql/it/CassandraqlResource.java
```diff
@@ -0,0 +1,111 @@
...
+        final Exchange exchange = consumerTemplate
+                .receive(createUrl(String.format("UPDATE employee SET name = '%s', address = '%s' WHERE id = %s",
+                        employee.getName(), employee.getAddress(), employee.getId())));
+        return exchange != null;
+    }
+
+    @Path("/deleteEmployee")
+    @POST
+    @Consumes(MediaType.TEXT_PLAIN)
+    public void updateEmployee(String id) throws Exception {
```
aldettinger, line 90: What about renaming to something like getEmployeeById ?

JiriOndrusek: I see now, that there is a bad method name vs path, which does not make sense, I'll fix it

aldettinger (approved):
This PR looks good. I find it interesting to leverage on matchers for Employee. Well done :+1:

integration-tests/cassandraql/src/main/java/org/apache/camel/quarkus/component/cassandraql/it/CassandraqlResource.java
JiriOndrusek: Correct functionality is to delete by Id, I renamed method to `deleteEmployeeById`.

jamesnetherton (approved):

jamesnetherton:
The `camel-master` branch needs some tidying so its expected to see some CI failures for docs. I'm working to fix that up.

zbendhiba (approved):

integration-tests/cassandraql/src/test/java/org/apache/camel/quarkus/component/cassandraql/it/CassandraqlTest.java
```diff
@@ -0,0 +1,135 @@
...
+@QuarkusTestResource(CassandraqlTestResource.class)
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+class CassandraqlTest {
+
+    private Employee sheldon = new Employee(1, "Sheldon", "Alpha Centauri");
+    private Employee leonard = new Employee(2, "Leonard", "Earth");
+    private Employee irma = new Employee(3, "Irma", "Jupiter");
+
+    @Test
+    @Order(1)
```
ppalaga, line 45: Let's have some academic discussion about test structuring ;)

I see that splitting into the methods helps structuring and thus understanding the code.

I see that `@Order` is intended to overcome the fact that the test methods depend on effects caused by their predecessors. This helps when the whole class is executed without any explicit method includes/excludes.

However, `@Order` does not express the dependency. Therefore, when the first test fails, the successor tests are still executed and will most probably fail. They will all be reported as failed and I am not sure if there is any way to see quickly from the report which one of them is the root cause.

Also, the test methods cannot be executed independently via `mvn clean test -Dtest=CassandraqlTest#testUpdate` (except for the first one). 

Given the above, I wonder whether this way of structuring brings enough benefits. I tend to see this kind of situation as end-to-end scenario that is best packed into as single test method where the structure is marked only by comments. (This is not a request to change this PR.) What do others think? Are there any other options to consider?

jamesnetherton: Yes, personally I'm not a fan of `@Order`. Like you say, most of our tests follow a pattern when if multiple steps are required to achieve some expected outcome, then they are bundled together in a single test case and marked by comments. E.g:

```
// Create foo

// Retrieve foo

// Delete foo

// Confirm foo was deleted
```

I'd like to 'close down' the `camel-master` branch and get the 3.8 upgrade merged to master soon. So if we can merge this now, we can have a follow up issue to restructure things later if we think its necessary.

ppalaga (approved):

integration-tests/cassandraql/src/test/java/org/apache/camel/quarkus/component/cassandraql/it/CassandraqlTest.java
ppalaga: This PR is good to merge. Thanks for your opinion.

PR #2222, opened 2021-02-09
Test AWS 2 SNS properly

ppalaga:
@llowinge something for you again :)


oscerd (approved):

zbendhiba:
Maybe you could rename the classes named `*Sqs*`, since they handle now both *SQS* and *SNS* ?

ppalaga:
> Maybe you could rename the classes named `*Sqs*`, since they handle now both _SQS_ and _SNS_ ?

You are right, but I have already done some changes in my workspace that make amending this PR very impractical. I'll try to do it in my next PR.

PR #2230, opened 2021-02-10
Allow running AWS 2 tests both grouped and isolated

ppalaga:
This is the idea how the new AWS 2 tests could be run both grouped (`integration-tests/aws2-grouped` for faster runs on our CI) and in isolation (`integration-tests-aws2/*` for methodologicaly cleaner, but slower testing for interested parties).

The grouped module dynamically puts all sources from the individual aws test modules to its `target/[test-]classes`. It concatenates `application.properties` and some service files. 

The `pom.xml` of the grouped module is not generated yet. I'd add that functionality later to `mvn process-resources -Pformat -N` that we already have in our workflow.

I am quite satisfied with this approach. What do others think?

cc @llowinge


oscerd (approved):

jamesnetherton (commented):
Seems ok to me. 

Can you drop some notes about how this works into a README if you've not done so already?

ppalaga:
> Can you drop some notes about how this works into a README if you've not done so already?

Yep, README is a good idea :)

ppalaga:
95d0012:

* Added readme
* Added the grouped module to its parent

Issue #2128, opened 2021-01-12
camel-spring-rabbitmq - new component


davsclaus:
In Camel 3.8 onwards there is a better rabbitmq component in the camel-spring-rabbitmq.

I wonder if we can have this work with quarkus.

A vertx based component is planned for the future, but in the mean time if we can get this to work then that would be good for folks using rabbitmq
https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-16026

PR #2233, opened 2021-02-11
camel-spring-rabbitmq - new component #2128

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/2128

There were some auto-generated changes (after `mvn process-resources -Pformat`). I put them into the second commit.


extensions/spring-rabbitmq/runtime/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-spring-rabbitmq</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-support-commons-logging</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-reflect</artifactId>
```
jamesnetherton, line 67: `kotlin-reflect` may actually not be required if you add a dependency on `camel-quarkus-support-spring`. It also helps to ensure our Spring backed extensions play well with the Quarkus Spring stuff.

oscerd (approved):

integration-tests/spring-rabbitmq/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- The following dependencies guarantee that this module is built after them. You can update them by running `mvn process-resources -Pformat -N` from the source tree root directory -->
```
jamesnetherton, line 102: Duplicate comment - FYI @ppalaga, I think there's a bug in the cq-maven-plugin that causes this.

davsclaus (approved):

extensions/aws-ec2/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -31,3 +31,4 @@ metadata:
   - "integration"
   status:
   - "stable"
+  - "deprecated"
```
ppalaga, line 34: These deprecation changes would deserve a separate commit. They belong to the Camel 3.8 upgrade.

JiriOndrusek: @ppalaga  I've put deprecation changed into separated commit in this PR, should I create another PR?

ppalaga: This commit is OK

extensions/spring-rabbitmq/runtime/pom.xml
JiriOndrusek: ok, thanks for the information, I'll try it  - it could be much better solution.

JiriOndrusek: you are right, there is a substitution in `camel-quarkus-support-spring`, which makes kotlin not required. I've replaced the dependencies.

integration-tests/spring-rabbitmq/pom.xml
JiriOndrusek: I've removed duplicity.

ppalaga: Good to know. It would be nice to keep an eye on this when running `mvn process-resources -Pformat` and give hints about situations when it happens.

ppalaga (approved):

integration-tests/spring-rabbitmq/pom.xml
ppalaga: I think I was able to pinpoint it: it happens when the newly inserted dependeny is placed before  the exiting ones. In that case the plugin chooses a bd insert location.

Thanks for notifying me!

Issue #2289, opened 2021-02-25
Azure storage datalake service support


ppalaga:

ppalaga:
A JVM-only extension with a test is underway https://github.com/apache/camel-quarkus/pull/2305

ppalaga:
Native mode did not look trivial at the first sight. I gave up after I saw some sort of MS identity library to load platform dependent native  libs from a static initalizer.

PR #2305, opened 2021-03-02
Azure Storage Data Lake JVM support #2289

ppalaga:
Rel. #2289


oscerd (approved):

extensions-jvm/azure-storage-datalake/integration-test/src/test/java/org/apache/camel/quarkus/component/azure/storage/datalake/it/AzureStorageDatalakeTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+            RestAssured.given()
+                    .delete("/azure-storage-datalake/filesystem/" + filesystem + "/path/" + filename)
+                    .then()
+                    .statusCode(204);
+
+            RestAssured.given()
+                    .delete("/azure-storage-datalake/filesystem/" + filesystem)
+                    .then()
+                    .statusCode(204);
```
aldettinger, line 110: I wonder if we should assert the deletions here.

aldettinger (approved):
Looks good to me.

jamesnetherton (approved):

extensions-jvm/azure-storage-datalake/integration-test/src/test/java/org/apache/camel/quarkus/component/azure/storage/datalake/it/AzureStorageDatalakeTest.java
ppalaga: Yes, we could, although I wonder whether that level of paranoia wouldn't be better suited for the tests in Camel. I think here in Camel Quarkus the main purpose of the tests is to ensure that the main execution paths work flawlessly on Quarkus incl. native mode. If the delete request does not throw an exception, then I think the functionality can be considered proven. But yes, checking the state of the remote storage would be even safer.

aldettinger: Indeed, I agree it's not really the scope of camel-quarkus test.

ppalaga:
JVM tests and azure passed - that should be enough

PR #2334, opened 2021-03-15
Fix wrong format of RouteBuilder

llowinge:


oscerd (approved):

docs/modules/ROOT/pages/user-guide/cdi.adoc
```diff
@@ -4,7 +4,7 @@
 CDI plays a central role in Quarkus and Camel Quarkus offers a first class support for it too.
 
 You may use `@Inject`, `@ConfigProperty` and similar annotations e.g. to inject beans and configuration values to
-your Camel `RouteBuilder`s. Here is the `RouteBuilder` from our `timer-log-cdi` xref:user-guide/examples.adoc[example]:
+your Camel `RouteBuilder` s. Here is the `RouteBuilder` from our `timer-log-cdi` xref:user-guide/examples.adoc[example]:
```
zbendhiba, line 7: Maybe drop the s in `RouteBuilder` s ?

ppalaga: Have you checked the result after generating the docs, @llowinge? Asking because, IIRC, single backtick does not work in AsciiDoctor in case there is no whitespace after the closing backtick. Two backticks work if I am not mistaken:

```
``RouteBuilder``s
```

llowinge: Thanks both for your tips. I've modified with double backticks, it works !

zbendhiba: @llowinge thanks
The "s "still bothers me. When I copy/paste, I don't like the visual format. We have a lonely s far from it's RouteBuilder.
I would like to drop the "s", but I'm not that fluent in English to determine if it's correct.

jamesnetherton: Yeah, agree. Lets drop the plural form and just have `RouteBuilder`. Otherwise it looks weird with the dangling 's'.

llowinge: Fixed. How many Engineers are needed to resolve one wrong letter 🤣

PR #2337, opened 2021-03-16
Disable OptaplannerTest.solveSync on CI

jamesnetherton:
OptaPlanner solve sync test seems to fail quite frequently on CI so disabling it until there's a solution.


aldettinger (approved):

zbendhiba (approved):

jamesnetherton:
Somewhat ironically, the async test failed too :smile:  

So I'll disable that one also.

PR #2338, opened 2021-03-16
Add configuration option to ignore unknown arguments

jamesnetherton:
I set the default behaviour to throw an exception if unknown args are found.

Maybe the default should be to log a warning instead?

### DIFF
--- a/extensions-core/main/runtime/src/main/java/org/apache/camel/quarkus/main/CamelMainConfig.java
+++ b/extensions-core/main/runtime/src/main/java/org/apache/camel/quarkus/main/CamelMainConfig.java
@@ -22,6 +22,7 @@ import io.quarkus.runtime.annotations.ConfigGroup;
 import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigPhase;
 import io.quarkus.runtime.annotations.ConfigRoot;
+import org.apache.camel.quarkus.core.CamelConfig.FailureRemedy;
 
 @ConfigRoot(name = "camel.main", phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
 public class CamelMainConfig {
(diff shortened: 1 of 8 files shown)


extensions-core/main/runtime/src/main/java/org/apache/camel/quarkus/main/CamelMainConfig.java
```diff
@@ -40,4 +46,27 @@
...
+     *
+     * FAIL - Prints the {@link CamelMain} usage statement and throws a {@link RuntimeException}
+     * IGNORE - Suppresses any warnings and the application startup proceeds as normal
+     * WARN - Prints the {@link CamelMain} usage statement but allows the application startup to proceed as normal
+     */
+    public enum UnknownArgumentFailMode {
+        FAIL,
+        IGNORE,
+        WARN
+    }
```
ppalaga, line 71: There is `org.apache.camel.quarkus.core.CamelConfig.FailureRemedy` which I assume is visible here. Shouldn't we re-use it?  
Wouldn't the `ArgumentConfig.onUnknown` field be a better place for the FAIL/IGNORE/WARN JavaDoc? - in that way user could see it on the extension page.

ppalaga (commented):
Nice test coverage, thanks! One suggestion inline.

extensions-core/main/runtime/src/main/java/org/apache/camel/quarkus/main/CamelMainConfig.java
jamesnetherton: Latest commit applies those suggestions.

oscerd (approved):

ppalaga:
> should be to log a warning instead?

I am not going to veto, if Luca favors that :)

ppalaga (approved):

jamesnetherton:
> I am not going to veto, if Luca favors that :)

I think it's actually my preference too. The warning is big and ugly enough that folks should hopefully notice something is wrong.

lburgazzoli:
Ah, I did it from the mobile app and it was mainly for the PR per se not specifically on the warn vs exception so @ppalaga you can put your veto :)

jamesnetherton:
Latest commit changes the default to `warn`.

PR #2342, opened 2021-03-17
Register classes with Solr Field annotations for reflection

jamesnetherton:


oscerd (approved):

zbendhiba (approved):

extensions/solr/deployment/src/main/java/org/apache/camel/quarkus/component/solr/deployment/SolrProcessor.java
```diff
@@ -35,4 +40,27 @@ FeatureBuildItem feature() {
...
+                    AnnotationTarget target = annotationInstance.target();
+                    AnnotationTarget.Kind kind = target.kind();
+                    if (kind.equals(AnnotationTarget.Kind.FIELD)) {
+                        ClassInfo classInfo = target.asField().declaringClass();
+                        return new ReflectiveClassBuildItem(false, true, classInfo.name().toString());
+                    } else if (kind.equals(AnnotationTarget.Kind.METHOD)) {
+                        ClassInfo classInfo = target.asMethod().declaringClass();
+                        return new ReflectiveClassBuildItem(true, false, classInfo.name().toString());
+                    } else {
+                        throw new RuntimeException(
```
aldettinger, line 60: As I'm reading it, would solr extend targets for the @Field annotation one day, then we would break the build ? what about a WARN log instead ?

aldettinger (approved):
Looks good. I just have a non-blocking question about @Field targets.

extensions/solr/deployment/src/main/java/org/apache/camel/quarkus/component/solr/deployment/SolrProcessor.java
jamesnetherton: If the targets are extended, then presumably it'd be our job to handle that? I.e our itests will fail and we make our extension compatible with the new Solr changes.

With a warning wouldn't we just be silently ignoring it?

aldettinger: If the targets are extended, there might be need to add some reflective build items too. I think our itests would fail when we would complement our tests with a new @Field annnotation on the new target.
I think camel-quarkus can throw when an updated solr libs is brought.

ppalaga: You know my general attitude to warnings ;)

In this particular case, as long as `org.apache.solr.client.solrj.beans.Field` has `@Target({FIELD, METHOD})`, I think it is correct to fail.

If we want to be paranoid, we should IMO (1) keep it failing in SolrProcessor and (2) add an assertion to our test that `org.apache.solr.client.solrj.beans.Field` 's `@Target` value has not changed to anything beyond `{FIELD, METHOD}`. But from a practical point of view, how likely it is that will ever change? Would adding a `CLASS` target make any conceivable sense? I am not implying any particular answer, I frankly do not know.

aldettinger: Indeed, a warning will likely be ignored and I don't think that kind of change would happen frequently.

PR #2373, opened 2021-03-24
Only invoke completed method on unknown arguments if the failure remedy is FAIL

jamesnetherton:


oscerd (approved):

ppalaga (approved):

jamesnetherton:
Converted to draft as there's something I need to check before merging.

jamesnetherton:
@ppalaga Do you mind taking a second look at this. I made some tweaks:

- I realised that if `on-failure` is `warn`, then you need to process all valid arguments, instead of stopping on the first invalid argument, so I fixed that.

- To make the warning message output a bit friendlier, I thought it'd be best to truncate long arguments (E.g like a classpath) and abbreviate with '...'.

ppalaga (approved):

aldettinger (approved):

integration-tests/main-command-mode/pom.xml
```diff
@@ -45,6 +45,10 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-timer</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-vertx-web</artifactId>
```
aldettinger, line 50: I guess the vertx web client is needed by the test harness.

jamesnetherton: Correct.

jamesnetherton:
The test failure is not related - hence merging this.

PR #2390, opened 2021-03-26
Quarkiverse dependency upgrades

ppalaga:


jamesnetherton:
I think we need to hold the google-cloud version bump. I did try it last week, but hit some problems with enforcer checks that I could not resolve quickly. I can open an issue to track.

oscerd (approved):

PR #2419, opened 2021-03-30
Fix intermittent failure in camel-quarkus-master-integration-test

jamesnetherton:


jamesnetherton:
FWIW investigating this uncovered something I was not aware of. The application is terminated if the `CamelContext` is stopped. Prior to CQ 1.8.0 if you chose to not use `camel-main`, this was not the case. The app stayed running. Maybe that's desired behaviour, I'm not sure.

I can open an issue if anything thinks it's worth looking into. Probably unlikely folks are going to be starting / stopping the `CamelContext` in their apps anyway + it's not a good practice.

oscerd (approved):

aldettinger (approved):
Yes, I think the spirit is more to start/stop containers. Anyway, we could change later if one come with strong rationale.

PR #2446, opened 2021-04-07
Fix xref-validator download failure

ppalaga:
…ref-validator/repository/archive.tar.gz?ref=12a2765fe934e2b661cafeeac568e7c221021bfe: Extracting tar content of undefined failed, the file appears to be corrupt: Invalid tar header. Maybe the tar is corrupted or it needs to be gunzipped?


oscerd (approved):
Cc @zregvart

ppalaga:
initial-mvn-install is enough for this one

PR #2452, opened 2021-04-08
chore: upgrade Nodejs and Yarn

zregvart:
Also switches to using `frontend-maven-plugin` to run Yarn. Not sure
about the problem encountered (error 134), I don't think it should be an
issue with this combination of versions. Can easily be reverted if
needed.


oscerd (approved):

ppalaga:
@zregvart is it really a good idea to have that yarn cache in git?

zregvart:
> @zregvart is it really a good idea to have that yarn cache in git?

If you'd like a deterministic/offline build, that's the only option. It makes the clean build (i.e. CI) much faster as well. This is what we do on the website and main Camel repository and so far no issues. Bit more on this can be found here: https://yarnpkg.com/features/pnp

ppalaga:
> If you'd like a deterministic/offline build, that's the only option

Could you please explain the determinism aspect? Is there no way to perform `yarn install` so that the result is the same for everybody?

>  It makes the clean build (i.e. CI) much faster as well.

This sounds interesting of course. But I wonder how much faster it can get? If I delete the cache locally, running `yarn install` takes about 15 seconds. OTOH downloading 20+ MB via git takes some time too. So let's assume the net gain can be up to 10 seconds. Do you agree with these figures, Zoran? (Maybe I am measuring wrong processes - I am a newbie in this area.)

My main concern is that git is not very effective at maintaining binary blobs. Those zip files in the yarn cache will change over time, and the overall size of the repo will get bigger and bigger.

I must say I do not feel persuaded, that I want that for saving 10 seconds on the CI. What do others think?

ppalaga:
I forgot to add that I rarely run the antora xref checks locally, bc . they are not included in the `-Dquickly` profile.

zregvart:
> Could you please explain the determinism aspect? Is there no way to perform `yarn install` so that the result is the same for everybody?

In part it depends on how one specifies the versions in `package.json`, as NPM allows loose versions, that's the reason behind the lock files. So a `npm install` or a `yarn install` might not fetch the same version on two occasions (baring the lock file). More details on this can be found here: https://classic.yarnpkg.com/blog/2017/05/31/determinism/ and here https://npm.github.io/npm-like-im-5/npm3/non-determinism.html. Yarn's PnP was rethink of the `node_modules` bit and having the cache in source control is more one step along that path.

An alternative could be to have a shared Yarn cache, that would need some work on the CI (to make sure cache is available and up to date).

ppalaga:
> > Could you please explain the determinism aspect? Is there no way to perform `yarn install` so that the result is the same for everybody?
> 
> In part it depends on how one specifies the versions in `package.json`, as NPM allows loose versions, that's the reason behind the lock files. So a `npm install` or a `yarn install` might not fetch the same version on two occasions (baring the lock file). More details on this can be found here: https://classic.yarnpkg.com/blog/2017/05/31/determinism/ and here https://npm.github.io/npm-like-im-5/npm3/non-determinism.html. Yarn's PnP was rethink of the `node_modules` bit and having the cache in source control is more one step along that path.

Isn't `yarn install --immutable` solving the reproducibility problem well enough? The docs says it is the replacement for `--frozen-lockfile`  that we used so far. `yarn install --immutable` takes 606ms if the `.yarn` folder is available and up to date. So for local development, the cache folder can just stay outside of git and `yarn install --immutable` invoked by the frontend plugin can care for keeping it up to date if necessary. If the dev does not delete it manually, the rebuilds should be fast. On the CI, it can be cached much like `~/.m2/repository`, no?

zregvart:
@ppalaga yes, as I've outlined in the last sentence of the last comment I made, this can be done as you describe using the shared Yarn cache. I think the way I've done it here is better, do you have any objections to doing it this way?

ppalaga:
My main objection is that this adds 20 MB of zips to git that are going to change over time thus bloating the repo. We'd do it for a relatively small benefit compared to the alternative mentioned above. But maybe I am a small minority with this opinion. What do others think?

jamesnetherton:
I have no strong preference for how we do this, but I lean towards Peters' approach if it can reduce repo bloatage.

ppalaga:
I'd prepare a new PR based on this one, but without adding the ZIP files to git. Would that work for you, @zregvart?

Issue #2449, opened 2021-04-07
camel-quarkus-dozer Error when running native executable


theseeker58:
Hi,
I get this exception when running native executable that uses Dozer to map objects in a route.
  Dest field type: java.util.Date:
com.github.dozermapper.core.converters.ConversionException:
java.lang.NoSuchMethodException: java.util.Date.<init>(long)
    at com.github.dozermapper.core.converters.DateConverter.convert(DateConverter.java:102)
    at com.github.dozermapper.core.converters.PrimitiveOrWrapperConverter.convert(PrimitiveOrWrapperConverter.java:84)
    at com.github.dozermapper.core.MappingProcessor.mapOrRecurseObject(MappingProcessor.java:496)
    at com.github.dozermapper.core.MappingProcessor.mapFromFieldMap(MappingProcessor.java:404)
    at com.github.dozermapper.core.MappingProcessor.mapField(MappingProcessor.java:354)
    at com.github.dozermapper.core.MappingProcessor.map(MappingProcessor.java:314)
    at com.github.dozermapper.core.MappingProcessor.mapToDestObject(MappingProcessor.java:263)
    at com.github.dozermapper.core.MappingProcessor.createByCreationDirectiveAndMap(MappingProcessor.java:236)
    at com.github.dozermapper.core.MappingProcessor.mapGeneral(MappingProcessor.java:209)
    ...
To make it working I had to register for reflection java.util.Date.
I guess if it's the right approach or if it should be made implicitly by the component itself, as it is stated as native.

Thanks

PR #2453, opened 2021-04-08
camel-quarkus-dozer Error when running native executable #2449

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/2449

 `java.util.Date` has to be registered for reflection as was reported.

There are several another classes, which has to be registered for reflection. Classes are used for conversion from `String` via `StringConstructorConverter`. (You can see required types in dozer [code](https://github.com/DozerMapper/dozer/blob/master/core/src/test/java/com/github/dozermapper/core/functional_tests/builder/PrimitiveTest.java#L138-L140)). Without these registration, there would be the following errors in native mode

> java.lang.NoSuchMethodException: java.net.URL.<init>(java.lang.String)
> java.lang.NoSuchMethodException: java.io.File.<init>(java.lang.String)

All cases are covered by unit tests. 

### DIFF
--- a/extensions/dozer/deployment/src/main/java/org/apache/camel/quarkus/component/dozer/deployment/DozerProcessor.java
+++ b/extensions/dozer/deployment/src/main/java/org/apache/camel/quarkus/component/dozer/deployment/DozerProcessor.java
@@ -110,6 +110,9 @@ class DozerProcessor {
                 Time[].class,
                 Timestamp[].class,
                 URL[].class,
+                Date.class,
+                File.class,
+                URL.class,
                 DozerThreadContextClassLoader.class,
                 ExpressionFactoryImpl.class));
 
--- a/integration-tests/dozer/src/main/java/org/apache/camel/quarkus/component/dozer/it/DozerResource.java
+++ b/integration-tests/dozer/src/main/java/org/apache/camel/quarkus/component/dozer/it/DozerResource.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.quarkus.component.dozer.it;
 
+import java.util.Calendar;
+import java.util.GregorianCalendar;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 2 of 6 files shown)


oscerd (approved):

extensions/dozer/deployment/src/main/java/org/apache/camel/quarkus/component/dozer/deployment/DozerProcessor.java
```diff
@@ -110,6 +110,9 @@ void configureCamelDozer(BuildProducer<ReflectiveClassBuildItem> reflectiveClass
                 Time[].class,
                 Timestamp[].class,
                 URL[].class,
+                Date.class,
+                File.class,
+                URL.class,
```
aldettinger, line 115: I'm quite beginner with dozer :) Is there really a finished set of classes to be registered as native ? Won't we end up registering every new classes that developers would eventually use in their scenarios ?

JiriOndrusek: @aldettinger 
Date is required for https://github.com/DozerMapper/dozer/blob/master/core/src/main/java/com/github/dozermapper/core/converters/DateConverter.java.
The other classes are special. You can see them for example here: https://github.com/DozerMapper/dozer/blob/master/core/src/test/java/com/github/dozermapper/core/functional_tests/builder/PrimitiveTest.java#L138-L140. No other classes should be required. Other classes are dynamically registered by DozerProcessor.

JiriOndrusek: I also added a custom class into the test to be sure, that there is no need to register it.

aldettinger: ok, I see looks good :+1:

aldettinger (approved):

Issue #2458, opened 2021-04-09
HBase testcontainer fails with UnknownHostException: <myhostname>.localdomain: Name or service not known


ppalaga:
When the container log is forwarded to stdout using `container.withLogConsumer(frame -> System.out.print(frame.getUtf8String()))`, I could see the following error messages. 

```
[INFO] [stdout] java.net.UnknownHostException: terpistone.localdomain: terpistone.localdomain: Name or service not known
[INFO] [stdout]         at java.net.InetAddress.getLocalHost(InetAddress.java:1506)
[INFO] [stdout]         at org.apache.hadoop.metrics2.impl.MetricsSystemImpl.getHostname(MetricsSystemImpl.java:547)
[INFO] [stdout]         at org.apache.hadoop.metrics2.impl.MetricsSystemImpl.configureSystem(MetricsSystemImpl.java:486)
[INFO] [stdout]         at org.apache.hadoop.metrics2.impl.MetricsSystemImpl.configure(MetricsSystemImpl.java:482)
[INFO] [stdout]         at org.apache.hadoop.metrics2.impl.MetricsSystemImpl.start(MetricsSystemImpl.java:189)
[INFO] [stdout]         at org.apache.hadoop.metrics2.impl.MetricsSystemImpl.init(MetricsSystemImpl.java:164)
[INFO] [stdout]         at org.apache.hadoop.metrics2.lib.DefaultMetricsSystem.init(DefaultMetricsSystem.java:54)
[INFO] [stdout]         at org.apache.hadoop.metrics2.lib.DefaultMetricsSystem.initialize(DefaultMetricsSystem.java:50)
[INFO] [stdout]         at org.apache.hadoop.hbase.metrics.BaseSourceImpl$DefaultMetricsSystemInitializer.init(BaseSourceImpl.java:54)
[INFO] [stdout]         at org.apache.hadoop.hbase.metrics.BaseSourceImpl.<init>(BaseSourceImpl.java:116)
[INFO] [stdout]         at org.apache.hadoop.hbase.io.MetricsIOSourceImpl.<init>(MetricsIOSourceImpl.java:46)
[INFO] [stdout]         at org.apache.hadoop.hbase.io.MetricsIOSourceImpl.<init>(MetricsIOSourceImpl.java:38)
[...]

PR #2459, opened 2021-04-09
Forward HBase testcontainer's log to stdout to see whether #2458 is o…

ppalaga:
…ccurring


oscerd (approved):

aldettinger:
Looks like this PR is triggering another issue in hase tests. Or maybe, it's part of the investigation ?

ppalaga:
> Looks like this PR is triggering another issue in hase tests. Or maybe, it's part of the investigation ?

Yes, see the last attempt

PR #2474, opened 2021-04-14
Revisit the documentation #2136 #2470 #2374

### LINKED ISSUE #2136: Document correct BoM usage
Issues like #2131 are a clear sign that we are lacking on documentation about dependency management. Let me file a new issue for that.

We should document at least:

* The easiest and most straightforward option is to use https://code.quarkus.io/ producing a project with the `quarkus-universe-bom` that aggregates `camel-quarkus-bom`. This has the advantage that the user does not need to care for the compatibility of the individual artifacts that may potentially come from various independent project. All artifacts managed by `quarkus-universe-bom` are supposed to work together flawlessly.
* Using `camel-quarkus-bom` instead of `quarkus-universe-bom`: As long as the user does not import any other BoM beyond `camel-quarkus-bom`, this is an easy and reliable option too. `camel-quarkus-bom` manages all supported Camel artifacts and it also imports `quarkus-bom`. This option can be used if the user plans to use only extensions from Quarkus and Camel Quarkus.
* Combinig `camel-quarkus-bom` or `quarkus-universe-bom` with each other manually: take care that you combine compatible versions. `camel-quarkus-bom` should typically be imported before `quarkus-universe-bom` - see the next point.
* Combinig `camel-quarkus-bom` with any other BoM: think carefully in which order you import them. The order of imports makes up the precedence. This matters for the artifacts managed by both BoMs.

### LINKED ISSUE #2374: Adjust the documentation after the merger of main into core
* https://camel.apache.org/camel-quarkus/latest/user-guide/bootstrap.html#_camel_main needs an adjustment
  - Should mention `quarkus.camel.main.enabled`
* https://camel.apache.org/camel-quarkus/latest/reference/extensions/main.html should mention that core is the replacement

### LINKED ISSUE #2470: Add info about Camel IDE plugins to the documentation
We list Quarkus specific plugins on https://camel.apache.org/camel-quarkus/latest/user-guide/first-steps.html#_ide_plugins but we do not mention any Camel specific ones. 
Here is the AsciiDoc source of the page https://github.com/apache/camel-quarkus/blob/main/docs/modules/ROOT/pages/user-guide/first-steps.adoc

ppalaga:
Fix #2136 
Fix #2470 
Fix #2374

Please follow the instructions in `docs/README.adoc` to get an HTML preview


docs/modules/ROOT/pages/user-guide/first-steps.adoc
```diff
@@ -101,6 +115,11 @@ The application has three compile dependencies:
 
 They are managed within the `camel-quarkus-bom` that is imported in `<dependencyManagement>`.
 
+[INFO]
+====
+More about xref:user-guide/dependency-management.adoc[BoMs].
```
jamesnetherton, line 120: Is the mixed case `BoMs` intended?

ppalaga: Yes, but I do not mind switching to BOM, if you like.

jamesnetherton: I'll leave it for you to decide. It's an acronym, so BOM is more correct. If you don't want to waste time tweaking it, then it's no big deal. PR looks good otherwise :+1:

ppalaga:
3b5c643 : `s/BoM/BOM/`

Issue #2421, opened 2021-03-31
Google Storage support


ppalaga:
New in Camel 3.9 https://camel.apache.org/components/latest/google-storage-component.html

Issue #1387, opened 2020-06-19
Turn off bean removal by ArC 


lburgazzoli:
By default ArC [removes unused beans](https://quarkus.io/guides/cdi-reference#remove_unused_beans) but that does not work very well with Camel as it does use `BeanManager` behind the scenes to look-up beans from the container and ArC can't be aware of what Camel need.

We probably need to discuss this topic with the Quarkus team and see if through a build item it could be possible to reverse the behaviour so not to remove beans unless that feature is turned on explicitly.

Taken from: https://github.com/apache/camel-quarkus/pull/1344#discussion_r440015225

jamesnetherton:
I'll close this as it's superseded by https://github.com/apache/camel-quarkus/issues/2822.

PR #2505, opened 2021-04-22
Google Storage support #2421

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/2421


oscerd (approved):

integration-tests/google-storage/src/test/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageTest.java
```diff
@@ -0,0 +1,202 @@
...
+    private static final String DEST_BUCKET = "camel_quarkus_test_dest_bucket";
+    private static final String TEST_BUCKET1 = "camel_quarkus_test_bucket1";
+    private static final String TEST_BUCKET2 = "camel_quarkus_test_bucket2";
+    private static final String TEST_BUCKET3 = "camel_quarkus_test_bucket3";
+    private static final String FILE_NAME_007 = "file007";
+    private static final String FILE_NAME_006 = "file006";
+
+    @BeforeEach
+    public void beforeEach() {
+        RestAssured.given().get("/google-storage/loadComponent").then().statusCode(200);
```
aldettinger, line 58: Do we really need do this before each test ?

tooling/scripts/test-categories.yaml
```diff
@@ -32,6 +32,7 @@ group-02:
   - elasticsearch-rest
   - google-bigquery
   - google-pubsub
+  - google-storage
```
aldettinger, line 35: group-01 or group-08 may achieve a greater balance https://github.com/apache/camel-quarkus/actions/runs/773244432

jamesnetherton:
Shouldn't we use the Quarkiverse google cloud bits given that it should have already solved the native part:

https://github.com/quarkiverse/quarkus-google-cloud-services/tree/main/storage

extensions/google-storage/runtime/src/main/java/org/apache/camel/quarkus/component/google/storage/UrlFetchTransportSubstitute.java
```diff
@@ -0,0 +1,33 @@
...
+import com.google.api.client.extensions.appengine.http.UrlFetchTransport;
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+
+@TargetClass(value = UrlFetchTransport.class)
+public final class UrlFetchTransportSubstitute {
+
+    @Substitute
+    protected Object buildRequest(String method, String url) throws IOException {
+        throw new IllegalStateException("Not used by camel extension");
```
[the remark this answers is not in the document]
JiriOndrusek: You are right, I'll fix it.

tooling/scripts/test-categories.yaml
JiriOndrusek: Sounds correct, thanks for the link

JiriOndrusek:
@jamesnetherton It makes totally sense. When I assigned this issue, there was no such extension (or I searched it badly). I'll rework it, thanks for noticing

integration-tests/google-storage/src/test/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageTest.java
JiriOndrusek: To be honest I don't like it either, but I need to configure component before usage. Better approach would be to use `@BeforeAll`, but than the method has to be static, which won't allow me to call `/google-storage/loadComponent`.

The other approach I see is to allow order of the tests and then call it only once in the first test. But this approach seems like a lesser evil to me. (there is an if statement, which skips second initialization) WDYT? @aldettinger

aldettinger: So, it's a bit like we would need to initialize the component for tests once before all tests are run. The only idea I have is `@BeforeAll`. static should not be an issue, but from your line I deduce that the quarkus junit extension would not have started enough at this stage. If that's the case, I don't have better idea. Thanks for explanations :+1:

zbendhiba: @JiriOndrusek can you load the configured component while the app starts?
something [like that ](https://github.com/apache/camel-quarkus/blob/e39f071c41353787308f2d8484b102445890ce16/integration-tests/hazelcast/src/main/java/org/apache/camel/quarkus/component/hazelcast/it/HazelcastRoutes.java#L80)

JiriOndrusek: thanks @zbendhiba  for the example. Using `@Produce` sounds great for this case. Exactly the way I was looking for. Thanks again.

JiriOndrusek: @aldettinger Exactly as you said, problem is not caused by the static call, but because it was called even sooner then Resource was prepared. But example from @zbendhiba looks very promising. I'll try it (with the bigger refactor to use quarkiverse extension)

extensions/google-storage/runtime/src/main/java/org/apache/camel/quarkus/component/google/storage/UrlFetchTransportSubstitute.java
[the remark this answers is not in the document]
JiriOndrusek: @aldettinger Class was removed during re-work

integration-tests/google-storage/src/test/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageTest.java
JiriOndrusek: @aldettinger I used approach according to @zbendhiba

tooling/scripts/test-categories.yaml
JiriOndrusek: @aldettinger  I moved test into group-01

JiriOndrusek:
@jamesnetherton I reworked extension to use quarkiverse .

There a few points I was facing:

1. Quarkiverse extension uses quarkus properties to create a storage client. This client was not autowired into component (because of https://github.com/apache/camel-quarkus/issues/1387). I tried to use `UnremovableBeanBuildItem`, but it had no effect. So I created class `GoogleStorageEnforcer`, which injects all storage clients, which marks them as used. 
2. To allow more different clients in one context, there is a way of doing it via producing more storage clients - it is described in doc.
3. There is  a warning during native build
> WARNING: Could not resolve io.grpc.netty.shaded.io.netty.channel.socket.nio.NioServerSocketChannel for reflection configuration. Reason: java.lang.ClassNotFoundException: io.grpc.netty.shaded.io.netty.channel.socket.nio.NioServerSocketChannel.

, but it doesn't make extension fail. 

4. There is a requirement to export `GOOGLE_PROJECT_ID` and `GOOGLE_APPLICATION_CREDENTIALS` to run tests with real account. It is described in README.adoc in the test project.

extensions/google-storage/deployment/src/main/java/org/apache/camel/quarkus/component/google/storage/deployment/GoogleStorageProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+        return new ExtensionSslNativeSupportBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    public AdditionalBeanBuildItem storageEnforcer() {
+        return new AdditionalBeanBuildItem(GoogleStorageEnforcer.class);
+    }
+
+    @BuildStep
+    void registerForReflection(CombinedIndexBuildItem combinedIndex,
```
jamesnetherton, line 60: Is this needed anymore? I was hoping the quarkiverse extension would take care of this.

integration-tests/google-storage/src/main/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageResource.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Path("/startPolling")
+    @POST
+    public void startPolling(@QueryParam(QUERY_BUCKET) String bucketName,
+            @QueryParam(QUERY_POLLING_ACTION) String pollingAction,
+            @QueryParam(QUERY_DESTINATION_BUCKET) String destinationBucket) {
+        // use another thread for polling consumer to demonstrate that we can wait before
+        // the message is sent to the queue
+        Executors.newSingleThreadExecutor().execute(() -> {
```
jamesnetherton, line 151: For this kind of thing, it's far easier to create a `RouteBuilder`, configure the `google-storage` consumer and direct any messages to a `seda` or `mock` endpoint which you can then retrieve via a `ConsumerTemplate` in the JAX-RS resource class.

It makes the test flow much better...

1. Create some storage object
2. Retrieve storage object

There's no need for executors, polling etc.

integration-tests/google-storage/src/test/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageTestResource.java
```diff
@@ -0,0 +1,63 @@
...
+public class GoogleStorageTestResource implements QuarkusTestResourceLifecycleManager {
+
+    public static final int PORT = 4443;
+    public static final String CONTAINER_NAME = "fsouza/fake-gcs-server";
+
+    private GenericContainer<?> container;
+
+    @Override
+    public Map<String, String> start() {
+        //if there is a real configuration, don't start container
```
jamesnetherton, line 35: We should probably add a call to `MockBackendUtils`. See example from AWS tests here:

https://github.com/apache/camel-quarkus/blob/c50a7ae86148f34e286e218bba8fbff48b4c0d68/integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/Aws2TestResource.java#L47-L48

jamesnetherton (commented):
I left a few comments. I think we spoke about the Quarkiverse project bumping their gcloud BOM version and maybe fixing some other things. If we've not done so, we should raise issues with that project and cross reference them here.

extensions/google-storage/deployment/src/main/java/org/apache/camel/quarkus/component/google/storage/deployment/GoogleStorageProcessor.java
JiriOndrusek: yes that is right it should be covered by quarkiverse. I'll recheck it again. I removed native things and did not cover build items.
As for upgrade of the version, my problem was caused by local error. Version of quarkiverse is using `1.113.4`, camel is using `1.113.4`. And it looks there is a newer version of quarkiverse, which is using much higher version of google cloud libraries, So I would say that current state is acceptable.

integration-tests/google-storage/src/main/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageResource.java
JiriOndrusek: No problem with it. I thought that not using seda|mock|direct libraries is better solution. I'll use route then.

integration-tests/google-storage/src/test/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageTestResource.java
JiriOndrusek: Ok, I'll look into it.

integration-tests/google-storage/src/main/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageResource.java
JiriOndrusek: I used a route builder to poll into direct and then receive message from it.

integration-tests/google-storage/src/test/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageTestResource.java
JiriOndrusek: I used `MockBackendUtils`

extensions/google-storage/deployment/src/main/java/org/apache/camel/quarkus/component/google/storage/deployment/GoogleStorageProcessor.java
JiriOndrusek: @jamesnetherton I'll report an issue to quarkiverse to upgrade their reference for `com.google.cloud:google-cloud-storage:jar`. Once this fix is provided, we can remove exclusion from camel component.

JiriOndrusek:
Please do not merge this PR yet. I need to look into a way of cleaning data after this test (which was removed during last refactor)

JiriOndrusek:
PR is ready to review/merge.

Issue for update of `google-cloud-storage` in quarkiverse extension is reported as https://github.com/quarkiverse/quarkus-google-cloud-services/issues/113
Issue for camel-quarkus to remove exclusion is reported as https://github.com/apache/camel-quarkus/issues/2526

docs/modules/ROOT/pages/reference/extensions/google-storage.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+== Additional Camel Quarkus configuration
+
+There are two different configuration approaches:
+
+* Google storage client can be defined via quarkus properties leveraging the Quarkiverse - Google Cloud Services - Storage (see https://github.com/quarkiverse/quarkus-google-cloud-services/blob/main/storage[extension git page]).
+Camel will autowire client into the Google-storage component.
+This configuration allows definition of only one storage client, therefore it isn't possible to define several different endpoints, which run together.
+
+* Provide client/clients for camel registry (e.g. CDI producer/bean) and reference them from endpoint.
```
aldettinger, line 55: I would stress the point here like (if I understand correctly):
When multiple storage clients are needed, quarkus properties should not be used. Instead, one or more storage client bean need to be registered in the registry (e.g. via a CDI bean producer). Each storage client bean could then be referenced from distinct endpoints.

extensions/google-storage/runtime/src/main/java/org/apache/camel/quarkus/component/google/storage/GoogleStorageEnforcer.java
```diff
@@ -0,0 +1,35 @@
...
+import com.google.cloud.storage.Storage;
+
+@ApplicationScoped
+public class GoogleStorageEnforcer {
+
+    //Quarkiverse StorageProducer creates storage based on properties.
+    //Because of https://github.com/apache/camel-quarkus/issues/1387, storage is removed and not applied to component.
+    //UnremovableBeanBuildItem does not fix this.
+    //Injecting all storages makes Quarkiverse StorageProducer to be used.
+    //We are injecting all storages, because client can define more storages. (so injecting only one could fail|)
```
aldettinger, line 32: small typo here `are need, It is possible could be used `

docs/modules/ROOT/pages/reference/extensions/google-storage.adoc
lburgazzoli: can we open an issue on the extension to check if configuring multiple client would be an option in the future ? 

quarkus has such option for jdbc, maybe it could be possible to have the same logic here too (and probably for other extensions)

integration-tests/google-storage/src/main/java/org/apache/camel/quarkus/component/google/storage/it/RealAccountPropertyProducer.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.ConfigSourceProvider;
+
+public class RealAccountPropertyProducer implements ConfigSourceProvider {
+
```
aldettinger, line 27: Can't we simply use env var from properties file, something like https://quarkus.io/guides/config-reference#combine-property-env-var ?

docs/modules/ROOT/pages/reference/extensions/google-storage.adoc
JiriOndrusek: @aldettinger Problem is unfortunately a little bit bigger here. Application properties has to be used always. Even if you define you own multiple clients.  Reason is that quarkiverse extension always creates client and if there are no properties, creation of client fails with 

>java.io.IOException: The Application Default Credentials are not available

It is unfortunate to require app properties to be filled even if it is not required, but I haven't found a way of making it worl (and to keep autowiring from camel component working) I plan to raise a ticket to quarkiverse extension asking  for optional configuration of client.

JiriOndrusek: @lburgazzoli I plan to open an issue to make configuration of client optional (see previous comment), but registration of multiple clients as you suggested makes even better sense. I'll create it in that way.

extensions/google-storage/runtime/src/main/java/org/apache/camel/quarkus/component/google/storage/GoogleStorageEnforcer.java
JiriOndrusek: I'll fix it.

integration-tests/google-storage/src/main/java/org/apache/camel/quarkus/component/google/storage/it/RealAccountPropertyProducer.java
JiriOndrusek: @aldettinger this is a nice idea. I already tried it and I was facing an issue with it. But now I think that it should be possible to figure it out. I'll try it once again. Thanks for suggestion!

docs/modules/ROOT/pages/reference/extensions/google-storage.adoc
aldettinger: Thanks for sharing. No big deal for the properties always needed as we'll have a separate issue.
However, I would stress that when reading the doc first, I did not understood why/when "Provide client/clients for camel registry (e.g. CDI producer/bean) and reference them from endpoint" is needed. Would you think changing the doc as below has value?
When multiple storage clients are needed, one or more storage client bean need to be registered in the registry (e.g. via a CDI bean producer). Each storage client bean could then be referenced from distinct endpoints.

JiriOndrusek: I'll rephrase the doc to provide a better explanation

extensions/google-storage/runtime/src/main/java/org/apache/camel/quarkus/component/google/storage/GoogleStorageEnforcer.java
JiriOndrusek: done

integration-tests/google-storage/src/main/java/org/apache/camel/quarkus/component/google/storage/it/RealAccountPropertyProducer.java
JiriOndrusek: @aldettinger it works now, thanks again.

docs/modules/ROOT/pages/reference/extensions/google-storage.adoc
JiriOndrusek: Doc is changed as requested.

Issue about multiple clients registration is created - https://github.com/quarkiverse/quarkus-google-cloud-services/issues/114
Follow-up isuue in camel-quarkus - https://github.com/apache/camel-quarkus/issues/2534

JiriOndrusek:
@aldettinger all requested changes are fixed.

integration-tests/google-storage/src/test/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+        //CreateDownloadLink - requires authentication
+        if (!GoogleStorageHelper.usingMockBackend()) {
+            executeOperation(TEST_BUCKET2, GoogleCloudStorageOperations.createDownloadLink,
+                    Collections.singletonMap(GoogleCloudStorageConstants.OBJECT_NAME, FILE_NAME_007 + "_copy"),
+                    startsWith("http"));
+        }
+    }
+
+    private void putObject(String sheldon, String testBucket1, String fileName007) {
```
aldettinger, line 148: Maybe the param names could be more generic here. testBucket ? filename ?

aldettinger (approved):
A last cosmetic proposal :) Beyond, it looks good to me. It was quite challenging to have the 2 testing modes for this extension. Well done :+1:

integration-tests/google-storage/src/test/java/org/apache/camel/quarkus/component/google/storage/it/GoogleStorageTest.java
JiriOndrusek: Fixed, probably generated method and I didn't notice wrong param names.

JiriOndrusek:
@aldettinger Thanks, fixed

aldettinger:
Hey @JiriOndrusek , were all the comments addressed ? Could this PR be merged ?

Issue #2241, opened 2021-02-12
AtlasMap: Use jandex to discover the types we need to register for reflection


zbendhiba:
Use jandex to discover the types we need to register for reflection.
See comment in [PR](https://github.com/apache/camel-quarkus/pull/2237)

PR #2509, opened 2021-04-23
AtlasMap: Use jandex to discover the types we need to register for re…

zbendhiba:
…flection #2241

### DIFF
--- a/extensions/atlasmap/deployment/src/main/java/org/apache/camel/quarkus/component/atlasmap/deployment/AtlasmapProcessor.java
+++ b/extensions/atlasmap/deployment/src/main/java/org/apache/camel/quarkus/component/atlasmap/deployment/AtlasmapProcessor.java
@@ -22,169 +22,28 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
 
-import io.atlasmap.actions.CollectionActions;
-import io.atlasmap.actions.DateFieldActions;
-import io.atlasmap.actions.ExpressionFieldAction;
-import io.atlasmap.actions.NumberFieldActions;
-import io.atlasmap.actions.ObjectFieldActions;
-import io.atlasmap.actions.StringComplexFieldActions;
-import io.atlasmap.actions.StringSimpleFieldActions;
-import io.atlasmap.converters.BigDecimalConverter;
-import io.atlasmap.converters.BigIntegerConverter;
-import io.atlasmap.converters.BooleanConverter;
-import io.atlasmap.converters.ByteConverter;
-import io.atlasmap.converters.CalendarConverter;
-import io.atlasmap.converters.CharBufferConverter;
-import io.atlasmap.converters.CharSequenceConverter;
-import io.atlasmap.converters.CharacterConverter;
-import io.atlasmap.converters.DateConverter;
-import io.atlasmap.converters.DoubleConverter;
-import io.atlasmap.converters.FloatConverter;
-import io.atlasmap.converters.GregorianCalendarConverter;
-import io.atlasmap.converters.IntegerConverter;
-import io.atlasmap.converters.LocalDateConverter;
-import io.atlasmap.converters.LocalDateTimeConverter;
-import io.atlasmap.converters.LocalTimeConverter;
-import io.atlasmap.converters.LongConverter;
-import io.atlasmap.converters.NumberConverter;
-import io.atlasmap.converters.ShortConverter;
-import io.atlasmap.converters.SqlDateConverter;
-import io.atlasmap.converters.SqlTimeConverter;
-import io.atlasmap.converters.SqlTimestampConverter;
-import io.atlasmap.converters.StringBufferConverter;
-import io.atlasmap.converters.StringBuilderConverter;
-import io.atlasmap.converters.StringConverter;
-import io.atlasmap.converters.ZonedDateTimeConverter;
 import io.atlasmap.core.DefaultAtlasContextFactory;
 import io.atlasmap.core.DefaultAtlasModuleInfo;
 import io.atlasmap.csv.module.CsvModule;
-import io.atlasmap.csv.v2.CsvDataSource;
-import io.atlasmap.csv.v2.CsvField;
 import io.atlasmap.dfdl.module.DfdlModule;
 import io.atlasmap.java.module.JavaModule;
-import io.atlasmap.java.v2.JavaField;
-import io.atlasmap.java.v2.Modifier;
-import io.atlasmap.java.v2.ModifierList;
 import io.atlasmap.json.module.JsonModule;
-import io.atlasmap.json.v2.JsonComplexType;
-import io.atlasmap.json.v2.JsonDataSource;
-import io.atlasmap.json.v2.JsonEnumField;
-import io.atlasmap.json.v2.JsonEnumFields;
-import io.atlasmap.json.v2.JsonField;
-import io.atlasmap.json.v2.JsonFields;
 import io.atlasmap.mxbean.AtlasContextFactoryMXBean;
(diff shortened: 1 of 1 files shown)


oscerd (approved):

extensions/atlasmap/deployment/src/main/java/org/apache/camel/quarkus/component/atlasmap/deployment/AtlasmapProcessor.java
```diff
@@ -361,6 +75,35 @@ NativeImageResourceBuildItem resource() {
...
+                        || n.startsWith("io.atlasmap.actions")
+                        || n.startsWith("io.atlasmap.converters")
+                        || n.startsWith("io.atlasmap.json.v2")
+                        || n.startsWith("io.atlasmap.java.v2")
+                        || n.startsWith("io.atlasmap.xml.v2")
+                        || n.startsWith("io.atlasmap.csv.v2")
+                        || n.startsWith("io.atlasmap.dfdl.v2"))
+                .toArray(String[]::new);
+
+        return new ReflectiveClassBuildItem(true, true, dtos);
```
ppalaga, line 104: An obligatory `true, true` comment from my side :) 

`new ReflectiveClassBuildItem(true, true` should be avoided as far as possible, because it bloats the amount of code the native compiler has to process thus substantially prolonging the native compilation and increasing the amount of RAM needed. 

It is very rare that a lib/framework uses both method and field reflection. Hence either `ReflectiveClassBuildItem(true, false` or `ReflectiveClassBuildItem(false, true` should be typically enough.

In this particular case, some of the classes were registered with `ReflectiveClassBuildItem(true, false` before this change and unless there is a known problem with that, we should prefer registering them like that.

So can we perhaps reduce number of `true, true` registrations in some way?

extensions/atlasmap/deployment/src/main/java/org/apache/camel/quarkus/component/atlasmap/deployment/AtlasmapProcessor.java
```diff
@@ -361,6 +75,35 @@ NativeImageResourceBuildItem resource() {
...
 
+    @BuildStep
+    void addDependencies(BuildProducer<IndexDependencyBuildItem> indexDependency) {
+        indexDependency.produce(new IndexDependencyBuildItem("io.atlasmap", "atlas-core"));
+        indexDependency.produce(new IndexDependencyBuildItem("io.atlasmap", "atlas-model"));
+        indexDependency.produce(new IndexDependencyBuildItem("io.atlasmap", "atlas-xml-model"));
+        indexDependency.produce(new IndexDependencyBuildItem("io.atlasmap", "atlas-java-model"));
+        indexDependency.produce(new IndexDependencyBuildItem("io.atlasmap", "atlas-json-model"));
+        indexDependency.produce(new IndexDependencyBuildItem("io.atlasmap", "atlas-csv-model"));
+        indexDependency.produce(new IndexDependencyBuildItem("io.atlasmap", "atlas-dfdl-model"));
```
lburgazzoli, line 86: Can we open an issue to the atlas map project to include a jandex index ?

extensions/atlasmap/deployment/src/main/java/org/apache/camel/quarkus/component/atlasmap/deployment/AtlasmapProcessor.java
```diff
@@ -361,6 +75,35 @@ NativeImageResourceBuildItem resource() {
...
+        // register Atlasmap model classes for serialize/deserialize
+        String[] dtos = index.getKnownClasses().stream().map(ci -> ci.name().toString())
+                .filter(n -> n.startsWith("io.atlasmap.v2")
+                        || n.startsWith("io.atlasmap.actions")
+                        || n.startsWith("io.atlasmap.converters")
+                        || n.startsWith("io.atlasmap.json.v2")
+                        || n.startsWith("io.atlasmap.java.v2")
+                        || n.startsWith("io.atlasmap.xml.v2")
+                        || n.startsWith("io.atlasmap.csv.v2")
+                        || n.startsWith("io.atlasmap.dfdl.v2"))
```
lburgazzoli, line 101: I think it is fine for the moment to filter base on the package but wonder if we can switch to a more efficient way, like registering for reflection based on interfaces and SPIs.

extensions/atlasmap/deployment/src/main/java/org/apache/camel/quarkus/component/atlasmap/deployment/AtlasmapProcessor.java
ppalaga: > Can we open an issue to the atlas map project to include a jandex index ?

+1 for doing it as long as AtlassMap has no index yet. However, that would not eliminate the need for having these IndexDependencyBuildItems, would it?

lburgazzoli: As far as I remember if a dependency includes a jandex index or a beans.xml, then it is automatically indexed but maybe something has changed

zbendhiba: I'll open one, unless you've already done !!

extensions/atlasmap/deployment/src/main/java/org/apache/camel/quarkus/component/atlasmap/deployment/AtlasmapProcessor.java
zbendhiba: How can I do that? I don't know how to look for it. Some of the classes extends an interface, and some just Serializable. And some needed Enums do not extend anything.
Also I was wondering if it's doable to split the Atlasmap into different packages, so that end user chooses if he/she needs java, XML, CSV ... Because we do register lot of classes for maybe just one type of conversion.

lburgazzoli: I think we need to explore using a mix of things, like to use `IndexView.getAllKnownImplementors` and/or `IndexView.getAllKnownSubclasses` for most of the bits and everything else as a special case. 

We could also use a mix of strategies like searching also from all the Serializable and enums withting the `io.atlasmap` package. I'd also explore the option to contribute to Atlasmap to make discovery easier.

I'm +1 to explore if we can split atlasmap extension so it's up to the user to decide what to include and we could also think about having an uber extension that includes everything as today (as transitive dep).

extensions/atlasmap/deployment/src/main/java/org/apache/camel/quarkus/component/atlasmap/deployment/AtlasmapProcessor.java
zbendhiba: Yeah (true, true) was the easiest way.
I'm working on this one to find the best way to reduce this.

zbendhiba:
I've made 2 changes 

- the `new ReflectiveClassBuildItem(true, true)` only for one of the classes : `DataSourceMetadata` 
- register the converters and actions differently

I created those issues : to follow up with other PRs  for camel-quarkus : 
- [Atlasmap : include jandex index](https://github.com/atlasmap/atlasmap/issues/2721)
- [Atlasmap : split Alasmap extension](https://github.com/apache/camel-quarkus/issues/2536)

zbendhiba:
The test failure is not related. Can I merge this, or do I wait for jandex index in Atlasmap. I think this PR includes better all the model classes

ppalaga:
Thanks, @zbendhiba !

Issue #2523, opened 2021-04-27
org.apache.camel.quarkus.component.lra.it.LraTest failed


ffang:
I found two issues in this test
1. the LRA port only bind to container but not to the localhost, so the http request to localhost always gets "SocketTimeout Read timed out"
2. the test logic isn't correct, like the test totally put 4 orders which needs to use 115 credit, but the current still assume it was 3 orders and needed 95 credit, hence we see IllegalStateException("Insufficient credit")

We don't see this error in CI because I think recently CI don't go that further.

PR #2524, opened 2021-04-27
fix org.apache.camel.quarkus.component.lra.it.LraTest failed #2523

ffang:

### DIFF
--- a/integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
+++ b/integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
@@ -33,31 +33,44 @@ public class LraTestResource implements QuarkusTestResourceLifecycleManager {
     private static final String LRA_IMAGE = "jbosstm/lra-coordinator:5.9.3.Final";
 
     private GenericContainer container;
+    private String hostname;
+    private int lraPort;
 
     @Override
     public Map<String, String> start() {
         try {
-            container = new GenericContainer(LRA_IMAGE)
-                    .withExposedPorts(LRA_PORT)
-                    .withCommand(
-                            "java",
-                            "-jar",
-                            "/deployments/lra-coordinator-swarm.jar",
-                            "-Djava.net.preferIPv4Stack=true",
-                            "-Dswarm.http.port=" + LRA_PORT)
-                    .waitingFor(Wait.forLogMessage(".*WFSWARM99999.*", 1));
-
-            container.start();
-
-            String hostname = null;
             if (SystemUtils.IS_OS_LINUX) {
-                hostname = "docker0";
+                hostname = "localhost";
+                container = new GenericContainer(LRA_IMAGE)
+                        .withNetworkMode("host")
+                        .withCommand(
+                                "java",
+                                "-jar",
+                                "/deployments/lra-coordinator-swarm.jar",
+                                "-Djava.net.preferIPv4Stack=true",
+                                "-Dswarm.http.port=" + LRA_PORT)
+                        .waitingFor(Wait.forLogMessage(".*WFSWARM99999.*", 1));
+                container.start();
+                lraPort = LRA_PORT;
             } else {
                 hostname = "host.docker.internal";
+                container = new GenericContainer(LRA_IMAGE)
+                        .withNetworkMode("bridge")
+                        .withExposedPorts(LRA_PORT)
+                        .withCommand(
+                                "java",
+                                "-jar",
+                                "/deployments/lra-coordinator-swarm.jar",
+                                "-Djava.net.preferIPv4Stack=true",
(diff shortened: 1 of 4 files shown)


oscerd (approved):

integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
```diff
@@ -36,8 +42,16 @@
     @Override
     public Map<String, String> start() {
         try {
+            String hostname = InetAddress.getLocalHost().getHostName();
```
jamesnetherton, line 45: Small nit - I don't see `hostname` used for anything so I guess it can be removed?

jamesnetherton:
Asside from the `LraTestResource` mods, I'm a bit confused by the changes to this test.

The general idea is that `CreditService` has a limit of 100 credits. The test then makes 3 'orders' amounting to 95 credits. It then attempts another order with a total of 20 to trigger the 'Insufficient Credit' scenario. So from what I can see, the assertions are valid and correct (I think I stole this idea from the Camel main project BTW).

jamesnetherton:
Another thought is that, without host networking, you may not be able to run this test properly. The LRA container needs the ability to 'call back' to the test application running on the host.

integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
ffang: Hi @jamesnetherton ,

Thanks for all the feedback!
I sent a new commit which removed unused code you mentioned above.

Freeman

ffang:
> Asside from the `LraTestResource` mods, I'm a bit confused by the changes to this test.
> 
> The general idea is that `CreditService` has a limit of 100 credits. The test then makes 3 'orders' amounting to 95 credits. It then attempts another order with a total of 20 to trigger the 'Insufficient Credit' scenario. So from what I can see, the assertions are valid and correct (I think I stole this idea from the Camel main project BTW).

OK, now I understand better about the testcase.

So the "java.lang.IllegalStateException: Insufficient credit" is expected, but this test still has the logic error.
Since it always creates new orders before checking the credit, so when checking valid order, the expected return value is 4 instead of 3. I adjusted the camel route definition(put checking credit before creating new order) to make the test pass.

Actually this problem also exists in the test counterpart(LRACreditIT#testCreditExhausted)  of camel-lra component, so I think we also need a change in camel project.

ffang:
> Another thought is that, without host networking, you may not be able to run this test properly. The LRA container needs the ability to 'call back' to the test application running on the host.

You raised a very good question here which pushed me to went through a bunch docker network documents to give the answer.

You must be on the linux machine so that withNetworkMode(host)(equals --network host, which means container shares the same network stack as host). However, this doesn't work on mac. Per the document here
https://docs.docker.com/network/host/
```
The host networking driver only works on Linux hosts, and is not supported on Docker Desktop for Mac, Docker Desktop for Windows, or Docker EE for Windows Server.
```
Moreover, when "--network host" is used, the "port forwarding"(-p container_port:host_port) related configuration will be ignored, which means for developer on mac(also on windows, if there is any), can't access the lra service in the container from the host. Since we can't guarantee all developers to use linux machine, I think we need the test runnable on major OS.
So I think we should use default network mode(bridge)

From this test, I don't see how LRA container needs the ability to 'call back' to the test application running on the host, since it works for me now on my mac, please test it with your machine if you are on linux. If in any case we need call back from container to host, with default network mode, we can also do it(using hostname docker0 for linux and host.docker.internal for mac and windows)

I think my new commit addressed all concerns we have so far, please take another look.

Thanks!
Freeman

jamesnetherton:
> the expected return value is 4 instead of 3

I believe it should be 3. We attempt to create 4 orders, but one will end up as cancelled and the LRA will be rolled back. If you're seeing 4 as the result, it means the compensation `direct:cancelOrder` route is not being invoked. That's why the LRA container needs access to the host application, because it triggers the compensation actions.

ffang:
> > the expected return value is 4 instead of 3
> 
> I believe it should be 3. We attempt to create 4 orders, but one will end up as cancelled and the LRA will be rolled back. If you're seeing 4 as the result, it means the compensation `direct:cancelOrder` route is not being invoked. That's why the LRA container needs access to the host application, because it triggers the compensation actions.

Thanks @jamesnetherton  for the quick reply!

Yes, I got your point.
Instead of using localhost for the callback host, we need to use host.docker.internal for mac, and docker0 for linux machine
We need change like
```
-                    String.format("http://localhost:%s", System.getProperty("quarkus.http.test-port", "8081")));
+                    String.format("http://%s:%s", "host.docker.internal",
+                            System.getProperty("quarkus.http.test-port", "8081")));
```
I will push another commit.
Freeman

Issue #2550, opened 2021-05-04
several integration-test failure in native mode


ffang:
I encountered test failures in integration-test for 
file
hazelcast
minio 
hbase
when running tests with -Pnative

For the hbase, since it's in extensions-jvm folder, IIUC, this means it supports jvm mode only, so I assume it shouldn't have native profile.

For the hazelcast, I got exceptions " java.lang.IllegalStateException: ICMP not supported in this platform: mac os x" on my mac. So I think disable it on mac should be sufficient
```
com.oracle.svm.core.util.UserError$UserException: Class initialization failed for com.hazelcast.internal.util.ICMPHelper. The class is requested for re-running (reason: Quarkus)
	at com.oracle.svm.core.util.UserError.abort(UserError.java:82)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.rerunInitialization(ConfigurableClassInitialization.java:385)
	at io.quarkus.runner.AutoFeature.beforeAnalysis(AutoFeature.zig:1155)
	at com.oracle.svm.hosted.NativeImageGenerator.lambda$runPointsToAnalysis$7(NativeImageGenerator.java:701)
	at com.oracle.svm.hosted.FeatureHandler.forEachFeature(FeatureHandler.java:70)
	at com.oracle.svm.hosted.NativeImageGenerator.runPointsToAnalysis(NativeImageGenerator.java:701)
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:563)
	at com.oracle.svm.hosted.NativeImageGenerator.lambda$run$0(NativeImageGenerator.java:476)
	at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(ForkJoinTask.java:1407)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
Caused by: java.lang.ExceptionInInitializerError
[...]

PR #2553, opened 2021-05-04
#2550 fix several native integration-test failure

ffang:


oscerd (approved):

extensions/hazelcast/deployment/src/main/java/org/apache/camel/quarkus/component/hazelcast/deployment/HazelcastProcessor.java
```diff
@@ -51,6 +52,10 @@
 
     private static final String FEATURE = "camel-hazelcast";
 
+    private static final String[] RUNTIME_INITIALIZED_CLASSES = new String[] {
+            "com.hazelcast.internal.util.ICMPHelper"
```
jamesnetherton, line 56: Small nit - If there's just one class, do we need an array?

aldettinger (approved):

extensions/hazelcast/deployment/src/main/java/org/apache/camel/quarkus/component/hazelcast/deployment/HazelcastProcessor.java
ffang: Hi @jamesnetherton ,

I can use a String instead, but put a String Array here just in case this list may get expanded.
Thanks!
Freeman

jamesnetherton (approved):

PR #2562, opened 2021-05-07
Add additional SOAP extension test coverage

jamesnetherton:


oscerd (approved):

integration-tests/soap/src/main/java/org/apache/camel/quarkus/component/soap/it/service/CustomerServiceService.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.net.URL;
+
+import javax.xml.namespace.QName;
+import javax.xml.ws.Service;
+import javax.xml.ws.WebEndpoint;
+import javax.xml.ws.WebServiceClient;
+import javax.xml.ws.WebServiceFeature;
+
+@WebServiceClient(name = "CustomerServiceService", wsdlLocation = "file:/home/james/Projects/git/master/camel/components/camel-soap/src/test/resources/org/apache/camel/dataformat/soap/CustomerService.wsdl", targetNamespace = "http://service.it.soap.component.quarkus.camel.apache.org/")
```
aldettinger, line 27: I wonder if it would be possible to write something like wsdlLocation="classpath:org/apache/camel/dataformat/soap/CustomerService.wsdl" ?

jamesnetherton: Ah good spot. I thought I caught all of the hard coded paths. I'll fix

jamesnetherton: Latest commit removes the `wsdlLocation` annotation value as it's not actually needed in the context of this change.

integration-tests/soap/src/main/java/org/apache/camel/quarkus/component/soap/it/multipart/MultiPartCustomerService.java
```diff
@@ -0,0 +1,58 @@
...
+@XmlSeeAlso({ ObjectFactory.class })
+@SOAPBinding(parameterStyle = SOAPBinding.ParameterStyle.BARE)
+public interface MultiPartCustomerService {
+
+    @WebMethod(action = "http://multipart.it.soap.component.quarkus.camel.apache.org/saveCustomerToo")
+    public void saveCustomerToo(
+
+            @WebParam(partName = "parameters", name = "saveCustomerToo", targetNamespace = "http://multipart.it.soap.component.quarkus.camel.apache.org/") SaveCustomer parameters,
+            @WebParam(partName = "product", name = "product", targetNamespace = "http://multipart.it.soap.component.quarkus.camel.apache.org/", header = true) Product product,
+            @WebParam(partName = "company", mode = WebParam.Mode.INOUT, name = "company", targetNamespace = "http://multipart.it.soap.component.quarkus.camel.apache.org/", header = true) javax.xml.ws.Holder<Company> company);
```
aldettinger, line 36: I see we have some holders but not sure there are used in tests. Is worth covering that ? I'm not sure.

aldettinger:
Would we need more tests or docs for native in another PR ?
https://github.com/apache/camel/blob/main/components/camel-soap/src/main/java/org/apache/camel/dataformat/soap/Soap11DataFormatAdapter.java#L128
https://github.com/apache/camel/blob/main/components/camel-soap/src/main/java/org/apache/camel/dataformat/soap/Soap11DataFormatAdapter.java#L218
https://github.com/apache/camel/blob/main/components/camel-soap/src/main/java/org/apache/camel/dataformat/soap/Soap12DataFormatAdapter.java#L131
https://github.com/apache/camel/blob/main/components/camel-soap/src/main/java/org/apache/camel/dataformat/soap/Soap12DataFormatAdapter.java#L240

aldettinger:
Great to have the fault issue caught :+1: 

Concerning holder, the better understanding I can come up with would be that the pointed change removed support for BeanInvocation. So I guess what's still supported is another feature, namely passing Holder as method argument. However, I'm not sure there is any quarkus/graalvm specifics to be tested here.

What do you think ?

jamesnetherton:
> So I guess what's still supported is another feature, namely passing Holder as method argument.

I can take another look into that. It's not 100% clear to me how it was supposed to work.

> However, I'm not sure there is any quarkus/graalvm specifics to be tested here

Me too :slightly_smiling_face:. I suspect it's probably ok in native mode, but can't know for sure without a test.

I think we have a decent test baseline now, so I'd like to follow up on that with another issue if that's acceptable.

PR #2581, opened 2021-05-11
Support @EndpointInject and @Produce 

### LINKED ISSUE #2539: Support @EndpointInject and @Produce
* `@org.apache.camel.EndpointInject` https://camel.apache.org/manual/latest/pojo-producing.html#POJOProducing-EndpointInject
* `@org.apache.camel.Produce` https://camel.apache.org/manual/latest/pojo-producing.html#POJOProducing-HidingtheCamelAPIsFromYourCodeUsingProduce

ppalaga:
Fix #2539

### DIFF
--- a/integration-tests/bean/src/main/java/org/apache/camel/quarkus/component/bean/BeanResource.java
+++ b/integration-tests/bean/src/main/java/org/apache/camel/quarkus/component/bean/BeanResource.java
@@ -16,8 +16,12 @@
  */
 package org.apache.camel.quarkus.component.bean;
 
+import java.util.List;
+import java.util.Map;
+
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import javax.inject.Named;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
@@ -26,8 +30,8 @@ import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 
-import org.apache.camel.ConsumerTemplate;
 import org.apache.camel.Exchange;
+import org.apache.camel.Produce;
 import org.apache.camel.ProducerTemplate;
 import org.apache.camel.quarkus.component.bean.model.Employee;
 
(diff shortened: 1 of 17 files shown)


ppalaga:
da0060b fixed the test failing due to using a bad class loader at build time.

integration-tests/bean/src/main/java/org/apache/camel/quarkus/component/bean/BeanResource.java
```diff
@@ -142,4 +142,27 @@ public String parameterBindingAnnotations(String statement, @PathParam("greeting
                 String.class);
     }
 
+    @Path("/produceInterface")
+    @POST
+    @Consumes(MediaType.TEXT_PLAIN)
+    @Produces(MediaType.TEXT_PLAIN)
+    public String produceInterface(String payload) {
+        produceInterface.sayHello(payload);
```
aldettinger, line 150: Couldn't we simply `return produceInterface.sayHello(payload);` as this is direct under the hood ?

aldettinger (approved):
I've left an optional question in case it makes sense. Overall, looks good to me.

integration-tests/bean/src/main/java/org/apache/camel/quarkus/component/bean/BeanResource.java
ppalaga: Hm... you made me thinking. The purpose of the test is to make sure that the message goes via the Camel route starting with `direct:produceInterface`. Would there be any certainty if we just returned `sayHello(payload)`? I guess yes, but to be really sure, we'd have to add some transformation of the message to the route so that we see that it really passed it.

ppalaga: I can try it in the next iteration, when implementing `@Consume`

ppalaga: I have added a commit in https://github.com/apache/camel-quarkus/pull/2586 implementing your proposal, @aldettinger. Thanks for suggesting it!

PR #2597, opened 2021-05-14
Revisit core, main and foundation integration tests #2362

### LINKED ISSUE #2362: Revisit core, main and foundation integration tests
Things to improve:

* Minimize the dependencies (timer, log might get removed from the core test)
* Consider splitting some of the modules to Increase isolation, e.g. timer, and log do not have their isolated modules
* Consider introducing a grouped test to execute all modules at once on the CI
* Make sure we have both consumer and producer assertions for `direct`

ppalaga:
Fix #2362 

* Created a new foundation group
* Split out timer, log, direct, etc. to separate test modules
* Minimized the dependencies in the individual modules
* Have both consumer and producer assertions for direct

There is some more potential to bring in bean, simple, the old foundation tests and maybe some parts of the main tests to the new foundation group.


oscerd (approved):

ppalaga:
0d55453:
* Fixed an issue in the grouping script
* Moved Simple and Bean tests to the foundation group

integration-test-groups/foundation/README.adoc
```diff
@@ -0,0 +1,52 @@
+== Azure isolated integration tests
+
+=== Azurite
```
jamesnetherton, line 3: I think this file can be deleted.

ppalaga: +1

PR #2600, opened 2021-05-17
Load a classpath resource using current thread's context class loader…

ppalaga:
… in Saxon test, because otherwise it cannot be found


jamesnetherton (approved):

oscerd (approved):

zbendhiba (approved):

zbendhiba:
@ppalaga And now the file is not loaded. Maybe just do like other tests of this class : put the content of the XML file directly as a String.

ppalaga:
> And now the file is not loaded.

Yeah, when loading from a class loader the slash needs be removed from `/myinput.xml`. Done in 833da24

ppalaga:
BTW this one can also be applied on `main` so I re-target it.

ppalaga:
I have cherry-picked the commit to https://github.com/apache/camel-quarkus/pull/2614 to save some CI resources

PR #2614, opened 2021-05-18
Test core languages in isolation

### LINKED ISSUE #2613: Test core languages in isolation
`camel-quarkus-core` brings support for the following languages:

* Constant
* ExchangeProperty
* File
* Header
* Ref
* Simple
* Tokenize

All of these except for the File language should be tested with minimal dependencies (core and direct).

Testing the File language makes only sense together with some component operating on `org.apache.camel.component.file.GenericFile`, such as File, FTP, etc. For file component, this was implemented in https://github.com/apache/camel-quarkus/pull/2559

### LINKED ISSUE #2616: Test data format EIPs
Already supported data formats should have tests for the given dataformat Java DSL method. 

E.g. for Bindy, in addition to the existing

```
        BindyDataFormat bindyCsvDataFormat = new BindyDataFormat();
        bindyCsvDataFormat.setClassType(CsvOrder.class);
        bindyCsvDataFormat.setType(BindyType.Csv.name());
        from("direct:marshal-csv-record").marshal(bindyCsvDataFormat);
        from("direct:unmarshal-csv-record").unmarshal(bindyCsvDataFormat);
```

we should also test the following (equivalent) notation:

```
        from("direct:marshal-csv-record").marshal().bindy(BindyType.Csv, CsvOrder.class);
        from("direct:unmarshal-csv-record").unmarshal().bindy(BindyType.Csv, CsvOrder.class);
```

ppalaga:
Fix #2613
Fix #2616


oscerd (approved):

integration-test-groups/foundation/core-languages/src/main/java/org/apache/camel/quarkus/core/languages/it/ConstantLanguageRoutes.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.apache.camel.builder.RouteBuilder;
+
+@Singleton
+public class ConstantLanguageRoutes extends RouteBuilder {
+
+    @Inject
```
aldettinger, line 30: Maybe the counter is not needed in this case ?

aldettinger:
Nice addition, the test structure is good. In terms of coverage, there is a single point I'm not sure about. The [tokenize-language](https://camel.apache.org/components/latest/languages/tokenize-language.html) page references an `xml` option which is false by default. Is it still in use ? Is it worth testing that ? I would be thinking about XML dependencies that might be needed for instance.

ppalaga:
> he [tokenize-language](https://camel.apache.org/components/latest/languages/tokenize-language.html) page references an `xml` option which is false by default. Is it still in use ? Is it worth testing that ?

I have added a test for `tokenizeXML()` which seems to be setting that xml attribute to true.

ppalaga:
Added a couple of fixes for #2616

PR #2644, opened 2021-05-21
Ensure camel-quarkus-integration-wiremock-support is test scope

jamesnetherton:


aldettinger:
Argh, I think in the case of digital ocean it will be needed at runtime.

ppalaga (approved):

Issue #2368, opened 2021-03-24
Store the association between Camel Quarkus Catalog items and Camel Catalog items


lburgazzoli:
This would be useful because as today in camel-k you can declare a dependency by using the artifact id, so as example, to include the direct component you can use

    kamel run integration.yaml -d camel-direct

With quarkus, you need to use the quarkus one, like:

    kamel run integration.yaml -d camel-quarkus-direct

But since the runtime is an implementation detail, it would be nice to let users declaring dependencies using the camel artifact id and to make it possible we need to have the information about the artifact the extension has been created from.

ppalaga:
A Camel Quarkus extension can depend on zero to *n* Camel artifacts. Do you have an idea where/how to store that info? 
Not sure adding a field to `org.apache.camel.tooling.model.ArtifactModel` is the best way to go. It would have little to no meaning in the context of Camel Core. Maybe we should add a free form map for this kind of ad hoc attributes that are only meaningful for Camel subprojects?

lburgazzoli:
I don't have any opinion about where to store such info, so generic attributes would work for me

ppalaga:
Created https://issues.apache.org/jira/browse/CAMEL-16557

ppalaga:
Blocked by https://github.com/apache/camel/pull/5471

ppalaga:
Looking at this in greater detail, the entries (components, languages and data formats) in camel quarkus catalog have zero or one camel artifact they come from. I am implementing something like this:
![image](https://user-images.githubusercontent.com/1826249/119336522-0a5a4c80-bc8e-11eb-88d3-fd06c62a718c.png)

ppalaga:
Fixed via a1260cb1ed0b19dbfe6010621532e9be326df489

PR #2661, opened 2021-05-25
Avoid overwriting the Qute json which breaks Camel K catalogs

orpiske:
This should solve apache/camel-k-runtime#677

Without this, the catalog seems to generate its own bogus qute.json. Then, during camel-k-runtime build, it gets picked instead of the one available in camel-quarkus-qute-component jar.

### DIFF
--- a/catalog/pom.xml
+++ b/catalog/pom.xml
@@ -4403,8 +4403,27 @@
                 <executions>
                     <execution>
                         <!-- prepare the catalog and update doc files, etc. -->
+                        <id>prepare-catalog</id>
                         <goals>
                             <goal>prepare-catalog-quarkus</goal>
+                        </goals>
+                        <phase>process-resources</phase>
+                        <configuration>
+                            <extensionDirectories>
+                                <extensionDirectory>${maven.multiModuleProjectDirectory}/extensions-core</extensionDirectory>
+                                <extensionDirectory>${maven.multiModuleProjectDirectory}/extensions</extensionDirectory>
+                                <extensionDirectory>${maven.multiModuleProjectDirectory}/extensions-jvm</extensionDirectory>
+                            </extensionDirectories>
+                            <skipArtifactIdBases>
+                                <skipArtifactIdBase>http-common</skipArtifactIdBase>
+                                <skipArtifactIdBase>qute</skipArtifactIdBase>
+                            </skipArtifactIdBases>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <!-- update doc files, etc. -->
+                        <id>update-doc-files</id>
+                        <goals>
                             <goal>check-extension-pages</goal>
                         </goals>
                         <phase>process-resources</phase>


oscerd (approved):

catalog/pom.xml
```diff
@@ -2979,7 +2979,7 @@
         </dependency>
         <dependency>
             <groupId>org.apache.camel.quarkus</groupId>
-            <artifactId>camel-quarkus-qute</artifactId>
+            <artifactId>camel-quarkus-qute-component</artifactId>
```
ppalaga, line 2982: These virtual deps only serve as module ordering hints for mvnd or mvn -T. Moreover they are 100% generated. If you run `mvn process-resources -Pformat -N` from the repo root (which recommended after changing dependencies), the change will get reverted. Could you please explain your motivation for the change?

jamesnetherton:
Looking at the CI failure - maybe we need to split the `prepare-catalog-quarkus` and `check-extension-pages` goals into separate executions with their own configuration. Otherwise the extension docs will get removed for `qute`.

orpiske:
> Looking at the CI failure - maybe we need to split the `prepare-catalog-quarkus` and `check-extension-pages` goals into separate executions with their own configuration. Otherwise the extension docs will get removed for `qute`.

Thanks. 

Do you have something like this in mind?

```
<execution>
                        <!-- prepare the catalog and update doc files, etc. -->
                        <goals>
                            <goal>prepare-catalog-quarkus</goal>
                        </goals>
                        <phase>process-resources</phase>
                        <configuration>
                            <extensionDirectories>
                                <extensionDirectory>${maven.multiModuleProjectDirectory}/extensions-core</extensionDirectory>
                                <extensionDirectory>${maven.multiModuleProjectDirectory}/extensions</extensionDirectory>
                                <extensionDirectory>${maven.multiModuleProjectDirectory}/extensions-jvm</extensionDirectory>
                            </extensionDirectories>
                            <skipArtifactIdBases>
                                <skipArtifactIdBase>http-common</skipArtifactIdBase>
                                <skipArtifactIdBase>qute</skipArtifactIdBase>
                            </skipArtifactIdBases>
                        </configuration>
                    </execution>
                    <execution>
                        <!-- prepare the catalog and update doc files, etc. -->
                        <goals>
                            <goal>check-extension-pages</goal>
                        </goals>
                        <phase>process-resources</phase>
                        <configuration>
                            <extensionDirectories>
                                <extensionDirectory>${maven.multiModuleProjectDirectory}/extensions-core</extensionDirectory>
                                <extensionDirectory>${maven.multiModuleProjectDirectory}/extensions</extensionDirectory>
                                <extensionDirectory>${maven.multiModuleProjectDirectory}/extensions-jvm</extensionDirectory>
                            </extensionDirectories>
                            <skipArtifactIdBases>
                                <skipArtifactIdBase>http-common</skipArtifactIdBase>
                            </skipArtifactIdBases>
                        </configuration>
                    </execution>
```

catalog/pom.xml
```diff
@@ -4416,6 +4416,7 @@
                             </extensionDirectories>
                             <skipArtifactIdBases>
                                 <skipArtifactIdBase>http-common</skipArtifactIdBase>
+                                <skipArtifactIdBase>qute</skipArtifactIdBase>
```
ppalaga, line 4419: Wouldn't this actually exclude the Qute component from the Camel Quarkus Catalog? - Is that what you want?

ppalaga (commented):
Some questions inline

catalog/pom.xml
orpiske: Thanks Peter. I actually want to prevent the catalog from including a bogus `qute.json` that is causing this issue https://github.com/apache/camel-k-runtime/issues/677

orpiske: Let me try without this one.

catalog/pom.xml
orpiske: I actually want to prevent the catalog from including a bogus `qute.json` that is causing this issue apache/camel-k-runtime#677. I'm adjusting the PR w/ your notes + @jamesnetherton' ones.

ppalaga: Reading https://github.com/apache/camel-k-runtime/issues/677 I am still not 100% sure what is the problem. `org.apache.camel.component.qute.QuteComponent` exists in `org.apache.camel.quarkus:camel-quarkus-qute-component:jar` Don't you need to add it to your class path?

orpiske: I think the one bundled in the catalog jar is taking precedence in the classpath. Thus causing the issue on 667.

ppalaga: Actually, we very recently started adding camel artifact coordinates to our catalog https://github.com/apache/camel-quarkus/issues/2368 I wonder whether that new `metadata: {camelArtifact: }` shouldn't be `org.apache.camel.quarkus:camel-quarkus-qute-component` for qute. Now it is probably empty, because it is a special case that is not handled properly.

orpiske: @ppalaga I'm leaning towards no, in this case, but I wouldn't know for sure. I think that, in the case of the Camel K catalog, and specifically in this problem, it is reading the `javaType` info and not the metadata.

orpiske:
@ppalaga @jamesnetherton I've test the updated PR, based on your comments and it worked for the Camel K Runtime. It is, indeed, caused by the bogus `qube.json` file on the catalog jar.

That said, I understand it may not be desirable from Camel Quarkus perspective. So I'd like to hear your insights: do you feel this is an appropriate approach or should I pursue a different strategy for dealing with this?

jamesnetherton:
@ppalaga how do you want to handle this? This would remove remove the broken `qute.json` from the catalog. Is there a simple means of getting a correct JSON file included?

ppalaga:
> @ppalaga how do you want to handle this? This would remove remove the broken `qute.json` from the catalog. Is there a simple means of getting a correct JSON file included?

I am still striving to understand what's wrong with qute.json from Camel K perspective. `javaType` being `org.apache.camel.component.qute.QuteComponent` is correct from CQ's point of view. The type exists in `org.apache.camel.quarkus:camel-quarkus-qute-component`. My working hypothesis is that Camel K does some wrong assumption about where to find that class (maybe it thinks it is in `org.apache.camel:camel-qute` which does not exist apparently). And that's why I say we should check/fix what we output to `metadata: {camelArtifact: }` of `qute.json` so that Camel K has an authoritative source for the Maven coordinates under which the `javaType` is supposed to exist.
I am all for clarifying this.

Removing `qute.json` from CQ catalog would not be correct. The extension and component exist and we support it. Removal would impact a couple of our tools that utilize the info from the catalog.

ppalaga:
I checked what `qute.json` looks like and indeed the `metadata` is missing. Also `componentProperties` and `properties` are missing. @orpiske could you please check which particular parts are required by Camel K?

orpiske:
> I checked what `qute.json` looks like and indeed the `metadata` is missing. Also `componentProperties` and `properties` are missing. @orpiske could you please check which particular parts are required by Camel K?

Thanks for looking at this. We need (using `activemq` one as a reference only):

1. The `javaType`: https://github.com/apache/camel/blob/main/components/camel-activemq/src/generated/resources/org/apache/camel/component/activemq/activemq.json#L10
2. The `scheme`: https://github.com/apache/camel/blob/main/components/camel-activemq/src/generated/resources/org/apache/camel/component/activemq/activemq.json#L15
3. The `groupId` and `artifactId`: https://github.com/apache/camel/blob/main/components/camel-activemq/src/generated/resources/org/apache/camel/component/activemq/activemq.json#L12-L13

ppalaga:
Agreed to fix differently: https://github.com/apache/camel-quarkus/issues/2701

Issue #2681, opened 2021-05-26
camel-quarkus-jacksonxml:JsonView annotations take no effect in native mode


ffang:
Can't control the xml payload(with selected properties), seems somehow JsonView doesn't work in native mode only

ffang:
It turns out to be that JsonView annotation relys on the reflection to load class like
```
@JsonView(Views.Weight.class)
    private int weight = 70;
```
And the class here(Both inner class and outer class should be marked as @RegisterForReflection during runtime)

ppalaga:
Fixed via 5d2ef77f7fb56379a77eaad55fbe8ab029f92e20

PR #2682, opened 2021-05-26
fix camel-quarkus-jacksonxml:JsonView annotations take no effect in n…

ffang:
…ative mode #2681

### DIFF
--- a/extensions/jacksonxml/deployment/src/main/java/org/apache/camel/quarkus/component/jacksonxml/deployment/JacksonxmlProcessor.java
+++ b/extensions/jacksonxml/deployment/src/main/java/org/apache/camel/quarkus/component/jacksonxml/deployment/JacksonxmlProcessor.java
@@ -16,8 +16,17 @@
  */
 package org.apache.camel.quarkus.component.jacksonxml.deployment;
 
+import java.util.function.Function;
+
+import com.fasterxml.jackson.annotation.JsonView;
 import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+import org.jboss.jandex.AnnotationValue;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.IndexView;
+import org.jboss.jandex.Type;
 
 class JacksonxmlProcessor {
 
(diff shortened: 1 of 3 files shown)


oscerd (approved):

ppalaga:
Could perhaps the extension auto-register the required classes?

jamesnetherton:
> Could perhaps the extension auto-register the required classes?

+1 I had the same thought. It'd make the UX better.

ppalaga:
You can use Jandex to find instances of `@JsonView` in the user code. Here is some overview https://peter.palaga.org/presentations/210407-quarkus-insights-how-to-write-quarkus-extensions/index.html#/jandex-example 
The pattern is used several times in Camel Quarkus code base, just grep for `CombinedIndexBuildItem`.

ffang:
Hi @ppalaga , @jamesnetherton ,

Just pushed a new commit which uses jandex to read JsonView annotations and add class to reflection list per your guys suggested.

Thanks!
Freeman

extensions/jacksonxml/deployment/src/main/java/org/apache/camel/quarkus/component/jacksonxml/deployment/JacksonxmlProcessor.java
```diff
@@ -28,4 +37,21 @@ FeatureBuildItem feature() {
...
+    @BuildStep
+    ReflectiveClassBuildItem registerJsonView(CombinedIndexBuildItem combinedIndex) {
+
+        IndexView index = combinedIndex.getIndex();
+        DotName JSON_VIEW = DotName.createSimple(JsonView.class.getName());
+        String[] jsonViews = index.getAnnotations(JSON_VIEW).stream().map(ai -> ai.value())
+                .filter(p -> AnnotationValue.Kind.ARRAY.equals(p.kind()))
+                .map((Function<? super AnnotationValue, ? extends String>) ai -> {
+                    Type[] annotationType = ai.asClassArray();
+                    return annotationType[0].name().toString();
```
ppalaga, line 49: Why are we taking only the first array element into account? IMO all elements need to be registered for reflection.

ffang: You are right, I will send another commit.
Thanks!

ffang: Hi @ppalaga ,

And here is the follow up PR
https://github.com/apache/camel-quarkus/pull/2709

Please take a look.
Freeman

PR #2698, opened 2021-06-01
Upgrade SmallRye Reactive Messaging Camel to 3.3.2

jamesnetherton:


oscerd (approved):

zbendhiba (approved):

pom.xml
```diff
@@ -113,7 +113,7 @@
         <retrofit.version>2.5.0</retrofit.version>
         <saaj.impl.version>1.5.3</saaj.impl.version>
         <scala-2.11.version>2.11.12</scala-2.11.version><!-- Spark -->
-        <smallrye.reactive.messaging.camel.version>2.9.0</smallrye.reactive.messaging.camel.version> <!-- keep in sync with Quarkus SmallRye Reactive Messaging -->
+        <smallrye.reactive.messaging.camel.version>3.3.2</smallrye.reactive.messaging.camel.version> <!-- keep in sync with Quarkus SmallRye Reactive Messaging -->
```
ppalaga, line 116: Should this be in sync with some version property in Quarkus or elsewhere? (We now have a tool that could do that semi-automatically.)

zbendhiba: @ppalaga what tool ?

jamesnetherton: Yeah it should be in sync with the Quarkus `smallrye-reactive-messaging.version` property.

ppalaga: > @ppalaga what tool ?

Items annotated with @sync can be updated by running `mvn cq:sync-versions -N -e`

Cited from 
https://github.com/apache/camel-quarkus/commit/7d238a1b8354502e1a277ffe0b7d136d857434df#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R55

ppalaga, line 116: Then, something like this could work:

```suggestion
        <smallrye.reactive.messaging.camel.version>3.3.2</smallrye.reactive.messaging.camel.version><!-- @sync io.quarkus:quarkus-bom:${quarkus.version} dep:io.smallrye.reactive:smallrye-reactive-messaging-provider -->
```

Note that the rule does not reference the version property, because the version property is not available in the flattened BOM they publish to Central. So the tool rather picks the version from a dependency that is known to use the desired version property.

jamesnetherton: Thanks, applied.

zbendhiba: thanks @ppalaga

PR #2699, opened 2021-06-01
Document usage of dashed query params with platform-http

jamesnetherton:


oscerd (approved):

aldettinger (approved):
Ok, I was in the idea that the doc could be located in camel but actually it's only vertx.
Well done @jamesnetherton .

ppalaga (approved):

Issue #2726, opened 2021-06-04
Fix JSON Jackson jacksonConversionPojo test


jamesnetherton:
Seem comment here:

https://github.com/apache/camel-quarkus/blob/f65f3b92610cdc577403fde5fbaf846ae7a58965/integration-tests/dataformats-json/src/main/java/org/apache/camel/quarkus/component/dataformats/jackson/json/JacksonJsonResource.java#L515-L520

ffang:
Hi @jamesnetherton,

FYI, I believe this has been fixed by https://github.com/apache/camel-quarkus/issues/2677 already.
Sorry, I forget to re-enable this test. I will send a PR to re-enable this test soon.

Best Regards
Freeman

ppalaga:
Any update, @ffang ?

PR #2739, opened 2021-06-07
Fix JSON Jackson jacksonConversionPojo test #2726

ffang:


oscerd (approved):

aldettinger:
Thanks for contribution @ffang. As far as I understand, it's all about testing the case where JAXB annotation support is configured on the whole context. So, there is a difficulty because we have tests sharing a single context that we want to be configured differently given the test we are currently executing.

If my understanding is correct, please find questions below:
1) Is it really needed to test this case in camel-quarkus ? I think the test is already covered in camel, is there anything specific to quarkus/graal in this situation ?
2) Couldn't we face situations where 2 JacksonJsonResource methods are called concurrently ? And then global options could be applied during another test as the context is not synchronized in other methods ?
3) When testing scenarios involving globalOptions(), would it be better to have a new CamelContext for each subset of global options ? Maybe another itest project or something like a quarkus test profile ?

ffang:
Hi @aldettinger ,

Thanks for the feedback, and please see my comments inline

> Thanks for contribution @ffang. As far as I understand, it's all about testing the case where JAXB annotation support is configured on the whole context. So, there is a difficulty because we have tests sharing a single context that we want to be configured differently given the test we are currently executing.
> 
> If my understanding is correct, please find questions below:
> 
>     1. Is it really needed to test this case in camel-quarkus ? I think the test is already covered in camel, is there anything specific to quarkus/graal in this situation ?
I just use the conservative way to add more tests, without assuming anything.

And this actually helps us find a issue in camel-quarkus native mode, 
this line
```
context.getGlobalOptions().put(JacksonConstants.TYPE_CONVERTER_MODULE_CLASS_NAMES,
                    JaxbAnnotationModule.class.getName());
```
Fasterxml Jackson uses reflection way to init JaxbAnnotationModule, so we need to add a ReflectiveClassBuildItem in camel-quarkus jackson extension
> 
>     2. Couldn't we face situations where 2 JacksonJsonResource methods are called concurrently ? And then global options could be applied during another test as the context is not synchronized in other methods ?
> 
Here the problem isn't the global options, but how camel-jackson JacksonTypeConverters works
```
@Converter(fallback = true)
    public <T> T convertTo(Class<T> type, Exchange exchange, Object value, TypeConverterRegistry registry) throws Exception {

        // only do this if enabled (disabled by default)
        if (!init && exchange != null) {
```
It only init once and after that all new global options  can't take effect(I think this is OK in real world camel usage, our test just share the same camel context which isn't very reasonable). That's why in the PR I just re-create the convert method to ensure any new global options before it can take effect.

And I think I put all global options  in sync block already.


>     3. When testing scenarios involving globalOptions(), would it be better to have a new CamelContext for each subset of global options ? Maybe another itest project or something like a quarkus test profile ?
Definitely you are correct, a new camelcontext for each testcase should be more isolated and clear. but I don't know how to do it easily in camel-quarkus test framework. So feel free to shed light or even point me an existing example.

Freeman

aldettinger:
Ok, great to know that the test is needed and that we caught up another native issue :+1: 

Related to point 3, I agree that having tests with a new context could help in such a situation. To my knowledge, there is no example for that in camel-quarkus. The only ideas I would have would be to create a new itest project or maybe investigate if something could be done with [quarkus test profile](https://quarkus.io/guides/getting-started-testing#testing_different_profiles).

I'm not sure about point 2, could you please confirm my understanding ? My understanding would be that we override the fallback converter, then we set the global options, then the converter is initialized on next call to `convertTo()` and from there will not change afterward. So, the overridden fallback converter is sometimes present when other tests are executed. But I guess that's no big deal because other tests are assumed not to use the fallback converter ?

ffang:
> Ok, great to know that the test is needed and that we caught up another native issue 👍
> 
> Related to point 3, I agree that having tests with a new context could help in such a situation. To my knowledge, there is no example for that in camel-quarkus. The only ideas I would have would be to create a new itest project or maybe investigate if something could be done with [quarkus test profile](https://quarkus.io/guides/getting-started-testing#testing_different_profiles).
> 
A new itest project for a specific testcase is a bit overkill IMO.

I will take a look at the quarkus test profile. And maybe we can create follow up issues for it if applicable, since I believe this is a general requirement across all integration tests.

> I'm not sure about point 2, could you please confirm my understanding ? My understanding would be that we override the fallback converter, then we set the global options, then the converter is initialized on next call to `convertTo()` and from there will not change afterward. So, the overridden fallback converter is sometimes present when other tests are executed. But I guess that's no big deal because other tests are assumed not to use the fallback converter ?

Yes, that's right.  If in the future other new added testcase need to use the converter we should use the same way(as this PR) to recreate the clean converter, provided  all test cases still have to share the same camel context.

aldettinger (approved):
Ok, I totally agree with the proposed view. Many thanks for explanation @ffang :+1:

ppalaga:
The same kind of questions like @aldettinger asked came to my mind. 

> > ```
> > 3. When testing scenarios involving globalOptions(), would it be better to have a new CamelContext for each subset of global options ? Maybe another itest project or something like a quarkus test profile ?
> > ```
> 
> 
> Definitely you are correct, a new camelcontext for each testcase should be more isolated and clear. but I don't know how to do it easily in camel-quarkus test framework. So feel free to shed light or even point me an existing example.

Yeah, having a separate itest module for each of the tested converter setups would probably be an overkill (although we have a couple of itest modules for e.g. camel-main). I wonder whether this is not a use case for a `io.quarkus.test.QuarkusUnitTest`. We have such tests e.g. in [Servlet extension](https://github.com/apache/camel-quarkus/tree/main/extensions/servlet/deployment/src/test/java/org/apache/camel/quarkus/component/servlet/test). A big advantage of those is that they are very lightweight and fast. You can have a special context setup per test class. But OTOH, they are executed only in JVM mode. So we maybe need to find a way how to keep a part of the scenario in a proper itest and move the rest to QuarkusUnitTest-style test.

I'd say if the second setup with `context.getGlobalOptions().put(JacksonConstants.TYPE_CONVERTER_MODULE_CLASS_NAMES,
                    JaxbAnnotationModule.class.getName());` uncovered the need for the reflection registration, then that setup should stay ghere and be hardwired for the whole itest module. The first setup (without TYPE_CONVERTER_MODULE_CLASS_NAMES) could then be transformed to a separate `io.quarkus.test.QuarkusUnitTest`. 

In that way we'd both have clean and readable tests without needing to configure the context in the middle of the flight and we still cover what we need in native mode. WDYT?

ffang:
> The same kind of questions like @aldettinger asked came to my mind.
> 
> > > ```
> > > 3. When testing scenarios involving globalOptions(), would it be better to have a new CamelContext for each subset of global options ? Maybe another itest project or something like a quarkus test profile ?
> > > ```
> > 
> > 
> > Definitely you are correct, a new camelcontext for each testcase should be more isolated and clear. but I don't know how to do it easily in camel-quarkus test framework. So feel free to shed light or even point me an existing example.
> 
> Yeah, having a separate itest module for each of the tested converter setups would probably be an overkill (although we have a couple of itest modules for e.g. camel-main). I wonder whether this is not a use case for a `io.quarkus.test.QuarkusUnitTest`. We have such tests e.g. in [Servlet extension](https://github.com/apache/camel-quarkus/tree/main/extensions/servlet/deployment/src/test/java/org/apache/camel/quarkus/component/servlet/test). A big advantage of those is that they are very lightweight and fast. You can have a special context setup per test class. But OTOH, they are executed only in JVM mode. So we maybe need to find a way how to keep a part of the scenario in a proper itest and move the rest to QuarkusUnitTest-style test.
> 
> I'd say if the second setup with `context.getGlobalOptions().put(JacksonConstants.TYPE_CONVERTER_MODULE_CLASS_NAMES, JaxbAnnotationModule.class.getName());` uncovered the need for the reflection registration, then that setup should stay ghere and be hardwired for the whole itest module. The first setup (without TYPE_CONVERTER_MODULE_CLASS_NAMES) could then be transformed to a separate `io.quarkus.test.QuarkusUnitTest`.
> 
> In that way we'd both have clean and readable tests without needing to configure the context in the middle of the flight and we still cover what we need in native mode. WDYT?

Thanks @ppalaga for the input!

I will take a close look at the tests used in servlet extension and see if we can split some of our testcases this way.

Freeman

ppalaga:
Could you please rebase and squash your commits, @ffang so that we can merge straight away once this passes the review?

ffang:
Hi @ppalaga ,

Please find squashed commit here
https://github.com/apache/camel-quarkus/pull/2739/commits/751063196f3a99e07906cf69b512f6ab970392e7

This contains all changes for this issue.
Cheers
Freeman

ffang:
somehow this PR was messed up. Close this one and sent a new PR
https://github.com/apache/camel-quarkus/pull/2790
We can continue the discussion there

Issue #2719, opened 2021-06-03
hazelcast integration-test failed in native mode(On Mac)


ffang:
Got error like
```
[camel-quarkus-integration-test-hazelcast-2.0.0-SNAPSHOT-runner:44470]        setup:   3,991.97 ms,  1.19 GB
com.oracle.svm.core.util.UserError$UserException: Incompatible change of initialization policy for com.hazelcast.internal.util.ICMPHelper: trying to change RUN_TIME from feature io.quarkus.runner.AutoFeature.beforeAnalysis to RERUN Quarkus
	at com.oracle.svm.core.util.UserError.abort(UserError.java:68)
	at com.oracle.svm.hosted.classinitialization.ClassInitializationConfiguration.insertRec(ClassInitializationConfiguration.java:99)
	at com.oracle.svm.hosted.classinitialization.ClassInitializationConfiguration.insertRec(ClassInitializationConfiguration.java:113)
	at com.oracle.svm.hosted.classinitialization.ClassInitializationConfiguration.insertRec(ClassInitializationConfiguration.java:113)
	at com.oracle.svm.hosted.classinitialization.ClassInitializationConfiguration.insertRec(ClassInitializationConfiguration.java:113)
	at com.oracle.svm.hosted.classinitialization.ClassInitializationConfiguration.insertRec(ClassInitializationConfiguration.java:113)
	at com.oracle.svm.hosted.classinitialization.ClassInitializationConfiguration.insertRec(ClassInitializationConfiguration.java:113)
	at com.oracle.svm.hosted.classinitialization.ClassInitializationConfiguration.insert(ClassInitializationConfiguration.java:64)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.rerunInitialization(ConfigurableClassInitialization.java:379)
	at io.quarkus.runner.AutoFeature.beforeAnalysis(AutoFeature.zig:1178)
	at com.oracle.svm.hosted.NativeImageGenerator.lambda$runPointsToAnalysis$7(NativeImageGenerator.java:701)
	at com.oracle.svm.hosted.FeatureHandler.forEachFeature(FeatureHandler.java:70)
	at com.oracle.svm.hosted.NativeImageGenerator.runPointsToAnalysis(NativeImageGenerator.java:701)
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:563)
[...]

PR #2747, opened 2021-06-08
fix hazelcast integration-test failed in native mode(On Mac) #2719

ffang:


oscerd (approved):

ppalaga (changes requested):
I'd very much prefer avoiding `--allow-incomplete-classpath` even on Mac. Why do we need it?

zbendhiba:
Can we report this one to Hazelcast ?

jamesnetherton:
> I'd very much prefer avoiding `--allow-incomplete-classpath` even on Mac. Why do we need it?

I was wondering that too. I think it's due to this:

https://github.com/hazelcast/hazelcast/blob/e3352af34221c58de14ed09dcde0edc9206098c8/hazelcast/src/main/java/com/hazelcast/internal/util/ThreadAffinityHelper.java#L56-L72

On Linux, some native libs bundled in the JAR are loaded for thread affinity support. On Mac you need to add a dependency:

https://github.com/hazelcast/hazelcast/blob/09347d52d29d40ce2e9ed6515c698ed773cd3061/hazelcast/pom.xml#L389-L396

zbendhiba:
> Can we report this one to Hazelcast ?

The issue I mean

ppalaga:
> On Linux, some native libs bundled in the JAR are loaded for thread affinity support. On Mac you need to add a dependency:
> 
> https://github.com/hazelcast/hazelcast/blob/09347d52d29d40ce2e9ed6515c698ed773cd3061/hazelcast/pom.xml#L389-L396

Indeed, it sounds like the place for the solution would be [Hazelcast Client](https://github.com/hazelcast/quarkus-hazelcast-client) themselves? Maybe they should have a Mac profile in their runtime module? Hard to say how well it would work with Quarkus Maven plugin

ppalaga:
This also brings forward the question whether we should have an ambition to support the native compilation on Mac (and Windows) at all? I mean what is the use case beyond demos? I guess close to nobody is running on Mac/Windows in production. Building in Docker on Mac/Windows should work, but that produces a Linux executable. It is true that such executable does not help when one wants to run native tests on Mac/Windows. So maybe that's the use case because of which we should try to support native compilation on Mac? I wonder if it would not be better if Quarkus JUnit extension could transparently boot the Linux executable via Docker and run the tests against against it? Hm... maybe it is too much work and it is easier to properly fix issues like this one. Sorry for the long essay :)

ffang:
Thanks @zbendhiba @jamesnetherton @ppalaga for the input!

I agree that we probably shouldn't put effort to support native build on mac or windows, very clear that nobody would use non-linux platform in production. Just from a developer on mac POV, I'd like to have a green native build on mac. So IMO, whatever the ways(DisableTests, properties configurations to hack/workaround, but not a real support in production, probably we can document this point) we can make a green build on mac, that would be a great experience for community developers/users on Mac.

Just my 2 cents
Freeman

zbendhiba:
@ffang I agree on document it at least. This workaround can benefit for other Mac users.

pivovarit:
To be honest, I don't really know how to solve it so that incomplete classpath is somehow acceptable without `--allow-incomplete-classpath` which I'd like to avoid as well.

Introducing the MacOS profile with the optional dependency sounds like something that could work, but is there anyone who could confirm that it's a legit solution for Quarkus?

ppalaga:
> Introducing the MacOS profile with the optional dependency sounds like something that could work, but is there anyone who could confirm that it's a legit solution for Quarkus?

I think the best way forward would be that somebody on Mac please checks whether adding a profile like the following here or in [Quarkus Hazelcast Client](https://github.com/hazelcast/quarkus-hazelcast-client/blob/master/runtime/pom.xml) solves the issue for her/him:

```
        <profile>
            <id>hazelcast-mac</id>
            <activation>
                <os><family>mac</family></os>
            </activation>
            <dependencies>
                <dependency>
                    <groupId>net.openhft</groupId>
                    <artifactId>affinity</artifactId>
                    <version>3.20.0</version>
                </dependency>
            </dependencies>
        </profile>
```

ffang:
Hello everyone,

Please take a look at the commits here
https://github.com/apache/camel-quarkus/pull/2747/commits/0a07b5bbfd9b2b10a6e1945e34666dab5a89aa65

The profile per you guys suggestd
```
<profile>
            <id>hazelcast-mac</id>
            <activation>
                <os><family>mac</family></os>
            </activation>
            <dependencies>
                <dependency>
                    <groupId>net.openhft</groupId>
                    <artifactId>affinity</artifactId>
                    <version>3.20.0</version>
                </dependency>
            </dependencies>
        </profile>

```
can address the issue "Discovered unresolved type during parsing: net.openhft.affinity.Affinity."

But I still need to add com.hazelcast.internal.util.ICMPHelper into Runtime Initialized class list. Actually this  was added by this commit 1580be6d3337a864847d9edbfdc6f788ebbc4c78(though I'm not sure if we have better ways to fix it), but was reverted by this later commit f8d6447138ea602da3995312291090448c703435 by @ppalaga 

Thanks!
Freeman

ppalaga:
> Hello everyone,
> 
> Please take a look at the commits here
> [0a07b5b](https://github.com/apache/camel-quarkus/commit/0a07b5bbfd9b2b10a6e1945e34666dab5a89aa65)
> 
> The profile per you guys suggestd
> 
> ```
> <profile>
>             <id>hazelcast-mac</id>
>             <activation>
>                 <os><family>mac</family></os>
>             </activation>
>             <dependencies>
>                 <dependency>
>                     <groupId>net.openhft</groupId>
>                     <artifactId>affinity</artifactId>
>                     <version>3.20.0</version>
>                 </dependency>
>             </dependencies>
>         </profile>
> ```
> 
> can address the issue "Discovered unresolved type during parsing: net.openhft.affinity.Affinity."

Nice to hear that.

> But I still need to add com.hazelcast.internal.util.ICMPHelper into Runtime Initialized class list. Actually this was added by this commit [1580be6](https://github.com/apache/camel-quarkus/commit/1580be6d3337a864847d9edbfdc6f788ebbc4c78)(though I'm not sure if we have better ways to fix it), but was reverted by this later commit [f8d6447](https://github.com/apache/camel-quarkus/commit/f8d6447138ea602da3995312291090448c703435) by @ppalaga

Could you please try to add it back so that we see whether it passes on our Linux CI?

There are suspiciously many commits in this PR. Maybe you pushed a wrong branch?

ffang:
> 
> Could you please try to add it back so that we see whether it passes on our Linux CI?
I've added it back in the commit here already.
https://github.com/apache/camel-quarkus/commit/0a07b5bbfd9b2b10a6e1945e34666dab5a89aa65

> 
> There are suspiciously many commits in this PR. Maybe you pushed a wrong branch?
Not sure what I did wrong, probably I should close this PR and send a new PR with clean commit there

ffang:
close this messed PR and sent a new one
https://github.com/apache/camel-quarkus/pull/2789
We can continue discussion there

PR #2765, opened 2021-06-11
Avoid duplicating code in BaseModel & FastCamelContext

jamesnetherton:
I stripped out most of the content from `FastCamelContext` except for CQ specific things.

I wasn't sure about methods like `getComponentParameterJsonSchema` so I left them as-is. Our impl is mostly the same as the one in `DefaultCamelContext`, except that it does a registry lookup, which I'm assuming is some kind of optimisation?


oscerd (approved):

zbendhiba:
@jamesnetherton thanks for this PR !

Issue #2671, opened 2021-05-25
Intermittent failures in MllpIT 


ppalaga:
Seen here https://github.com/apache/camel-quarkus/pull/2665/checks?check_run_id=2665270714#step:9:6969
```
[INFO] Running org.apache.camel.quarkus.component.mllp.it.MllpIT
May 25, 2021 1:57:08 PM org.jboss.threads.Version <clinit>
INFO: JBoss Threads version 3.2.0.Final
======================================================================================
  native.image.path was not set, making a guess for the correct path of native image
  guessed path: /home/runner/work/camel-quarkus/camel-quarkus/integration-tests/mllp/target/camel-quarkus-integration-test-mllp-2.0.0-M1-SNAPSHOT-runner
======================================================================================
Executing [/home/runner/work/camel-quarkus/camel-quarkus/integration-tests/mllp/target/camel-quarkus-integration-test-mllp-2.0.0-M1-SNAPSHOT-runner, -Dquarkus.http.port=43777, -Dquarkus.http.ssl-port=46297, -Dtest.url=http://localhost:43777, -Dquarkus.log.file.path=/home/runner/work/camel-quarkus/camel-quarkus/integration-tests/mllp/target/quarkus.log, -Dquarkus.log.file.enable=true]
2021-05-25 13:57:10,589 INFO  [org.apa.cam.qua.cor.CamelBootstrapRecorder] (main) Bootstrap runtime: org.apache.camel.quarkus.main.CamelMainRuntime
2021-05-25 13:57:10,598 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main) Routes startup summary (total:3 started:3)
2021-05-25 13:57:10,598 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main)     Started route1 (mllp://localhost:43777)
2021-05-25 13:57:10,598 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main)     Started route2 (direct://validMessage)
2021-05-25 13:57:10,599 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main)     Started route3 (direct://invalidMessage)
2021-05-25 13:57:10,599 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main) Apache Camel 3.10.0 (camel-1) started in 5ms (build:0ms init:2ms start:3ms)
[...]

PR #2769, opened 2021-06-11
Avoid port clashes where it may already be reserved via the build-helper-maven-plugin

jamesnetherton:
Relates to findings in #2671.


oscerd (approved):

zbendhiba (approved):
Nice !!

jamesnetherton:
Just realised that I messed up the placeholders in the log message. I used JBoss logging style instead of SLF4J. I'll fix...

aldettinger (approved):
Good catch, I guess it might have already class some test resources.

PR #2775, opened 2021-06-14
Adapt Observability docs page to Quarkus 2.0.0

jamesnetherton:


jamesnetherton:
Companion PR for the examples is here https://github.com/apache/camel-quarkus-examples/pull/41.

oscerd (approved):

ppalaga (commented):
I wonder whether we should point out somewhere that the new URIs are since CQ 2.0 and Quarkus 2.0. and that they have been different with older versions. There is https://github.com/quarkusio/quarkus/wiki/Migration-Guide-2.0#legacy-redirect-for-non-application-endpoints-such-as-health mentioning `quarkus.http.redirect-to-non-application-root-path` but that one was removed from Quarkus main https://github.com/quarkusio/quarkus/commit/e94ac2b8a2c973b81a23498b288fba4f6df7e76a#diff-72a2dbd079ddb1a497d6fb409bee853c04a9d7710dce7abc34761faad0563aaeL52-L72

There is still [quarkus.http.non-application-root-path](url) property that can be used to get back to the old behavior by setting `quarkus.http.non-application-root-path = ${quarkus.http.root-path}`

jamesnetherton:
> I wonder whether we should point out somewhere that the new URIs are since CQ 2.0 and Quarkus 2.0

I was thinking maybe we do that in the release announcement blog post? Or should we start a doc page where we start gathering stuff like this together. Perhaps like Camel does with its [migration guide](https://camel.apache.org/manual/latest/camel-3x-upgrade-guide.html)?

ppalaga:
Yes, migration guide is a good idea, but regardless of that, I'd personally like to find a minimal amount of that info also in the docs. Something like

```
The observability paths are `/q/...` ... . 

Note that the `/q/...` prefix was added in xref:migration-quide/2.0.0.adoc[Camel Quarkus 2.0/Quarkus 2.0]
```

I think it would make the `current` version of the docs more valuable to end users.

jamesnetherton:
Added the requested additional notes and the starting point for some migration docs.

ppalaga:
Thanks!

PR #2818, opened 2021-06-21
Tested and documented the specification of custom beans in applicatio…

### LINKED ISSUE #2778: Test specifying custom beans in application.properties
I do not think we have a test for this already
https://camel.apache.org/components/latest/others/main.html#_specifying_custom_beans

aldettinger:
…n.properties 

fix #2778

### DIFF
--- a/docs/modules/ROOT/pages/reference/extensions/core.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/core.adoc
@@ -92,6 +92,18 @@ quarkus.native.resources.includes = mysimple.txt
 
 More information about selecting resources for inclusion in the native executable can be found at xref:user-guide/native-mode.adoc#embedding-resource-in-native-executable[Embedding resource in native executable].
 
+==== Configuring a custom bean via properties in native mode
+When specifying a custom bean via properties in native mode with configuration like `#class:*` or `#type:*`, it may be needed to register some classes for reflection manually.
+
+For instance, the custom bean definition below involves the use of reflection for bean instantiation and setter invocation:
+[source,properties]
+---
+camel.beans.customBeanWithSetterInjection = #class:org.example.PropertiesCustomBeanWithSetterInjection
+camel.beans.customBeanWithSetterInjection.counter = 123
+---
+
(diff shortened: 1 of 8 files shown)


oscerd (approved):

integration-test-groups/foundation/core/pom.xml
```diff
@@ -43,13 +43,12 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
-
-        <!-- test dependencies -->
         <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-junit5</artifactId>
-            <scope>test</scope>
         </dependency>
```
ppalaga, line 49: I see that having junit in application class path simplifies the code a bit by not having to pass complex objects over the wire, but I wonder whether the pros outweight the cons: it is not only junit, it is quarkus-junit5 with the whole set of its transitive deps, including quarkus-core-deployment, ASM, jANSI and other funny stuff. Letting the native compiler to analyze them may not only bring strange errors, but can also prolong the native compilation (because it is more code). Having said that, I'd vote for not keeping junit in test scope and using assertions in `src/test/java`. I know we have this pattern in several other test modules, and I even may have approved it in the past. Anyway, as long as there are better alternatives, we should consider preferring them. 

So what are the alternatives?

resteasy-jaxb/resteasy-jackson would solve the problem of passing complex objects over the wire, but I think we should avoid even this one here in the core test module, where we strive for a maximal methodological purity (read: minimal dependency set).

What else?

How about adding an ad hoc `toString()` to `PropertiesCustomBeanWithConstructorParameterInjection` and other beans and returning it from the respective JAX-RS endpoint? I think it would serve the purpose quite well. It would allow for both having minimal app dependencies and moving the assertions to the test method. WDYT?

aldettinger: Yeah, keeping only JUnit 5 deps in compile may also be a candidate. I know we have already some test serializing object to JSON but then there is this case where quarkus-jackson register those object for reflection and can falsify test result.
In this case, let me see if I can come up with something like toString()/fromString() method.

aldettinger: so, I've updated the bean with the toString() so poms are not changed.

ppalaga (approved):
Perfect, thanks a lot, @aldettinger!

ppalaga:
Could you please have a look at the conflicts, @aldettinger ?

Issue #2651, opened 2021-05-24
Class loader issues in AvroRpcTest with Quarkus 2.0.0.Alpha3


ppalaga:
A variant of https://github.com/apache/camel-quarkus/issues/2514 

`quarkus.test.flat-class-path=true` is a workaround

ppalaga:
Related: https://github.com/quarkusio/quarkus/issues/17463

JiriOndrusek:
Problem is caused by issue in jetty. Wrong classloader is set in jetty. Issue was reported as https://github.com/eclipse/jetty.project/issues/6399

ppalaga:
I wonder what we can do for 2.0.0 that will happen before Camel 3.11? Could you @JiriOndrusek perhaps check the feasibility of having the Undertow based `org.apache.avro.ipc.Server` impl. here in Camel Quarkus for now? We could do it properly once we have the necessary changes in Camel.

JiriOndrusek:
Change in Camel is proposed as https://issues.apache.org/jira/browse/CAMEL-16740 (PR is created - https://github.com/apache/camel/pull/5714)

@ppalaga I'll try to find a way of fixing it in meantime in camel-quarkus only, as you suggested.

PR #2845, opened 2021-06-24
Replaced jetty with undertow in avro-rpc

JiriOndrusek:
fixes  https://github.com/apache/camel-quarkus/issues/2651

Adds support of new SPI - https://github.com/apache/camel/pull/5714


oscerd (approved):

jamesnetherton:
While this no doubt works fine, I wonder if starting & managing our own Undertow server at runtime is an optimal solution.

Can we not leverage the existing `quarkus-undertow` APIs like `ServletBuildItem` etc, and register the Servlet + any other required bits at build time? Then let Quarkus manage the server in a more optimised way.

I'm happy to be convinced otherwise if others disagree, but managing our own HTTP server feels wrong to me.

Maybe we can have this as an interim solution and follow up with improvements.

JiriOndrusek:
@jamesnetherton your proposal sounds right, I'll investigate this approach - whether it would work without a problem

JiriOndrusek:
I'm starting POC about build time approach.

Idea is to ask user for `quarkus.http.port`, let quarkus start a server with our servlet and then do not start http server in the component and let everything on the default one.

Current solution (in this PR) can still be used in meantime.

ppalaga:
> Idea is to ask user for quarkus.http.port, 

Not 100% sure what you mean, but regarding host and port I think we should do the same as we did e.g. in gRPC , i.e. ignore settings done in the component and rely solely on Quarkus config. https://camel.apache.org/camel-quarkus/latest/reference/extensions/grpc.html#_additional_camel_quarkus_configuration

JiriOndrusek:
Superseded by https://github.com/apache/camel-quarkus/pull/2859

PR #2859, opened 2021-06-29
Class loader issues in AvroRpcTest with Quarkus 2.0.0.Alpha3 #2651

JiriOndrusek:
fixes  https://github.com/apache/camel-quarkus/issues/2651
supersedes https://github.com/apache/camel-quarkus/pull/2845

Adds support of new SPI - https://github.com/apache/camel/pull/5714

This enhancement uses vertx for the http server. Servers are started for required ports ad are managed via quarkus.
Tests are still using jetty http server - but only in scope test.

Unfortunatelly there is no public access to these methods (which have to be used in http server) - [HttpTranscier#readBuffers](https://github.com/justinsb/avro/blob/master/src/java/org/apache/avro/ipc/HttpTransceiver.java#L73), [HttpTransciever#writeBuffers](https://github.com/justinsb/avro/blob/master/src/java/org/apache/avro/ipc/HttpTransceiver.java#L94) and [HttpTransciever#writeLength](https://github.com/justinsb/avro/blob/master/src/java/org/apache/avro/ipc/HttpTransceiver.java#L104). For that reason, vertx request handler has to leverage [RespondServlet](https://github.com/justinsb/avro/blob/master/src/java/org/apache/avro/ipc/ResponderServlet.java), which requires creation of very simple `HttpServletRequest` and `HttpServletResponse`, which basically wraps `byte[]`.
It makes sense to ask whether it is possible to make these methods public (or for example through creations of class called VertxResponder). If yes, `VertxHttpServer.java ` could be simplified - but that would be covered by follow-up issue.

@ppalaga It seems, that we don't need to use any blocking feature of vertx here, camel component waits for the response of the server. But there is a small issue here. Vertx logs warnings if thread is blocked for more than 2000 ms. I'm not sure whether this could affect users. (but it doesn't affect the functionality) WDYT?


ppalaga:
@JiriOndrusek is there any reason why this PR is against `camel-main`? Wouldn't `main` be fine too?

extensions/avro-rpc/deployment/src/main/java/org/apache/camel/quarkus/component/avro/rpc/deployment/AvroRpcProcessor.java
```diff
@@ -37,7 +41,8 @@ FeatureBuildItem feature() {
     }
 
     @BuildStep
-    ReflectiveClassBuildItem registerForReflection(CombinedIndexBuildItem combinedIndex) {
+    ReflectiveClassBuildItem registerForReflection(CombinedIndexBuildItem combinedIndex,
+            BuildProducer<ReflectiveClassBuildItem> reflectiveClassProducer) {
```
ppalaga, line 45: reflectiveClassProducer does not seem to be used

extensions/avro-rpc/runtime/src/main/java/org/apache/camel/quarkus/component/avro/rpc/spi/HttpServletResponseWithBytes.java
```diff
@@ -0,0 +1,252 @@
...
+    @Override
+    public void addCookie(Cookie cookie) {
+
+    }
+
+    // --------------------- following methods are not implemented and not used ----------------------------
+
+    @Override
+    public boolean containsHeader(String s) {
+        return false;
```
ppalaga, line 56: Maybe better throw an UnsupportedOperationException here and below?

extensions/avro-rpc/runtime/src/main/java/org/apache/camel/quarkus/component/avro/rpc/spi/VertxHttpServer.java
```diff
@@ -0,0 +1,73 @@
...
+    public void start() {
+        Vertx vertx = Arc.container().instance(Vertx.class).get();
+
+        vertx.createHttpServer().requestHandler(rc -> {
+            rc.bodyHandler(handler -> {
+                HttpServletRequestFromBytes req = new HttpServletRequestFromBytes(handler.getBytes(), rc.method().name());
+                HttpServletResponseWithBytes resp = new HttpServletResponseWithBytes();
+
+                try {
+                    new ResponderServlet(responder).service(req, resp);
```
ppalaga, line 53: org.apache.avro.ipc.ResponderServlet should be stateless, so one instance per VertxHttpServer should be enough, no?

JiriOndrusek:
@ppalaga  yes, you are right. I forgot that the current main already uses camel 3.11. I rebased PR.

ppalaga (commented):
This looks quite elegant!

> Tests are still using jetty http server - but only in scope test.

I cannot find the scope change for the jetty artifact. Is it there?

> Unfortunatelly there is no public access to these methods (which have to be used in http server) - [HttpTranscier#readBuffers](https://github.com/apache/avro/blob/master/lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpTransceiver.java#L97), [HttpTransciever#writeBuffers](https://github.com/apache/avro/blob/master/lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpTransceiver.java#L69) and [HttpTransciever#writeLength](https://github.com/apache/avro/blob/master/lang/java/ipc/src/main/java/org/apache/avro/ipc/HttpTransceiver.java#L126). For that reason, vertx request handler has to leverage [RespondServlet](https://github.com/apache/avro/blob/master/lang/java/ipc/src/main/java/org/apache/avro/ipc/ResponderServlet.java), which requires creation of very simple `HttpServletRequest` and `HttpServletResponse`, which basically wraps `byte[]`.
> It makes sense to ask whether it is possible to make these methods public (or for example through creations of class called VertxResponder). If yes, `VertxHttpServer.java ` could be simplified - but that would be covered by follow-up issue.

+1 this PR is OK for now, we can ask Avro to change the visibility and improve our code once they do. Actually when asking for some changes there, an avro-ipc artifact without the Servlet API dependency would be nice.

> @ppalaga It seems, that we don't need to use any blocking feature of vertx here, camel component waits for the response of the server. But there is a small issue here. Vertx logs warnings if thread is blocked for more than 2000 ms.

My understanding is that ResponderServlet.service() entails the execution of the whole Camel route. We never know whether the code in the route is or is not going to block. Hence we have to handle it as blocking. @davsclaus may want to correct me if I am wrong.

extensions/avro-rpc/runtime/src/main/java/org/apache/camel/quarkus/component/avro/rpc/spi/HttpServletResponseWithBytes.java
JiriOndrusek: Yes this would be better.

extensions/avro-rpc/deployment/src/main/java/org/apache/camel/quarkus/component/avro/rpc/deployment/AvroRpcProcessor.java
JiriOndrusek: I thought I removed it, so I'll do it again.

extensions/avro-rpc/runtime/src/main/java/org/apache/camel/quarkus/component/avro/rpc/spi/VertxHttpServer.java
JiriOndrusek: Makes sense, I'll do it.

integration-tests/avro-rpc/pom.xml
```diff
@@ -66,6 +66,11 @@
             <artifactId>camel-quarkus-integration-test-support</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro-ipc-jetty</artifactId>
+            <scope>test</scope>
+        </dependency>
```
JiriOndrusek, line 73: @ppalaga Here is a jetty dependency for tests scope used by tests.

jamesnetherton:
> We never know whether the code in the route is or is not going to block

Yeah - we had to deal with the same problem in the `platform-http` consumer. So we process the request on a separate thread with `vertx.executeBlocking`.

extensions/avro-rpc/deployment/src/main/java/org/apache/camel/quarkus/component/avro/rpc/deployment/AvroRpcProcessor.java
JiriOndrusek: done

extensions/avro-rpc/runtime/src/main/java/org/apache/camel/quarkus/component/avro/rpc/spi/VertxHttpServer.java
JiriOndrusek: done

extensions/avro-rpc/runtime/src/main/java/org/apache/camel/quarkus/component/avro/rpc/spi/HttpServletResponseWithBytes.java
JiriOndrusek: done

extensions/avro-rpc/runtime/src/main/java/org/apache/camel/quarkus/component/avro/rpc/spi/VertxHttpServer.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    public void start() {
+        Vertx vertx = Arc.container().instance(Vertx.class).get();
+
+        vertx.createHttpServer().requestHandler(rc -> {
+            rc.bodyHandler(handler -> {
+                HttpServletRequestFromBytes req = new HttpServletRequestFromBytes(handler.getBytes(), rc.method().name());
+                HttpServletResponseWithBytes resp = new HttpServletResponseWithBytes();
+                vertx.executeBlocking(
```
JiriOndrusek, line 51: @ppalaga  I changed implementation to be blocking.

JiriOndrusek:
Follow-up issue https://github.com/apache/camel-quarkus/issues/2861
(in case that avro brings a better API for this case - https://issues.apache.org/jira/browse/AVRO-3166)

integration-tests/avro-rpc/pom.xml
ppalaga: I am blind, sorry.

ppalaga (approved):
Looks good to me. @jamesnetherton would you like to throw a final look?

I have filed https://github.com/apache/camel-quarkus/issues/2862 for the OpenStack failure

jamesnetherton (approved):
Looks ok to me, so let's merge as-is.

I still think there's potentially some room to optimise things a bit, but I need to dig into the Quarkus Vert.x internals before creating any follow-up issues.

PR #2865, opened 2021-07-01
Fix usage of incorrect @Produces annotation

jamesnetherton:


oscerd (approved):

aldettinger:
Looks good to me. One question however, @Unremovable are useless because this is managed by camel-quarkus-core ?

jamesnetherton:
> Looks good to me. One question however, @Unremovable are useless because this is managed by camel-quarkus-core ?

AFAIK - if the bean has an explicit name, it's unremovable. There is some logic in core that I think is supposed to make component beans unremovable, but I'm not sure if it's working. Hence how I stumbled on this issue while experimenting.

aldettinger (approved):
ok, thanks for sharing.

ppalaga:
Oh, I see it was `@javax.ws.rs.Produces`. I guess `@javax.enterprise.inject.Produces` would be correct but superfluous in this situation?

PR #2879, opened 2021-07-05
kafka : impossible to authenticate with oauth2

### LINKED ISSUE #2878: camel-quarkus-kafka : Kafka oauth authentication issue with kafka-oauth-client 0.8.x versions
See description in [quarkus-18398](https://github.com/quarkusio/quarkus/issues/18398)
It would be cool if we can add an integration test for this use case, if it's possible to config testcontainers for oAuth2.

zbendhiba:
fixes #2878

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,7 +101,7 @@
         <netty3.version>${netty3-version}</netty3.version><!-- Spark -->
         <minio.version>${minio-version}</minio.version><!-- Keep in sync with quarkiverse-minio and Camel, preferring Camel assuming that it is more likely to be up to date -->
         <mvel2.version>${mvel-version}</mvel2.version>
-        <nimbus-jose-jwt.version>4.41.1</nimbus-jose-jwt.version><!-- Mess in hdfs transitive deps -->
+        <nimbus-jose-jwt.version>9.10</nimbus-jose-jwt.version><!-- It's placed here because of mess in hdfs transitive deps, please upgrade when it's possible -->
         <okhttp.version>${squareup-okhttp-version}</okhttp.version>
         <okio.version>${squareup-okio-version}</okio.version>
         <protobuf.version>${protobuf-version}</protobuf.version>


jamesnetherton (approved):

pom.xml
```diff
@@ -101,7 +101,7 @@
         <netty3.version>${netty3-version}</netty3.version><!-- Spark -->
         <minio.version>${minio-version}</minio.version><!-- Keep in sync with quarkiverse-minio and Camel, preferring Camel assuming that it is more likely to be up to date -->
         <mvel2.version>${mvel-version}</mvel2.version>
-        <nimbus-jose-jwt.version>4.41.1</nimbus-jose-jwt.version><!-- Mess in hdfs transitive deps -->
+        <nimbus-jose-jwt.version>9.10</nimbus-jose-jwt.version><!-- It's placed here because of mess in hdfs transitive deps, please upgrade when it's possible -->
```
ppalaga, line 104: If this is impacting Kafka, it would be nice to mention Kafka in the comment.

If the version 9.10 comes from some Kafka pom, then we should check whether it is possible to use `@sync` to update it automatically.

zbendhiba: Okay. It's not a kafka client pom, but usage of Strimzi, with oAuth2. The dependency is added for this purpose only : https://quarkus.io/guides/kafka#authenticating-with-oauth
 
That bug will maybe appear too on the Camel-K side, since they're using the Strimzi-oauth-client too.
When upgrading the client from 0.7.x to 0.8.x, the Strimzi oAuth client moves from Keyclock to Nimbus jose. so this configuration fails to start.

```
camel.component.kafka.security-protocol = SASL_SSL
camel.component.kafka.sasl-mechanism = OAUTHBEARER
camel.component.kafka.sasl-jaas-config = org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
        oauth.client.id="<YOUR-SERVICE-ACCOUNT-ID-HERE>" \
       oauth.client.secret="<YOUR-SERVICE-ACCOUNT-SECRET-HERE>" \
       oauth.token.endpoint.uri="<TOKEN_ENDPOINT_URI>" ;
camel.component.kafka.additional-properties[sasl.login.callback.handler.class] = io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler
```
And when I checked that version we have is way too old. I was wondering if we can't just push this one just for hdfs purpose ?

zbendhiba: To make sure it works, we need to have an integration tests with testcontainers + Strimzi that has oAuth2, I don't know if it's easy to setup

ppalaga: I thought `io.strimzi:kafka-oauth-client` would be managed in Quarkus BOM, but it is not. It it was managed there, we could use our tooling to keep our `nimbus-jose-jwt.version` in sync with theirs https://github.com/strimzi/strimzi-kafka-oauth/blob/main/pom.xml#L90 
I wonder how much sense would it make for quarkus to manage `io.strimzi:kafka-oauth-client`. WDYT?

ppalaga: In any case the comment should say something like `mess in hdfs transitive deps, used also by io.strimzi:kafka-oauth-client recommended by https://quarkus.io/guides/kafka#authenticating-with-oauth`

zbendhiba:
@ppalaga  fix here https://github.com/apache/camel-quarkus/pull/2896
FYI, the oauth Strimzi client version is not related to kafka client versions

PR #2881, opened 2021-07-06
Leverage catalog metadata to discover unremovable types

jamesnetherton:


extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/catalog/BuildTimeCamelCatalog.java
```diff
@@ -0,0 +1,86 @@
...
+                .map(this::componentModel)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+
+        // For components we need to combine the component options with the endpoint options
+        componentModels.stream()
+                .flatMap(componentModel -> componentModel.getComponentOptions().stream())
+                .forEach(options::add);
+
+        componentModels.stream()
```
aldettinger, line 62: this statement may be a duplicate ?

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/catalog/SchemaResource.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(type, name, className);
+    }
+
+    @Override
+    public String toString() {
+        return "SchemaInfo{" + "type='" + type + '\'' + ", name='" + name + '\'' + ", packageName='" + className + '\'' + '}';
```
aldettinger, line 107: neat pick :) But maybe we could say "SchemaResource" instead of "SchemaInfo"

jamesnetherton: > neat pick :) But maybe we could say "SchemaResource" instead of "SchemaInfo" ?

Yeah, I renamed that class quite a few times :grinning:. I'll fix...

aldettinger (approved):
Really nice, it will save a lot of headache :)
I was thinking about corner cases like FileLanguage extends SimpleLanguage and I think we are fine in this respect. Also, we still have some UnremovableBeanBuildItem in camel quarkus and I think they are needed.

ppalaga:
Nice work, thanks @jamesnetherton !

Issue #2755, opened 2021-06-09
Refactor registration of serialization once support is present in Quarkus


JiriOndrusek:
Once following change is part of quarkus (https://github.com/quarkusio/quarkus/pull/15380), usage of serialization configuration should be revisited and replaced with buildItems.

Currently there are 2 extensions using it: `nitrite` and `sql`.

Consider making automatic registration of basic types and of `DefaultExchangeHolder` with its hierarchy. 

(Here is a branch with a try of such refactor without buildItems - https://github.com/JiriOndrusek/camel-quarkus/tree/sql-aggregator-core-changes-not-wanted)

Issue #2255, opened 2021-02-16
Nitrite - follow up - simplify serialization configuration


JiriOndrusek:
Current implementation contains GraalVM native configuration file for serialization.

Once there is a better support for serialization on quarkus side - see https://github.com/quarkusio/quarkus/issues/14530,
nitrite extension should be refactored.

PR #2904, opened 2021-07-15
Use of serialization feature of Quakus (includes Sql and Nitrite)

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/2755
fixes https://github.com/apache/camel-quarkus/issues/2255

~In his change I propose registration for serialization of several basic classes. But because this is not required by the most of the extensions, I suppose that  creating a buildItem and based registration on it, would be better. @ppalaga @jamesnetherton WDYT?~
~Another problem is `BigInteger`, which doesn't work - I have to find why.~


jamesnetherton:
> I suppose that creating a buildItem and based registration on it

Yeah, that's how I would have implemented it. Otherwise every extension that wants serialization support has to duplicate the required steps.

I guess the registration of the basic Java types ideally belongs in Quarkus itself.

JiriOndrusek:
I agree that basic stuff would be in quarkus - in the future. (I plan to create a PR there)
But some stuff like `org.apache.camel.support.DefaultExchangeHolder` would stay here.

lburgazzoli:
oh that's nice.

btw we should really add a task to revisit DefaultExchangeHolder as it has been designed when serialization was cool but now it should be better to use json/avro/protobof

zbendhiba:
> btw we should really add a task to revisit DefaultExchangeHolder as it has been designed when serialization was cool but now it should be better to use json/avro/protobof

+1

jamesnetherton:
@JiriOndrusek please rebase on `main` as Quarkus has been bumped to 2.1.0.CR1.

JiriOndrusek:
I'll rebase
and here is an issue about `DefaultExchangeHolder` - https://issues.apache.org/jira/browse/CAMEL-16805

docs/modules/ROOT/pages/reference/extensions/core.adoc
```diff
@@ -104,6 +104,27 @@ camel.beans.customBeanWithSetterInjection.counter = 123
...
+* `java.lang.Long`
+* `java.lang.Number`
+* `java.lang.String`
+* `java.math.BigInteger`
+* `java.util.Date`
+* `java.util.HashMap`
+* `java.util.LinkedHashMap`
+* `org.apache.camel.support.DefaultExchangeHolder`
+
+Developers can register other classes for serialization via the `ReflectiveClassBuildItem`.
```
ppalaga, line 126: `ReflectiveClassBuildItem` is for extension developers and this is rather a part of user guide. I think we should better omit this sentence or mention some way how application devs can register additional classes for serialization. Don't we have a property for that? (Or I wonder whether `@RegisterForReflection` shouldn't be modify to allow that too?)

Wouldn't [native-mode.adoc](https://camel.apache.org/camel-quarkus/latest/user-guide/native-mode.html) be a better place for this whole section about Serialization in native mode?

docs/modules/ROOT/pages/reference/extensions/core.adoc
```diff
@@ -224,6 +245,13 @@ A comma separated list of Ant-path style patterns to match class names that shou
 | `string`
 | 
 
+|icon:lock[title=Fixed at build time] [[quarkus.camel.native.reflection.serialization-enabled]]`link:#quarkus.camel.native.reflection.serialization-enabled[quarkus.camel.native.reflection.serialization-enabled]`
+
+If `true`, basic classes are registered for serialization; otherwise Java serialization will not work in native mode.
```
ppalaga, line 250: It would be nice to list the classes here as well. 

> otherwise Java serialization will not work in native mode. 

Isn't that a bit of an overstatement? It would work, but only for the registered classes, no?

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/spi/CamelSerializationBuildItem.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.core.deployment.spi;
+
+import io.quarkus.builder.item.MultiBuildItem;
+
+/**
+ * A build item that does not carry any data but it is used to signal that registration for serialization is requested.
```
ppalaga, line 22:
```suggestion
 * Signal that basic classes listed in {@code org.apache.camel.quarkus.core.deployment.CamelSerializationProcessor.baseSerializationClasses() should be registered for serialization.
```

integration-test-groups/foundation/core/src/main/java/org/apache/camel/quarkus/core/MySerializationObject.java
```diff
@@ -0,0 +1,76 @@
...
+                _string != null &&
+                _float != null &&
+                _double != null &&
+                _bigInteger != null &&
+                _hashMap != null &&
+                _linkedHashMap != null &&
+                _char != null &&
+                _boolean != null &&
+                _byte != null;
+    }
```
ppalaga, line 74: I'd vote for being more paranoid here - we should check whether the fields have the same values as set by `initValues()`

ppalaga (approved):
Nice work, thanks @JiriOndrusek! Some comments and questions inline.

docs/modules/ROOT/pages/reference/extensions/core.adoc
JiriOndrusek: yes, you are right . Native-mode.adoc seems to be better place to write about it and I'll try to omit the buildItem and add link to the native registration via properties.

docs/modules/ROOT/pages/reference/extensions/core.adoc
JiriOndrusek: Add listing classes - I thought that link to the list would be enough, but I can list them here as well.
Yes it is a overstatement, I'll change it.

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/spi/CamelSerializationBuildItem.java
JiriOndrusek: I'll do it.

integration-test-groups/foundation/core/src/main/java/org/apache/camel/quarkus/core/MySerializationObject.java
JiriOndrusek: We can be more paranoid here. My idea was, that if serialization ends without an error, it works. (this should be probably covered somewhere in GraalVM tests). But no problem with being more strict and check the real values.

jamesnetherton:
Is there any work remaining on this one? It'd be good to get it merged ASAP as there's a few other things that depend on it.

JiriOndrusek:
Please leave me about an hour to add changes pointed by @ppalaga

docs/modules/ROOT/pages/reference/extensions/core.adoc
JiriOndrusek: done

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/spi/CamelSerializationBuildItem.java
JiriOndrusek: done

docs/modules/ROOT/pages/reference/extensions/core.adoc
JiriOndrusek: done

integration-test-groups/foundation/core/src/main/java/org/apache/camel/quarkus/core/MySerializationObject.java
JiriOndrusek: done

JiriOndrusek:
@jamesnetherton  All items are fixed. There is no more remaining work on this PR.

docs/modules/ROOT/pages/user-guide/native-mode.adoc
```diff
@@ -111,3 +111,28 @@ in {@code application.properties} - e.g.
...
+* `java.lang.Long`
+* `java.lang.Number`
+* `java.lang.String`
+* `java.math.BigInteger`
+* `java.util.Date`
+* `java.util.HashMap`
+* `java.util.LinkedHashMap`
+* `org.apache.camel.support.DefaultExchangeHolder`
+
+Developers can configure classes for serialization via the serialization configuration file
```
jamesnetherton, line 135: We can address this in a follow up issue / PR but we shouldn't we mention `@RegisterForReflection(serialization = true)` as the preferred way of registering user classes for serialization? `-H:SerializationConfigurationResources` is kind of a low level option which should be avoided if possible.

Maybe this has been discussed elsewhere already but I wonder if we should introduce a config property that lets you configure arbitrary classes for serialization like we have for reflection?

jamesnetherton (approved):

docs/modules/ROOT/pages/user-guide/native-mode.adoc
JiriOndrusek: Oh I understand one previous comment wrongly, I'll reword it with @RegisterForReflection(serialization = true)

JiriOndrusek: done

Issue #2721, opened 2021-06-04
camel-quarkus-minio: "quarkus.minio.url" is mandatory and must be a valid url


lburgazzoli:
Create from https://camel.zulipchat.com/#narrow/stream/257299-camel-k/topic/Minio.20Source.20Kamlet.20to.20Knative.20Sequence

Running the following kamelet binding using camel-k:

```yaml
apiVersion: camel.apache.org/v1alpha1
kind: KameletBinding
metadata:
  name: minio-to-tester-binding
  namespace: hello-eventing
spec:
  source:
    ref:
      kind: Kamelet
      apiVersion: camel.apache.org/v1alpha1
      name: minio-source
    properties:
      accessKey: "dev-access-key"
      bucketName: "trigger"
      endpoint: "http://minio.minio.svc.cluster.local:9000"
      secretKey: "dev-secret-key"
  sink:
    uri:  http://tester-sequence-kn-sequence-0-kn-channel.hello-eventing.svc.cluster.local
```

Leads to:

```
io.quarkiverse.minio.client.MinioProducer_ProducerMethod_produceMinioClient_a74f1d65447af9f0051b3615caa3aca9cac0ca25_ClientProxy.arc$delegate(MinioProducer_ProducerMethod_produceMinioClient_a74f1d65447af9f0051b3615caa3aca9cac0ca25_ClientProxy.zig:67)
    at io.quarkiverse.minio.client.MinioProducer_ProducerMethod_produceMinioClient_a74f1d65447af9f0051b3615caa3aca9cac0ca25_ClientProxy.bucketExists(MinioProducer_ProducerMethod_produceMinioClient_a74f1d65447af9f0051b3615caa3aca9cac0ca25_ClientProxy.zig:3843)
    at org.apache.camel.component.minio.MinioEndpoint.bucketExists(MinioEndpoint.java:175)
    at org.apache.camel.component.minio.MinioEndpoint.doStart(MinioEndpoint.java:95)
    at org.apache.camel.support.service.BaseService.start(BaseService.java:115)
    at org.apache.camel.support.service.ServiceHelper.startService(ServiceHelper.java:113)
    at org.apache.camel.impl.engine.RouteService.doWarmUp(RouteService.java:186)
    at org.apache.camel.impl.engine.RouteService.warmUp(RouteService.java:121)
    ... 28 more
"quarkus.minio.url" is mandatory and must be a valid url
```

jamesnetherton:
[...]

PR #2911, opened 2021-07-19
camel-quarkus-minio: quarkus.minio.url is not mandatory

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/2721


oscerd (changes requested):
This will break any dependant project using quarkus extension as runtime. If I set the minio url at camel level in a kamelet I don't have to set the minio url as quarkus property.

JiriOndrusek:
@oscerd I'm not sure if I understand. My fix is only setting property ''quarkus.minio.allow-empty" to true, which has no effect on cases, where minio client is initialized from quarkus properties:

- if all other quarkus minio parameters are null, no minio client will be created (without it, it would throw a validation exception)

From this perspective,  everything that worked before, would work after this fix. 
Only difference is in the case, where all quarkus parameters are null (url, server_secret and server_access) - case of kamelet - it wouldn't throw a validation exception, but return `null` as a client. Autowiring capability of the component then would use minio client from the context.

JiriOndrusek:
Better explanation of `quarkus.minio.allow-empty` is - do not throw a validation exception but return `null` as a minio client (so it won't throw an exception during autowiring, if all parameters are `null`)

oscerd:
The title was a bit misleading.. it looks like quarkus.minio.url is always mandatory, which should be wrong.

JiriOndrusek:
Sorry, I didn't see this, I'll change title.

ppalaga (approved):

PR #2912, opened 2021-07-19
Add information about BOMs and precedence of them

mmuzikar:
More info is given to the users reading the docs regarding what's in the different BOMs and how the precedence of BOM might affect the users.

### DIFF
--- a/docs/modules/ROOT/pages/user-guide/dependency-management.adoc
+++ b/docs/modules/ROOT/pages/user-guide/dependency-management.adoc
@@ -5,7 +5,7 @@ A specific Camel Quarkus release is supposed to work only with a specific Quarku
 == Quarkus BOM 
 
 The core Quarkus dependencies are managed via `io.quarkus:quarkus-bom`. 
-Any additional dependencies for related Quarkus artifacts are not contained in this BOM. 
+This BOM doesn't manage the additional Quarkus extensions (Camel Quarkus, Kubernetes, ...), so there are other BOMs available for you to use in your projects that contain import `quarkus-bom`.
 
 == Quarkus Universe BOM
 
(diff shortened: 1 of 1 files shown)


oscerd (approved):

docs/modules/ROOT/pages/user-guide/dependency-management.adoc
```diff
@@ -2,6 +2,11 @@
 
 A specific Camel Quarkus release is supposed to work only with a specific Quarkus release.
 
+== Quarkus BOM 
+
+The core Quarkus dependencies are managed via `io.quarkus:quarkus-bom`. 
+Any additional dependencies for related Quarkus artifacts are not contained in this BOM.
```
ppalaga, line 8: Sorry, I cannot follow what you mean with "additional dependencies for related Quarkus artifacts". Could you please try to rephrase it?

docs/modules/ROOT/pages/user-guide/dependency-management.adoc
```diff
@@ -81,6 +86,9 @@ When combining `camel-quarkus-bom` or `quarkus-universe-bom` with any other BOM,
 think carefully in which order you import them,
 because the order of imports defines the precedence.
 
+ie. if `quarkus-universe-bom` is imported before `camel-quarkus-bom` then the version defined in
+`quarkus-universe-bom` will take priority. And might lead to incompatible versions of libraries being used.
```
ppalaga, line 90: I think the `quarkus-universe-bom` and `camel-quarkus-bom` combo is not a good example. I wonder why would anybody want to do that given that `camel-quarkus-bom` is included in `quarkus-universe-bom`? 

How about this:

```suggestion
I.e. if `my-foo-bom` is imported before `camel-quarkus-bom` then the versions defined in
`my-foo-bom` will take the precedence. This might or might not be what you want, depending on whether there are any overlaps between `my-foo-bom` and `camel-quarkus-bom` and depending on whether those versions with higher precedence work with the rest of the artifacts managed in `camel-quarkus-bom`.
```

ppalaga (approved):

ppalaga:
Thanks @mmuzikar!

Issue #2005, opened 2020-11-11
Investigate replacing bouncycastle support extension with Quarkus bouncycastle 


jamesnetherton:
Quarkus now has a bouncycastle extension. Maybe we can use it, remove our own extension and contribute anything that's missing if required.

ppalaga:
Also related to bouncy castle: https://github.com/apache/camel-quarkus/issues/2673

sberyozkin:
> Quarkus now has a bouncycastle extension

Not yet - right now Quarkus only indirectly facilitates the creation of BC providers in its `quarkus-security` extension; perhaps we can indeed introduce a dedicated extension

PR #2915, opened 2021-07-20
Laveraging quarkus-security

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/2005

Quarkus can only  register boucyCastleProvider in current state. So I used it instead of our code and removed duplicate code (which exists in quarkus).

### DIFF
--- a/extensions-support/bouncycastle/deployment/pom.xml
+++ b/extensions-support/bouncycastle/deployment/pom.xml
@@ -39,6 +39,10 @@
             <artifactId>quarkus-arc-deployment</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-security-deployment</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-support-bouncycastle</artifactId>
         </dependency>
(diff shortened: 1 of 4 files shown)


ppalaga (approved):
Looks good, thanks @JiriOndrusek!

@JiriOndrusek is there any potential to move some of the stuff from here to quarkus-security?

oscerd (approved):

JiriOndrusek:
@ppalaga I'm thinking about moving some registrations for reflection into quarkus-security, because some similar stuff is already there (see https://github.com/quarkusio/quarkus/blob/main/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java#L149)
But I'm not sure whether it will be accepted. If some of that code makes it into qaurkus, it should be removed from here...

ppalaga:
> @ppalaga I'm thinking about moving some registrations for reflection into quarkus-security, because some similar stuff is already there (see https://github.com/quarkusio/quarkus/blob/main/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java#L149)
> But I'm not sure whether it will be accepted. If some of that code makes it into qaurkus, it should be removed from here...

+1, could you please file an issue on our side so that we don't forget about that?

JiriOndrusek:
Issue is created as https://github.com/apache/camel-quarkus/issues/2922

PR #2919, opened 2021-07-20
Support JMS ObjectMessage

jamesnetherton:


oscerd (approved):

jamesnetherton:
Having dug into the serialization support code a bit more. I think my docs are a bit misleading. In this case, there's actually no need for users to explicitly enable it via the config property because the extension(s) produce `CamelSerializationBuildItem`. I'll amend the wording...

jamesnetherton:
Latest commit fixes the wording of the documentation to something that more accurately describes the serialization support.

jamesnetherton:
Latest commit adds the `org.w3c.dom.Node` statement to SJMS & SJMS2 docs (I mistakenly thought those components did not support it).

ppalaga:
Perfect, thanks!

PR #2921, opened 2021-07-21
Salesforce - Mock existing integration tests

### LINKED ISSUE #2667: salesforce : add test coverage
- Investigate if it's possible to use Wiremock to mock APIs
- Add tests for BULK API 2.0
- Add tests for REST streaming API
- Add tests for Platform events
- Add more test coverafge for REST API
- Add more test coverage for REST Bulk API

zbendhiba:
Fixes #2667

This PR covers just one part of the 2667 gitHub issue, which is mocking the existing examples. There will be another PR for more test coverage.
 
The mocking scenario is the same used in some tests in the Camel project, but using wiremock. We can mock if using template, so the Route that starts in this project, is out of the scope, and is loaded only if salesforce credentials are set with environment variables.

### DIFF
--- a/integration-tests/salesforce/src/main/java/org/apache/camel/quarkus/component/salesforce/SalesforceRoutes.java
+++ b/integration-tests/salesforce/src/main/java/org/apache/camel/quarkus/component/salesforce/SalesforceRoutes.java
@@ -20,9 +20,7 @@ import java.util.Optional;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Named;
-import javax.ws.rs.Produces;
 
-import io.quarkus.arc.Unremovable;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.salesforce.AuthenticationType;
 import org.apache.camel.component.salesforce.SalesforceComponent;
--- a/integration-tests/salesforce/src/main/java/org/apache/camel/quarkus/component/salesforce/SalesforceRoutes.java
+++ b/integration-tests/salesforce/src/main/java/org/apache/camel/quarkus/component/salesforce/SalesforceRoutes.java
@@ -44,9 +44,6 @@ public class SalesforceRoutes extends RouteBuilder {
     @ConfigProperty(name = "SALESFORCE_CLIENTSECRET", defaultValue = "clientSecret")
     String clientSecret;
 
-    @Produces
-    @ApplicationScoped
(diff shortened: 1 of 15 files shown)


integration-tests/salesforce/src/main/java/org/apache/camel/quarkus/component/salesforce/SalesforceRoutes.java
```diff
@@ -16,13 +16,66 @@
...
+    @ConfigProperty(name = "SALESFORCE_CLIENTID", defaultValue = "clientId")
+    String clientId;
+
+    @ConfigProperty(name = "SALESFORCE_CLIENTSECRET", defaultValue = "clientSecret")
+    String clientSecret;
+
+    @Produces
+    @ApplicationScoped
+    @Unremovable
+    @Named("salesforce")
```
ppalaga, line 50:
```suggestion
    @Named("salesforce")
```

You probably want `@javax.enterprise.inject.Produces` instead of `@javax.ws.rs.Produces`. The fact that it works swhows, that both is redundant :). `@ApplicationScoped` is inherited from the class and `@Unremovable` is implicit for `@Named`

zbendhiba: @ppalaga good catch

ppalaga: All the fame belongs to @jamesnetherton from whom I learned all of this recently :)

ppalaga (commented):
Nice work, thanks @zbendhiba 
Could you please update the README adding the info, that the test is by default run against a mock and how to make sure that it is run against a real Salesforce much like we have it in other tests using Wiremock?

jamesnetherton (approved):

zbendhiba:
> Nice work, thanks @zbendhiba
> Could you please update the README adding the info, that the test is by default run against a mock and how to make sure that it is run against a real Salesforce much like we have it in other tests using Wiremock?

@ppalaga  thanks for catching this. I've just pushed a fix for the Readme file

ppalaga:
Perfect, thanks!

zbendhiba:
@ppalaga I can merge this one ?

ppalaga:
Oh, sorry, haven't I approved?

PR #2928, opened 2021-07-21
Use static issues to report nightly sync workflow failures

jamesnetherton:
I think this should be enough for the sync jobs to reuse a single issue for failure reporting. Lets see how things go over the next few days.


ppalaga (approved):

ppalaga:
Thanks @jamesnetherton !

Issue #2814, opened 2021-06-21
[Camel 3.11] Solr cloud integration tests are failing


jamesnetherton:
With Camel 3.11 tests that use the `solrCloud` endpoint scheme are failing. Seems there's an endless attempt trying to connect to ZooKeeper.

```
2021-06-21 10:41:58,525 INFO  [org.apa.zoo.ZooKeeper] (executor-thread-0) Initiating client connection, connectString=localhost:49466/solr/collection1 sessionTimeout=45000 watcher=org.apache.solr.common.cloud.SolrZkClient$ProcessWatchWithExecutor@1471977f
2021-06-21 10:41:58,525 INFO  [org.apa.zoo.ClientCnxnSocket] (executor-thread-0) jute.maxbuffer value is 1048575 Bytes
2021-06-21 10:41:58,526 INFO  [org.apa.zoo.ClientCnxn] (executor-thread-0) zookeeper.request.timeout value is 0. feature enabled=false
2021-06-21 10:41:58,526 INFO  [org.apa.sol.com.clo.ConnectionManager] (executor-thread-0) Waiting for client to connect to ZooKeeper
2021-06-21 10:41:58,526 INFO  [org.apa.zoo.ClientCnxn] (executor-thread-0-SendThread(localhost:49466)) Opening socket connection to server localhost/127.0.0.1:49466.
2021-06-21 10:41:58,527 INFO  [org.apa.zoo.ClientCnxn] (executor-thread-0-SendThread(localhost:49466)) SASL config status: Will not attempt to authenticate using SASL (unknown error)
2021-06-21 10:41:58,527 INFO  [org.apa.zoo.ClientCnxn] (executor-thread-0-SendThread(localhost:49466)) Socket connection established, initiating session, client: /127.0.0.1:40638, server: localhost/127.0.0.1:49466
2021-06-21 10:41:58,529 WARN  [org.apa.zoo.ClientCnxn] (executor-thread-0-SendThread(localhost:49466)) Session 0x0 for sever localhost/127.0.0.1:49466, Closing socket connection. Attempting reconnect except it is a SessionExpiredException.: java.io.IOException: Packet len 1213486160 is out of range!
	at org.apache.zookeeper.ClientCnxnSocket.readLength(ClientCnxnSocket.java:121)
	at org.apache.zookeeper.ClientCnxnSocketNIO.doIO(ClientCnxnSocketNIO.java:84)
[...]

Issue #2941, opened 2021-07-23
Solr: Investigate, why native tests fail in openjdk 11 (not in graalvm jdk)


JiriOndrusek:
Solr test executed in open jdk 11 fails in native mode:

Java version:
```
openjdk version "11.0.11" 2021-04-20
OpenJDK Runtime Environment 18.9 (build 11.0.11+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.11+9, mixed mode)

```

Error:

```
2021-07-23 14:49:54,961 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-0) HTTP Request to /solr/ssl/streaming failed, error id: bd19e18d-73b9-4af0-b644-77f92e4cdf4d-7: org.jboss.resteasy.spi.UnhandledException: org.apache.camel.CamelExecutionException: Exception occurred during execution on the exchange: Exchange[]
	at org.jboss.resteasy.core.ExceptionHandler.handleApplicationException(ExceptionHandler.java:106)
	at org.jboss.resteasy.core.ExceptionHandler.handleException(ExceptionHandler.java:372)
	at org.jboss.resteasy.core.SynchronousDispatcher.writeException(SynchronousDispatcher.java:218)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:519)
	at org.jboss.resteasy.core.SynchronousDispatcher.lambda$invoke$4(SynchronousDispatcher.java:261)
	at org.jboss.resteasy.core.SynchronousDispatcher.lambda$preprocess$0(SynchronousDispatcher.java:161)
	at org.jboss.resteasy.core.interception.jaxrs.PreMatchContainerRequestContext.filter(PreMatchContainerRequestContext.java:364)
	at org.jboss.resteasy.core.SynchronousDispatcher.preprocess(SynchronousDispatcher.java:164)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:247)
	at io.quarkus.resteasy.runtime.standalone.RequestDispatcher.service(RequestDispatcher.java:73)
	at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch(VertxRequestHandler.java:138)
	at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler$1.run(VertxRequestHandler.java:93)
[...]

PR #2942, opened 2021-07-23
[Camel 3.11] Solr cloud integration tests are failing #2814

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/2814

I used docker-compose approach for the cloud test resource. Previously used solr container (in case of cloud configuration), even when started with zookeeper, has a problem with accessing Zookeeper - problem mentioned [here](https://stackoverflow.com/questions/58001396/zookeeper-solr-in-testcontainers). Usage of docker compose allows defining Solr's and Zookeeper's cloud in a way, that everything works. Moreover testcontaioner's implementation of docker-compose creates automaically a simple container (`ambassador` - https://www.testcontainers.org/modules/docker_compose/), which works as a proxy to internal docker's network and exposes required ports to outside. It is possible to use random ports and (from my perspective) having cluster configuration in a yaml file makes it easily understandable.

Another issue reported as https://github.com/apache/camel-quarkus/issues/2941

### DIFF
--- a/extensions/solr/deployment/src/main/java/org/apache/camel/quarkus/component/solr/deployment/SolrProcessor.java
+++ b/extensions/solr/deployment/src/main/java/org/apache/camel/quarkus/component/solr/deployment/SolrProcessor.java
@@ -25,6 +25,7 @@ import io.quarkus.deployment.builditem.ExtensionSslNativeSupportBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.RuntimeInitializedClassBuildItem;
+import org.apache.zookeeper.ClientCnxnSocketNIO;
 import org.jboss.jandex.AnnotationTarget;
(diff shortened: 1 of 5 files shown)


oscerd (approved):

zbendhiba (approved):

extensions/solr/deployment/src/main/java/org/apache/camel/quarkus/component/solr/deployment/SolrProcessor.java
```diff
@@ -65,6 +66,8 @@ void registerForReflection(BuildProducer<ReflectiveClassBuildItem> reflectiveCla
                     }
                 })
                 .forEach(reflectiveClass::produce);
+
+        reflectiveClass.produce(new ReflectiveClassBuildItem(false, false, ClientCnxnSocketNIO.class.getName()));
```
ppalaga, line 70: We are registering the same class in HBaseProcessor. This is perhaps not enough for a separate support extension yet.

ppalaga (approved):
Nice, thanks @JiriOndrusek !

ppalaga (approved):

Issue #2860, opened 2021-06-30
AWS2 ddb-streams integration tests failures


mmuzikar:
The integration tests timeout on the ddb-streams part. Rarely 1 stream record of gets through but never the expected 3. 
Interesting fact is that the ddb tests pass just fine, but the streams always fail so the changes are definitely getting created but not observed by camel.

jamesnetherton:
I spent a little bit of time looking at this. I cannot replicate it locally. However, on an underpowered machine it seems to always fail.

One thing I observed was that if you `Thread.sleep()` at the beginning of the test, it will always pass. Which makes me wonder if some how the ddb-streams consumer is getting itself into a state where it never sees any of the events, possibly due to the [`lastSeenSequenceNumber`](https://github.com/apache/camel/blob/main/components/camel-aws/camel-aws2-ddb/src/main/java/org/apache/camel/component/aws2/ddbstream/Ddb2StreamConsumer.java#L43).

ppalaga:
Sounds like we could do some tricks to make the test pass. Not necessarily Thread.sleep() but some sort of condition that will make the producer requests wait until the consumer has started properly. But I am still wondering whether this is not an issue in the lastSeenSequenceNumber logic?

PR #2946, opened 2021-07-26
Do not merge ! Disabling tests except aws-ddb

JiriOndrusek:
do not merge, this is for testing purposes only. (to see result of build for aws-ddb)

Investigation of https://github.com/apache/camel-quarkus/issues/2860


lburgazzoli:
wonder if when a pr should not be merged, best would be to open a draft pr (afaik co triggers as well)

JiriOndrusek:
@lburgazzoli  I noticed, that with draft PRs, the build is not run completely (or I could be wrong about it)

lburgazzoli:
> @lburgazzoli I noticed, that with draft PRs, the build is not run completely (or I could be wrong about it)
> 

Oh, didn't know, I guess we should investigate if there are any difference between the type of PRs.
Btw, I don't have any objection on warning people with the PR description :) but a draft pr would be much more robust

JiriOndrusek:
My bad, it seems that CI starts with draft PRs, I'm keeping this PR as draft, because it is aligned with its purpose.

JiriOndrusek:
I saw the test results, closing PR.

PR #2951, opened 2021-07-27
AWS2 ddb-streams integration tests failures #2860

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/2860

Problem was caused by the fact that on slower machines, route containing consumer is started, but consumer isn't receiving (not started). Then it could easily happen, that some events are ignored because of that.

I added a periodical check, that some random (init event) are received. These init events have to be created periodically to be sure that consumer is running in the moment of event's creation.

### DIFF
--- a/integration-test-groups/aws2/aws2-ddb/src/test/java/org/apache/camel/quarkus/component/aws2/ddb/it/Aws2DdbTest.java
+++ b/integration-test-groups/aws2/aws2-ddb/src/test/java/org/apache/camel/quarkus/component/aws2/ddb/it/Aws2DdbTest.java
@@ -16,9 +16,11 @@
  */
 package org.apache.camel.quarkus.component.aws2.ddb.it;
 
+import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


oscerd (approved):

aldettinger (approved):
Looks good to me. Note that I was not able to replicate the issue locally, yet this potential fix makes sense IMHO :+1:

JiriOndrusek:
@aldettinger Issue can not be replicated manually or at least I wasn't able to (without tweaking some component code). I was testing it on a machine provided by the reporter. The same behavior could reproduced with the use of `Thread.sleep` in the aws2-ddb consumer.

JiriOndrusek:
Build error doesn't seem to be related to the code change: `Non-resolvable import POM: Could not find artifact org.apache.camel.quarkus:camel-quarkus-bom:pom:2.1.0-SNAPSHOT`

aldettinger:
I agree, the build issue must be related to post release steps. So let's merge.

Issue #2914, opened 2021-07-20
Provide a common set of JMS tests to test different combinations of JMS extension, client & broker


jamesnetherton:
It'd be good if we can have a common set of integration tests for JMS / AMQP and be able to reuse them to test different combinations of JMS client like Artemis, Qpid JMS & others. We already have a decent baseline for JMS testing in the `messaging` itest module so most of those should be reused.

The question is, what's the best way of achieving this?

Perhaps the cleanest way is to create new itest modules for each combination we want to test. The downside being more bloat added to the project + increased CI build times.

PR #2963, opened 2021-07-29
Introduce a build time optimized FastComponentNameResolver

jamesnetherton:


oscerd (approved):

davsclaus (approved):

ppalaga (commented):
Does this prevent end users from adding their own components via service loader? For some reason, I do not see any invocation of `ComponentNameResolver.resolveNames()` in Camel or Camel Quarkus source tree to figure out myself.

jamesnetherton:
> Does this prevent end users from adding their own components via service loader?

No, the components will still get registered. But I did realise something with the build time catalog that I wrote recently, in that doing discovery via these [annotations](https://github.com/apache/camel-quarkus/blob/main/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelUnremovableBeansProcessor.java#L47-L51) may not work with old style components because it's a Camel 3 construct. Hence my question on [Zulip](https://camel.zulipchat.com/#narrow/stream/257302-camel-quarkus/topic/CamelServiceBuildItem) about `CamelServiceBuildItem`.

davsclaus:
Yeah it looks like I added this API in 2020 for some use-case but we didn't need it anyway, so its not in use currently. However it can be used to find which components are available on the classpath in Camel if needed (maybe for logging, diagnose, jmx management etc.)

jamesnetherton:
> it can be used to find which components are available on the classpath in Camel if needed

I was thinking of using it for something I have planned as part of https://github.com/apache/camel-quarkus/issues/2914. But the `DefaultComponentNameResolver` doesn't actually work in native mode at the moment because the component service resources are not added to the native image (which circles back to my [Zulip](https://camel.zulipchat.com/#narrow/stream/257302-camel-quarkus/topic/CamelServiceBuildItem) question). And is partly why I created this PR.

ppalaga:
Thanks for the context, it is much clearer to me!

ppalaga (approved):

aldettinger (approved):

Issue #1433, opened 2020-06-30
[Serialization] Jolt: Remove the DeepCopySubstitution when/if graal supports Object serialization


aldettinger:
In #1421, the `DeepCopySubstitution` class was introduced to workaround [graal#460](https://github.com/oracle/graal/issues/460). Let's see what happen on the graal side, and remove the `DeepCopySubstitution` if possible.

aldettinger:
GraalVM 21.0.0 should include experimental support for objects serialization. Let's have a look when it is released.

aldettinger:
Quarkus 2.1 should include [serialization support](https://github.com/quarkusio/quarkus/issues/14530).

PR #2972, opened 2021-08-02
jolt: remove the DeepCopySubstitution in favor of Quarkus serializati…

aldettinger:
…on support #1433


extensions/jolt/deployment/src/main/java/org/apache/camel/quarkus/component/jolt/deployment/JoltProcessor.java
```diff
@@ -36,5 +36,15 @@ void registerReflectiveClasses(BuildProducer<ReflectiveClassBuildItem> producer)
...
+
+        // A JOLT Defaultr transformation spec is a JSON content and it needs to be serialized at some point.
+        // As such, we need to register all JSON base types and super types for serialization.
+        producer.produce(new ReflectiveClassBuildItem(false, false, false, true, java.lang.Boolean.class.getName()));
+        producer.produce(new ReflectiveClassBuildItem(false, false, false, true, java.lang.Number.class.getName()));
+        producer.produce(new ReflectiveClassBuildItem(false, false, false, true, java.lang.Double.class.getName()));
+        producer.produce(new ReflectiveClassBuildItem(false, false, false, true, java.lang.Integer.class.getName()));
+        producer.produce(new ReflectiveClassBuildItem(false, false, false, true, java.lang.String.class.getName()));
+        producer.produce(new ReflectiveClassBuildItem(false, false, false, true, java.util.AbstractList.class.getName()));
+        producer.produce(new ReflectiveClassBuildItem(false, false, false, true, java.util.ArrayList.class.getName()));
```
ppalaga, line 48: There is CamelSerializationBuildItem and https://github.com/apache/camel-quarkus/blob/main/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSerializationProcessor.java#L42-L71 that perhaps do 99% of what we need here?
As far as I can see only ArrayList is missing in CamelSerializationProcessor. Maybe we could add it there and save a couple of lines here?

aldettinger: Make sense to consider JSON types as common for camel. I've pushed an update in this direction.

ppalaga (approved):
Thanks @aldettinger!

PR #2982, opened 2021-08-03
Polish dependency management

ppalaga:


oscerd (approved):

zbendhiba (approved):

extensions-support/reactor-netty/runtime/pom.xml
```diff
@@ -62,6 +62,12 @@
         <dependency>
             <groupId>io.projectreactor.netty</groupId>
             <artifactId>reactor-netty</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>io.netty</groupId>
+                    <artifactId>netty-resolver-dns-native-macos</artifactId>
```
aldettinger, line 68: So, we are disabling netty dns resolvers on mac O/S ? Or maybe there is a kind of non native fallback dns ?

ppalaga: Yeah, my assumption is that this native resolver artifact does not help in GraalVM native mode anyway and that there is some plain java fallback. That's why I asked @ffang to check on his Mac.

aldettinger: ok, thanks for explanation :+1:

aldettinger (approved):

PR #2998, opened 2021-08-06
Salesforce : Expand producer test coverage

zbendhiba:

### DIFF
--- a/integration-tests/salesforce/src/test/java/org/apache/camel/quarkus/component/salesforce/SalesforceTest.java
+++ b/integration-tests/salesforce/src/test/java/org/apache/camel/quarkus/component/salesforce/SalesforceTest.java
@@ -156,35 +156,33 @@ class SalesforceTest {
                 .body()
                 .asString();
 
-        if (accountId != null) {
-            // get Account basic info
-            SObjectBasicInfo accountBasicInfo = RestAssured.given()
-                    .get("/salesforce/basic-info/account")
-                    .then()
-                    .statusCode(200)
-                    .extract()
-                    .body()
-                    .as(SObjectBasicInfo.class);
-            assertNotNull(accountBasicInfo);
-            List<RecentItem> recentItems = accountBasicInfo.getRecentItems();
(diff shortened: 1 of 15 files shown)


oscerd (approved):

integration-tests/salesforce/src/test/java/org/apache/camel/quarkus/component/salesforce/SalesforceTest.java
```diff
@@ -81,4 +102,119 @@ public void testBulkJobApi() {
...
+            RestAssured.get("/salesforce/account/" + accountId)
+                    .then()
+                    .statusCode(200)
+                    .body(
+                            "Id", not(emptyString()),
+                            "AccountNumber", not(emptyString()));
+
+            // delete the account
+            // Clean up
+            RestAssured.delete("/salesforce/account/" + accountId)
```
aldettinger, line 184: Would the delete not be invoked due to crash/exception, I think subsequent tests should be fine as the accountId would not be the same.

zbendhiba: yes, thanks @aldettinger. I'd fix that

zbendhiba: The latest commit addresses this unnecessary not null test on the accountId. Thanks for catching this

aldettinger: @zbendhiba I was more thinking about cases like the "get" before the "delete" that would throw a network exception. And, in this rare case, there should only be an account leak and subsequent tests should be fine.

I think, it should be fine. CI is only running mock tests and worst case one would need to cleanup the salesforce account from time to time on manual test.

aldettinger (approved):

ppalaga (approved):

PR #3000, opened 2021-08-09
Introduce a common set of extendable messaging tests

jamesnetherton:
Not sure if this is the best / most maintainable solution, but it provides a way to have a set of 'common' JMS extension tests that can be reused for ActiveMQ, AMQP, JMS, SJMS & SJMS2 + test different combinations of JMS broker & client.

The README in integration-tests/messaging explains how it works.


oscerd (approved):

zhfeng (approved):
Nice work !

tooling/scripts/validate-github-workflows.groovy
```diff
@@ -30,7 +30,7 @@ final Set<String> executedBaseNames = [] as Set
 
 // Add any ignored itest modules here. Or prefix the module name with '#' to disable it
 // hystrix - https://github.com/apache/camel-quarkus/issues/1146
-final List<String> excludedModules = ['hystrix', 'nats'] as List
+final List<String> excludedModules = ['hystrix', 'messaging', 'nats'] as List
```
ppalaga, line 33: Wouldn't integration-tests-support be a better place for the messaging module, given that it is not a real integration test? We could  avoid excluding it here and in tooling/test-list/pom.xml then. But you maybe have some reason to keep `messaging` under `integration-tests`?

ppalaga (approved):
Nice work, thanks @jamesnetherton. One little suggestion inline, that you may or may not want to accept. 

I wonder how well will these new test-jars behave inside platform. Have you tested it by any chance?

tooling/scripts/validate-github-workflows.groovy
ppalaga: BTW, do we still need to exclude nats here? I mean it is there in test-categories.yaml

jamesnetherton: > But you maybe have some reason to keep `messaging` under `integration-tests`?

Mainly to avoid having to navigate away from the main itests area if you want to add / amend tests. If it were just  few tests that we were maintainting, I'd probably have put it under test-support. Since its a catalog of tests that we'll likely add to in future, I thought i made sense to keep them in the standard location.

jamesnetherton:
> I wonder how well will these new test-jars behave inside platform. Have you tested it by any chance?

Yes, I did a basic platform test, the generated itest modules looked fine and I ran tests successfully.

ppalaga:
The activemq test passed for me locally. Let's merge

ppalaga:
... and congrats to PR #3000!

PR #3004, opened 2021-08-10
Avoid deprecated BuildProducer injection

jamesnetherton:


oscerd (approved):

aldettinger (approved):

ppalaga:
Is field injection really deprecated? Definitely a good step because the state of processor objects is not kept between build step invocations anyway. Any source of that info? Constructor param injection used to work too. I wonder whether that one or anything else is also deprecated.

aldettinger:
I've not found any release notes/blogs about it but there is a log message about deprecation when building such projects.

PR #3005, opened 2021-08-10
Salesforce: Expand Consumer integration tests : add tests for Streami…

zbendhiba:
…ng API

### DIFF
--- a/integration-tests/salesforce/src/test/java/org/apache/camel/quarkus/component/salesforce/SalesforceIntegrationTest.java
+++ b/integration-tests/salesforce/src/test/java/org/apache/camel/quarkus/component/salesforce/SalesforceIntegrationTest.java
@@ -23,7 +23,9 @@ import io.restassured.RestAssured;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
 
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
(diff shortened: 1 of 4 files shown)


integration-tests/salesforce/src/test/java/org/apache/camel/quarkus/component/salesforce/SalesforceIntegrationTest.java
```diff
@@ -33,16 +35,17 @@
 public class SalesforceIntegrationTest {
 
     @Test
-    public void testChangeDataCaptureEvents() {
+    public void testCDCAndStreamingEvents() {
```
aldettinger, line 38: Is there any requirements/simplifications to test CDC and streaming in the same place ?

zbendhiba: @aldettinger  There is no requirement. 
Just that all the consumers CDC, Streaming without RAW payload, and streaming with payload, will get the same created Account.
If we want to test those separately, I guess I need to order them or run in different classes.
Is this answering your question ?

aldettinger: Yes, I think I get it. The key point is that the Salesforce events mechanism is kind of shared by design. So I don't think there is a need to split, it looks clear now.

aldettinger (approved):

PR #3020, opened 2021-08-19
Test camel-xchange with kraken until binance issues are resolved

jamesnetherton:


oscerd (approved):

extensions/xchange/deployment/src/main/java/org/apache/camel/quarkus/component/xchange/deployment/XchangeProcessor.java
```diff
@@ -97,9 +97,9 @@ void registerForReflection(BuildProducer<ReflectiveClassBuildItem> reflectiveCla
         String[] dtoClasses = index.getKnownClasses()
                 .stream()
                 .map(classInfo -> classInfo.name().toString())
-                .filter(className -> className.startsWith("org.knowm.xchange.dto"))
+                .filter(className -> className.matches("^org\\.knowm\\.xchange.*dto.*"))
```
aldettinger, line 100: I was thinking about being more restrictive like:
.filter(className -> className.startsWith("org.knowm.xchange.dto"))
.filter(className -> className.startsWith("org.knowm.xchange.kraken.dto"))
But not sure it makes sense finally.

aldettinger (approved):

extensions/xchange/deployment/src/main/java/org/apache/camel/quarkus/component/xchange/deployment/XchangeProcessor.java
jamesnetherton: The problem is that users can choose from a multitude of different crypto exchanges and there is no real consistency among the xchange libs in terms of how the package names are defined. The only common element is that they start with `org.knowm.xchange` and have `dto` somewhere in the remainder of the name.

aldettinger: Indeed, we need some flexibility here :+1:

Issue #3025, opened 2021-08-23
kamelet: Test the kamelet location option


aldettinger:
Test the kamelet component query parameter named `location` that [has been added](https://github.com/apache/camel/commit/712a9a5e39a4ee51b4ab6ffde01b6f3bce79e8fd#diff-3ff692c0d442774944dd89f52d38dec27a65aa6ded4089c3ab6d361e1f1b87ee) . It should be possible to load a kamelet from an external file provided at runtime.

PR #3026, opened 2021-08-23
Test setting the kamelet from an external file provided at runtime #3025

aldettinger:

### DIFF
--- a/integration-tests/kamelet/pom.xml
+++ b/integration-tests/kamelet/pom.xml
@@ -60,6 +60,10 @@
             <artifactId>camel-quarkus-bean</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-xml-io-dsl</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-resteasy</artifactId>
         </dependency>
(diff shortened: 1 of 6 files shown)


integration-tests/kamelet/src/main/java/org/apache/camel/quarkus/component/kamelet/it/KameletRoutes.java
```diff
@@ -62,6 +62,9 @@ public void configure() throws Exception {
         from("direct:chain")
                 .to("kamelet:echo/1?prefix=Camel Quarkus&suffix=Chained")
                 .to("kamelet:echo/2?prefix=Hello&suffix=Route");
+
+        from("direct:kamelet-location-at-runtime")
+                .kamelet("upper?location=file:src/test/resources/upper-kamelet.xml");
```
jamesnetherton, line 67: Is that location going to work on `quarkus-platform`?

aldettinger: Yeah good catch, it should not work on platform. Maybe it is feasible to find another way, worst case the test may be removed.

aldettinger: The test is now using classpath, should be ok on quarkus-platform.

integration-tests/kamelet/pom.xml
```diff
@@ -59,6 +59,10 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-bean</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-xml-io-dsl</artifactId>
```
jamesnetherton, line 64: Can you add the `-deployment` dependency for this.

jamesnetherton (commented):
Found another small thing to fix. Otherwise LGTM.

integration-tests/kamelet/pom.xml
aldettinger: Fixed, a `process-resources -P format -N` from root folder was missing.

Issue #3028, opened 2021-08-24
Unsatisfying test coverage of binding mode in rest component


VratislavHais:
Rest tests does not cover binding mode in org.apache.camel.component.rest.RestProducer class. Goal of this issue is to increase the coverage as much as possible.

PR #3052, opened 2021-08-30
[#3028] Increase test coverage of a binding mode of camel-rest component

VratislavHais:
Resolves https://github.com/apache/camel-quarkus/issues/3028

### DIFF
--- a/integration-tests/rest/src/main/java/org/apache/camel/quarkus/component/rest/it/RestResource.java
+++ b/integration-tests/rest/src/main/java/org/apache/camel/quarkus/component/rest/it/RestResource.java
@@ -68,4 +68,26 @@ public class RestResource {
         return producerTemplate.requestBodyAndHeaders(
                 "rest:get:/rest/template/{messageStart}/{messageEnd}?host=localhost:" + port, null, headers, String.class);
     }
+
+    @Path("/producer/binding/mode/json")
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public Person restProducerBindingModeJson(@QueryParam("port") int port) {
+        String query = "rest:get:/rest/binding/json/producer" +
+                "?bindingMode=json" +
+                "&outType=org.apache.camel.quarkus.component.rest.it.Person" +
+                "&host=localhost:" + port;
+        return producerTemplate.requestBody(query, null, Person.class);
+    }
+
+    @Path("/producer/binding/mode/xml")
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public Person restProducerBindingModeXml(@QueryParam("port") int port) {
+        String query = "rest:get:/rest/binding/xml/producer" +
+                "?bindingMode=xml" +
+                "&outType=org.apache.camel.quarkus.component.rest.it.Person" +
+                "&host=localhost:" + port;
+        return producerTemplate.requestBody(query, null, Person.class);
+    }
 }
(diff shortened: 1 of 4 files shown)


oscerd (approved):

vkasala:
@ppalaga @jamesnetherton Can someone of you take a look?

integration-tests/rest/src/test/java/org/apache/camel/quarkus/component/rest/it/RestTest.java
```diff
@@ -25,9 +25,11 @@
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.matchesPattern;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 @QuarkusTest
 class RestTest {
+    private static final Person person = new Person("John", "Doe", 64);
```
aldettinger, line 32: Is it a requirement to use the same object for json/xml consumer/producer binding ? Otherwise, I would use distinct objects, and maybe with different value.

integration-tests/rest/src/main/java/org/apache/camel/quarkus/component/rest/it/RestResource.java
```diff
@@ -68,4 +68,26 @@ public String restProducer(@QueryParam("port") int port) {
...
     }
+
+    @Path("/producer/binding/mode/json")
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public Person restProducerBindingModeJson(@QueryParam("port") int port) {
+        String query = "rest:get:/rest/binding/json/producer" +
+                "?bindingMode=json" +
+                "&outType=org.apache.camel.quarkus.component.rest.it.Person" +
+                "&host=localhost:" + port;
```
ppalaga, line 79: Looking at this, I am am why we need a JAX-RS endpoint to test the rest component? This is perhaps a way to pass `bindingMode=json`, but it it can also be passed via REST DSL, which we already did before this PR, see https://github.com/apache/camel-quarkus/pull/3052/files#diff-58cbc36531a538cc13ef975a1889ac4ec3b28c54786b5896118916846558a592R64-R71 for JSON and https://github.com/apache/camel-quarkus/pull/3052/files#diff-58cbc36531a538cc13ef975a1889ac4ec3b28c54786b5896118916846558a592R78-R85 for XML. 
Is this PR adding anything new the above two exist?

integration-tests/rest/src/test/java/org/apache/camel/quarkus/component/rest/it/RestTest.java
VratislavHais: it's definitely not a requirement. I just thought it would be a little better to create the object only once. I'll change it as you suggest.

integration-tests/rest/src/main/java/org/apache/camel/quarkus/component/rest/it/RestResource.java
VratislavHais: Hello @ppalaga,
the above two test only Consumer while addition in this PR covers Producer. As you can see within this test coverage result https://fuse-next-jenkins-csb-fuse-qe.apps.ocp4.prod.psi.redhat.com/job/Integration.next/job/camel-quarkus/job/code-coverage/43/Camel_20JaCoCo_20report/org.apache.camel.component.rest/RestProducer.java.html#L272 
method `createBindingProcessor()` is almost entirely untested. With those changes the coverage increase greatly.

ppalaga: Thanks for explaining, @vhais!

PR #3064, opened 2021-09-02
[Website] fix attribute syntax

djencks:


ppalaga:
Why is this nececessary, given that it works as expected on the website? 
image

djencks:
Asciidoctor emits a warning about the missing attribute. With Antora 3 I plan to configure warnings to fail the build. Today, you can see the warnings if you look at the build output, although they don’t tell you what file the problem is in.

David Jencks

ppalaga:
Thanks for the explanation @djencks!

ppalaga (approved):

Issue #3053, opened 2021-08-30
Sql - enable test with different databases


JiriOndrusek:
Enable camel-quarkus sql tests to be run with a different databases. Support following db types: https://github.com/quarkusio/quarkus/tree/main/extensions/jdbc.

These tests can be executed only through parameters and with use of the external DB. No driver or test DB will be added into the tests.

PR #3066, opened 2021-09-03
Sql - enable test with different databases #3053

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/3053

Adds a possibility to execute tests against different db types. The full list of supported databases follows (supported types are defined via [jdbc extensions](https://github.com/quarkusio/quarkus/tree/main/extensions/jdbc).)

-  h2
- postgresql
- mssql
- db2
- mysql
- mariadb
- derby
- oracle

Default behavior is not changed and `h2` in dev-service mode is used.

If database can run in dev-service mode, it could be used. If environment variables are set, external db will be used. See more details in `README.adoc` file.

Test for stored procedure is disabled for different types of db except `H2` - this could be fixed in the future.


oscerd (approved):

JiriOndrusek:
It could be possible to execute tests against several db types by default (at least dev-service db types which don't require license). What do you think about it, should it be investigated @ppalaga , @jamesnetherton ?

integration-tests/sql/pom.xml
```diff
@@ -143,6 +139,125 @@
...
+            <activation>
+                <property>
+                    <name>SQL_JDBC_DB_KIND</name>
+                    <value>postgresql</value>
+                </property>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-jdbc-postgresql</artifactId>
```
aldettinger, line 168: I wonder if some thing like quarkus-jdbc-${SQL_JDBC_DB_KIND} could be used. It would be less lines in the pom file and maybe even fail when a wrong value is provided. I'm just curious, it's not blocking at all.

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlRoutes.java
```diff
@@ -114,7 +123,9 @@ private Path createTmpFileFrom(String file) throws IOException {
             while ((c = is.read()) >= 0) {
                 baos.write(c);
             }
-            fos.write(baos.toByteArray());
+            String content = new String(baos.toByteArray());
```
aldettinger, line 126: Is it possible to remove this replaceAll if we use distinct files for projectsViaClasspath and projectsViaFile ?
Otherwise, would a comment make it clear that we reuse the same file in order load the sql request from file or from classpath ?

integration-tests/sql/src/test/java/org/apache/camel/quarkus/component/sql/it/SqlTest.java
```diff
@@ -224,16 +228,20 @@ private ValidatableResponse postMapWithParam(String toUrl, String param, String
...
-            await().atMost(5, TimeUnit.SECONDS).until(() -> RestAssured
-                    .get("/sql/route/" + routeId + "/status")
-                    .then()
-                    .extract().asString(), equalTo(expectedOutput));
+    @SuppressWarnings("unchecked")
+    public static org.hamcrest.Matcher<java.util.Map<String, Object>> matchMapIgnoringCase(Map<String, Object> map) {
+        Matcher m = null;
+        for (Map.Entry<String, Object> entry : map.entrySet()) {
+            if (entry.getKey().toUpperCase().equals("PROCESSED")) {
+                //it is boolean type and different dbs return different representations of boolean
```
aldettinger, line 236: Couldn't we have a loose assertion checking that it maches "0" or "false" ?

jamesnetherton:
> It could be possible to execute tests against several db types by default (at least dev-service db types which don't require license). What do you think about it, should it be investigated @ppalaga , @jamesnetherton ?

Probably, because otherwise we have a bunch of profiles that are never tested on CI. So there's a risk they get stale, stop working and nobody notices.

Quarkus has `@TestProfile` which could perhaps be used for this use case (at the expense of some test execution speed). We don't have to do this for the initial impl, but we should look to clean things up a bit eventually.

aldettinger:
> It could be possible to execute tests against several db types by default (at least dev-service db types which don't require license). What do you think about it, should it be investigated @ppalaga , @jamesnetherton ?

I would say testing against each db types is up to quarkus. I think camel-sql is using jdbc, so nothing special per db types IMHO.

integration-tests/sql/README.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+- Postgresql (container) - add `-DSQL_JDBC_DB_KIND=postgresql`
+- MySQL (container) - add `-DSQL_JDBC_DB_KIND=mysql`
+- MariaDB (container) - add `-DSQL_JDBC_DB_KIND=mariadb`
+- H2 (in-process) used by default
+- Apache Derby (in-process) - add `-DSQL_JDBC_DB_KIND=derby`
+- DB2 (container) (requires license acceptance) - add `-DSQL_JDBC_DB_KIND=db2`
+- MSSQL (container) (requires license acceptance) - add `-DSQL_JDBC_DB_KIND=mssql`
+
+For more information about dev-service mode, see https://quarkus.io/guides/datasource#dev-services[documentation].
```
zbendhiba, line 24: I wasn't aware there were that much kind of databases now with dev services.
I guess we can run the same tests only with these kinds of databases. Am I right?

JiriOndrusek: Yes,only dev services databases, which doesn't require a licence file to be present.

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlHelper.java
```diff
@@ -0,0 +1,48 @@
...
+public class SqlHelper {
+
+    private static Set<String> BOOLEAN_AS_NUMBER = new HashSet<>() {
+        {
+            add("db2");
+            add("mssql");
+            add("oracle");
+
+        }
+    };
```
ppalaga, line 31: This is OK in a test, but it's good to know that double bracket initialization may have a performance impact due to introducing a new type https://www.youtube.com/watch?v=wgJWs14YcEs&t=1504s More about [ mono | bi | poly | mega ]morphic call sites https://www.youtube.com/watch?v=wgJWs14YcEs&t=483s
`new HashSet<>(Arrays.asList("db2", "mssql", "oracle"))` might be generally a better choice

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlRoutes.java
ppalaga, line 126:
```suggestion
            String content = new String(baos.toByteArray(), StandardCharsets.UTF_8);
```

You never know what the current machine's default charset is.

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlRoutes.java
```diff
@@ -114,7 +123,9 @@ private Path createTmpFileFrom(String file) throws IOException {
...
+            content = content.replaceAll("projectsViaClasspath", "projectsViaFile");
+            fos.write(content.getBytes());
```
ppalaga, line 128:
```suggestion
            fos.write(content.getBytes(StandardCharsets.UTF_8));
```

integration-tests/sql/src/test/java/org/apache/camel/quarkus/component/sql/it/SqlTest.java
```diff
@@ -67,6 +71,7 @@ public void testSqlComponent() {
     }
 
     @Test
+    @DisabledIfSystemProperty(named = "SQL_JDBC_DB_KIND", matches = "[^h][^2].*")
```
ppalaga, line 74: If you think this can be made to work for other DBs, please file an issue and paste the link here.

ppalaga (commented):
Very nice, @JiriOndrusek! Some nitpicks inline.

integration-tests/sql/pom.xml
JiriOndrusek: I started with this approach (similar to  quarkus-jdbc-${SQL_JDBC_DB_KIND} ) , I'll try to return to it

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlRoutes.java
JiriOndrusek: I'll add it there - this code was present before current change, so I didn't see a need of commenting it. But I agree that it would make code easier to understand

integration-tests/sql/src/test/java/org/apache/camel/quarkus/component/sql/it/SqlTest.java
JiriOndrusek: I agree, I chose this way, because the assertions look complex. I'll try to cover boolean with it and we will see, whether its complexity is acceptable.

integration-tests/sql/README.adoc
```diff
@@ -0,0 +1,45 @@
+== SQL integration tests
+
+=== Default database type
+
+When the tests are executed without any special configuration, dev-service `H2` database is used (more details will follow).
+
+=== Dev-service databases
+
+As is described  in the https://quarkus.io/guides/datasource#dev-services[documentation], several database types could be started in dev-service mode.
+Running the tests against a database in dev-service mode could be achieved by addition of build property `SQL_JDBC_DB_KIND`. Example of usage:
```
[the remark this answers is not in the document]
JiriOndrusek: I'll change that (I started with env var and forgot about it:) )

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlHelper.java
JiriOndrusek: Thanks for the information, I  didn't know that.

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlRoutes.java
JiriOndrusek: I'll fix it.

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlRoutes.java
JiriOndrusek: I'll fix it.

integration-tests/sql/src/test/java/org/apache/camel/quarkus/component/sql/it/SqlTest.java
JiriOndrusek: Makes sense, I'll fix it.

integration-tests/sql/pom.xml
JiriOndrusek: done

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlRoutes.java
JiriOndrusek: done

integration-tests/sql/src/test/java/org/apache/camel/quarkus/component/sql/it/SqlTest.java
JiriOndrusek: Looks better then I thought! Done

integration-tests/sql/README.adoc
JiriOndrusek: done

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlHelper.java
JiriOndrusek: done

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlRoutes.java
JiriOndrusek: done

integration-tests/sql/src/test/java/org/apache/camel/quarkus/component/sql/it/SqlTest.java
JiriOndrusek: done

integration-tests/sql/src/main/java/org/apache/camel/quarkus/component/sql/it/SqlRoutes.java
JiriOndrusek: done

JiriOndrusek:
@ppalaga , @aldettinger all requested changes are done.

aldettinger (approved):
Looks nicer and I've also seen the stored proc part follow up issue https://github.com/apache/camel-quarkus/issues/3080.

Well done for this PR @JiriOndrusek

ppalaga (approved):
@JiriOndrusek looks very good, thanks. Could you please file the followup for running all the DB kind tests on the CI?

JiriOndrusek:
Issue https://github.com/apache/camel-quarkus/issues/3083 created.

PR #3069, opened 2021-09-06
Fixed reference to Camel-Kamelets-catalog

oscerd:
I think the reference is wrong, at least at the moment, there is no camel-quarkus-kamelets-catalog.


zbendhiba (approved):

aldettinger (approved):
Need for sure. Please check #3070 in case the generated doc is missing.

PR #3073, opened 2021-09-07
Expand OpenApi Java test coverage

jamesnetherton:


integration-tests/openapi-java/src/test/java/org/apache/camel/quarkus/component/openapijava/it/common/OpenApiTest.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    @ParameterizedTest
+    @EnumSource(OpenApiContentType.class)
+    public void openApiEndpointSecurity(OpenApiContentType contentType) {
+        RestAssured.given()
+                .header("Accept", contentType.getMimeType())
+                .get("/openapi")
+                .then()
+                .contentType(ContentType.JSON)
```
aldettinger, line 80: We maybe need a clarification here.
So, we expect a response in YAML or JSON ? But the returned content type is always JSON ?

oscerd (approved):

integration-tests/openapi-java/src/test/java/org/apache/camel/quarkus/component/openapijava/it/common/OpenApiTest.java
jamesnetherton: The YAML response is intercepted & transformed to JSON so that we don't have to do custom verifcations on the response.

aldettinger: Ok, now I remind the YamlToJsonFilter from previous review. Thanks @jamesnetherton .

aldettinger (approved):

integration-tests/openapi-java/src/main/java/org/apache/camel/quarkus/component/openapijava/it/OpenApiRoutes.java
```diff
@@ -129,15 +144,34 @@ public void configure() throws Exception {
...
-
-                .get("/security/openid")
-                .route()
-                .setBody().constant("/security/openid")
-                .endRest()
-                .securityDefinitions()
-                .openIdConnect("openId", "https://secure.apache.org/fake/openid-configuration")
                 .end();
+
+        if (openApiVersion.equals("3.0.0")) {
```
ppalaga, line 149: Would it perhaps be worth to make this forward-compatible with versions newer than 3.0.0?

integration-tests/openapi-java/src/test/java/org/apache/camel/quarkus/component/openapijava/it/common/OpenApiTest.java
```diff
@@ -0,0 +1,224 @@
...
+                .statusCode(200)
+                .body(
+                        "description", containsInAnyOrder("Winter fruit", "Tropical fruit"),
+                        "name", containsInAnyOrder("Apple", "Pineapple"));
+    }
+
+    @ParameterizedTest
+    @EnumSource(OpenApiContentType.class)
```
ppalaga, line 58: Elegant!

integration-tests/openapi-java/src/main/java/org/apache/camel/quarkus/component/openapijava/it/OpenApiRoutes.java
jamesnetherton: TBH - the minor & patch parts of the version string are a bit of an irrelevance. There's some logic in camel to check the major version prefix and that's how it decides whether to produce output for OpenApi V2 or V3.

integration-tests/openapi-java/pom.xml
```diff
@@ -154,10 +171,28 @@
...
                                 <goals>
                                     <goal>integration-test</goal>
                                     <goal>verify</goal>
                                 </goals>
+                                <configuration>
+                                    <systemPropertyVariables>
+                                        <test.profile>openapi.v3</test.profile>
+                                    </systemPropertyVariables>
+                                </configuration>
                             </execution>
```
ppalaga, line 196: This will cost some work to make it work on the platform. I guess there is no way to have multiple OpenApi versions in a single app (so that we can we can avoid this Maven config)?

Is there no need to have the same for surefire?

ppalaga: > Is there no need to have the same for surefire?

That's done via `@TestProfile` that works only in JVM mode?

jamesnetherton: > > Is there no need to have the same for surefire?
> 
> That's done via `@TestProfile` that works only in JVM mode?

Correct, the additional bits added to the `pom.xml` were to support native mode. So shouldn't be an issue for platform as native is not tested there (AFAIK).

ppalaga: Their CI does not, but IMO, we should take care to have native runable anyway. I sometimes do run native tests there. OK, so it needs porting https://github.com/apache/camel-quarkus/issues/3075

ppalaga (approved):

Issue #2590, opened 2021-05-13
Test camel-quarkus-aws2-s3 together with quarkus-amazon-s3


ppalaga:
If the end user adds quarkus-amazon-s3 to an app together with camel-quarkus-aws2-s3 the app should work even if he has to configure them separately. We should add a test for this. There is some potential to clash also in case of other AWS services supported by Quarkus https://github.com/quarkusio/quarkus/tree/main/extensions/amazon-services Maybe in the end the simplest solution would be to make our extensions depend on those from Quarkus. In that way it would not be possible to use them separately and there would be no need for separate tests.

zhfeng:
@ppalaga I checked tests  with quarkus-amazon-s3 and basically they work since the ```S3Client``` could be wired automatically.
```
2021-09-11 00:25:34,837 INFO  [org.apa.cam.qua.cor.CamelBootstrapRecorder] (main) Bootstrap runtime: org.apache.camel.quarkus.main.CamelMainRuntime
2021-09-11 00:25:34,886 INFO  [org.apa.cam.mai.BaseMainSupport] (main) Auto-configuration summary
2021-09-11 00:25:34,887 INFO  [org.apa.cam.mai.BaseMainSupport] (main)     camel.component.aws2-s3.uriEndpointOverride=http://127.0.0.1:33272
2021-09-11 00:25:34,887 INFO  [org.apa.cam.mai.BaseMainSupport] (main)     camel.component.aws2-s3.region=us-east-1
2021-09-11 00:25:34,887 INFO  [org.apa.cam.mai.BaseMainSupport] (main)     camel.component.aws2-s3.accessKey=xxxxxx
2021-09-11 00:25:34,887 INFO  [org.apa.cam.mai.BaseMainSupport] (main)     camel.component.aws2-s3.secretKey=xxxxxx
2021-09-11 00:25:34,887 INFO  [org.apa.cam.mai.BaseMainSupport] (main)     camel.component.aws2-s3.overrideEndpoint=true
2021-09-11 00:25:34,909 INFO  [org.apa.cam.mai.MainAutowiredLifecycleStrategy] (main) Autowired property: amazonS3Client on component: aws2-s3 as exactly one instance of type: software.amazon.awssdk.services.s3.S3Client (io.quarkus.amazon.s3.runtime.S3ClientProducer_ProducerMethod_client_c3e44358d8bba829ee2f6ae63d6f9ef0b4609f3a_ClientProxy) found in the registry
[...]

PR #3077, opened 2021-09-08
Fix #2745 Expand AWS S3 test coverage

### LINKED ISSUE #2745: Expand AWS S3 test coverage
The following scenarios tested in Camel should be also tested in Camel Quarkus to make sure that they work also in native mode: 

 I did an investigation from QE POV:

* Via producer:
  * [Multipart upload (Camel)](https://github.com/apache/camel/blob/camel-3.9.0/components/camel-aws/camel-aws2-s3/src/test/java/org/apache/camel/component/aws2/s3/integration/S3MultipartUploadOperationIntegrationTest.java#L72)
  * [copyObject (Camel)](https://github.com/apache/camel/blob/camel-3.9.0/components/camel-aws/camel-aws2-s3/src/test/java/org/apache/camel/component/aws2/s3/integration/S3CopyObjectCustomerKeyOperationIntegrationTest.java#L95)
  * [listBuckets (Camel)](https://github.com/apache/camel/blob/camel-3.9.0/components/camel-aws/camel-aws2-s3/src/test/java/org/apache/camel/component/aws2/s3/integration/S3ListObjectsOperationIntegrationTest.java#L65)
  * [deleteBucket (Camel)](https://github.com/apache/camel/blob/camel-3.9.0/components/camel-aws/camel-aws2-s3/src/test/java/org/apache/camel/component/aws2/s3/integration/S3ListObjectsOperationIntegrationTest.java#L103)
  * [downloadLink (Camel)](https://github.com/apache/camel/blob/camel-3.9.0/components/camel-aws/camel-aws2-s3/src/test/java/org/apache/camel/component/aws2/s3/integration/S3CreateDownloadLinkOperationIntegrationTest.java#L78)
  * [getObjectRange (Camel)](https://github.com/apache/camel/blob/camel-3.9.0/components/camel-aws/camel-aws2-s3/src/test/java/org/apache/camel/component/aws2/s3/integration/S3ObjectRangeOperationIntegrationTest.java#L84)
  * [Using KMS to encrypt data (Camel - Localstack)](https://github.com/apache/camel/blob/camel-3.9.0/components/camel-aws/camel-aws2-s3/src/test/java/org/apache/camel/component/aws2/s3/localstack/S3SimpleEncryptedUploadOperationLocalstackTest.java)
* Via consumer:
  * As customer i would like to use the same AWS client for Camel and plain Quarkus (see <https://quarkus.io/guides/amazon-s3>) - so they are aligned and configured once

zhfeng:
* Multipart upload
* copyObject
* listBuckets
* deleteBucket
* downloadLink
* getObjectRange

### DIFF
--- a/integration-test-groups/aws2/aws2-s3/src/main/java/org/apache/camel/quarkus/component/aws2/Aws2S3Resource.java
+++ b/integration-test-groups/aws2/aws2-s3/src/main/java/org/apache/camel/quarkus/component/aws2/Aws2S3Resource.java
@@ -18,19 +18,16 @@ package org.apache.camel.quarkus.component.aws2;
 
 import java.io.BufferedReader;
 import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.net.URI;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
-import java.nio.file.StandardCopyOption;
+import java.nio.file.StandardOpenOption;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.UUID;
 import java.util.stream.Collectors;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -52,9 +49,6 @@ import org.apache.camel.ProducerTemplate;
 import org.apache.camel.component.aws2.s3.AWS2S3Constants;
 import org.apache.camel.component.aws2.s3.AWS2S3Operations;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
-import org.jboss.resteasy.annotations.providers.multipart.MultipartForm;
-import org.jboss.resteasy.plugins.providers.multipart.InputPart;
-import org.jboss.resteasy.plugins.providers.multipart.MultipartFormDataInput;
 import software.amazon.awssdk.core.ResponseInputStream;
 import software.amazon.awssdk.services.s3.model.Bucket;
 import software.amazon.awssdk.services.s3.model.GetObjectResponse;
@@ -135,32 +129,17 @@ public class Aws2S3Resource {
         return objects.stream().map(S3Object::key).collect(Collectors.toList());
     }
 
-    @Path("s3/upload")
+    @Path("s3/upload/{key}")
     @POST
-    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    @Consumes(MediaType.TEXT_PLAIN)
     @Produces(MediaType.TEXT_PLAIN)
-    public String multipartUpload(@MultipartForm MultipartFormDataInput input) throws Exception {
-
-        Map<String, List<InputPart>> formDataMap = input.getFormDataMap();
-        String fileName = formDataMap.get("fileName").get(0).getBodyAsString();
-
-        if (fileName == null) {
-            throw new FileNotFoundException();
-        }
-
+    public String upload(@PathParam("key") String key, String content) throws Exception {
         File file = File.createTempFile("aws2-", ".tmp");
+        Files.writeString(file.toPath(), content, StandardOpenOption.WRITE);
+        int partSize = 5 * 1024 * 1024;
 
-        List<InputPart> inputParts = formDataMap.get("file");
-        for (InputPart inputPart : inputParts) {
-            InputStream inputStream = inputPart.getBody(InputStream.class, null);
-            Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
-        }
-
-        input.close();
-
-        String key = fileName + UUID.randomUUID().toString();
         producerTemplate.sendBodyAndHeader(
-                componentUri() + "?multiPartUpload=true&autoCreateBucket=true",
+                componentUri() + "?multiPartUpload=true&partSize=" + partSize + "&autoCreateBucket=true",
                 file,
                 AWS2S3Constants.KEY,
                 key);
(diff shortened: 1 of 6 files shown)


zhfeng:
The KMS encypt test and some other consumer tests could leverage with the #2590. So I will get futher investigating.

oscerd (approved):

integration-test-groups/aws2/aws2-s3/src/main/java/org/apache/camel/quarkus/component/aws2/Aws2S3Resource.java
```diff
@@ -97,18 +123,152 @@ public Response read(@PathParam("key") String key) throws Exception {
...
+        String fileName = formDataMap.get("fileName").get(0).getBodyAsString();
+
+        if (fileName == null) {
+            throw new FileNotFoundException();
+        }
+
+        File file = File.createTempFile("aws2-", ".tmp");
+
+        List<InputPart> inputParts = formDataMap.get("file");
+        for (InputPart inputPart : inputParts) {
```
aldettinger, line 154: Ok, we take only the last part and the test contains a single part.

integration-test-groups/aws2/aws2-s3/src/main/java/org/apache/camel/quarkus/component/aws2/Aws2S3Resource.java
```diff
@@ -97,18 +123,152 @@ public Response read(@PathParam("key") String key) throws Exception {
...
+            InputStream inputStream = inputPart.getBody(InputStream.class, null);
+            Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
```
aldettinger, line 156: So we are upload only the second part ?

integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/Aws2S3Test.java
```diff
@@ -125,4 +129,156 @@ public void consumer() {
...
+    private void createObject(String oid, String blobContent) {
+        RestAssured.given()
+                .contentType(ContentType.TEXT)
+                .body(blobContent)
+                .post("/aws2/s3/object/" + oid)
+                .then()
+                .statusCode(201);
+    }
+
+    private String[] getAllObject(String bucket) {
```
aldettinger, line 260: Maybe getAllObjects for the method name ?

zhfeng: Thanks for pointing it out !

integration-test-groups/aws2/aws2-s3/src/main/java/org/apache/camel/quarkus/component/aws2/Aws2S3Resource.java
zhfeng: This test is migrated from the camel one ```S3MultipartUploadOperationIT.java``` which runs a test to upload a file. I just check the S3 MultipartUpload operation in camel-aws2-s3 component. It looks like that it splits a file content into pieces by partSize. Then upload them in S3 in one request. I think this mulitpart upload does not  like the HTTP_MULTIPART_FORM. Anyway, I'm not an expert of AWS S3 so it mghit be some misunderstanding here.

aldettinger: Not an expert too, trying to sort it out :) From the readings, please find a proposal below:
- The test send a simple string content and a key to this method (so we don't use @ MultipartForm)
- The string content is put in a `aws2-*.tmp` temp file
- We send this temp file and key with `multiPartUpload=true`
- Optionally, we try to make AWS2 library to chunk the content with low partSize for instance `partSize=1`

Would such a test makes sense ?

zhfeng: Thanks @aldettinger - it makes sense.

zhfeng: It looks like we can set ```partSize=1```` which throws
```
Caused by: software.amazon.awssdk.services.s3.model.S3Exception: Your proposed upload is smaller than the minimum allowed object size.
```

zhfeng: Acoording to https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html
```
Each part must be at least 5 MB in size, except the last part. There is no size limit on the last part of your multipart upload.
```

I think we should add this limit in the camel-aws2-s3 document.

zhfeng: And ```getObject``` after uploading, there is a ```chunk-signature``` fragment in the response result. I'm not sure that it could be handled by camel-aws2-s3 component or by users themself ? In current upload test, I just strip it.

aldettinger: Concerning this PR, I think uploading a small file without partSize should be enough. We cover createMultipartUploadRequest, createMultipartUpload, uploadPart and completeMultipartUpload.

Concerning follow up issues to improve camel-aw2-s3 component. Checking partSize >= 5MB is a good idea, ideally the 5MB value should come from amazon in case it changes. Concerning getObject/chunk-signature, I have no idea. Maybe opening camel ticket would allow to collect feedback.

aldettinger (approved):

aldettinger:
Let's merge when we have follow up/tracker issues.

zhfeng:
OK, I create https://github.com/apache/camel-quarkus/issues/3095 to track the large size upload issue.

Issue #2595, opened 2021-05-14
AWS2 Lambda integration timeouts on Localstack


ppalaga:
I am disabling `Aws2LambdaTest` and `Aws2LambdaIT`.

PR #3102, opened 2021-09-15
Fix lambda tests on localstack #2595

aldettinger:


oscerd (approved):

aldettinger:
AWS lambda tests were not working previously on localstack. Switching to python solves the issue as python seems to be present built-in in localstack container.

jamesnetherton (approved):

ppalaga (approved):
How, how simple and elegant!

zbendhiba (approved):
Nice !

aldettinger:
On CI, the lambda test is working in isolation but not grouped. I cannot reproduce this behavior locally, so I may need to push few investigation commits on this PR.

PR #3107, opened 2021-09-16
[WIP] Fix #2590 to test together with quarkus-amazon-s3

### LINKED ISSUE #2590: Test camel-quarkus-aws2-s3 together with quarkus-amazon-s3

zhfeng:
1. it needs to add ```quarkus.s3.sync-client.type=apache``` in application.properties because we are using ```software.amazon.awssdk:apache-client```.  And the ```software.amazon.awssdk:url-connection-client``` is the default one in Quarkus.
2. it needs to inject ```S3Client``` instance explicitly in test.


aldettinger:
So, more than testing that camel-quarkus-aws2-s3 could be used next to quarkus-amazon-s3, this PR is making camel-quarkus-aws2-s3 depends upon quarkus-amazon-s3 ?
In this way, will camel-quarkus-aws2-s3 users need to manually inject a S3Client and define relevant properties on their own ?

oscerd (changes requested):
To me the extension should not depend on quarkus Amazon S3. We constantly have trouble with quarkus extension on Camel-k. So where it is possible we should avoid using pure quarkus extensions

ppalaga:
> To me the extension should not depend on quarkus Amazon S3. We constantly have trouble with quarkus extension on Camel-k. So where it is possible we should avoid using pure quarkus extensions

Yep, -1 for now. We are discussing with Quarkus team how their extensions could change to allow us to configure the clients via  Camel configuration.

zhfeng:
Thanks all for reviewing and this PR is just for a test purpose to see if the quarkus amazon S3 client is working with the camel. And @ppalaga  is there any Quarkus issue for discussing this problem ?

oscerd:
If it is just for testing the dependency should be with scope test

aldettinger:
> If it is just for testing the dependency should be with scope test

Yes, and maybe the dependency should be on itests only, not extension.
And Maybe there is need to have 2 dedicated itest projects ? like:

1. The existing one testing only camel-quarkus-aws2-s3
2. Another one testing camel-quarkus-aws2-s3 used in the same project with quarkus-amazon-s3

zhfeng:
@aldettinger Is it similar with the camel infinspan itest ?
https://github.com/apache/camel-quarkus/blob/b856df58078471e027c7d8e7dbc586ba8d37b87f/integration-tests/infinispan/src/test/java/org/apache/camel/quarkus/component/infinispan/InfinispanServerTestResource.java#L58-L74

aldettinger:
@zhfeng I see the point, 2 tests might not be needed. I think the bottom line is to test an application using both quarkus-amazon-s3 and camel-quarkus-aws2-s3. A single itest project showing both configurations working ala infinispan might be ok.

aldettinger:
@jamesnetherton @ppalaga Does the autowiring approach make sense if documented like in this PR ? 

If yes, then I would aligned with the documentation proposed in https://github.com/apache/camel-quarkus/pull/3116

ppalaga:
I looked at #3116 and I expressed my thoughts there https://github.com/apache/camel-quarkus/pull/3116#pullrequestreview-760579842 
So far our extension does not depend on the Quarkus one, I think we need two test apps: one depending only on our extension and one depending on both ours and Quarkus extensions. The second one can perhaps be a light weight test leveraging `@org.junit.jupiter.api.extension.RegisterExtension`

aldettinger:
I think there is good ideas from #3133 that this PR could align on.

zhfeng:
OK, I will create a new PR to relpace this one.

jamesnetherton:
Superseeded by #3135.

PR #3110, opened 2021-09-16
use Antora 3.0.0-alpha.9 and fail build on warn to check xrefs

djencks:
This upgrades the partial site build to use Antora 3.0.0-alpha.9 to support the symlinks now used in the main build. Xref checking is now done by failing the build on warnings. Since missing includes are an error, the spring-boot includes now need to be satisfied in the partial build.


oscerd (approved):

ppalaga (approved):

ppalaga:
Thanks a lot, @djencks !

Issue #3114, opened 2021-09-17
Finish expanding of AWS DDB tests


JiriOndrusek:
Follows https://github.com/apache/camel-quarkus/issues/2776.

Document that feature `Batch consumer` is not implemented because of https://issues.apache.org/jira/browse/CAMEL-16837

Expand extension's tests to cover:

- Our DDB extension should be able to work when quarkus-amazon-dynamodb is in the class path.

JiriOndrusek:
Thanks @jamesnetherton for taking care of this issue.

PR #3116, opened 2021-09-17
Finish expanding of AWS DDB tests #3114

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/3114

### DIFF
--- a/extensions/aws2-ddb/runtime/src/doc/configuration.adoc
+++ b/extensions/aws2-ddb/runtime/src/doc/configuration.adoc
@@ -1,0 +1,50 @@
+The aws2-ddb extension is using a DynamoDB client under the hood.
+The underlying DynamoDB client could either be created by this extension itself or by the `quarkus-amazon-dynamobdb`.
+
+=== Client created by this extension
+
+In that case fill following properties:
+
+
+[source,properties]
+----
+camel.component.aws2-ddb.access-key=${AWS_ACCESS_KEY}
+camel.component.aws2-ddb.secret-key=${AWS_SECRET_KEY}
+camel.component.aws2-ddb.region=${AWS_REGION:us-east-1}
+----
+
+or
+
+[source,properties]
+----
+camel.component.aws2-ddbstream.access-key=${AWS_ACCESS_KEY}
+camel.component.aws2-ddbstream.secret-key=${AWS_SECRET_KEY}
+camel.component.aws2-ddbstream.region=${AWS_REGION:us-east-1}
+----
+
+=== Client created by `quarkus-amazon-dynamodb`
+
+To use client created by `quarkus-amazon-dynamodb` extension, follow the  https://quarkus.io/guides/amazon-dynamodb#configuring-dynamodb-clients[documentation] with some restrictions:
+
+- Apache client type has to be selected
+[source,properties]
+----
+quarkus.dynamodb.sync-client.type=apache
+----
+
+- All required dependencies are already part of the extension.
+There is no need to add them.
+
+- `DynamoDbClient` has to be injected somewhere in the application - this triggers creation of client via `quarkus-amazon-dynamodb`:
+
+[source,java]
+----
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
+
+@Inject
+DynamoDbClient dynamoDB;
+----
+
+
+
+
--- a/extensions/aws2-ddb/runtime/src/doc/limitations.adoc
+++ b/extensions/aws2-ddb/runtime/src/doc/limitations.adoc
@@ -1,0 +1,3 @@
+Feature `Batch consumer` (see https://camel.apache.org/components/latest/aws2-ddbstream-component.html#_batch_consumer[documentation])
+is not available in quarkus aws2-ddb extension because of the issue
+https://issues.apache.org/jira/browse/CAMEL-16837[CAMEL-16837].
--- a/integration-test-groups/aws2/aws2-ddb/src/main/resources/application.properties
+++ b/integration-test-groups/aws2/aws2-ddb/src/main/resources/application.properties
@@ -15,10 +15,13 @@
 ## limitations under the License.
 ## ---------------------------------------------------------------------------
 
-camel.component.aws2-ddb.access-key=${AWS_ACCESS_KEY}
-camel.component.aws2-ddb.secret-key=${AWS_SECRET_KEY}
-camel.component.aws2-ddb.region=${AWS_REGION:us-east-1}
+#client for aws2-ddb is created by quarkus extension io.quarkus:quarkus-amazon-dynamodb
+quarkus.dynamodb.aws.region=${AWS_REGION:us-east-1}
+quarkus.dynamodb.aws.credentials.type=static
+quarkus.dynamodb.aws.credentials.static-provider.access-key-id=${AWS_ACCESS_KEY}
+quarkus.dynamodb.aws.credentials.static-provider.secret-access-key=${AWS_SECRET_KEY}
+quarkus.dynamodb.sync-client.type=apache
 
-camel.component.aws2-ddbstream.access-key=${AWS_ACCESS_KEY}
-camel.component.aws2-ddbstream.secret-key=${AWS_SECRET_KEY}
-camel.component.aws2-ddbstream.region=${AWS_REGION:us-east-1}
+quarkus.dynamodb.endpoint-override=${AWS_CONTAINER_CREDENTIALS_FULL_URI:}
+
+#client for aws2-ddbstreams is created by camel component
--- a/integration-test-groups/aws2/aws2-ddb/pom.xml
+++ b/integration-test-groups/aws2/aws2-ddb/pom.xml
@@ -55,6 +55,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-resteasy-jackson</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-amazon-dynamodb</artifactId>
+        </dependency>
 
         <!-- test dependencies -->
         <dependency>
--- a/integration-test-groups/aws2/aws2-ddb/src/main/java/org/apache/camel/quarkus/component/aws2/ddb/it/Aws2DdbResource.java
+++ b/integration-test-groups/aws2/aws2-ddb/src/main/java/org/apache/camel/quarkus/component/aws2/ddb/it/Aws2DdbResource.java
@@ -46,6 +46,7 @@ import org.apache.camel.component.aws2.ddb.Ddb2Constants;
 import org.apache.camel.component.aws2.ddb.Ddb2Operations;
 import org.apache.camel.util.CollectionHelper;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
 import software.amazon.awssdk.services.dynamodb.model.AttributeAction;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValueUpdate;
@@ -73,6 +74,12 @@ public class Aws2DdbResource {
 
     @Inject
     ProducerTemplate producerTemplate;
+
+    /**
+     * Triggers creation of dynamoDb client via quarkus extension
+     */
+    @Inject
+    DynamoDbClient dynamoDB;
 
     @SuppressWarnings("serial")
     @Path("/item/{key}")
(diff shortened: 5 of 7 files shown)


oscerd (approved):

aldettinger:
CI is failing, looks like it is related to this PR.

JiriOndrusek:
@aldettinger yes, fails are related to this PR, I'll look into it

JiriOndrusek:
@aldettinger, @ppalaga  current failure is caused by doc plugin. I don't see any relation to my PR (but I'm not sure, because I'm adding some doc to the extension), do you know what is the real reason of the failure?

aldettinger:
> @aldettinger, @ppalaga current failure is caused by doc plugin. I don't see any relation to my PR (but I'm not sure, because I'm adding some doc to the extension), do you know what is the real reason of the failure ?

@JiriOndrusek I don't think it's related. It happened on other PRs as well.

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
```diff
@@ -0,0 +1,60 @@
+There are 2 ways of DynamoDB client creation:
```
aldettinger, line 1: What would think about the wording below ? If I get it right at all :)
The aws2-ddb extension is using a DynamoDB client under the hood. The underlying DynamoDB client could either be created by this extension itself or by the quarkus-amazon-dynamobdb.

JiriOndrusek: Nice:) this explains the reality much better, I'll use it

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+o use client created by `quarkus-amazon-dynamodb` extension, follow the  https://quarkus.io/guides/amazon-dynamodb#configuring-dynamodb-clients[documentation] with some restrictions:
+
+- Apache client type has to be selected
+[source,properties]
+----
+quarkus.dynamodb.sync-client.type=apache
+----
+
+- All required dependencies are already part of the extension. There is no need to add them.
```
aldettinger, line 34: If there is no need to add them, maybe we don't show the example below ?

JiriOndrusek: That makes sense, I'll remove it immediately.

JiriOndrusek: done

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
JiriOndrusek: done

aldettinger:
Comments help a lot @JiriOndrusek. From the context I have, I think the approach is ok but let's check with others and make sure this PR and https://github.com/apache/camel-quarkus/pull/3107 are aligned.

@jamesnetherton @ppalaga Does the autowiring approach make sense if documented like in this PR ?

aldettinger (approved):

integration-test-groups/aws2/aws2-ddb/src/main/resources/application.properties
```diff
@@ -15,10 +15,16 @@
...
+#client for aws2-ddb is created by quarkus extension io.quarkus:quarkus-amazon-dynamodb
+quarkus.dynamodb.aws.region=${AWS_REGION:us-east-1}
+quarkus.dynamodb.aws.credentials.type=static
+quarkus.dynamodb.aws.credentials.static-provider.access-key-id=${AWS_ACCESS_KEY}
+quarkus.dynamodb.aws.credentials.static-provider.secret-access-key=${AWS_SECRET_KEY}
+quarkus.dynamodb.sync-client.type=apache
 
+quarkus.dynamodb.endpoint-override=${AWS_CONTAINER_CREDENTIALS_FULL_URI:}
+
+#client for aws2-ddbstreams is created by camel component
```
zhfeng, line 27: if we want to create a client by camel componet, I think it needs to add ```camel.component.aws2-ddbstream.autowired-enabled=false```

zhfeng (changes requested):

integration-test-groups/aws2/aws2-ddb/src/main/resources/application.properties
JiriOndrusek: You are right, I missed the fact, that ddbstream test is autowiring the client created by quarkus extension/ I'll fix it.

JiriOndrusek: @zhfeng I investigated it and it is correct. Let me explain:

Quarkus extension is creating `DynamoDbClient`, which is autowired here: https://github.com/apache/camel/blob/main/components/camel-aws/camel-aws2-ddb/src/main/java/org/apache/camel/component/aws2/ddb/Ddb2Configuration.java#L39

But ddbstreams is autowring `DynamoDbStreamsClient` here: https://github.com/apache/camel/blob/main/components/camel-aws/camel-aws2-ddb/src/main/java/org/apache/camel/component/aws2/ddbstream/Ddb2StreamConfiguration.java#L44

Which is not autowired because stream client doesn't exist. Therefore client for stream tests is created by component (as is described in properties)

I'm doing one small code change though. There is no need to define client properties for ddbstream client in `application.properties`, because it is created in `integration-tests-support-aws` and values in application.properties doesn't matter.

JiriOndrusek: @zhfeng if you agree, please let me know

aldettinger: Ok, we have 2 components at play here and they have different types and they are autowired by types. Make sense to me.

extensions/aws2-ddb/runtime/src/doc/limitations.adoc
```diff
@@ -0,0 +1,3 @@
+Feature `Batch consumer` (see https://camel.apache.org/components/latest/aws2-ddbstream-component.html#_batch_consumer[documentation])
+is not available in quarkus aws2-ddb extension because of the issue
+https://issues.apache.org/jira/browse/CAMEL-16837[CAMEL-16837].
```
ppalaga, line 3:
```suggestion
The https://camel.apache.org/components/latest/aws2-ddbstream-component.html#_batch_consumer[batch consumer] feature is not available in Camel Quarkus AWS2 DynamoDB extension because of https://issues.apache.org/jira/browse/CAMEL-16837[CAMEL-16837].
```

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
```diff
@@ -0,0 +1,50 @@
+The aws2-ddb extension is using a DynamoDB client under the hood.
+The underlying DynamoDB client could either be created by this extension itself or by the `quarkus-amazon-dynamobdb`.
```
ppalaga, line 2:
```suggestion
The underlying DynamoDB client could either be created by this extension itself or by the `quarkus-amazon-dynamodb` from Quarkus project.
```

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
```diff
@@ -0,0 +1,50 @@
...
+or
+
+[source,properties]
+----
+camel.component.aws2-ddbstream.access-key=${AWS_ACCESS_KEY}
+camel.component.aws2-ddbstream.secret-key=${AWS_SECRET_KEY}
+camel.component.aws2-ddbstream.region=${AWS_REGION:us-east-1}
+----
+
+=== Client created by `quarkus-amazon-dynamodb`
```
ppalaga, line 25:
```suggestion
=== Client created by `quarkus-amazon-dynamodb` extension
```

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+To use client created by `quarkus-amazon-dynamodb` extension, follow the  https://quarkus.io/guides/amazon-dynamodb#configuring-dynamodb-clients[documentation] with some restrictions:
```
ppalaga, line 27:
```suggestion
To use the client created by `quarkus-amazon-dynamodb` extension, follow the  https://quarkus.io/guides/amazon-dynamodb#configuring-dynamodb-clients[documentation] with some restrictions:
```

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+- Apache client type has to be selected
```
ppalaga, line 29:
```suggestion
- Apache client type has to be selected
+
```

A single `+` sign on a line makes following code block belong to the previous list item

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
```diff
@@ -0,0 +1,50 @@
...
+[source,properties]
+----
+quarkus.dynamodb.sync-client.type=apache
+----
+
```
ppalaga, line 34:
```suggestion
+
```

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
```diff
@@ -0,0 +1,50 @@
...
+- All required dependencies are already part of the extension.
+There is no need to add them.
+
```
ppalaga, line 37:
```suggestion
```

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
```diff
@@ -0,0 +1,50 @@
...
+To use client created by `quarkus-amazon-dynamodb` extension, follow the  https://quarkus.io/guides/amazon-dynamodb#configuring-dynamodb-clients[documentation] with some restrictions:
+
+- Apache client type has to be selected
+[source,properties]
+----
+quarkus.dynamodb.sync-client.type=apache
+----
+
+- All required dependencies are already part of the extension.
+There is no need to add them.
```
ppalaga, line 36: It looks like the opposite is true, isn't it? 
```suggestion
- You need to add the `io.quarkus:quarkus-amazon-dynamodb` to your application.
```

ppalaga (commented):
Could you please explain the idea behind `copyEnvPropertyAs()` and setting only `quarkus.dynamodb.*` properties in `application.properties`?

So we test only the situation when both our extension and quarkus extension are in the app class path, but we still assume that creating the client both ways is hit? I think that even if this is true currently, it is very easy for this test to get flawed over time. I think as long as our extension does not depend on the Quarkus one, we need two test applications: 

1. one using only our extension (as it was before this PR)
2. one using both our and Quarkus extension

The test no. 2. could perhaps be a light weight `@org.junit.jupiter.api.extension.RegisterExtension` test in `camel-quarkus-aws2-ddb-deployment` module. The Quarkus extension would have to be added with `<scope>test</scope>` to the pom. In the end we only need to test that our extension picks the client from the CDI container. Of course the question is whether/how we can start localstack from a `@RegisterExtension` test. `@QuarkusTestResource` won't work there, so we perhaps can use testcontainers directly. 
If a `@RegisterExtension` test isn't be possible, I'd be for a separate integration test module having only a JVM test class.
WDYT?

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+- `DynamoDbClient` has to be injected somewhere in the application - this triggers creation of client via `quarkus-amazon-dynamodb`:
```
jamesnetherton, line 38: Wouldn't it be autowired? When camel does the lookup of `DynamoDbClient` against ARC, that should trigger creation of the bean (assuming the relevant config properties were provided).

jamesnetherton (approved):

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
JiriOndrusek: Unfortunatelly quarkus extension is looking for injection points, see https://github.com/quarkusio/quarkus/blob/main/extensions/amazon-services/common/deployment/src/main/java/io/quarkus/amazon/common/deployment/AbstractAmazonServiceProcessor.java#L64

integration-test-groups/aws2/aws2-ddb/src/main/resources/application.properties
zhfeng: Thanks @JiriOndrusek and it makes sense.

zhfeng (approved):

extensions/aws2-ddb/runtime/src/doc/configuration.adoc
zhfeng: So it there any room to improve in Quarkus ?  Add an opiton to force create the client ?

jamesnetherton:
Superseeded by #3133.

PR #3119, opened 2021-09-20
Fixes #3109 replicate more of the main camel website build: …

### LINKED ISSUE #3109: Docs xref check failures
Lots of xref check failures, probably due to things being reorganised on the Camel side.

To reproduce:

```
cd docs
mvn clean verify
```

djencks:
use @djencks/asciidoctor-jsonpath and include quarkus examples

The quarkus website component was already using the jsonpath extension, which requires the quarkus-examples component, but the use did not cause a failure until the extension was added to the partial build so that the relevant parts of the main build worked.

Including the extension in the partial build is a good idea since it's used in this component, but there's too much dependence on replicating the main build.  I'm looking for a better solution long term.

### DIFF
--- a/docs/.pnp.js
+++ b/docs/.pnp.js
@@ -40,6 +40,7 @@ function $$SETUP_STATE(hydrateRuntimeState, basePath) {
             ["@antora/cli", "npm:3.0.0-alpha.9"],
             ["@antora/site-generator-default", "npm:3.0.0-alpha.9"],
             ["@djencks/asciidoctor-antora-indexer", "npm:0.0.6"],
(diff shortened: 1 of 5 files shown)


oscerd (approved):

jamesnetherton:
The `license-maven-plugin` is a bit picky about the format of the license. It expects the `/*` comment style instead of `//`. If you run `mvn process-resources -Pformat -N` from the `docs` directory, it'll auto apply the license headers for you.

djencks:
Thanks! That fixed it.  I think this can be merged now but not all the ci tasks have run, although I think the website-related one(s) have.  Perhaps a cq expert should merge?

jamesnetherton (approved):

PR #3124, opened 2021-09-22
Improve Kafka test coverage

### LINKED ISSUE #2627: Kafka - add test coverage
Add test for these functionalities for camel-quarkus-kafka extension :
- Capability to use [manual commit with Kafka consumer](https://camel.apache.org/components/latest/kafka-component.html#_using_manual_commit_with_kafka_consumer)
- Customize [Kafka header propagation](https://camel.apache.org/components/latest/kafka-component.html#_kafka_headers_propagation)
- Usage of Strimzi container
- Serialization/Deserialization of keys,headers, etc.

zbendhiba:
Fixes #2627

### DIFF
--- a/integration-tests/kafka/src/main/java/org/apache/camel/quarkus/component/kafka/CamelKafkaRoutes.java
+++ b/integration-tests/kafka/src/main/java/org/apache/camel/quarkus/component/kafka/CamelKafkaRoutes.java
@@ -21,10 +21,21 @@ import javax.enterprise.inject.Produces;
 import javax.inject.Named;
 
 import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.kafka.KafkaConstants;
+import org.apache.camel.component.kafka.KafkaManualCommit;
 import org.apache.camel.processor.idempotent.kafka.KafkaIdempotentRepository;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 
 public class CamelKafkaRoutes extends RouteBuilder {
+
+    private final static String KAFKA_CONSUMER_MANUAL_COMMIT = "kafka:manual-commit-topic"
+            + "?groupId=group1&sessionTimeoutMs=30000&autoCommitEnable=false"
+            + "&allowManualCommit=true&autoOffsetReset=earliest";
+
+    private final static String SEDA_FOO = "seda:foo";
+    private final static String SEDA_SERIALIZER = "seda:serializer";
+    private final static String SEDA_HEADER_PROPAGATION = "seda:propagation";
+
     @ConfigProperty(name = "camel.component.kafka.brokers")
     String brokers;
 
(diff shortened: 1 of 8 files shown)


oscerd (approved):

integration-tests/kafka/src/main/java/org/apache/camel/quarkus/component/kafka/CamelKafkaRoutes.java
```diff
@@ -46,5 +65,33 @@ public void configure() throws Exception {
...
+        // Kafka consumer that use Manual commit and performs the manual commit
+        from(KAFKA_CONSUMER_MANUAL_COMMIT)
+                .routeId("foo")
+                .to(SEDA_FOO)
+                .process(e -> {
+                    KafkaManualCommit manual = e.getIn().getHeader(KafkaConstants.MANUAL_COMMIT, KafkaManualCommit.class);
+                    manual.commitSync();
+                });
+
+        // Kafka consumer that use Maunal commit and doesn't perform the manual commit
```
aldettinger, line 78: A small typo there ?

zbendhiba: oh yes !

integration-tests/kafka/src/test/java/org/apache/camel/quarkus/component/kafka/it/CamelKafkaTest.java
```diff
@@ -86,4 +89,127 @@ void testQuarkusKafkaClientFactoryNotConfigured() {
                 .statusCode(200)
                 .body(is("true"));
     }
+
+    @Test
```
aldettinger, line 93: Ok, I think I got it. My understanding is that the trick to start/stop foo/bar routes is just to have the message committed or not.
So, I wonder if it would be clearer with a single route that would call `manual.commitSync();` each 2 messages only ?
So the first message would be committed, the second not and so on....

zbendhiba: And the idea is to verify at some point if it's committed, or never ? I'm trying to understand how to replace the last startup

aldettinger: Thinking aloud:
We would have a kind of counter in the process() statement
The process statement would do something like that setBody(counter)
Send "content" to kafka, assert seda as seen counter = 1 and counter 2 (so delivered uncommitted and then delivered committed).
That way, we wouldn't need the last startup ? Would it cover everything needed thought ?

zbendhiba: Yeah it could work. Let me  try this.
I'm not either fan of starting and stopping Routes.

zbendhiba: I pushed a fix, to have only one single Route.
I still need to stop and restart this one to redeliver the message

aldettinger: I had the idea that it would work like jdbc, amq commits. But maybe in the context of kakfa it's something else. Maybe it's just an asynchronous commit that has the time to be executed with the %2 testing method. If any, does camel-kafka tests also stop/start routes ?

aldettinger: Ok, then I would say let's do this also.

aldettinger (approved):

aldettinger:
Well done @zbendhiba

ppalaga:
Ignoring the failing docs checks

PR #3125, opened 2021-09-23
Improve Kafka integration with Quarkus dev services

jamesnetherton:


oscerd (approved):

extensions/kafka/runtime/src/main/doc/usage.adoc
```diff
@@ -0,0 +1,9 @@
+=== Quarkus Kafka Dev Services
+
```
aldettinger, line 2: Do you think there would be advantages to have a single sentence per line ? Not sure how easy it would be for review.

aldettinger (approved):
Looks good to me. Just a question to open a debate :)

extensions/kafka/runtime/src/main/doc/usage.adoc
jamesnetherton: Yes, I can change it.

jamesnetherton:
I modifed things to use the sentence per line 'best practice'. IMO it's aesthetically awful and I'd prefer we abandoned it. But happy for this PR to be the guinea pig for it.

aldettinger:
I see the point, the source adoc becomes less aesthetic. I think I could get used to that on the long run if I'm convinced it help. And I understand it could help with merging conflicts. From a review perspective, I think we won't be able to catch but that's no blocker. 

I would say, let's merge as is not to block this PR and then maybe @zbendhiba @zhfeng @JiriOndrusek @ppalaga @ffang  and others would have comment.

zbendhiba:
> I modifed things to use the sentence per line 'best practice'. IMO it's aesthetically awful and I'd prefer we abandoned it. But happy for this PR to be the guinea pig for it.

Let's see how it looks like in the camel website.

PR #3144, opened 2021-09-30
Upgrade Camel to 3.12.0

jamesnetherton:
Note - 3.12.0 is still staged and not yet GA. So if we merge this before we cut 2.3.0, then some tricks will be needed to perform the release. @ppalaga should be able to provide more info on how that works.


oscerd (approved):

ppalaga (approved):

ppalaga:
> some tricks will be needed to perform the release.

The only trick I remember is to make sure before releasing that the staging repos are not present in the root pom https://github.com/apache/camel-quarkus/pull/3144/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R283-R339 and have those same repos in your local `~/.m2/settings.xml`. But I agree with Claus and James, that having Camel 3.12 is not critical at this time.

jamesnetherton:
The docs are broken (again), but I'm going to merge this and address that elsewhere.

Issue #3137, opened 2021-09-29
Enable `Aws2TestEnvContext` to handle setting up Quarkus AWS configuration properties


jamesnetherton:
Currently it only sets Camel component options and these are used in the aws quarkus-client grouped tests `Aws2TestEnvCustomizer` to set the corresponding Quarkus config.

It'd be better if the Quarkus properties were set in `Aws2TestEnvCustomizer`. The camel component properties aren't really needed when we test the Quarkus client scenario.

PR #3158, opened 2021-10-05
Revert Aws2TestEnvContext changes in order to extend aws lambda function version operations test coverage

aldettinger:


aldettinger:
I was not able to find a solution concerning https://github.com/apache/camel-quarkus/commit/cb4f00f44a7bdb613dbd174da32a5cef913a8b0e, so I've reverted. Let's see whether is successful and then we can re-open #3137.

aldettinger:
Ok looks good, let's re-open the issue then.

PR #3170, opened 2021-10-08
Aws sqs test fix

VratislavHais:
This PR fixes the problem with `sqsDeleteMessage` test. Unfortunately when testing I have discovered that test for delay queue is broken. I am still investigating it. It looks like the problem is on AWS side as everything works as expected in CQ a camel.

### DIFF
--- a/integration-test-groups/aws2/aws2-ddb/src/main/java/org/apache/camel/quarkus/component/aws2/ddb/it/Aws2DdbStreamRoutes.java
+++ b/integration-test-groups/aws2/aws2-ddb/src/main/java/org/apache/camel/quarkus/component/aws2/ddb/it/Aws2DdbStreamRoutes.java
@@ -46,6 +46,7 @@ public class Aws2DdbStreamRoutes extends RouteBuilder {
         from("aws2-ddbstream://" + streamTableName
                 + "?streamIteratorType=FROM_LATEST")
                         .id("aws2DdbStreamRoute")
+                        .autoStartup(false)
                         .process(e -> {
                             Record record = e.getMessage().getBody(Record.class);
                             StreamRecord item = record.dynamodb();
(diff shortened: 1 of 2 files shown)


zbendhiba:
We can open a separate issue to fix the queue delay later

aldettinger:
Beyond previous comments from Zineb, I think that the aggregated changes looks good.

However, I wonder whether those commits should be squashed ? For instance one commit is adding method "deleteQueueAndListQueues" and then another commit is deleting this very same method.

Also the CI build is failing, maybe running `mvn clean process-resources -P format` would help.

zbendhiba:
> Are you able to reproduce the failures on your localhost? I have tried it both with localstack and with AWS credentials and the test is passing. The failing test is `Aws2DdbStreamTest`

Is it related to this PR : https://github.com/apache/camel-quarkus/pull/3175 ?
If yes, maybe just rebase from main

zbendhiba:
> I don't understand. It has been rebased with newest changes and yet the tests (not sqs tests) are failing...

Let me try this locally on my machine

zbendhiba:
@VratislavHais did rebase and it didn't work
I reset the branch to be equal to main, and added the commit on top of it, and it's still not passing the CI.
I don't know what's the problem really

aldettinger:
@VratislavHais If we are not sure this PR is causing the build issue. I would propose to rebase and push without the fix, have a CI build, then push the fix and have another CI build. It may be a good experiment.

If we are sure the PR is causing the issue on CI, then would it be possible to comment parts of the PRs and have intermediate CI builds to check which part would be problematic ?

If we are not able to correct, maybe implementing test in another way would be a way to avoid long investigation on CI builds ?

And yeah, you are right those issues occurring on CI only are really hard to deal with. Sadly, it happens from time to time.

aldettinger:
As far as I remember, we are running on RHEL 7 and RHEL 8 and the test passes on our machines. CI builds seem to occur on Ubuntu 20.04.3.

VratislavHais:
@aldettinger I believe this is the problem: https://pastebin.com/xzbUFnbv 
I have tried increasing the size of table for shards (as suggested by google) from 10 to 20 but it did not help. I'll continue my investigation.

aldettinger:
@JiriOndrusek, @VratislavHais, @zbendhiba The build is passing now, well done for this hard investigation :+1: 
Do you think it's valuable to merge this PR at this stage ? Or was it just to experiment the build and then we are waiting for a cleaner/long-term fix ?

JiriOndrusek:
Fix of the `aws2-ddb` test, which is part of this PR, is long term fix - it fixes way how route is started. Although it is not required for ddb, it makes sense (and makes ddb test working in case that local tests are taking a lot of time - which can happen during grouped run).

PR #3185, opened 2021-10-14
Improve the User guide

ppalaga:


ppalaga:
0e9dbde: replaced "has three compile dependencies" with "has two compile dependencies"

Issue #3131, opened 2021-09-27
Google-Storage Extension: Cannot set serviceAccountKey because of Quarkiverse extension


oscerd:
I think this part of the quarkiverse extension

https://github.com/quarkiverse/quarkus-google-cloud-services/blob/main/storage/runtime/src/main/java/io/quarkiverse/googlecloudservices/storage/runtime/StorageProducer.java#L35

In this way we cannot use different configuration coming from:

https://github.com/apache/camel/blob/main/components/camel-google/camel-google-storage/src/main/java/org/apache/camel/component/google/storage/GoogleCloudStorageConnectionFactory.java

Because it will be always override by the credentials provider coming from the extension.

jamesnetherton:
This will have been fixed by https://github.com/apache/camel-quarkus/pull/3188.

PR #3188, opened 2021-10-15
Remove quarkiverse-google-cloud-services dependencies

jamesnetherton:
Raising this in draft initially to start a conversation.

I [commented](https://github.com/apache/camel-quarkus/issues/3131#issuecomment-938614747) in a recent issue related to the google-storage extension that we could maybe live without the Quarkiverse Google Cloud Services extensions.

So I looked into what changes would be required to remove it and use the plain Camel components. Turns out it seems to not be much work.


jamesnetherton:
Thanks for the feedback. Seems the general consensus is that we go ahead and merge this 👍

PR #3216, opened 2021-10-20
2.3.x new indexer/jsonpath syntax

djencks:
See apache/camel-website#648.


djencks:
There's a chicken-egg problem. I've adjusted the local partial antora build so it will work once the 3.11.x branches of this work in camel and camel-spring-boot are applied, but meanwhile the checks on this one will fail.  I guess I can add another commit to make the PR check pass and remove it again before merging.

PR #3246, opened 2021-11-01
Assert that Component DSL and Endpoint DSL work for AtlasMap

ppalaga:


oscerd (approved):

ppalaga:
I do not think this PR impacts the documentation in any way. I do not see any meaningful message in the log that would explain what's wrong. Any idea what's wrong @djencks ?

djencks:
Someone broke the user manual by using an unescaped ${body}, it has nothing to do with your PR. The current way of doing partial 
site builds is not maintainable, I hope to start some discussion on alternatives soon.

davsclaus:
Trying to fix that weird antora xref error about ${body}
https://github.com/apache/camel/commit/ff147620a0ff99f8efd67be9c9a7a47c4dd2d760

djencks:
If what is inside the {}follows asciidoc syntax for an attribute Asciidoctor will interpret it as one, so an ${} expression will need escaping. If there's a '.' It can't be an attribute name so no escaping is needed.

Issue #3258, opened 2021-11-03
Cannot run tests against alternative BOMs


ppalaga:
In https://github.com/apache/camel-quarkus/pull/3134 we wanted to make it possible to run our tests against alternative BOMs, such as Quarkus Platform BOMs. The current solution does not work for two reasons:

1. `camel-quarkus-bom-test` version used in `camel-quarkus-build-parent-it` is `${project.version}` - so `camel-quarkus-bom-test` would have to be installed locally to be able to run the test.
2. Virtual dependencies in the tests, such as 

```
        <dependency>
            <groupId>org.apache.camel.quarkus</groupId>
            <artifactId>camel-quarkus-direct-deployment</artifactId>
            <version>${project.version}</version>
            <type>pom</type>
            <scope>test</scope>
            <exclusions>
                <exclusion>
                    <groupId>*</groupId>
                    <artifactId>*</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

cannot be resolved too.

For 1. I think the best approach would be to make the `camel-quarkus-bom-test` version in `camel-quarkus-build-parent-it` parametrized, defaulting to `${project.version}`

For 2. I think we should move the virtual deps in tests to a new profile that would be active by default, but still de-activable via a property passed through CLI.

PR #3278, opened 2021-11-09
Fixup 218bef4 Special fix for Solr - Cannot run tests against

ppalaga:
alternative BOMs #3258


oscerd (approved):

aldettinger (approved):
Played a bit with grep and this is the only occurrence I've found too.

PR #3279, opened 2021-11-09
Fix #3276 to make sure all objects have been deleted after each test …

### LINKED ISSUE #3276: AWS Integration test doesn't clean bucket
After running integration test `camel-quarkus-integration-test-aws2-s3` the bucket isn't cleaned.

zhfeng:
…in AWS2 S3

### DIFF
--- a/integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/Aws2S3Test.java
+++ b/integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/Aws2S3Test.java
@@ -26,8 +26,11 @@ import io.restassured.RestAssured;
 import io.restassured.http.ContentType;
 import org.apache.camel.quarkus.test.support.aws2.Aws2TestResource;
 import org.apache.commons.lang3.RandomStringUtils;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
 
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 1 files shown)


oscerd (approved):

ppalaga (approved):
Looks good,. thanks @zhfeng 

Could you please squash the two commits?

integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/Aws2S3Test.java
```diff
@@ -36,6 +37,19 @@
...
+    public void after() {
+
+        // Make sure all create objects have been deleted
+        final String[] objects = RestAssured.given()
+                .get("/aws2/s3/object-keys")
+                .then()
+                .statusCode(200)
+                .extract()
+                .body().as(String[].class);
+        Assertions.assertTrue(objects.length == 0);
```
aldettinger, line 50: Is it good idea to put assertions in @AfterEach ?
Would each test be successful only when objects.length ==0, maybe we could assert at the end of each test ?
Would objects.length==0 show a situation where the tests execution is not satisfying, maybe throwing an exception is better ?

ppalaga: > Is it good idea to put assertions in @AfterEach ?

My understanding is that this `@AfterEach` assertion is intended much more as a check that the test has no leaks than an ordinary assertion checking the behavior of the application under test. Hence +1 for me.

> Would each test be successful only when objects.length ==0, maybe we could assert at the end of each test ?

`@AfterEach` has the advantage that it holds for all current and future test methods. If a method gets added in the future that leaks objects, the author will get notified early.

> Would objects.length==0 show a situation where the tests execution is not satisfying, maybe throwing an exception is better ?

Well, the assert method throws an AssertionError. Not sure what else could we do for making the test execution fail?

zhfeng: Good point ! The other issue could be if one of the test leaks the object, and remain tests which running after it will be all failed even if there are no leak objects with them. This might make a confusion to the author. So what about

```java
    private int objects_num_before;
    private int objects_num_after;

    @BeforeEach
    public void before() {
        objects_num_before = getObjects().length;
    }

    @AfterEach
    public void after(TestInfo testInfo) {
        
        // Make sure all create objects have been deleted after test
        objects_num_after = getObjects().length;
        if (objects_num_after != objects_num_before) {
            Assertions.fail("There is object leak after " + testInfo.getTestMethod().get().getName());
        }
    }

    private String[] getObjects() {
        final String[] objects = RestAssured.given()
                .get("/aws2/s3/object-keys")
                .then()
                .statusCode(200)
                .extract()
                .body().as(String[].class);

        return objects;
    }
```

aldettinger: Ok, I see. The intent is like a general assertion that would apply to the set of tests as a whole (not each test individually). So, yes maybe objects_num_after == objects_num_before could be a good fit. Or maybe assume objects_num_before =0. Or maybe record each object ids in a list and check they are not present. All that kind of solutions should be ok I guess.

aldettinger (approved):

PR #3298, opened 2021-11-13
update xrefs from camel 3.12.x to 3.13.x

djencks:
See apache/camel-website#670


djencks:
Current plan seems to be to leave camel 3.12.x on the website until cq 2.4.x is retired: this would replace the need for this PR.

PR #3299, opened 2021-11-13
Upgrade Camel to 3.13.0

jamesnetherton:

PR #3304, opened 2021-11-16
Upgrade camel 3.13.0 dj

djencks:
This gets the local antora build to work.

I do not recommend applying this: I think it would be greatly preferable to remove the local partial build ASAP.

What I did:

- copy package.json, antora-playbook.yml, and util from camel-website.
- make minimal modifications to get the build to work.
- prune out as many sources and start_paths as possible.
- update the pom to use the correct yarn version and yarn scripts.

The groovy script to update stuff in the playbook has no effect since the annotations aren't there.


jamesnetherton:
> I do not recommend applying this: I think it would be greatly preferable to remove the local partial build ASAP.

@djencks So can I just disable the xref checks until there is some alternate solution in place?

I don't have much time at the moment to dedicate to fixing up docs. So my preference would be to do the minimal thing required to move the build along, get #3299 merged, and proceed to cut the 2.5.0 release later this week. I can create an issue to track any future fixes.

djencks:
I recommend removing the xref check in cq and keeping an eye on the real website build.

jamesnetherton:
I disabled the xref checks and opened https://github.com/apache/camel-quarkus/issues/3307 to track improvements.

Issue #3179, opened 2021-10-12
`quarkus.camel.main.shutdown.timeout`  doesn't work as intended


iqnev:
The property 

> quarkus.camel.main.shutdown.timeout

  doesn't work when I tried to configure my application. Also I  tried with the following code to set graceful shutdown but It doesn't work, too.
```
camelContext
        .getShutdownStrategy()
        .setTimeout(66); 
```

While I was debugging my Camel application I realized, that the default graceful shutdown strategy is 

> NoShutdownStrategy

In Apache Camel the default strategy  is https://camel.apache.org/manual/latest/graceful-shutdown.html#_implementing_custom_component_or_shutdownstrategy

> DefaultShutdownStrategy

JiriOndrusek:
@iqnev It is possible to set graceful shudown via property
```
camel.main.shutdownTimeout = 15
```
or via camel context.
```
camelContext
        .getShutdownStrategy()
        .setTimeout(66);
```

In development mode `NoShutdownStrategy` is used.

iqnev:
@JiriOndrusek Thanks! I will try it.

PR #3310, opened 2021-11-19
Graceful shutdown strategy used as default one

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/3179

Adds following behavior:
- in development mode, if no graceful shutdown timeout is set, there is no change (`noShutdownStrategy` is used)
- in other cases `gracefulShutdownStrategy` (same a is used in the Camel) is used
@ppalaga I think that this should be mentioned in some doc, do you have a suggestion where to write it?

### DIFF
--- a/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
+++ b/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
@@ -25,6 +25,8 @@ import java.util.stream.Collectors;
 import io.quarkus.arc.deployment.BeanContainerBuildItem;
 import io.quarkus.arc.deployment.BeanDiscoveryFinishedBuildItem;
 import io.quarkus.arc.deployment.SyntheticBeansRuntimeInitBuildItem;
+import io.quarkus.deployment.IsDevelopment;
+import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.Consume;
 import io.quarkus.deployment.annotations.ExecutionTime;
@@ -100,6 +102,24 @@ public class CamelContextProcessor {
         }
 
         return new CamelContextBuildItem(context);
+    }
+
+    /**
+     * This step customizes camel context for development mode.
+     *
+     * @param recorder the recorder
+     * @param producer producer of context customizer build item
+     */
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep(onlyIf = IsDevelopment.class)
+    public void devModeCamelContextCustomizations(
+            CamelContextRecorder recorder,
+            BuildProducer<CamelContextCustomizerBuildItem> producer) {
+        String val = CamelSupport.getOptionalConfigValue("camel.main.shutdownTimeout", String.class, null);
+        if (val == null) {
+            //if no graceful timeout is set in development mode, graceful shutdown is replaced with no shutdown
+            producer.produce(new CamelContextCustomizerBuildItem(recorder.createNoShutdownStrategyCustomizer()));
+        }
     }
 
     /**
--- a/docs/modules/ROOT/pages/migration-guide/2.6.0.adoc
+++ b/docs/modules/ROOT/pages/migration-guide/2.6.0.adoc
@@ -1,0 +1,19 @@
+= Camel Quarkus 2.6.0 Migration Guide
+
+The following guide outlines how to adapt your code to changes that were made in Camel Quarkus 2.6.0 & Quarkus 2.6.0.Final.
+
+== Graceful shutdown strategy is used by default
+
+In previous releases, graceful shutdown strategy (default strategy in Camel, see https://camel.apache.org/manual/graceful-shutdown.html[documentation] ) wasn't used by default.
+Shutdown wasn't controlled by any strategy.
+
+This is no longer the case.
+Graceful shutdown strategy is enabled by default (see https://camel.apache.org/manual/graceful-shutdown.html[documentation]) with one exception only.
+If timeout for graceful shutdown is not set and application runs in a development mode, no shutdown strategy is used (behavior for the development mode without timeout for graceful timeout wasn't changed).
+
+`DefaultShutdownStrategy` could be configured via `application.properties`.
+For example:
+```
+#set graceful timeout to 15 seconds
+camel.main.shutdownTimeout = 15
+```
--- a/docs/modules/ROOT/pages/migration-guide/index.adoc
+++ b/docs/modules/ROOT/pages/migration-guide/index.adoc
@@ -4,6 +4,7 @@ We do frequent releases, a release almost every month, and even though we strive
 
 Listed here are guides on how to migrate between major versions and anything of significance to watch for when upgrading from minor versions.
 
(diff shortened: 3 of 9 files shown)


oscerd (approved):

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
```diff
@@ -102,6 +106,28 @@ CamelContextBuildItem context(
...
+            CamelContextRecorder recorder,
+            Capabilities capabilities,
+            BuildProducer<CamelContextCustomizerBuildItem> producer) {
+        if (capabilities.isPresent(CamelCapabilities.MAIN)) {
+            String val = CamelSupport.getOptionalConfigValue("camel.main.shutdownTimeout", String.class, null);
+            if (val == null) {
+                //if no graceful timeout is set in development mode, graceful shutdown is replaced with no shutdown
+                producer.produce(new CamelContextCustomizerBuildItem(recorder.createNoShutdownStrategyCustomizer()));
+            }
+        }
```
aldettinger, line 128: Shouldn't we also set no shutdown strategy in DEV mode when no running with main ?

aldettinger:
@JiriOndrusek Looks good, still few questions. Logically, it should make `quarkus.camel.main.shutdown.timeout=xxx` work. I don't see how to automatic test this but wonder if you have been able to check this manually on your machine ?

jamesnetherton:
> I don't see how to automatic test this

As a minimum, we could probably use [`QuarkusDevModeTest`](https://github.com/apache/camel-quarkus/blob/5ad13f2be3d6d8ccb771dcea4c37a7d0c600b24e/integration-tests/main-devmode/src/test/java/org/apache/camel/quarkus/main/CamelMainRoutesIncludePatternWithNoPrefixDevModeTest.java#L41) and verify that the correct shutdown strategy is configured on the `CamelContext`, depending on whether `camel.main.shutdownTimeout` is set.

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
JiriOndrusek: makes sense, I'll fix it.

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
```diff
@@ -102,6 +106,28 @@ CamelContextBuildItem context(
...
+    /**
+     * This step customizes camel context for development mode.
+     *
+     * @param recorder     the recorder
+     * @param capabilities the registered capabilities
+     * @param producer     producer of context customizer build item
+     */
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep(onlyIf = IsDevelopment.class)
+    public void developmenMode(
```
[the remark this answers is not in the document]
JiriOndrusek: I'll fix it.

JiriOndrusek:
I'll try to add a test.
(@aldettinger I tested it locally with an app)
What do you think about a doc change regarding this?

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
JiriOndrusek: fixed

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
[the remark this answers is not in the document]
JiriOndrusek: fixed

JiriOndrusek:
@jamesnetherton , @aldettinger I added 3 tests:

1. in core (not dev mode) - verifying that `DefaultShutdownStrategy` is used.
2. in main-devmode - `CamelContextNoShutdownStrategyDevModeTest` -  if `camel.main.shutdownTimeout` is set, `DefaulShutdownStrategy` is used 
3. in main-devmode - `CamelContextNoShutdownStrategyDevModeTest` -  if `camel.main.shutdownTimeout` is not set, `NoShutdownStrategy` is used

aldettinger:
@JiriOndrusek It looks better :+1: 

Do we have a test for core in DEV mode ? I mean without using camel-main in DEV mode. My current reading is that we will use the default shutdown strategy whereas no shutdown strategy would be preferrable ?

About the doc change: Are you talking about stating in the doc that the shutdown strategy has changed ? IMHO, I don't think it is needed.

JiriOndrusek:
@aldettinger I would say that the use of `DefaultShudownStrategy` is preferable in all cases, with one exception - dev mode - because of quick redeploy. (but if in dev-mode `camel.main.shutdownTimeout` is set, default shutdown strategy will be used)

As for documentation. Current status was, that `NoShutdownStrategy` was used in all cases. This fix brings a change, which could affect some users. This is the reason, why I would say that doc should be updated - somewhere. (that default shutdown strategy was changed - mainly for not dev-mode)

aldettinger:
> I would say that the use of DefaultShudownStrategy is preferable in all cases, with one exception - dev mode - because of quick redeploy. (but if in dev-mode camel.main.shutdownTimeout is set, default shutdown strategy will be used)

I agree. Maybe I missread that with the current implementation, `DefaultShudownStrategy` is used in DEV mode when camel main is disabled. Do you agree with that ?

> As for documentation. Current status was, that NoShutdownStrategy was used in all cases. This fix brings a change, which could affect some users. This is the reason, why I would say that doc should be updated - somewhere. (that default shutdown strategy was changed - mainly for not dev-mode)

Ok, it's clearer now. I think James has an idea that this could be in the migration guide. I think it is a good option.

JiriOndrusek:
@aldettinger 
```
DefaultShudownStrategy is used in DEV mode when camel main is disabled. Do you agree with that ?
```
Not entirely correct. `DefaultShutdownStrategy` is used in dev mode only if `camel.main.shutdownTimeout` is not null. (without taking into account whether camel main is disabled) - may not be the solution which we want. So in case that different condition should be applied, I'll change it.

JiriOndrusek:
I hope that following statement will make the current behavior clear.

Current implementation is using only `NoShutdownStrategy`.
 This fix changes strategy to `DefaultShutdownStrategy`,
Only if `camel.main.shutdownTimeout` is not set and dev mode is enabled -  `NoShutdownStrategy` remains.

JiriOndrusek:
I'll add change into the migration doc -  https://github.com/apache/camel-quarkus/tree/main/docs/modules/ROOT/pages/migration-guide

aldettinger (approved):
@JiriOndrusek I'm +1 on adding the documentation to the migration guide

And indeed, I see it that now it does not take into account whether camel main is disabled. So the test coverage is good :+1:

JiriOndrusek:
@aldettinger documentation is added.

aldettinger:
The migration guide is definitely the right place. We could merge this PR from my point of view :+1:

ppalaga (approved):

jamesnetherton (approved):

PR #3311, opened 2021-11-19
Bump Optaplanner version to 8.14.0.Final

zbendhiba:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
         <camel.sb.docs.branch>camel-spring-boot-${camel.major.minor}.x</camel.sb.docs.branch><!-- The stable camel-spring-boot branch on which our Antora docs depends -->
         <cassandra-quarkus.version>1.1.1</cassandra-quarkus.version><!-- https://repo1.maven.org/maven2/com/datastax/oss/quarkus/cassandra-quarkus-bom/ -->
         <debezium.version>1.6.1.Final</debezium.version><!-- May go back to Camel's ${debezium-version} when they are in sync https://repo1.maven.org/maven2/io/debezium/debezium-bom/ -->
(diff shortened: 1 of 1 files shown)


oscerd (changes requested):
This won't work i guess, camel needs to stay on 8.9 until we stop supporting jdk 8. And this is will happen after 3.14

zbendhiba:
> This won't work i guess, camel needs to stay on 8.9 until we stop supporting jdk 8. And this is will happen after 3.14

When you say this won't work, is there any gap in functionalities between those 2 versions? As all Quarkus extensions we use are JDK 11 minimum. 

Or maybe it doesn't make sense to upgrade to a version behind camel 3, as it could be problematic ?

What will happen exactly if we stay on 8.9 when we generate the project from code.quarkus.io ? Will we have 8.9 or 8.13 ? It's still not clear to me. @ppalaga @jamesnetherton @aldettinger

oscerd:
As far as I know, there shouldn't be breaking change between 8.9 and 8.10-8.13, the upgrade should be smooth. What I don't know, is what happen if you use camel component based on 8.9 in the context of Camel-Quarkus, importing Optaplanner 8.13 from his BOM. Not sure this will be smooth, so I would be careful, but I might be too much worried.

jamesnetherton:
> What will happen exactly if we stay on 8.9 when we generate the project from code.quarkus.io ? Will we have 8.9 or 8.13 ? It's still not clear to me. @ppalaga @jamesnetherton @aldettinger

AFAIK platform will prioritise the version provided by the participants. So I think we'll get 8.13.

https://github.com/quarkusio/quarkus-platform/blob/main/generated-platform-project/quarkus-camel/bom/pom.xml#L7137-L7156

The platform JVM mode tests seem to be passing, so maybe the version bump is ok for us.

zbendhiba:
If we are okay to merge this. 
I will upgrade to the latest version bumped into platform, which is 8.14.0.Final, and rebase main.

oscerd (approved):

Issue #2978, opened 2021-08-03
[Camel 3.12] New feature: route configurations


davsclaus:
https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-16757

camel-quarkus may need to add some code to support this for build time optimizations.

There is a new RouteConfigurationBuilder (extends RouteBuilder) that should preferred to be added before any routes. However camel-core takes usually care of this.

A good start is to make a quarkus example of the spring boot
https://github.com/apache/camel-spring-boot-examples/tree/main/routes-configuration

It uses route configurations in java, xml, and yaml

aldettinger:
Related camel PR: https://github.com/apache/camel/pull/5917/files#diff-e510105c62e50b5f9cfe0a983a17e3df523e09668e568380805ccaab6dd277ff

aldettinger:
@davsclaus It is working against the [current test coverage](https://github.com/apache/camel-quarkus/commit/a2f9cf851641d797b7cabca7074cfeb3ab128580).

Would you please have any precise ideas about:
- Possible build time optimizations ?
- Cases where RouteConfigurationBuilders should be added before RouteBuilders on camel-quarkus side ?

davsclaus:
Same build time as regular RouteBuilder classes
Add them before regular RouteBuilder

camel-core uses routes loader to load/add routes, and it has logic such as
```
    default void loadRoutes(Collection<Resource> resources) throws Exception {
        Collection<RoutesBuilder> builders = findRoutesBuilders(resources);
        // add configuration first before the routes
        for (RoutesBuilder builder : builders) {
            if (builder instanceof RouteConfigurationsBuilder) {
                getCamelContext().addRoutesConfigurations((RouteConfigurationsBuilder) builder);
            }
        }
        for (RoutesBuilder builder : builders) {
            getCamelContext().addRoutes(builder);
        }
    }
```

aldettinger:
The default case where camel main is enabled is ok. Under the hood, we use a RoutesConfigurer that ultimately adds the discovered routes in the right order, i.e. route configurations before classical route builders. This is the case in all current itests and the re-ordering is happening correctly.

The case where camel main is disabled in another beast. The route builders order depends upon a mix of route discovery, route known by the container and route provided by the registry. More investigation is needed but likely route builder re-ordering will be needed.

davsclaus:
When would camel main be disabled or enabled?

For Camel K I would assume we boot with camel main.
[...]

PR #3316, opened 2021-11-22
core: Add RouteConfigurationsBuilder before regular RoutesBuilder whe…

aldettinger:
…n camel main is disabled #2978

### DIFF
--- a/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
+++ b/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
@@ -149,7 +149,7 @@ public class CamelContextProcessor {
     @BuildStep(onlyIfNot = CamelMainEnabled.class)
     @Record(value = ExecutionTime.RUNTIME_INIT, optional = true)
     /* @Consume(SyntheticBeansRuntimeInitBuildItem.class) makes sure that camel-main starts after the ArC container is
-     * fully initialized. This is required as under the hoods the camel registry may look-up beans form the
+     * fully initialized. This is required as under the hoods the camel registry may look-up beans from the
      * container thus we need it to be fully initialized to avoid unexpected behaviors. */
     @Consume(SyntheticBeansRuntimeInitBuildItem.class)
     public CamelRuntimeBuildItem runtime(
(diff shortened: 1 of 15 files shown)


oscerd (approved):

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelContextRecorder.java
```diff
@@ -90,6 +93,8 @@ public void customize(RuntimeValue<CamelContext> context, RuntimeValue<CamelCont
         return new RuntimeValue((CamelContextCustomizer) context -> context.setShutdownStrategy(new NoShutdownStrategy()));
     }
```
aldettinger, line 95: I've marked those recorded methods as deprecated. Maybe it's even ok to remove them ? I don't know what we have done in the past for that kind of situations.

zbendhiba: If it's for internal use only, I'd vote for removing them. 
If the user can use them, I'd say maybe deprecate and remove when introducing 3.14, the next LTS ?

jamesnetherton: Agreed - I think for internal stuff in the past like recorder methods, we have removed them without deprecation. It's not public API.

aldettinger: Okay I've removed the useless methods. Thanks all.

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
```diff
@@ -165,17 +165,12 @@ public CamelRuntimeBuildItem runtime(
             List<CamelRuntimeTaskBuildItem> runtimeTasks,
             CamelConfig config) {
 
-        for (CamelRoutesBuilderClassBuildItem item : routesBuilderClasses) {
-            // don't add routes builders that are known by the container
-            if (containerBeans.getClasses().contains(item.getDotName())) {
-                continue;
-            }
-
-            recorder.addRoutes(context.getCamelContext(), item.getDotName().toString());
```
aldettinger, line 174: I've spread the logic from the loop above in the code block below. So, the logic is now executed only when config.routesDiscovery.enabled = true.

However, to my current understanding, when config.routesDiscovery.enabled = false, then the routesBuilderClasses list is empty. As such, the logic in the loop above was not executed when config.routesDiscovery.enabled = false.

zhfeng (approved):
LGTM and well done !

zbendhiba (approved):

PR #3319, opened 2021-11-23
JFR Native support

jamesnetherton:


oscerd (approved):

docs/modules/ROOT/pages/reference/extensions/jfr.adoc
```diff
@@ -34,38 +38,75 @@ Please refer to the above link for usage and configuration details.
...
+----
+quarkus.native.enable-vm-inspection=true
+----
+
+
 == Additional Camel Quarkus configuration
 
 [width="100%",cols="80,5,15",options="header"]
 |===
 | Configuration property | Type | Default
```
aldettinger, line 84: It's not really linked to this PR. I've noticed that the default value column is empty.

aldettinger (approved):
Looks good to me.

zbendhiba (approved):

docs/modules/ROOT/pages/reference/extensions/jfr.adoc
jamesnetherton: It's because I did not specify any default values on the config items. Camel has its own internal defaults which I mention in the config property documentation.

Maybe I should specify the defaults in `RuntimeCamelJfrConfig` so that it's clear in the docs?

aldettinger: Having the table filled with the default value would looks clearer from doc perspective. However, I read it like the true default values come from Camel, so maybe we should not maintain that in camel-quakus ?
I don't know if it's possible to point to camel default values.

zhfeng (approved):

docs/modules/ROOT/pages/reference/extensions/jfr.adoc
jamesnetherton: I did some experiments and have decided to leave things as-is and not duplicate the defaults that are already defined in Camel.

I may follow up with another PR later to refactor the way the JFR config works to simplify it a bit.

Issue #3260, opened 2021-11-03
Sql test using derby doesn't start dev service and shows class loading issue if stored procedure is called


JiriOndrusek:
Follows  https://github.com/apache/camel-quarkus/issues/3080

- Dev service for derby doesn't start docker container
- If stored procedure via java class is added, it's execution shows class loading problem.

PR #3324, opened 2021-11-23
Sql test using derby doesn't start dev service and stored procedure is not working on derby db

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/3260

- I used a workaround for an issue in quarkus - https://github.com/quarkusio/quarkus/issues/21639
- Test still can't be executed with external derby db (hance the condition on the stored procedure test). Problem is caused by the fact, that stored procedure is a java class, which has to be uploaded to the external derby  instance. With devServices (which uses in-memory db) this is not a problem.
- I used different approach for the stored procedure test - creation of temp file. Which is sufficient for the test.

### DIFF
--- a/integration-tests/sql/pom.xml
+++ b/integration-tests/sql/pom.xml
@@ -156,5 +156,22 @@
                 </dependency>
             </dependencies>
         </profile>
+        <profile>
+            <id>derbyDevServiceWorkaround</id>
+            <activation>
+                <property>
+                    <name>cq.sqlJdbcKind</name>
+                    <value>derby</value>
+                </property>
+            </activation>
+            <dependencies>
+                <dependency>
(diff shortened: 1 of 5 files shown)


aldettinger (approved):
Well done, it looks good.

Maybe we'll need a follow up for quarkusio/quarkus#21639 i, camel-quarkus 

> Test still can't be executed with external derby db

Is it an issue end users could face or maybe just a test harness issue ? Then, should we have a follow up to fix that too ?

JiriOndrusek:
@aldettinger I created a follow-up issue https://github.com/apache/camel-quarkus/issues/3327
At the first glance I think that it won't be possible to use external derby db for testing stored procedures, but I might be wrong.

jamesnetherton (approved):

PR #3337, opened 2021-11-25
Ban com.google.code.findbugs:jsr305 unconditionally 

### LINKED ISSUE #3335: Ban com.google.code.findbugs:jsr305 unconditionally
`com.google.code.findbugs:jsr305` is a pure dev UX dependency that might be useful when compiling code containing those annotations, but is completely useless when it is only a transitive dependency.
We exclude it where we can in Camel Quarkus, but we still need to fix some holes.

We should ban it via enforcer plugin or similar

See also https://github.com/quarkusio/quarkus-platform-bom-generator/issues/95

ppalaga:
Fix #3335

### DIFF
--- a/extensions-support/google-http-client/runtime/pom.xml
+++ b/extensions-support/google-http-client/runtime/pom.xml
@@ -58,6 +58,12 @@
         <dependency>
             <groupId>com.google.http-client</groupId>
             <artifactId>google-http-client</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.google.code.findbugs</groupId>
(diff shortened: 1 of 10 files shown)


oscerd (approved):

jamesnetherton (approved):

aldettinger (approved):

davsclaus:
We need similar "fixes" in the core camel project as well - i assume we can do the same for the camel components that an extension has been done for, and do the same there to with maven exclude

davsclaus:
Created a ticket for camel core project
https://issues.apache.org/jira/browse/CAMEL-17237

Issue #3340, opened 2021-11-26
[JDK17]kudu:integration test failed in native mode


ffang:
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 14.436 s <<< FAILURE! - in org.apache.camel.quarkus.component.kudu.it.KuduIT
[ERROR] org.apache.camel.quarkus.component.kudu.it.KuduIT  Time elapsed: 14.435 s  <<< ERROR!
java.lang.IllegalStateException: Could not hack the kudu tablet server hostname resolution
	at org.apache.camel.quarkus.component.kudu.it.KuduInfrastructureTestHelper.overrideTabletServerHostnameResolution(KuduInfrastructureTestHelper.java:87)
	at org.apache.camel.quarkus.component.kudu.it.KuduTest.setup(KuduTest.java:51)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at io.quarkus.test.junit.QuarkusTestExtension.interceptBeforeAllMethod(QuarkusTestExtension.java:635)
	at org.junit.jupiter.engine.execution.ExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(ExecutableInvoker.java:115)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.lambda$invoke$0(ExecutableInvoker.java:105)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
[...]

PR #3341, opened 2021-11-26
fix  [JDK17]kudu:integration test failed in native mode #3340

ffang:

### DIFF
--- a/integration-tests/kudu/pom.xml
+++ b/integration-tests/kudu/pom.xml
@@ -30,6 +30,11 @@
     <name>Camel Quarkus :: Integration Tests :: Kudu</name>
     <description>Integration tests for Camel Quarkus Kudu extension</description>
 
+    <properties>
+        <!-- This is to allow the "deep reflection" we do in KuduInfrastructureTestHelper -->
+        <argLine>--add-opens java.base/java.net=ALL-UNNAMED</argLine>
+    </properties>
+
     <dependencies>
          <dependency>
             <groupId>io.quarkus</groupId>
@@ -94,24 +99,6 @@
                                 </goals>
                             </execution>
                         </executions>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
-        <!-- Required by KuduInfrastructureTestHelper.overrideTabletServerHostnameResolution -->
-        <profile>
-            <id>jdk16-workarounds</id>
-            <activation>
-                <jdk>[16,)</jdk>
-            </activation>
-            <build>
-                <plugins>
-                    <plugin>
-                        <groupId>org.apache.maven.plugins</groupId>
-                        <artifactId>maven-surefire-plugin</artifactId>
-                        <configuration>
-                            <argLine>--add-opens java.base/java.net=ALL-UNNAMED</argLine>
-                        </configuration>
                     </plugin>
                 </plugins>
             </build>


oscerd (approved):

aldettinger (approved):

zbendhiba (approved):

zbendhiba:
@ffang can this be added in the same profile as jdk16-workarounds ?
there is a similar line here : https://github.com/apache/camel-quarkus/blob/cb9c92008820642be374fa136f9b53a4677777d6/integration-tests/kudu/pom.xml#L113
Maybe it will be good to refactor the config. WDYT ?

ppalaga (changes requested):
A custom argline is not an end user-friendly solution. Is there no Quarkus BuildItem for this?
If we keep relying on argline, then we need to document this on the given extension page,

jamesnetherton:
> A custom argline is not an end user-friendly solution.

I think in this case, it's acceptable because we unfortunately need this hack in the itests:

https://github.com/apache/camel-quarkus/blob/main/integration-tests/kudu/src/main/java/org/apache/camel/quarkus/component/kudu/it/KuduInfrastructureTestHelper.java#L64

It's not required for users to do this in their apps AFAIK.

ffang:
> @ffang can this be added in the same profile as jdk16-workarounds ? there is a similar line here :
> 
> https://github.com/apache/camel-quarkus/blob/cb9c92008820642be374fa136f9b53a4677777d6/integration-tests/kudu/pom.xml#L113
> 
> Maybe it will be good to refactor the config. WDYT ?

ffang:
> > A custom argline is not an end user-friendly solution.
> 
> I think in this case, it's acceptable because we unfortunately need this hack in the itests:
> 
> https://github.com/apache/camel-quarkus/blob/main/integration-tests/kudu/src/main/java/org/apache/camel/quarkus/component/kudu/it/KuduInfrastructureTestHelper.java#L64
> 
> It's not required for users to do this in their apps AFAIK.

Yes, this issue indeed from KuduInfrastructureTestHelper itself, which tries to access no-public field from java.net.InetAddress with reflection. This won't happen in users apps.

ppalaga:
Thanks for the explanation @jamesnetherton and @ffang . Could we please have a big fat command near the `argLine` saying that this is due to the nasty things we do in `KuduInfrastructureTestHelper`?

ppalaga:
About where to put `<argLine>--add-opens java.base/java.net=ALL-UNNAMED</argLine>` given that java 11 is our baseline, where this should work and given that we need it for both plugins in all profiles, can't we just add 

```
<properties>
  <argLine>--add-opens java.base/java.net=ALL-UNNAMED</argLine>
</properties>
```

in the top profile-less context (and remove the old `jdk16-workarounds` profile altogether)?

ffang:
> About where to put `<argLine>--add-opens java.base/java.net=ALL-UNNAMED</argLine>` given that java 11 is our baseline, where this should work and given that we need it for both plugins in all profiles, can't we just add
> 
> ```
> <properties>
>   
>   <argLine>--add-opens java.base/java.net=ALL-UNNAMED</argLine>
> </properties>
> ```
> 
> in the top profile-less context (and remove the old `jdk16-workarounds` profile altogether)?

Hi @ppalaga ,

I may miss your point, but if we add
```
</dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <!-- This is to allow the "deep reflection" we do in KuduInfrastructureTestHelper -->
+                    <argLine>--add-opens java.base/java.net=ALL-UNNAMED</argLine>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>

```
in the pom.xml out of profiles, this can't affect the maven-failsafe-plugin configuration in native profile. The maven-failsafe-plugin configuration in native profile can't pick up this argLine and will fail with JDK17 in native mode. We need to add this argLine in native profile anyway. 

Freeman

ppalaga:
I spoke about `<argLine>` in `<properties>`, not plugin config.  We should be able to hit both plugins that way.

ffang:
> I spoke about `<argLine>` in `<properties>`, not plugin config. We should be able to hit both plugins that way.

That's right, I will revise the PR soon.

Thanks!
Freeman

ffang:
revised and squashed commits here
https://github.com/apache/camel-quarkus/pull/3341/commits/43d45606d89f07f6f44e5d998903cb8145e75263

ppalaga (approved):
Thanks @ffang!

PR #3359, opened 2021-12-07
Workaround for NoSuchFileException: .../target/classes when executing a

ppalaga:
codegen task https://github.com/quarkusio/quarkus/issues/21718

### DIFF
--- a/integration-tests/avro/pom.xml
+++ b/integration-tests/avro/pom.xml
@@ -62,6 +62,7 @@
             <plugin>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
+                <extensions>true</extensions><!-- Workaround for https://github.com/quarkusio/quarkus/issues/21718 -->
                 <executions>
                     <execution>
                         <id>generate-code</id>
--- a/integration-tests/protobuf/pom.xml
+++ b/integration-tests/protobuf/pom.xml
@@ -62,6 +62,7 @@
             <plugin>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
+                <extensions>true</extensions><!-- Workaround for https://github.com/quarkusio/quarkus/issues/21718 -->
                 <executions>
                     <execution>
                         <id>quarkus-generate-code</id>


oscerd (approved):

integration-tests/avro/pom.xml
```diff
@@ -62,6 +62,7 @@
             <plugin>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
+                <extensions>true</extensions><!-- Workaround for https://github.com/quarkusio/quarkus/issues/21718 -->
```
zbendhiba, line 65: What is the default version of extensions ? Why we don't have a value for it ?
On the doc here, it seems to be set to true on each snippet of code. 

I'm trying to understand. Otherwise, LGTM.

ppalaga: > What is the default version of extensions ? 

`false`

> Why we don't have a value for it ?

`true` is only needed if a plugin should act as Maven extension - i.e. have access to more Maven APIs than usual plugins.
Here I am trying to eliminate the failure of second `mvnd` invocation, that happens only with projects where we invoke `quarkus-maven-plugin` 's generate-code` goal. Without `<extensions>true</extensions>` some cache inside `quarkus-maven-plugin` does not get cleared via `AbstractMavenLifecycleParticipant`, because `AbstractMavenLifecycleParticipant` are only honored in Maven extensions.

This helps to get rid of the failure, but it is still a good question whether the plugin works correctly in the rest of the test modules (even though it does not fail there). 

WDYT @aloubyansky do we need to have `<extensions>true</extensions>` also in test modules where second mvnd invocation does not fail? - i.e. does the `build` mojo assume/require a fresh bootstrap on every maven invocation?

aloubyansky: `build` may re-use the one bootstraped in `generate-code`. Otherwise, in general, every module build should have its own bootstrap.

ppalaga:
Let's merge like this for now. If there is a solution for https://github.com/quarkusio/quarkus/issues/21718 that does not require `<extensions>true</extensions>`, then this commit can be reverted. Otherwise, we will probably have to add `<extensions>true</extensions>` everywhere.

Issue #3327, opened 2021-11-24
Sql derby - test exernal db if possible, remove workaround because of quarkus issue #21639


JiriOndrusek:
Follows https://github.com/apache/camel-quarkus/issues/3260

1. It should be investigated whether it is possible to run stored procedure est with external derby db
2. Issue https://github.com/quarkusio/quarkus/issues/21639 is already fixed, workaround (profile) can be removed.

JiriOndrusek:
@ppalaga @aldettinger 
I'm investigating a possibility of running stored procedure test on external derby db. I now see that it would need to require  manual steps from user:

1. Jar with stored procedure has to be uploaded into the external file system, which db has access to. Then it is possible to add the jar into db classpath and then execute it as a stored procedure. 
2. Another possible complication: Camel quarkus has to be build with Java 11+. If external derby db contains only Java 8-, it won't be possible to use a jar created as a part of cq build. User would need to create the procedure jar with the right java.

Considering previous limitations, I would say that disabed stored procedure test for external derby is the right solution and it makes no sense to enable such test for external Derby db. What do you think?

Here are some commands required to make stored procedure on external db work:

Java class for procedure
```
package test;

public class App 
{
    public static void testProc(int a, int b) {
         ...
}
```

Commands to be executed via sql:
```
CALL sqlj.install_jar('/home/test-1.0-SNAPSHOT.jar', 'App' , 0);

call syscs_util.syscs_set_database_property('derby.database.classpath', 'DOCKERDB.APP');

CREATE PROCEDURE ADD_NUMS(IN a INTEGER, IN b INTEGER) PARAMETER STYLE JAVA LANGUAGE JAVA EXTERNAL NAME 'test.App.testProc';

call ADD_NUMS2(1, 2);
```

aldettinger:
I think the jar would not forcely need to be rebuilt by quarkus, it could be stored as is in github. But then we should maybe have one compiled in java 8 and one in java 11. If it' s not possible to automate the db setup, then a test is pointless from my point of view.

Maybe we could find a docker image with a java stored proc already configured ? And just ensure that camel-quarkus can call it.

JiriOndrusek:
@aldeinger it might be possible to use docker image (I suppose), upload jar there and test it. (I was able to do it manually with docker derby). I will try to automatize the process. With docker based test, we will use external derby db, even if it is not users's own derby. It should make coverage bigger. Thank you for suggestion!

[...]

Issue #3083, opened 2021-09-09
Sql - enable  all the DB kind tests on the CI


JiriOndrusek:
Follows [PR](https://github.com/apache/camel-quarkus/pull/3066)

Enable CI testing of more db kinds for sql extension tests.
(should be done for all db kinds with dev services and without license requoirement)

PR #3370, opened 2021-12-09
Sql test using derby doesn't start dev service and shows class loadin…

JiriOndrusek:
…g issue if stored procedure is called #3260

fixes https://github.com/apache/camel-quarkus/issues/3327

Needs https://github.com/quarkusio/quarkus/issues/22058 to work correctly.

I introduced an option to start sql test with derby started by docker. - to simulate external database (and not dev service).
There are 2 things which may need some changes:

1. To run stored procedure on external derby db, there has to be a jar which is uploaded there. I created module `sql-derby` for that purpose and I put it on the same hierarchy level as `sql` test module. I'm not sure if this is the right approach, another option is to put both `sql` and `sql-derby` in one module.
2. I'm facing a problem with generated port number for docker container. If I let docker to generate its own random port, I'm not able propagate it back to datasource definition. Therefore there are 2 properties, which forces usage of docker derby db:
```
export SQL_USE_DERBY_DOCKER=true
export SQL_USE_DERBY_PORT="#port"
```
On the other hand in current moment, when there is no CI automation for derby (see https://github.com/apache/camel-quarkus/issues/3083) it could stay in that way.

WDYT? @aldettinger

### DIFF
--- a/integration-tests/sql/README.adoc
+++ b/integration-tests/sql/README.adoc
@@ -43,3 +43,28 @@ $Env:SQL_JDBC_PASSWORD="#password"
 ```
 
 Oracle database could be used as external db. In that case use parameter `-Dcq.sqlJdbcKind=oracle`.
+
+=== External Derby database
+
+To execute tests against external Derby database, stored procedure has to be uploaded into the database classpath.
+Jar with stored procedure for the derby database is creaed by module `sql-derby`.
+Jar could be uploaded via following commands through `ij`:
+```
+CALL sqlj.install_jar('/PATH_TO_JAR/camel-quarkus-integration-test-sql-derby-stored-procedure-*.jar', 'AddNumsProcedure' , 0)
+
+CALL syscs_util.syscs_set_database_property('derby.database.classpath', 'APP.ADDNUMSPROCEDURE')
+```
+
+=== External Derby database via Docker
+
+To avoid manual upload of the jar, test can automatically use external derby database created via docker.
+To execute the tests against external derby database, set the environment variable `SQL_USE_DERBY_DOCKER` to value `true`:
+
+```
+export SQL_USE_DERBY_DOCKER=true
+```
+
+or for windows:
+
+```
+$Env:SQL_USE_DERBY_DOCKER = "true"
--- a/integration-tests-support/pom.xml
+++ b/integration-tests-support/pom.xml
@@ -48,6 +48,7 @@
         <module>kafka</module>
         <module>mongodb</module>
         <module>process-executor-support</module>
+        <module>sql-derby-support</module>
         <module>test-support</module>
         <module>mock-backend</module>
         <module>wiremock</module>
--- a/integration-tests-support/sql-derby-support/README.adoc
+++ b/integration-tests-support/sql-derby-support/README.adoc
@@ -1,0 +1,3 @@
+== SQL integration tests
+
+This module creates jar with stored procedure used in test from `sql` module (for derby database).
--- a/integration-tests-support/sql-derby-support/pom.xml
+++ b/integration-tests-support/sql-derby-support/pom.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one or more
+    contributor license agreements.  See the NOTICE file distributed with
+    this work for additional information regarding copyright ownership.
+    The ASF licenses this file to You under the Apache License, Version 2.0
+    (the "License"); you may not use this file except in compliance with
+    the License.  You may obtain a copy of the License at
+
+         http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">    
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-integration-tests-support</artifactId>
+        <version>2.17.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>camel-quarkus-integration-tests-support-sql-derby</artifactId>
+    <name>Camel Quarkus :: Integration Tests :: Support :: Sql Derby</name>
+
+    <properties>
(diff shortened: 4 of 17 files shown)


JiriOndrusek:
I think that better approach (step 1) is to created sql-derby as a module in integration-tests-support, I'll try it

integration-tests/sql/README.adoc
```diff
@@ -42,4 +42,32 @@ $Env:SQL_JDBC_USERNAME="#username"
...
+=== External Derby database
+
+To execute tests against external Derby database, stored procedure has to be uploaded into the database classpath.
+Jar with stored procedure for the derby database is creaed by module `sql-derby`.
+Jar could be uploaded via following commands through `ij`:
+```
+CALL sqlj.install_jar('/PATH_TO_JAR/camel-quarkus-integration-test-sql-derby-stored-procedure-*.jar', 'AddNumsProcedure' , 0)
+
+CALL syscs_util.syscs_set_database_property('derby.database.classpath', 'APP.ADDNUMSPROCEDURE')
+```
```
aldettinger, line 56: In such case, I wonder whether pointing to a documentation owned by derby would be good ? From time to time, I found us modifying docs for things that under camel-quarkus control.

aldettinger:
It looks a big part of the difficulty comes from the setup of a derby database using a stored proc from a jar.
I wonder whether such a setup could be already present in an already built docker image ? Maybe derby testers would already have such kind of setup present in a docker image ?

In case case, this is good to see this path being dug. Well done for that @JiriOndrusek :+1:

integration-tests/sql/README.adoc
JiriOndrusek: I agree that linkto some doc would help us to avoid modyfication od this readme. But on the other hand, the same code is used in docker initialization in this change. So we would need to keep it updated either way. (But of course in case that there is no better way of initialization of stored procedure in derby, it is still worth of investigation)

JiriOndrusek:
Ufortunately even if quarkus issue is fixed, test still fails in native. See this comment for a stacktrace - https://github.com/quarkusio/quarkus/issues/22058#issuecomment-989825112

I'll keep digging into this.

integration-tests/sql/README.adoc
aldettinger: Indeed, it's clear now that this is the itest doc and not the extension doc. So, yes I also think we need to keep this.

zbendhiba:
@JiriOndrusek  I'm not sure if I understand this PR correctly. Why do we need this external jar for native mode ?
Is it for a Test Resource needs, or is it required for the code of the app ?

JiriOndrusek:
@zbendhiba  I'll try to explain. From what I remember:

1. We want to add a test for derby against external db.
2. External db is started by the test via docker image.
3. To be able to execute stored procedure in the derby db, jar with the implementation of that procedure has to be uploaded into  the db.
```
CALLLL sqlj.install_jar('/PATH_TO_JAR/camel-quarkus-integration-test-sql-derby-stored-procedure-*.jar', 'AddNumsProcedure' , 0)

CALL syscs_util.syscs_set_database_property('derby.database.classpath', 'APP.ADDNUMSPROCEDURE')
```

(In case of the dev service, derby runs with the same class loader and there is no need to upload this procedure. It is there because it is a dependecy of the module.)
My answer is that we need this jar for both JVM and native and it is used by the test resource to initialize derby (started via docker) to contain our stored procedure.

zbendhiba:
Maybe I'm wrong. Can you achieve this using it only  at QuarkusTestResourceLifecycleManager level? Or is the problem come from the fact of the existence of a library in the pom.xml file ?
Or can we push the jar inside the Docker image using docker compose?
My point is that something needed just for TestResource should not be included in the native build. Unless I'm wrong about how  QuarkusTestResourceLifecycleManager works

JiriOndrusek:
@zbendhiba your suggestion sounds good. I used this way and didn't think that it is against best practices in native. If it is possible to put a jar inside docker image, it would be solution.
Current fix uploads a jar into docker container
```
  container = new FixedHostPortGenericContainer("az82/docker-derby")
                    .withFixedExposedPort(SqlHelper.getDerbyDockerPort(), 1527)
                    .withCopyFileToContainer(
                            MountableFile.forClasspathResource("derby/" + jars[0].getName()),
                            "/dbs/storedProcedure.jar")
                    .waitingFor(Wait.forListeningPort());
```

Do I understand correctly, that having jar with stored procedure among dependencies cause quarkus in native to inspect and process it, 
Maybe I can use a current solution, where the jar won't be in dependency, but will be copied (or created) in this module via maven plugin. Would it help?

zbendhiba:
> @zbendhiba your suggestion sounds good. I used this way and didn't think that it is against best practices in native. If it is possible to put a jar inside docker image, it would be solution. Current fix uploads a jar into docker container
> 
> ```
>   container = new FixedHostPortGenericContainer("az82/docker-derby")
>                     .withFixedExposedPort(SqlHelper.getDerbyDockerPort(), 1527)
>                     .withCopyFileToContainer(
>                             MountableFile.forClasspathResource("derby/" + jars[0].getName()),
>                             "/dbs/storedProcedure.jar")
>                     .waitingFor(Wait.forListeningPort());
> ```
> 
> Do I understand correctly, that having jar with stored procedure among dependencies cause quarkus in native to inspect and process it, Maybe I can use a current solution, where the jar won't be in dependency, but will be copied (or created) in this module via maven plugin. Would it help?

No I think it's fine. I was looking at related issues, and was thinking we have a native error because of this.

Thanks for your answer

jamesnetherton:
You'll want to rebase and change the target branch to `main` since Quarkus has been bumped to 2.7.x.

JiriOndrusek:
I rebased this PR to `main`. I can locally run tests with external derby via docker in **JVM**  mode Native mode fails with 
```
ERROR [io.qua.run.Application] (main) Failed to start application (with profile prod): ERROR 08006: Insufficient data while reading from the network - expected a minimum of 6 bytes and received only 0 bytes.  The connection has been terminated.
```

But native mode is not expected to be successful, see https://github.com/quarkusio/quarkus/issues/22058#issuecomment-989714178

JiriOndrusek:
Derby dev service doesn't work because of quarkus error - https://github.com/quarkusio/quarkus/issues/23083

JiriOndrusek:
@jamesnetherton The CI build fails with:
```
Run [[ -z $(git status --porcelain | grep -v antora.yml) ]] || { echo 'There are uncommitted changes'; git status; exit 1; }
There are uncommitted changes
HEAD detached at pull/3370/merge
	modified:   pom.xml
```

But if I build the branch locally (even without `-Dquickly`) the is no modified file. Do you know what could be wrong?

jamesnetherton:
Do you need to have the target branch as `quarkus-main`? `main` has been updated to Quarkus 2.7.x.

JiriOndrusek:
Thanks, this was the mistake, I rebase it on git repo but not on this PR. thanks a lot.

JiriOndrusek:
I slightly reworked and retested derby tests. By default derby dev service is used,
For external db testing -> it is possible to use docker image with the test (by simple `export SQL_USE_DERBY_DOCKER=true` or use external db , which requires several manual configurations because of stored procedure (described in README.adoc). PR can be merged.

jamesnetherton:
@JiriOndrusek can you update `camel-quarkus-integration-tests-support-sql-derby` to 2.17.0-SNAPSHOT and we can merge this.

JiriOndrusek:
Done, can be merged.

Issue #3302, opened 2021-11-16
Better control what we manage in our BOM


ppalaga:
Issues with the current setup:

1. Third party BOMs like Azure or Debezium may import tons of stuff none of our extensions needs
2. Two third party BOMs (e.g. Azure vs. Debezium) may both partly manage some group (e.g. io.dropwizard.metrics) in different versions which may lead to inconsitencies and incompatibilities in our own effective BOM
3. Our own entries in the BOM may become stale (no extension requires them anymore)

Perhaps we should have a tool that would be able to check whether the effective BOM is consistent and whether each of the entries is required by at least one of our extensions. We may want to flatten the BOM and remove any unneeded entries.

ppalaga:
This came up as a result of the discussion in https://github.com/quarkusio/quarkus-platform/pull/371/files#r749595546

ppalaga:
4. We should get more control over the transitive dependencies. We should probably manage as much as possible, perhaps on cost of removing some extensions that are hard to adapt. E.g. our extensions are currently pulling as much as six different versions of `org.xerial.snappy:snappy-java`. It could probably be one by managing it properly in our BOM.

ppalaga:
4.1: We must manage all `org.apache.camel:camel-*` transitives. We currently miss `org.apache.camel:camel-cluster` and `org.apache.camel:camel-http-base`

jamesnetherton:
In general I see a lot of the BOM imports as being quite problematic. Especially since the ones for the third party extensions import `quarkus-bom`.

Currently we have this mix of `quarkus-bom` versions being resolved in the build:

1.3.3.Final
2.0.0.Final
2.1.0.Final
2.2.3.Final
2.5.0.Final
2.5.1.Final
2.6.0.CR1

The main issue is that if something is removed from `quarkus-bom`, it ends up being silently added back again by one of the older BOM versions. Thus giving the potential for messing up our dependency management.

ppalaga:
> Currently we have this mix of `quarkus-bom` versions being resolved in the build:
> 
> 1.3.3.Final 2.0.0.Final 2.1.0.Final 2.2.3.Final 2.5.0.Final 2.5.1.Final 2.6.0.CR1

I think all Quarkus Platform participants (including us) should stop importing Quarkus BOM into their BOMs. That would solve some problems on our side and also in the Platform. Have you ever thought of such a policy @aloubyansky?

aloubyansky:
[...]

PR #3371, opened 2021-12-09
Ensure consistent version of `software.amazon.awssdk` dependencies

jamesnetherton:
Since the move of Quarkus Amazon Services to Quarkiverse & given the dependency management issues I mentioned [here](https://github.com/apache/camel-quarkus/issues/3302#issuecomment-989648677) and [here](https://camel.zulipchat.com/#narrow/stream/257302-camel-quarkus/topic/BOM.20imports.20for.203rd.20party.20extensions), I don't see a better way of ensuring we are using a consistent & known version of the AWS dependencies.

Alternatively, I could move [this BOM import](https://github.com/apache/camel-quarkus/blob/main/poms/bom-test/pom.xml#L50-L56) into `camel-quarkus-bom`. But `quarkus-amazon-services-bom` also includes the Quarkus AWS extensions, and I don't think we want those being imported into our core BOM. Hence I opted to import the plain `software.amazon.awssdk` BOM.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
         <animal-sniffer.version>1.18</animal-sniffer.version>
         <antlr3.version>3.5.2</antlr3.version><!-- Spark, Stringtemplate and probably others -->
         <avro.version>1.11.0</avro.version><!-- @sync io.quarkus:quarkus-bom:${quarkus.version} dep:org.apache.avro:avro -->
-        <awssdk.version>2.17.90</awssdk.version><!-- TODO: Remove this https://github.com/apache/camel-quarkus/issues/3366 -->
+        <awssdk.version>2.17.90</awssdk.version><!-- @sync io.quarkiverse.amazonservices:quarkus-amazon-services-parent:${quarkiverse-amazonservices.version} prop:awssdk.version -->
         <aws-java-sdk.version>1.11.714</aws-java-sdk.version>
         <azure-sdk-bom.version>1.0.5</azure-sdk-bom.version><!-- Keep in sync with camel-azure component versions -->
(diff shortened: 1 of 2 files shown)


zbendhiba:
@jamesnetherton For my understanding, this imports the software.amazon.awssdk of Quarkiverse only for the tests ?

oscerd (approved):

jamesnetherton:
> @jamesnetherton For my understanding, this imports the software.amazon.awssdk version of Quarkiverse only for the tests ?

No, also for non-test usage too. I.e so that our Camel AWS extensions use a consistent version for all `software.amazon.awssdk` dependencies.

We only import `quarkus-amazon-services-bom` in `camel-quarkus-bom-test` because we have [these aws2-quarkus-client tests](https://github.com/apache/camel-quarkus/tree/main/integration-test-groups/aws2-quarkus-client), and they need access to the Quarkiverse AWS extensions.

Hope that makes sense.

zbendhiba:
> > @jamesnetherton For my understanding, this imports the software.amazon.awssdk version of Quarkiverse only for the tests ?
> 
> No, also for non-test usage too. I.e so that our Camel AWS extensions use a consistent version for all `software.amazon.awssdk` dependencies.
> 
> We only import `quarkus-amazon-services-bom` in `camel-quarkus-bom-test` because we have [these aws2-quarkus-client tests](https://github.com/apache/camel-quarkus/tree/main/integration-test-groups/aws2-quarkus-client), and they need access to the Quarkiverse AWS extensions.
> 
> Hope that makes sense.

ok, I think what disturbs me is the comment to sync from Quarkiverse. But I guess, since Amazon services are part of the Quarkus platform, I guess you're right, this is the version to sync from.

zbendhiba (approved):

PR #3378, opened 2021-12-10
Ban log4j-core and log4j-slf4j-impl 

### LINKED ISSUE #3377: Ban log4j-core and log4j-slf4j-impl
There is https://access.redhat.com/security/cve/CVE-2021-44228 in log4j-core 

The main point is that log4j-core is not needed on Quarkus at all.

On Quarkus, jboss logmanager is used as a logging backend.

Apps and libs may choose to use log4j as a logging facade to pass their log messages, but my understanding is that log4j-core is not needed in that case.

log4j-api should be enough.

Hence we could ban & exclude log4j-core as a mitigation of the CVE.

(I assume log4j-core depends on log4j-api, not the other way round)

Does any of our extensions depend on log4j-core currently? yes, but it seem to be only nsq and corda.

ppalaga:
Fix #3377

### DIFF
--- a/extensions-jvm/corda/runtime/pom.xml
+++ b/extensions-jvm/corda/runtime/pom.xml
@@ -52,8 +52,22 @@
             <artifactId>camel-quarkus-core</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.jboss.logmanager</groupId>
+            <artifactId>log4j2-jboss-logmanager</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.camel</groupId>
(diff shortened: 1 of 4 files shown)


oscerd (approved):

zbendhiba (approved):
thanks @ppalaga

aldettinger (approved):

ppalaga:
`camel-quarkus-integration-test-avro-rpc` is passing for me locally on Temurin-17.0.1+12 so I classify the CI faílure as a non related accident.

PR #3385, opened 2021-12-15
POC/WIP local build setup

djencks:
See https://github.com/apache/camel-website/issues/720


aldettinger:
@djencks I don't know the doc part that much and can't run it locally because of setup issues. Yet, let me do my best to review.
One question thought, is there a simple way to double check that this PR does not break the doc ?

docs/pom.xml
```diff
@@ -85,45 +85,45 @@
...
-                                </goals>
-                                <phase>verify</phase>
-                                <configuration>
-                                    <arguments>run checks</arguments>
-                                </configuration>
-                            </execution>
-                            -->
-                        </executions>
-                    </plugin>
+<!--                    <plugin>-->
```
aldettinger, line 88: This whole section is commented now. So, is it relevant to keep it ?

djencks: There's no point in keeping it's current state, but I left it there to help me think about whether there's a plausible way to do a partial docs build as part of the maven build... so far I don't see a good way.

djencks:
> @djencks I don't know the doc part that much and can't run it locally because of setup issues. Yet, let me do my best to review.

Could you elaborate on the setup issues?  The instructions are in my dev list post on local partial builds. If the instructions are unclear or don't work I'd like to know :-)

> One question thought, is there a simple way to double check that this PR does not break the doc ?

I'm not quite sure what you are asking. The only sure way to check is to do a website PR build as described here: https://camel.apache.org/manual/improving-the-documentation.html#_creating_a_documentation_pull_request. Unfortunately this is not simple :-)  Doing a "full local build" as described in the dev list post is almost as good and much simpler.  However, since the changes here don't change anything about the stuff used in the actual website build, they are unlikely to break it.

docs/pom.xml
aldettinger: ok I see.

aldettinger:
@djencks the setup issue on my machine has no link with this specific PR. It's just I can't generate the doc locally independent from this PR.

I was asking a bout a simple way to check the PR. And it looks there is not. So I guess it's ok to merge and see how it's going :)

aldettinger (approved):

Issue #3366, opened 2021-12-08
[Quarkus 2.6.0.Final] Remove `software.amazon.awssdk:apache-client` dependency override 


jamesnetherton:
In Quarkus 2.6.0.CR1 there's an invalid dependency declaration in `quarkus-bom` for the AWS `apache-client`.

So for now, I've worked around it by specifying the dependency in our own BOM.

When 2.6.0.Final is available, we should be able to remove it.

aldettinger:
The workaround evoked in the previous must be [here](https://github.com/apache/camel-quarkus/commit/77cee8366620ca5e774c38080b8cf9b98027a4da).
I guess we would need to revert only the change in `poms/bom/pom.xml`.

ppalaga:
Now I see what is going on: the exclusion of `commons-logging` in the removed apache-client management entry makes the difference. We actually need to ban commons-logging altogether.

Using `mvn quarkus:dependency-tree` instead of `mvn dependency:tree` helped me to figure out what's going on.

PR #3388, opened 2021-12-15
⬆️ Upgrade Quarkus 2.6.0.Final

zbendhiba:


oscerd (approved):

zbendhiba:
I didn't check if versions were updated in the quarkus-bom. I'm upgrading the kotlin version and double-check all the versions we usually sync from the quarkus bom

aldettinger (approved):
Great. It should also open the path for #3366.

ppalaga (approved):
`mvn cq:sync-versions -N -e` does not show any upgrades anymore, so we are good

zbendhiba:
> `mvn cq:sync-versions -N -e` does not show any upgrades anymore, so we are good

yes I need to use this !! thanks

PR #3391, opened 2021-12-15
⏪️ Revert software.amazon.awssdk:apache-client dependency override

### LINKED ISSUE #3366: [Quarkus 2.6.0.Final] Remove `software.amazon.awssdk:apache-client` dependency override

zbendhiba:
Fixes #3366

### DIFF
--- a/poms/bom/pom.xml
+++ b/poms/bom/pom.xml
@@ -6055,18 +6055,6 @@
                 <artifactId>json-smart</artifactId>
                 <version>${json-smart.version}</version>
             </dependency>
-            <!-- TODO: Remove this https://github.com/apache/camel-quarkus/issues/3366 -->
-            <dependency>
-                <groupId>software.amazon.awssdk</groupId>
-                <artifactId>apache-client</artifactId>
-                <version>${awssdk.version}</version>
-                <exclusions>
-                    <exclusion>
-                        <groupId>commons-logging</groupId>
-                        <artifactId>commons-logging</artifactId>
-                    </exclusion>
-                </exclusions>
-            </dependency>
             <dependency>
                 <groupId>org.antlr</groupId>
                 <artifactId>antlr</artifactId>


oscerd (approved):

poms/bom/pom.xml
```diff
@@ -6055,18 +6048,6 @@
...
-                <groupId>software.amazon.awssdk</groupId>
-                <artifactId>apache-client</artifactId>
-                <version>${awssdk.version}</version>
-                <exclusions>
-                    <exclusion>
-                        <groupId>commons-logging</groupId>
-                        <artifactId>commons-logging</artifactId>
-                    </exclusion>
-                </exclusions>
-            </dependency>
```
ppalaga, line 6069: I think this is the only part we should remove now that we upgrade to Quarkus 2.6. I think `software.amazon.awssdk:bom` in version synced with quarkiverse amazonservices is still needed to keep the versions of software.amazon.awssdk dependencies consistent.

aldettinger: Ok, my note in #3366 is misleading then. Sorry about that.

zbendhiba: ok, let me change this.

zbendhiba: should be fine by now

aldettinger (approved):

ppalaga (approved):
Thanks @zbendhiba!

zbendhiba:
I can't reproduce those errors locally. As it is AWS2 related, it's maybe related to these changes ?

ppalaga:
I can try to reproduce

aldettinger:
`Discovered unresolved method during parsing: org.apache.commons.logging.impl.LogFactoryImpl.handleThrowable(java.lang.Throwable)`
A native build issue, I would expect it to be reproducible.

@zbendhiba Are you building locally with native-image 21.3.0 as CI does ?

ppalaga:
Replaced by https://github.com/apache/camel-quarkus/pull/3406

PR #3430, opened 2022-01-03
Update cron.adoc

erayerdem:
from("cron:tab?schedule=0/1+*+*+*+*+?")
                .setBody().constant("event")
                .log("${body}");  

cron job doesn't works with just cron component dependence .  we must add quarts component for cron component

### DIFF
--- a/docs/modules/ROOT/pages/reference/extensions/cron.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/cron.adoc
@@ -37,4 +37,14 @@ Or add the coordinates to your existing project:
 </dependency>
 ----
 
+[source,xml]
+----
+<dependency>
+      <groupId>org.apache.camel.quarkus</groupId>
+      <artifactId>camel-quarkus-quartz</artifactId>
+    </dependency>
+----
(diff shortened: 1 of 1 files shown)


aldettinger:
Hello @erayerdem and many thanks for stepping in. From the [camel documentation](https://camel.apache.org/components/3.14.x/cron-component.html), the cron component is just a generic interface component. An implementation should be selected, I wonder if it is safe to assume that it will always be quartz ?

Also, note that the doc files are generated. Given that, I wonder whether it would be a good idea to add the `src/main/doc` directly in the cron extension, for instance as it was done in the [xslt extension](https://github.com/apache/camel-quarkus/blob/main/extensions/xslt/runtime/src/main/doc/configuration.adoc). There we could have a doc describing that a cron implementation need to be chosen, for instance quartz ?

aldettinger:
@erayerdem Do you think the doc update is not needed anymore ? If you don't have time, let me know, I could try to write something.

erayerdem:
@aldettinger , thanks for interested . u can write

Issue #3421, opened 2021-12-24
[Quarkus 2.7.0] :  CI : Found Banned Dependency: commons-logging:commons-logging


zbendhiba:
On quarkus-main CI build, enforcer fail for banned artifact commons-logging:commons-logging
I can't reproduce this locally. Can't from dependency:tree neither

```
2021-12-24T10:24:54.6836239Z [INFO] --< org.apache.camel.quarkus:camel-quarkus-integration-test-aws2-grouped >--
2021-12-24T10:24:54.6838378Z [INFO] Building Camel Quarkus :: Integration Tests :: AWS2 Grouped 2.7.0-SNAPSHOT [1078/1375]
2021-12-24T10:24:54.6839724Z [INFO] --------------------------------[ jar ]---------------------------------
2021-12-24T10:24:55.7581887Z [INFO] 
2021-12-24T10:24:55.7584504Z [INFO] --- maven-clean-plugin:3.1.0:clean (default-clean) @ camel-quarkus-integration-test-aws2-grouped ---
2021-12-24T10:24:55.7593236Z [INFO] 
2021-12-24T10:24:55.7596623Z [INFO] --- maven-enforcer-plugin:3.0.0-M3:enforce (camel-quarkus-enforcer-rules) @ camel-quarkus-integration-test-aws2-grouped ---
2021-12-24T10:24:58.0366083Z [WARNING] Rule 3: org.apache.maven.plugins.enforcer.BannedDependencies failed with message:
2021-12-24T10:24:58.0374042Z Found Banned Dependency: commons-logging:commons-logging:jar:1.2
2021-12-24T10:24:58.0381333Z Use 'mvn dependency:tree' to locate the source of the banned dependencies.
```

PR #3438, opened 2022-01-04
Exclude commons-logging from htmlunit-driver

jamesnetherton:
Relates to #3421.


ppalaga (approved):
Looks good. In case we see some ClassNotFoundExceptions, we may need to add org.jboss.logging:commons-logging-jboss-logging to the affected test modules.

jamesnetherton:
Test failures are unrelated to this change...

Issue #3447, opened 2022-01-12
aws2-quarkus-client Verify that no client except quarkus one is used during test


JiriOndrusek:
During investigation of https://github.com/apache/camel-quarkus/issues/3436, I discovered that our sqs test is using sqs client created by camel component in the early part of the test. Which caused that the test for above issue was falsly successful.

I fixed that for sqs tests. It has to be extended to all other *-quarkus-client tests. (I tried to fix all the tests, but it seems, that it causes some errors during execution and needs deeper investigation)

PR #3448, opened 2022-01-12
deadLetterChannel does not work in BaseRouteBuilder configure method 3436

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/3436

Following issue: https://github.com/apache/camel-quarkus/issues/3447

PR #3458, opened 2022-01-13
aws2-quarkus-client Verify that no client except quarkus one is used 

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/3447

PR takes care of the non-sqs tests  (sqs is fixed in https://github.com/apache/camel-quarkus/pull/3448)

- All the tests are enforced to use only quarkus client.
- Error in `aws2-s3` test is fixed (Error is happening only with the correctly configured client from the registry. See https://issues.apache.org/jira/browse/CAMEL-17478)
- `Aws2-s3` is refactored to not duplicate code from the non-quarkus client test.


oscerd (approved):

integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/s3/it/Aws2S3Test.java
```diff
@@ -41,6 +44,9 @@
     private int objects_num_before;
     private int objects_num_after;
 
+    @Aws2Client(LocalStackContainer.Service.S3)
```
aldettinger, line 47: We inject an Aws2Client that is not used elsewhere. Is there a trick to be explained here ?

integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/s3/it/Aws2S3Test.java
```diff
@@ -41,6 +44,9 @@
...
+    S3Client client;
```
aldettinger, line 48: We inject a client without using it directly. Is there a trick to explain ?

JiriOndrusek: It is requirement for client created via quarkus, that the client has to be injected (doesn't need to be used). The injection point is detected by quarkus and client is created. If there is no injection point, client is not created even if all required paameters are present (https://camel.apache.org/camel-quarkus/2.6.x/reference/extensions/aws2-s3.html#_optional_integration_with_quarkus_amazon_s3)

aldettinger (approved):
As far as I get it looks needed indeed. Well done @JiriOndrusek.

There is a single non blocking question if you have time and I think we could merge.

integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/s3/it/Aws2S3Test.java
JiriOndrusek: sorry I missed it from my older attempts, I'll remove this one variable

aldettinger: Ok, I think you are right that it's slated for removal.

On the other side, reading again the doc section pointed in previous comment, I have another question somewhere else in the PR :)

integration-test-groups/aws2-quarkus-client/aws2-s3/src/main/java/org/apache/camel/quarkus/component/aws2/s3/it/Aws2S3QuarkusClientInvoker.java
```diff
@@ -16,9 +16,15 @@
...
 
-@NativeImageTest
-class Aws2S3QuarkusClientIT extends Aws2S3QuarkusClientTest {
+import software.amazon.awssdk.services.s3.S3Client;
+
+@ApplicationScoped
+public class Aws2S3QuarkusClientInvoker {
+
+    @Inject
+    S3Client s3;
```
aldettinger, line 28: So, we inject the S3Client here. But the class is not @Unremovable as stated in the doc https://camel.apache.org/camel-quarkus/2.6.x/reference/extensions/aws2-s3.html#_optional_integration_with_quarkus_amazon_s3.

What do you think ?

JiriOndrusek: To be honest I'm not sure. My guess is that in some cases, quarkus my optimize code and remove unused injections. But our tests does not allow it. So it works without it. (but it may be required with different configuration)

aldettinger: Ok, Quarkus might not detect that this bean is needed to be removed in our test configuration then. Thanks for sharing.

PR #3512, opened 2022-02-02
Upgrade formatter-maven-plugin to 2.17.1

jamesnetherton:


oscerd (approved):

poms/build-parent/pom.xml
```diff
@@ -97,6 +97,7 @@
                     <version>${formatter-maven-plugin.version}</version>
                     <configuration>
                         <configFile>${maven.multiModuleProjectDirectory}/tooling/eclipse-formatter-config.xml</configFile>
+                        <lineEnding>KEEP</lineEnding>
```
ppalaga, line 100: What's the meaning of this change? Would the plugin keep Windows EOLs if end users enter them?

ppalaga: Oh really? Keep is certainly better than user's platform. Is there no way to force Unix EOLs?

jamesnetherton: >  Is there no way to force Unix EOLs?

Yes, I guess it'd make more sense to use `<lineEnding>LF</lindending>`.

jamesnetherton: Latest commit forces Unix EOLs.

PR #3555, opened 2022-02-16
Improve available port discovery in tests

jamesnetherton:


integration-tests-support/test-support/src/main/java/org/apache/camel/quarkus/test/AvailablePortFinder.java
```diff
@@ -103,4 +124,19 @@ private static boolean isQuarkusReservedPort(int port) {
         }
         return false;
     }
+
+    private static String getCallerClassName() {
+        return StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE)
+                .walk(s -> s.map(StackWalker.StackFrame::getClassName)
+                        .filter(className -> !className.equals(AvailablePortFinder.class.getName()))
```
ppalaga, line 131: Shouldn't we make AvailablePortFinder class final, to better match the assumption made on the above line?

jamesnetherton: > Shouldn't we make AvailablePortFinder class final

The class declaration is:

```java
public final class AvailablePortFinder {
}
```

aldettinger: Filtering on className should be enough then.

integration-tests-support/test-support/src/main/java/org/apache/camel/quarkus/test/AvailablePortFinder.java
```diff
@@ -103,4 +124,19 @@ private static boolean isQuarkusReservedPort(int port) {
...
+                        .findFirst()
```
aldettinger, line 132: So `findFirst()` returns the first class directly using AvailablePortFinder ? Would it work when we use AvailablePortFinder 2 times from the same class to reserve 2 distinct ports (if make sense) ?

integration-tests-support/test-support/src/main/java/org/apache/camel/quarkus/test/AvailablePortFinder.java
```diff
@@ -54,15 +57,23 @@ private AvailablePortFinder() {
...
             try (ServerSocket ss = new ServerSocket()) {
                 ss.setReuseAddress(true);
                 ss.bind(new InetSocketAddress((InetAddress) null, 0), 1);
 
                 int port = ss.getLocalPort();
                 if (!isQuarkusReservedPort(port)) {
-                    LOGGER.info("getNextAvailable() -> {}", port);
-                    return port;
+                    String callerClassName = getCallerClassName();
+                    String value = RESERVED_PORTS.putIfAbsent(port, callerClassName);
```
ppalaga, line 72: Is this making some assumptions about whether this class can concurrently exist in multiple JVMs or multiple class loaders within a single maven invocation? IIRC surefire is forking a new JVM for each Maven module. So if building with mvnd, the AvailablePortFinder class may exist in several concurrent JVMs.

integration-tests-support/test-support/src/main/java/org/apache/camel/quarkus/test/AvailablePortFinder.java
ppalaga: > The class declaration is:
> 
> ```java
> public final class AvailablePortFinder {
> }
> ```

I am blind. Sorry for the noise.

integration-tests-support/test-support/src/main/java/org/apache/camel/quarkus/test/AvailablePortFinder.java
aldettinger: Ok, actually it's Map<int,String> so the value would appear 2 times in the map => OK

aldettinger (approved):

integration-tests-support/test-support/src/main/java/org/apache/camel/quarkus/test/AvailablePortFinder.java
jamesnetherton: Can you elaborate a bit more? It's not entirely clear to me why it matters in this context.

All it's doing is trying to track which class reserved a specific port. Arguably we don't need to do that. It just provides some nice logging output and a way to clear out stale entries from the `RESERVED_PORTS` map.

ppalaga: Thanks for explaining. If it is just for logging, indeed, multiple RESERVED_PORTS instances in concurrent per-module JVMs do not matter.

ppalaga (approved):

PR #3568, opened 2022-02-22
Increase azure-storage-blob extension test coverage

jamesnetherton:


oscerd (approved):

integration-test-groups/azure/azure-resources.sh
```diff
@@ -69,6 +70,14 @@ function deleteResources() {
     az group delete --name ${RESOURCE_GROUP} --yes
 }
 
+if ! which az > /dev/null; then
```
aldettinger, line 73: Great idea to check. Does which output on stderr in this case ? Maybe "2>" could be tried.

jamesnetherton: I'll do that so that all output is dumped to `/dev/null`.

integration-test-groups/azure/azure-storage-blob/src/test/java/org/apache/camel/quarkus/component/azure/storage/blob/it/AzureStorageBlobTest.java
```diff
@@ -63,47 +87,443 @@ private static BlobContainerClient blobContainer() {
...
+                    .body(BLOB_CONTENT)
+                    .post("/azure-storage-blob/blob/create")
+                    .then()
+                    .statusCode(201);
+
+            // Download file
+            String downloadPath = RestAssured.get("/azure-storage-blob/blob/download")
+                    .then()
+                    .statusCode(200)
+                    .body(endsWith("target/test"))
```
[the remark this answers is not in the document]
jamesnetherton: Yeah, good spot. It would fail because the REST endpoint returns the absolute file path, which on Windows will naturally have different separators. So it'll need 'nixifying' before being returned and the test assertion can remain as-is.

aldettinger (approved):
Beyond agreed upon comment, this PR looks good. Well done for the great job @jamesnetherton

zbendhiba (approved):

Issue #3571, opened 2022-02-23
jsonpath: JSON keys are unquoted when using writeAsString in native mode


aldettinger:

aldettinger:
Backported to 2.7.x: https://github.com/apache/camel-quarkus/commit/0b222aa6a99b8dee160fe4bff13d895d42bab8e1

PR #3576, opened 2022-02-23
Fix JSON keys are unquoted when using writeAsString in native mode #3571

aldettinger:


aldettinger:
I labelled with backport/2.7.x. Will the backport be automatic or is there something to do manually ?

ppalaga:
> I labelled with backport/2.7.x. Will the backport be automatic or is there something to do manually ?

No, for now the flag serves just forgetful humans.

ppalaga (approved):

PR #3585, opened 2022-03-02
Add test coverage for consuming email with attachments

jamesnetherton:


oscerd (approved):

integration-tests/mail/src/main/java/org/apache/camel/quarkus/component/mail/MockMailbox.java
```diff
@@ -16,40 +16,90 @@
...
 import org.jvnet.mock_javamail.Mailbox;
 
-@Path("/mock/{username}")
+@Path("/mock")
 @ApplicationScoped
 public class MockMailbox {
 
+    @DELETE
+    @Path("/clear")
+    @Produces(MediaType.TEXT_PLAIN)
```
aldettinger, line 48: It is possible to @Produces with a void method ?

jamesnetherton: Nope - too much copy / paste. I'll remove it.

jamesnetherton:
> I wonder if there would be interest in testing with a secure flavor, I mean imaps, pop3s and on.

Yes, in general the extension needs more test coverage to cover functionality for each of the supported mail protocols.

jamesnetherton:
Looks like there's an NPE to resolve (which is not occurring for me locally)....

jamesnetherton:
I have reworked the tests to switch from `mock-javamail` to using a containerized mock mail server. Hopefully this helps to fix things. It also opens the door to testing with additional protocols and with security enabled etc.

jamesnetherton:
Actually after doing 50 test runs there still seems to be some instability. Not sure if it's the way I have composed the tests or some issue in `camel-mail`. I'm going to abandon this for now....

PR #3588, opened 2022-03-02
Manage mvel via cq:sync-versions to allow additional consistency checks

ppalaga:


oscerd (approved):

ppalaga:
The weather test failure seems to have nothing to do with the current change and it is passing for me locally.

Issue #3584, opened 2022-03-02
Intermittent failure of FileTests


jamesnetherton:
Regular offender for CI failures:

```
2022-03-01T12:54:16.4609416Z org.awaitility.core.ConditionTimeoutException: org.apache.camel.quarkus.component.file.it.FileTest expected "Hello1" but was "" within 10 seconds.
2022-03-01T12:54:16.4610283Z 	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:164)
2022-03-01T12:54:16.4611019Z 	at org.awaitility.core.AbstractHamcrestCondition.await(AbstractHamcrestCondition.java:86)
2022-03-01T12:54:16.4612118Z 	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:939)
2022-03-01T12:54:16.4612630Z 	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:645)
2022-03-01T12:54:16.4613190Z 	at org.apache.camel.quarkus.component.file.it.FileTest.idempotent(FileTest.java:137)
2022-03-01T12:54:16.4613758Z 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
2022-03-01T12:54:16.4614376Z 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
```

aldettinger:
The same kind of issue happens in  `FileTest.filter`:

```
org.awaitility.core.ConditionTimeoutException: org.apache.camel.quarkus.component.file.it.FileTest expected "Hello2" but was "" within 10 seconds.
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:164)
	at org.awaitility.core.AbstractHamcrestCondition.await(AbstractHamcrestCondition.java:86)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:939)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:645)
	at org.apache.camel.quarkus.component.file.it.FileTest.filter(FileTest.java:170)
```

After investigations, the race condition could be made more probable with manual changes below:
```
         String result = mockEndpoint.getExchanges().stream().map(e -> e.getIn().getBody(String.class))
                 .collect(Collectors.joining(SEPARATOR));
+        Log.info("Calling getFromMock/" + mockId + " result = " + result.length());
 
[...]

PR #3596, opened 2022-03-07
file: fix a race condition where an exchange is missed due to a call …

aldettinger:
…to mockEndpoint.reset() #3584


oscerd (approved):

zhfeng (approved):
Nice work!

jamesnetherton (approved):

zbendhiba (approved):

PR #3597, opened 2022-03-08
Add Azure Core HTTP Client Vert.x extension

jamesnetherton:
This is mostly based on the existing Azure SDK HTTP Clients.

I tested using the existing set of CQ Azure integration tests with Azurite + the real Azure APIs and everything passes. I also included the Azure SDK test suite of around 280 tests and they also pass.

FWIW - I'm a bit nervous about introducing a custom client. We do not exercise every individual Azure API in our tests. Only those that the Camel components use. So there's some potential for bugs that have gone undetected. Also, some of the Azure extensions are JVM mode only and have no proper tests yet.

### DIFF
--- a/extensions-support/azure-core-http-client-vertx/deployment/src/test/java/org/apache/camel/quarkus/support/azure/core/http/vertx/VertxHttpClientBuilderTests.java
+++ b/extensions-support/azure-core-http-client-vertx/deployment/src/test/java/org/apache/camel/quarkus/support/azure/core/http/vertx/VertxHttpClientBuilderTests.java
@@ -1,0 +1,181 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.support.azure.core.http.vertx;
+
+import java.net.InetSocketAddress;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+
+import com.azure.core.http.HttpClient;
+import com.azure.core.http.HttpMethod;
+import com.azure.core.http.HttpRequest;
+import com.azure.core.http.ProxyOptions;
+import com.azure.core.util.Configuration;
+import com.github.tomakehurst.wiremock.WireMockServer;
+import com.github.tomakehurst.wiremock.client.WireMock;
+import com.github.tomakehurst.wiremock.core.WireMockConfiguration;
+import io.vertx.core.Vertx;
+import io.vertx.ext.web.client.WebClientOptions;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+import reactor.test.StepVerifier;
+
+import static org.apache.camel.quarkus.support.azure.core.http.vertx.VertxHttpClientTestResource.PROXY_PASSWORD;
+import static org.apache.camel.quarkus.support.azure.core.http.vertx.VertxHttpClientTestResource.PROXY_USER;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Tests {@link VertxHttpClientBuilder}.
+ */
+public class VertxHttpClientBuilderTests {
+    private static final String COOKIE_VALIDATOR_PATH = "/cookieValidator";
+    private static final String DEFAULT_PATH = "/default";
+    private static final String DISPATCHER_PATH = "/dispatcher";
+
+    private static final WireMockServer server = new WireMockServer(
+            WireMockConfiguration.options().dynamicPort().disableRequestJournal());
+    private static final Vertx vertx = Vertx.vertx();
+
+    private static String defaultUrl;
+
+    @BeforeAll
+    public static void setupWireMock() {
+        // Mocked endpoint to test building a client with a prebuilt client.
+        server.stubFor(WireMock.get(COOKIE_VALIDATOR_PATH).withCookie("test", WireMock.matching("success"))
+                .willReturn(WireMock.aResponse().withStatus(200)));
+
+        // Mocked endpoint to test building a client with a timeout.
+        server.stubFor(WireMock.get(DEFAULT_PATH).willReturn(WireMock.aResponse().withStatus(200)));
+
+        // Mocked endpoint to test building a client with a dispatcher and uses a delayed response.
+        server.stubFor(WireMock.get(DISPATCHER_PATH).willReturn(WireMock.aResponse().withStatus(200)
+                .withFixedDelay(5000)));
+
+        server.start();
+
+        defaultUrl = "http://localhost:" + server.port() + DEFAULT_PATH;
+    }
+
+    @AfterAll
+    public static void afterAll() throws InterruptedException {
+        if (server.isRunning()) {
+            server.shutdown();
+        }
+        CountDownLatch latch = new CountDownLatch(1);
+        vertx.close(x -> latch.countDown());
+        latch.await();
+    }
+
+    @Test
+    public void buildWithConfigurationNone() {
+        HttpClient client = new VertxHttpClientBuilder(vertx)
+                .configuration(Configuration.NONE)
+                .build();
+        try {
+            StepVerifier.create(client.send(new HttpRequest(HttpMethod.GET, defaultUrl)))
+                    .assertNext(response -> assertEquals(200, response.getStatusCode()))
+                    .verifyComplete();
+        } finally {
+            ((VertxHttpClient) client).close();
+        }
+    }
+
+    @Test
+    public void buildWithDefaultConnectionOptions() {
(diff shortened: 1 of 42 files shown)


oscerd (approved):

ppalaga (approved):
Wow, great work, @jamesnetherton !

@cescoffier would you like to review?

cc @maxandersen @vkasala

extensions-support/azure-core-http-client-vertx/deployment/src/test/java/org/apache/camel/quarkus/support/azure/core/http/vertx/VertxHttpClientBuilderTests.java
```diff
@@ -0,0 +1,164 @@
...
+
+        defaultUrl = "http://localhost:" + server.port() + DEFAULT_PATH;
+    }
+
+    @AfterAll
+    public static void shutdownWireMock() {
+        if (server.isRunning()) {
+            server.shutdown();
+        }
+        vertx.close();
```
cescoffier, line 82: This is async. In general, it's better to wait for the complete termination. 

```java
CoundDownLatch latch = new CoundDownLatch(1);
vertx.close(x -> latch.countDown());
latch.await();
```

extensions-support/azure-core-http-client-vertx/deployment/src/test/java/org/apache/camel/quarkus/support/azure/core/http/vertx/VertxHttpClientBuilderTests.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Test
+    public void buildWithConfigurationNone() {
+        HttpClient client = new VertxHttpClientBuilder(vertx)
```
cescoffier, line 87: It might be beneficial to close the client.
It will be closed automatically when vert.x is closed, but you only close vert.x after all the tests.

extensions-support/azure-core-http-client-vertx/deployment/src/test/java/org/apache/camel/quarkus/support/azure/core/http/vertx/VertxHttpClientHttpClientTests.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @BeforeAll
+    public static void getWireMockServer() {
+        server.start();
+    }
+
+    @AfterAll
+    public static void shutdownWireMockServer() {
+        server.shutdown();
+        vertx.close();
```
cescoffier, line 39: Same comment, it is generally better to wait for the termination.

extensions-support/azure-core-http-client-vertx/runtime/src/main/java/org/apache/camel/quarkus/support/azure/core/http/vertx/BufferedVertxHttpResponse.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private final Buffer body;
+
+    BufferedVertxHttpResponse(HttpRequest request, io.vertx.ext.web.client.HttpResponse response, Buffer body) {
+        super(request, response);
+        this.body = body;
+    }
+
+    @Override
+    public Flux<ByteBuffer> getBody() {
```
cescoffier, line 39: Does the usage of Mono and Flux come from the Azure SDK?

extensions-support/azure-core-http-client-vertx/runtime/src/main/java/org/apache/camel/quarkus/support/azure/core/http/vertx/BufferedVertxHttpResponse.java
```diff
@@ -0,0 +1,72 @@
...
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+public final class BufferedVertxHttpResponse extends VertxHttpAsyncResponse {
+
+    private final Buffer body;
+
+    BufferedVertxHttpResponse(HttpRequest request, io.vertx.ext.web.client.HttpResponse response, Buffer body) {
+        super(request, response);
+        this.body = body;
```
cescoffier, line 35: We can do something better here (not sure it's needed).

We could just get the body handler or the read stream associated with it and returns truly async Mono and Flux. It would only be useful if:

1. the operation needs to be lazy
2. the body size is huge

extensions-support/azure-core-http-client-vertx/runtime/src/main/java/org/apache/camel/quarkus/support/azure/core/http/vertx/BufferedVertxHttpResponse.java
```diff
@@ -0,0 +1,72 @@
...
+        return Mono.defer(() -> {
+            if (this.body == null || this.body.length() == 0) {
+                return Mono.empty();
+            }
+            return Mono.just(this.body.getBytes());
+        });
+    }
+
+    @Override
+    public Mono<InputStream> getBodyAsInputStream() {
```
cescoffier, line 59: That is funny... InputStream is blocking by nature, so returning a Mono of InputStream is rather odd, but possible for sure.

extensions-support/azure-core-http-client-vertx/runtime/src/main/java/org/apache/camel/quarkus/support/azure/core/http/vertx/VertxHttpAsyncResponse.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class VertxHttpAsyncResponse extends VertxHttpResponse {
+
+    VertxHttpAsyncResponse(HttpRequest request, HttpResponse response) {
+        super(request, response);
+    }
+
+    @Override
+    public Flux<ByteBuffer> getBody() {
+        Buffer responseBody = getVertxHttpResponse().bodyAsBuffer();
```
cescoffier, line 35: It might be better to stream the body.

aldettinger (approved):
This is a single extension support for already 6 extensions and it comes with benefits. Well done :+1: 

For tests, as far as we have a good coverage of Camel features, it should be ok. I would not expect users to reuse this HTTP client on their own. The JVM only extensions side of the question is a bit more delicate, they are not tested today, but still this PR can break something without noticing.

cescoffier (changes requested):
I made a few comments, nothing critical.

The main question is whether we want to improve the laziness as, at the moment, it buffers the response and builds Mono/Flux from something already in memory.

extensions-support/azure-core-http-client-vertx/runtime/src/main/java/org/apache/camel/quarkus/support/azure/core/http/vertx/BufferedVertxHttpResponse.java
jamesnetherton: Yes, it's using Project Reactor.

extensions-support/azure-core-http-client-vertx/runtime/src/main/java/org/apache/camel/quarkus/support/azure/core/http/vertx/BufferedVertxHttpResponse.java
jamesnetherton: Originally, I was thinking along the same lines. When I started testing with massive responses (10s of gigabytes) it seems Azure / Azure SDK handles it as `HTTP 206 Partial Content`, so the overall body is read and assembled in chunks. Thus I left things as-is.

> We could just get the body handler or the read stream associated with it

Do you have an example of that?

extensions-support/azure-core-http-client-vertx/runtime/src/main/java/org/apache/camel/quarkus/support/azure/core/http/vertx/BufferedVertxHttpResponse.java
jamesnetherton: Yes, it's handled the same way in the other Azure client implementations.

extensions-support/azure-core-http-client-vertx/runtime/src/main/java/org/apache/camel/quarkus/support/azure/core/http/vertx/BufferedVertxHttpResponse.java
cescoffier: @tsegismont - do you have an example using `pipe`? If not I can try to implement one.

zhfeng (approved):

extensions-support/azure-core-http-client-vertx/runtime/src/main/java/org/apache/camel/quarkus/support/azure/core/http/vertx/BufferedVertxHttpResponse.java
tsegismont: I would not recommend to use pipe with a custom implementation of `WriteStream`. If you face the need to do this, it means that you should probably be using the raw `HttpClient` which is more flexible than the `WebClient` in this area.

One thing you need to know about the `pipe` body codec: it is only when the whole content has been read that the `WebClient.send` callback is invoked. So you can't, for example, inspect the HTTP status code or headers before deciding what to do with the content.

I you choose to go the `HttpClient` route, I think there are examples in the Vert.x docs for consumption of HTTP bodies.

cescoffier: Thanks @tsegismont!

So I believe we are good with the current solution for now. If later we need real streaming we would switch to the http client.

zbendhiba (approved):

JonathanGiles:
Hi all! I'm curious if this was something that should be considered as a contribution back into the main azure-sdk-for-java repo to make it more accessible to more developers? It's your call, but I'm curious if it is worth consideration. Thanks!

cescoffier:
@JonathanGiles We are considering it. We are discussing where the split should be at the moment. We are leaning toward extracting the `azure-core-http-client` runtime code into a separate module that would be contributed to the Azure SDK (or wherever it makes sense).

JonathanGiles:
@cescoffier if you decide to contribute to the Azure SDK for Java repo, I'm happy to help facilitate that. It would naturally fit alongside our other web clients for netty, okhttp, and jdk httpclient.

Issue #3579, opened 2022-02-24
Improve quartz test coverage


JiriOndrusek:
Add tests for:
- use of CronTrigger
- uses SimpleTrigger 
- use of quartz.properties file for configuration
- quart scheduler in jmx
- delayed start of quartz scheduler
- cluster mode
- getting headers from quartz
- specifycation of timezone
- missfire instructions
- use of QuartzScheduledPollCoonsumerScheduler

PR #3602, opened 2022-03-08
3579 Improve quartz test coverage

JiriOndrusek:
Issue: https://github.com/apache/camel-quarkus/issues/3579

Requires https://issues.apache.org/jira/browse/CAMEL-17764 (already fixed) for proper native clustered test, hence `camel-main` branch is sed as the base.

I faced several problems, which may not be resolved in this PR completely:

1. Clustering. Quarkus in their extension forces users to use quarkus's `application.properties` to define clustering. (see https://quarkus.io/guides/quartz) This approach allows to define and configure quartz scheduler and avoid use of `c3p0` library (which is used by quartz clustering capability). To keep an option to define clustering in the quartz itself (using `quartz.properties`)  I  had to add `c3p0` and improve extension to allow native run with this library.

Currently there are a few limitations. 
- Oracle db type won't work in native, because it requires registration for reflection (possible) which references `oracle.sql.BLOB` which I'm not sure how to correctly handle (there is a **todo** in `QuartzProcessor`)
- Quarkus sets property `org.quartz.jobStore.useProperties` to `true` by substitution, so it can not be changed. Which is the reason why this issue has to be fixed in the camel codebase - https://issues.apache.org/jira/browse/CAMEL-17764 (already fixed)

2. I used a fixed port for postgrresql db, because there is no obvious way, how to use variable in `quartz.properties` (where port has to be passed for db configuration, Unfortunately agroal DS can not be used, because the only way how to pass DS int oquartz is by JNDI - quarkus supports only CDI)

3. With this PR users can use quarkus's clustering (tested by quarkus) or Quartz's clustering (tested by this PR).

To summarize:

- c3p0 dependency was added
- clustering by `quartz.properties` is tested
- so far oracle won't work in native
- fixed port is used for postgresql db in the test (I used postresql db, because quarkus is using h2 for testing -> which brings more coverage)


oscerd (approved):

jamesnetherton:
Before I review properly, can this really only work on `camel-main`? If it's possible to have it working with Camel 3.15, lets do this PR against `main` and make any further adjustments once we upgrade to 3.16.

JiriOndrusek:
The fix I rely on was merge a few hours ago. Therefore I think that it is required to be on 3.16 (without the fix, one test will fail) - I can disable it for purpose of `main` branch, if needed.

JiriOndrusek:
> Quartz has [ConnectionProvider](https://www.quartz-scheduler.org/api/2.3.0/org/quartz/utils/ConnectionProvider.html#getConnection--). So we can probably create a QuarkusQuartzConnectionProvider and look up the Agroal DataSource from the CDI BeanManager either by name or type.

I'll look into that option, using Agroal DS would remove a lot of code and limitations.

JiriOndrusek:
> PR needs a rebase on the latest camel-main changes. Then it'll be easier to review.

@jamesnetherton  There is no need to review current code without ConnectionProvider, it should make code much easier. I'll let you know once it is used in this PR (I'll rebase PR with this change). Thanks for your help.

JiriOndrusek:
Replaced by https://github.com/apache/camel-quarkus/pull/3609

Issue #3627, opened 2022-03-16
file: FileTest is flaky


aldettinger:
The build has failed once on [camel-quarkus CI](https://github.com/apache/camel-quarkus/runs/5509250742?check_suite_focus=true) and once in [quarkus-platform CI](https://github.com/quarkusio/quarkus-platform/runs/5569244554?check_suite_focus=true#step:6:24599) with error message as below:

```
Error:  Tests run: 10, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 23.503 s <<< FAILURE! - in org.apache.camel.quarkus.component.file.it.FileTest
Error:  charset  Time elapsed: 10.127 s  <<< ERROR!
org.awaitility.core.ConditionTimeoutException: org.apache.camel.quarkus.component.file.it.FileTest expected "Hello World 你好" but was "" within 10 seconds.
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:164)
	at org.awaitility.core.AbstractHamcrestCondition.await(AbstractHamcrestCondition.java:86)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:939)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:645)
	at org.apache.camel.quarkus.component.file.it.FileTest.charset(FileTest.java:93)
```

The frequency of this build issue seems quite low, still problematic.

There could be a link with #3584 .

aldettinger:
One occurence more: https://github.com/apache/camel-quarkus/runs/5579549052?check_suite_focus=true#step:9:10012

Issue #3511, opened 2022-02-02
CamelTestSupport style of testing 


ppalaga:
Based on the demand heart in https://github.com/apache/camel-quarkus/issues/1981 and elsewhere, we should provide a way for testing in the CamelTestSupport way known from other Camel runtimes.

This may include at least the following:

1. Offer our own CamelTestSupport class  overriding createCamelContext() override, so that end users do not need to copy it over and over again:

```
@QuarkusTest
public class TransactedTest extends CamelTestSupport {
    @Inject
    CamelContext context;

    @Override
    protected CamelContext createCamelContext() throws Exception {
        return this.context;
    }
}
```
 
2. We should have a couple of tests of this kind to both test that it works and for demo purposes. 
3. We should document this kind of testing (including the warning that it would not work in native mode).

aldettinger:
Some users reported that demonstrating injection of MockEndpoint and usage of AdviceWith would be of interest. They see those camel tests as complement for integration tests, not replacement.

PR #3628, opened 2022-03-16
Fix #3432 add a build step to support source location

### LINKED ISSUE #3432: Camel 3.15 - Add build step to capture route source location:line-number
https://issues.apache.org/jira/browse/CAMEL-17420

In quarkus we can have a build step that can capture the source location : line-number for all the DSLs as this can be done during build phase, and therefore has no cost overhead at runtime.

zhfeng:
I assume that the config should be ```camel.quarkus.source-location-enabled = true``` according to

https://camel.apache.org/components/3.15.x/eips/message-history.html#_enabling_source_location_information

### DIFF
--- a/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
+++ b/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelContextProcessor.java
@@ -111,6 +111,20 @@ public class CamelContextProcessor {
     }
 
     /**
+     * Enable source location if camel.quarkus.source-location-enabled=true
+     *
+     * @param recorder the recorder
+     * @param producer producer of context customizer build item
+     */
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep(onlyIf = SourceLocationEnabled.class)
+    public void enableSourceLocation(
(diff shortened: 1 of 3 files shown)


jamesnetherton:
> I assume that the config should be `camel.quarkus.source-location-enabled = true`

There is also `camel.main.sourceLocation​Enabled`, which I assume is the same thing.

poms/bom/pom.xml
```diff
@@ -2508,6 +2508,11 @@
                 <artifactId>camel-telegram</artifactId>
                 <version>${camel.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.apache.camel</groupId>
+                <artifactId>camel-test-junit5</artifactId>
```
jamesnetherton, line 2513: Not sure if we want to have this in the BOM yet given that https://github.com/apache/camel-quarkus/issues/3511 is unresolved. WDYT @ppalaga?

The convention used in these `QuarkusUnitTest` type tests is to create a `RouteBuilder` impl in the test class and it will get picked up automatically by the CQ bootstrap.

zhfeng: OK, I remove it and will take a look at #3511 .

zhfeng:
The failure test is related to #3627 and @jamesnetherton I will merge this PR if you have no more comment.

PR #3639, opened 2022-03-18
Camel quarkus 3532

aldettinger:


jamesnetherton (commented):
I'm ok with it. But, since we're not doing an announcement for 2.8.0-M1 and its not being released on the platform, I don't think we've publicised that these extensions are going to be removed. So we should at least drop a note into the 2.8.0 migration guide.

zhfeng (approved):

aldettinger:
> 

Yes indeed.

Another point, I was also thinking about the 2.7.x branch. Those extensions were also half removed in the [2.7.x branch](https://github.com/apache/camel-quarkus/commit/232c57ef386ab35ed8073bcb656173fc93c5c058).

So, I guess they would disappear in 2.7.1 anyway. And so, this PR would need to be backported in 2.7.x also. But it's like removing extensions from an LTS branch.

aldettinger:
> > this PR would need to be backported in 2.7.x also
> 
> The alternative would be to revert the removal commit on `2.7.x `and reinstate the extensions.

Ok, let's do that. It may come with a price to build stability on the 2.7.x LTS branch. If the price is too high, we could still remove those extensions later on.

Issue #3530, opened 2022-02-07
`FileTest.charset` integration test fails on Windows


jamesnetherton:
The expected UT8 string is being read as `Hello World ??` so the test assertion fails.

aldettinger:
Should be fixed by https://github.com/apache/camel-quarkus/pull/3647. Let's re-open if the charset part prove to be broken again..

PR #3647, opened 2022-03-21
file: Ensure FileTest.charset is fixed under Windows #3530

aldettinger:

### DIFF
--- a/integration-tests/file/src/test/java/org/apache/camel/quarkus/component/file/it/FileTest.java
+++ b/integration-tests/file/src/test/java/org/apache/camel/quarkus/component/file/it/FileTest.java
@@ -37,8 +37,6 @@ import org.apache.camel.quarkus.core.util.FileUtils;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.DisabledOnOs;
-import org.junit.jupiter.api.condition.OS;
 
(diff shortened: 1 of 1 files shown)


aldettinger:
@jamesnetherton Concerning #3530, I have the hope that it is fixed as side of effect of the #3627 fix. Let's see in this PR if that's actually the case.

aldettinger:
Looks there is an issue:
```
org.awaitility.core.ConditionTimeoutException: org.apache.camel.quarkus.component.file.it.FileTest expected "Hello2" but was "" within 10 seconds.
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:164)
	at org.awaitility.core.AbstractHamcrestCondition.await(AbstractHamcrestCondition.java:86)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:939)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:645)
	at org.apache.camel.quarkus.component.file.it.FileTest.filter(FileTest.java:157)
```

Let's have another run to see whether it's flaky or consistent.

aldettinger:
On both run, the charset related test has not failed. We don't see the expected UTF-8 string being read as Hello World ??.
So I would at least merge this one and close the ticket.

Now, those file tests seems to still be flaky. So, yet another ticket should be re-opened for that I think.
@jamesnetherton What do you think ? Maybe you have identified other tasks for the ticket at hand ?

PR #3666, opened 2022-03-26
Better control what we manage in our BOM 

### LINKED ISSUE #3302: Better control what we manage in our BOM

ppalaga:
Fix #3302

This fixes issues 1, 2, 3, 4.1 and 5 of #3302

The best way to to review the effects of the constraints filtering is to compare `poms/bom/src/main/generated/flattened-full-pom.xml` with `poms/bom/src/main/generated/flattened-reduced-verbose-pom.xml`

`flattened-full-pom.xml` is just flattened, whereas `flattened-reduced-verbose-pom.xml` is flattened with superfluous entries removed. 

I am not quite sure about the many FHIR jars, that were eliminated. I wonder whether we should whitelist them?

`poms/bom/src/main/generated/flattened-reduced-pom.xml` is like `poms/bom/src/main/generated/flattened-reduced-verbose-pom.xml` but without item origin comments.


jamesnetherton:
> I am not quite sure about the many FHIR jars, that were eliminated. I wonder whether we should whitelist them?

For now, whitelist them. We can think about making some of them optional in the Camel component.

oscerd (approved):

ppalaga:
853b5b7 :
* FHIR libs are back
* Kafka connect-runtime is back

Probably still not perfect

ppalaga:
Ah the Camel 3.16 staging repo does not exist anymore and the artifacts were still not synced to central :-(

ppalaga:
In between I could perhaps try to test this in 2.7.x branch.

ppalaga:
Camel 3.16 seems to be in Central now.

extensions-support/azure-core-http-client-vertx/deployment/pom.xml
```diff
@@ -101,11 +101,6 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>io.projectreactor</groupId>
-            <artifactId>reactor-test</artifactId>
-            <scope>test</scope>
-        </dependency>
```
ppalaga, line 108: This does not seem to be required. Is it really a leftover from some experiments, @jamesnetherton? 
Otherwise I'd have to find a way how to manage the version.

jamesnetherton: Yes, probably a leftover. The test suite seems to run ok without it.

aldettinger:
While looking at the `poms/bom/src/main/generated/flattened-reduced-pom.xml`, it looks that the version is the same with a given `groupId` except near the end of the file for:
 + com.azure
 + com.microsoft.azure
 + com.datastax.oss
 + com.hazelcast

Is there cases where it is not possible/needed to align ?

ppalaga:
> While looking at the `poms/bom/src/main/generated/flattened-reduced-pom.xml`, it looks that the version is the same with a given `groupId` except near the end of the file for:
> 
> * com.azure
> * com.microsoft.azure

I checked in `flattened-reduced-verbose-pom.xml` where each entry has a comment saying which original bom it comes from. 
And all `com.azure:*` and `com.microsoft.azure:*` come from the same `com.azure:azure-sdk-bom:1.0.5`
So the different versions should be fine.

> * com.datastax.oss

Similar: they are from `com.datastax.oss.quarkus:cassandra-quarkus-bom:1.1.2` and the transitively imported `com.datastax.oss:java-driver-bom:4.14.0`

> * com.hazelcast

Same: `com.hazelcast:quarkus-hazelcast-client-bom:3.0.0`

PR #3676, opened 2022-03-29
Update antora.yml to point at 3.16.x branches

jamesnetherton:
This reverts commit a4664f908041ac11874ec78d8ed3c6c618fb23e9.


zbendhiba (approved):

davsclaus (approved):

oscerd (commented):
Late to the party, but thanks, we can add back gcp secret manager in this way I guess.

PR #3681, opened 2022-03-29
paho: Add test coverage for file persistence #3680

aldettinger:

### DIFF
--- a/integration-tests/paho/src/main/java/org/apache/camel/quarkus/component/paho/PahoResource.java
+++ b/integration-tests/paho/src/main/java/org/apache/camel/quarkus/component/paho/PahoResource.java
@@ -26,6 +26,7 @@ import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
@@ -87,4 +88,11 @@ public class PahoResource {
         return mqex.getMessage();
     }
 
+    @Path("/readThenWriteWithFilePersistenceShouldSucceed")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public String readThenWriteWithFilePersistenceShouldSucceed(@QueryParam("message") String message) throws Exception {
+        producerTemplate.requestBody("paho:withFilePersistence?retained=true&cleanSession=false&persistence=FILE", message);
+        return consumerTemplate.receiveBody("paho:withFilePersistence?cleanSession=false&persistence=FILE", 5000, String.class);
+    }
 }
--- a/extensions/paho/deployment/src/main/java/org/apache/camel/quarkus/component/paho/deployment/PahoProcessor.java
+++ b/extensions/paho/deployment/src/main/java/org/apache/camel/quarkus/component/paho/deployment/PahoProcessor.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.quarkus.component.paho.deployment;
 
+import java.nio.channels.FileChannel;
+import java.nio.channels.FileLock;
 import java.util.ResourceBundle;
 
 import io.quarkus.deployment.annotations.BuildProducer;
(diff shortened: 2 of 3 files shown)


ppalaga (approved):

zhfeng:
@aldettinger Do you think this test could be also for ```paho-mqtt5``` ?

aldettinger:
@zhfeng I've added this test because paho mqtt3 has a class using reflection https://github.com/eclipse/paho.mqtt.java/blob/master/org.eclipse.paho.client.mqttv3/src/main/java/org/eclipse/paho/client/mqttv3/internal/FileLock.java#L33. This class is used when camel is configured with FilePersistence.

So, If there are mqtt5 file classes using reflection in somes camel scenarios, I would say it's worth taking a look.

zhfeng:
Thanks @aldettinger yeah, v5 uses the same FileLock impl https://github.com/eclipse/paho.mqtt.java/blob/master/org.eclipse.paho.mqttv5.client/src/main/java/org/eclipse/paho/mqttv5/client/internal/FileLock.java. I will add one.

integration-tests/paho/src/main/java/org/apache/camel/quarkus/component/paho/PahoResource.java
```diff
@@ -87,4 +88,11 @@ public String mqttExceptionDuringReconnectShouldSucceed() {
         return mqex.getMessage();
     }
 
+    @Path("/readThenWriteWithFilePersistenceShouldSucceed")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public String readThenWriteWithFilePersistenceShouldSucceed(@QueryParam("message") String message) throws Exception {
+        producerTemplate.requestBody("paho:withFilePersistence?retained=true&cleanSession=false&persistence=FILE", message);
```
zhfeng, line 95: if I remove ```cleanSession=false``` , the native test starts to fail but the jvm test is OK. I just wonder if it is a workaround for the native test or it is a limitination?

aldettinger: Indeed, I have the same with mqtt3. In native with cleanSession = true, there is a NullPointerException. So far I've been able to investigate at this stage, it appears that the clientDir is deleted.

The deletion show up only in native mode, but I suspect it is a more general race condition. Perharps even in camel-paho itself. At this stage, I was not able to fully qualify this bug. cleanSession = false allow to test with file persistence without this bug.

zhfeng: yeah, the similar situation with mqtt5 but I have to add ```cleanStart=false```. I took some debug works by adding some ```system.out.println``` and [clientDir.delete()](https://github.com/eclipse/paho.mqtt.java/blob/master/org.eclipse.paho.mqttv5.client/src/main/java/org/eclipse/paho/mqttv5/client/persist/MqttDefaultFilePersistence.java#L298) return ```false``` in jvm mode but ```true``` in native mode.

It might be worth to raise a upstream CAMEL issue track this problem.

aldettinger: Hey, great catch. Would you by chance be able to check whether the `clientDir` is empty before deletion in native and JVM ? It might be a reason to return false.

Tracking this issue is definitely a good idea. Should the issue be opened in camel-quarkus or camel, I don't know. What do you think ?

aldettinger: Ok, that would be a threading issue that we just see more often in native mode then. Really nice @zhfeng :+1:

zhfeng: Ha, intersting ! ```java.nio.file.Files.delete(Path)``` throws ```java.nio.file.DirectoryNotEmptyException```. There is ```.lck``` in  the directory. It looks like the client still holds a ```FileLock``` when clear() method is invoked.

aldettinger: A theroy could be that the lock is not being released (in native mode only then):
https://github.com/eclipse/paho.mqtt.java/blob/master/org.eclipse.paho.mqttv5.client/src/main/java/org/eclipse/paho/mqttv5/client/internal/FileLock.java#L76

Maybe, I messed up the reflection config and `fileLock.getClass().getMethod("release",new Class[]{});` is throwing ?

zhfeng: @aldettinger The root cause is we are missing ```RandomAccessFile``` reflection config which fails to get the file lock.
So adding
```
p.produce(new ReflectiveClassBuildItem(true, false, RandomAccessFile.class));
```
resloves the problem.

aldettinger: Oh, great catch. I take a note to test this with mqtt3 too :)

PR #3699, opened 2022-04-05
Test Azure Blob with autowiredEnabled=false

jamesnetherton:


integration-test-groups/azure/azure-storage-blob/src/main/java/org/apache/camel/quarkus/component/azure/storage/blob/it/AzureStorageBlobProducers.java
```diff
@@ -46,4 +47,11 @@ public BlobServiceClient createBlobClient() throws Exception {
                 .httpLogOptions(new HttpLogOptions().setLogLevel(HttpLogDetailLevel.BODY_AND_HEADERS).setPrettyPrintBody(true))
                 .buildClient();
     }
+
+    @Named("azure-storage-blob-managed-client")
+    public BlobComponent azureBlobComponentWithManagedClient() {
+        BlobComponent component = new BlobComponent();
+        component.setAutowiredEnabled(false);
```
aldettinger, line 54: Out of curiosity. We need both `component.setAutowiredEnabled(false);` on the component and `autowiredEnabled=false` on the endpoint uri ? Or maybe only one of them would be sufficient ?

aldettinger (approved):

integration-test-groups/azure/azure-storage-blob/src/main/java/org/apache/camel/quarkus/component/azure/storage/blob/it/AzureStorageBlobProducers.java
aldettinger: Okay, good to know. Many thanks :+1:

PR #3691, opened 2022-04-01
camel-quarkus-paho: Add reflection config for RandomAccessFile

zhfeng:
follows up https://github.com/apache/camel-quarkus/pull/3681

PR #3700, opened 2022-04-06
camel-quarkus-paho: Add reflection config for RandomAccessFile (#3691)

zhfeng:


ppalaga:
@zhfeng thanks for taking care for the backporting! Just please note that backporting individual PRs in different order than they are merged to main may make it harder to backport other important stuff. So it is better to let the person maintaining the given maintenance branch backport once a week or so thus picking commits in proper order. You may want to label your PR with `backport/2.7.x` label to ensure that the backporter does not skip it. It did not do any harm in this case, I just wanted to share the backporter's view :)

PR #3729, opened 2022-04-14
Fix #3728 AWS S3 integration test should remove all objects in finall…

### LINKED ISSUE #3728: AWS S3 integration test should remove all objects in finally block

zhfeng:
…y block

### DIFF
--- a/integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/s3/it/Aws2S3Test.java
+++ b/integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/s3/it/Aws2S3Test.java
@@ -164,22 +164,24 @@ class Aws2S3Test {
         final String oid = UUID.randomUUID().toString();
         final String blobContent = "Hello KMS " + oid;
 
-        // Create
-        RestAssured.given()
-                .contentType(ContentType.TEXT)
-                .body(blobContent)
-                .post("/aws2/s3/object/" + oid + "?useKms=true")
-                .then()
-                .statusCode(201);
+        try {
+            // Create
+            RestAssured.given()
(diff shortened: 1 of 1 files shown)


oscerd (approved):

integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/s3/it/Aws2S3Test.java
```diff
@@ -164,47 +164,51 @@ public void testKms() throws Exception {
...
-        // Delete
-        deleteObject(oid);
+            // Read
+            RestAssured.get("/aws2/s3/object/" + oid + "?useKms=true")
+                    .then()
+                    .statusCode(200)
+                    .body(is(blobContent));
+        } finally {
+            // Delete
+            deleteObject(oid);
```
essobedo, line 183: Should be done in an AfterEach method instead of a  finally block

zhfeng: The oid is a random generated id and there are some other tests such as ```copyObjectDeleteBucket``` which has a different test logic. There's also a ```@AfterEach``` method to check the objects leaking.

ppalaga: @essobedo thanks for the review! 
We were hit by high AWS bills in the past when tests were either missing cleanup code or when something failed midair and the cleanup code was not invoked. Using try-catch helps to address both: missing cleanup code is easier to spot and existing cleanup code is harder to skip. Thus +1 for this change from my side.

ppalaga (approved):

PR #3709, opened 2022-04-07
Fix #3656 Improve camel-quarkus-paho-mqtt5 test coverage

zhfeng:

PR #3731, opened 2022-04-14
Fix #3730 improve paho-mqtt5 ssl tests

### LINKED ISSUE #3730: Improve paho-mqtt5 SSL test
- close the input stream after copying the .jks file
- use tmp file for client keystore.

zhfeng:

### DIFF
--- a/integration-tests/paho-mqtt5/src/main/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5Resource.java
+++ b/integration-tests/paho-mqtt5/src/main/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5Resource.java
@@ -16,10 +16,12 @@
  */
 package org.apache.camel.quarkus.component.paho.mqtt5.it;
 
+import java.io.File;
 import java.io.InputStream;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.util.concurrent.TimeUnit;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -68,13 +70,25 @@ public class PahoMqtt5Resource {
     public String consumePahoMessage(
             @PathParam("protocol") String protocol,
             @PathParam("queueName") String queueName) {
-        if ("ssl".equals(protocol)) {
-            setKeyStore(keystore, password);
-        }
-        String result = consumerTemplate.receiveBody("paho-mqtt5:" + queueName + "?brokerUrl=" + brokerUrl(protocol), 5000,
-                String.class);
-        if ("ssl".equals(protocol)) {
-            removeKeyStore(keystore);
+        String tmpKeystore = null;
+        String sslClientProps = "";
+        String result;
+
+        try {
+            if ("ssl".equals(protocol)) {
+                tmpKeystore = setKeyStore(keystore);
+                sslClientProps = "&sslClientProps.com.ibm.ssl.keyStore=" + tmpKeystore +
+                        "&sslClientProps.com.ibm.ssl.keyStorePassword=" + password +
+                        "&sslClientProps.com.ibm.ssl.trustStore=" + tmpKeystore +
+                        "&sslClientProps.com.ibm.ssl.trustStorePassword=" + password;
+            }
+            result = consumerTemplate.receiveBody(
+                    "paho-mqtt5:" + queueName + "?brokerUrl=" + brokerUrl(protocol) + sslClientProps, 5000,
+                    String.class);
+        } finally {
+            if ("ssl".equals(protocol) && tmpKeystore != null) {
+                removeKeyStore(tmpKeystore);
+            }
         }
         return result;
     }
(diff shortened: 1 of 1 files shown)


integration-tests/paho-mqtt5/src/main/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5Resource.java
```diff
@@ -68,13 +70,17 @@ public class PahoMqtt5Resource {
...
+            tmpKeystore = setKeyStore(keystore, password);
         }
+
         String result = consumerTemplate.receiveBody("paho-mqtt5:" + queueName + "?brokerUrl=" + brokerUrl(protocol), 5000,
                 String.class);
-        if ("ssl".equals(protocol)) {
-            removeKeyStore(keystore);
+
+        if ("ssl".equals(protocol) && tmpKeystore != null) {
+            removeKeyStore(tmpKeystore);
```
ppalaga, line 83: I still wonder why removing the keyStore is important. Could you please explain the motivation? Would other tests' expectations in this Maven module somehow clash with those system props being set permanently? 
Tests in other modules should not matter because surefire always bootstraps a new JVM for running tests within a module.

If the keystore removal is really important, than we should perhaps enclose receiveBody in a `try` block and call `removeKeyStore()` in `finally` like you do below.

integration-tests/paho-mqtt5/src/main/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5Resource.java
```diff
@@ -159,18 +167,22 @@ private String brokerUrl(String protocol) {
...
+            Files.copy(in, Paths.get(tmpKeystore), StandardCopyOption.REPLACE_EXISTING);
+
+            System.setProperty("javax.net.ssl.keyStore", tmpKeystore);
+            System.setProperty("javax.net.ssl.keyStorePassword", password);
+            System.setProperty("javax.net.ssl.trustStore", tmpKeystore);
+            System.setProperty("javax.net.ssl.trustStorePassword", password);
 
-        try {
-            Files.copy(in, Paths.get(keystore));
         } catch (Exception e) {
```
ppalaga, line 182: I think it would be better to fail with a clear message if we cannot copy.

```suggestion
        } catch (Exception e) {
            throw new RuntimeException("Could not copy "+ keystore +" from the classpath to " + tmpKeystore);
```

integration-tests/paho-mqtt5/src/main/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5Resource.java
zhfeng: Well, @jamesnetherton suggest to unset the keystore properties in https://github.com/apache/camel-quarkus/pull/3709#discussion_r845767127 and I think it is reasonable to clean the environment properties after testing. I'm not very sure that surefire always lauchs a new JVM for each test ? e.g. with ```reuseForks=true```

integration-tests/paho-mqtt5/src/main/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5Resource.java
zhfeng: I wll add this message.

integration-tests/paho-mqtt5/src/main/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5Resource.java
ppalaga: My understanding of reuseForks is that it controls whether JVMs can be reused for individual test classes within the same Maven module. 

To make sure, I have added 

```
System.out.println("=== pid(" + getClass().getSimpleName() + ") " + ManagementFactory.getRuntimeMXBean().getName());
```
to `Base64Test` and `BeanValidatorTest` and I have run 
```
cd integration-tests && mvn clean test -pl :camel-quarkus-integration-test-base64,:camel-quarkus-integration-test-bean-validator
...
=== pid(Base64Test) 111907@terpistone
...
=== pid(BeanValidatorTest) 112141@terpistone
```

So the two tests from two different Maven modules were run in two different JVMs. 

> it is reasonable to clean the environment properties after testing.

+1 on generally wanting to cleanup after the test. The questions in this particular case are:

1. Whether the cleanup does what it is supposed to do
2. Whether there might be any unwanted side effects. 
3. Whether this is a good practice to be followed by end users

For 1., it depends how we define our intention. Removal of the system props themselves is effective for sure, but I doubt that by doing that we also remove the SSL context instantiated somewhere deep in Paho. Should that be our goal? I don't know, probably not. But if we do not remove the SSL context whose state relies on the props, should we remove the props (and the key store file) at all? Maybe better not before the the test app is shut down?

For 2., `javax.net.ssl.*` properties are not owned by us. They are owned by the JVM. There may exist third party code sensitive to the changes. My impression from what I have read about `javax.net.ssl.*` in the past is, that they are intended to be passed to a JVM at startup and that they are not so much intended to be updated at runtime. Will we break any third party code relying on `javax.net.ssl.*` properties? Probably not (otherwise we'd see failures in the test). 

For 3., I'd say our end users should not be setting `javax.net.ssl.*` properties at runtime for reasons named in 2. Removing the key store file and the props may break something in their app. If we keep the code in the test, I think we should at least add a comment saying something like "This may break code relying on `javax.net.ssl.*` properties. Do not do this in production.".
 
Maybe passing the props from a TestResource to the app when it is starting and removing the key store file after the app is shut down from `QuarkusTestResourceLifecycleManager.stop()` would address all these questions?

integration-tests/paho-mqtt5/src/main/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5Resource.java
ppalaga: Sorry, should be `throw new RuntimeException("Could not copy "+ keystore +" from the classpath to " + tmpKeystore, e);` - we do not want to swallow the exception.

integration-tests/paho-mqtt5/src/main/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5Resource.java
zhfeng: Thanks a lot @ppalaga! I revisit the [paho-mqtt5-component.sslClientProps](https://camel.apache.org/components/3.14.x/paho-mqtt5-component.html#_component_option_sslClientProps) and it seems that we can pass with them ```com.ibm.ssl.keyStore``` and ```com.ibm.ssl.keyPassword``` in uri. It should be better to avoid setting ```javax.net.ssl.*``` system properties. I will try it at first.

zhfeng: yeah, it works.

ppalaga (approved):
Great work, thank, @zhfeng !

aldettinger:
Great to see this improvement. Nice jobs @ppalaga and @zhfeng :+1:

PR #3747, opened 2022-04-25
Fix MockEndpoint usage in Infinispan tests

jamesnetherton:
Noticed some of the Infinispan mock endpoint assertions were having no effect, so this change fixes that.


zhfeng (approved):

zbendhiba (approved):

integration-tests/infinispan/src/main/java/org/apache/camel/quarkus/component/infinispan/InfinispanResources.java
```diff
@@ -198,16 +232,24 @@ public void putAllAsync(@QueryParam("component") String component)
...
+        try {
+            IntStream.of(1, 10).forEach(value -> {
+                Map<String, Object> headers = getCommonHeaders(component);
+                headers.put("MessageId", messageId);
+                template.sendBodyAndHeaders(uri, "Message " + value, headers);
+            });
+
+            mockEndpoint.assertIsSatisfied(5000);
+        } finally {
+            mockEndpoint.reset();
```
aldettinger, line 251: Manually calling `reset()` on MockEndpoint generated a race condition in another test. Due to [this](https://github.com/apache/camel/blob/main/components/camel-mock/src/main/java/org/apache/camel/component/mock/MockEndpoint.java#L319). I don't think it applies to the case at hand. I'm just wondering whether it would generally be better practice to have single usage mock endpoint where possible. In this PR, maybe we could have single use endpoint by integrating component name in the mock uri ?

It's not a blocker. Just an idea to share.

aldettinger (approved):

integration-tests/infinispan/src/main/java/org/apache/camel/quarkus/component/infinispan/InfinispanResources.java
jamesnetherton: >  I'm just wondering whether it would generally be better practice to have single usage mock endpoint where possible. In this PR, maybe we could have single use endpoint by integrating component name in the mock uri

Would that maybe not be a good idea at the moment given that this is not fixed till Camel 3.17?

https://issues.apache.org/jira/browse/CAMEL-17948

Just so I understand the problem, is it that `reset` can still be in progress while some other test is trying to do assertions on the mock endpoint?

jamesnetherton: I've tweaked things to `use single use endpoint by integrating component name`. There are some other extension tests like this that I'd like to modify. So makes sense to have a consistent approach.

aldettinger: As far as I understand, I was talking about another situation than CAMEL-17948.
As far as I remember, the situation was that `reset` was called before or maybe during a call to getReceivedExchanges().

Anyway, the implemented solution seems good here. Let's just recall that invoking `mock.reset()` should be done with care or not at all :)

essobedo: Hi, if [CAMEL-17948](https://issues.apache.org/jira/browse/CAMEL-17948) is not clear enough for you, let me know, I will try to clarify it a little bit more

Issue #3758, opened 2022-04-28
paho: add test case for RFC3986 style urls


aldettinger:

aldettinger:
It's present in the 2.9.x branch. So, I've moved to 2.9.0 milestone.

PR #3759, opened 2022-04-28
paho: add test case for RFC3986 style urls #3758

aldettinger:

### DIFF
--- a/integration-tests/paho/src/main/java/org/apache/camel/quarkus/component/paho/PahoResource.java
+++ b/integration-tests/paho/src/main/java/org/apache/camel/quarkus/component/paho/PahoResource.java
@@ -125,6 +125,21 @@ public class PahoResource {
                 String.class);
     }
 
+    @Path("/sendReceiveWithRfc3986AuthorityShouldSucceed")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public String sendReceiveWithRfc3986AuthorityShouldSucceed(@QueryParam("message") String message) {
+
+        // Change the brokerUrl to an RFC3986 form
+        String tcpUrl = ConfigProvider.getConfig().getValue("paho.broker.tcp.url", String.class);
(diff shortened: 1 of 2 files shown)


jamesnetherton (approved):

oscerd (approved):

zbendhiba (approved):

integration-tests/paho/src/main/java/org/apache/camel/quarkus/component/paho/PahoResource.java
```diff
@@ -125,6 +125,21 @@ public String readThenWriteWithFilePersistenceShouldSucceed(@QueryParam("message
...
     }
 
+    @Path("/sendReceiveWithRfc3986AuthorityShouldSucceed")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public String sendReceiveWithRfc3986AuthorityShouldSucceed(@QueryParam("message") String message) {
+
+        // Change the brokerUrl to an RFC3986 form
+        String tcpUrl = ConfigProvider.getConfig().getValue("paho.broker.tcp.url", String.class);
+        tcpUrl = tcpUrl.replaceAll("tcp://([^:]*):(.*)", "tcp://user:password@$1:$2");
```
zhfeng, line 135: ```user``` and ```password``` will be override by ```camel.component.paho.username``` and ```camel.component.paho.password```. I think ```PahoTestResource.java``` returns these two properties. Please check it.

aldettinger: @zhfeng Many thanks for review :+1:  As it stands, the purpose of this test is not to have `user` and `password` actually used. We are using dummy `user` and `password` values because the purpose is just to check that we don't hit [those lines](https://github.com/eclipse/paho.mqtt.java/blob/master/org.eclipse.paho.client.mqttv3/src/main/java/org/eclipse/paho/client/mqttv3/internal/NetworkModuleService.java#L137-L140) when using RFC 3986 style URL.

Does it make more sense now ? What do you think ?

zhfeng (approved):

PR #3777, opened 2022-05-06
Fix verification of AWS S3 download links

jamesnetherton:


integration-test-groups/aws2/aws2-s3/src/test/java/org/apache/camel/quarkus/component/aws2/s3/it/Aws2S3Test.java
```diff
@@ -304,23 +306,12 @@ public void downloadLink() throws Exception {
...
-                        .statusCode(200)
-                        .body(is(blobContent));
-            }
-
+            RestAssured.given()
+                    .port(downloadUri.getPort())
+                    .get(URLDecoder.decode(downloadLink, StandardCharsets.UTF_8))
+                    .then()
+                    .statusCode(200)
+                    .body(is(blobContent));
```
zhfeng, line 314: So the donwload link could be availiable in the localstack since you add ```S3Presigner```?

Issue #3689, opened 2022-03-31
Split camel-quarkus-integration-test-compression into separate zipfile, lzf and zip-deflater tests


ppalaga:
We should create a new group for the new split tests under integration-test-groups and have a grouped test under integration-tests.
Much like we we did with mongodb.

PR #3791, opened 2022-05-16
Split compression related tests into a test group #3689

aldettinger:


jamesnetherton:
Maybe also add an entry for `sortModulesPath` in the root `pom.xml` for the `cq-maven-plugin`.

oscerd (approved):

pom.xml
```diff
@@ -900,6 +901,15 @@
...
                                         </mergePom>
+                                        <mergePom>
+                                            <sourcePoms>
+                                                <directory>${basedir}/integration-test-groups/compression</directory>
+                                                <includes>
+                                                    <include>*/pom.xml</include>
+                                                </includes>
+                                            </sourcePoms>
+                                            <destinationPom>${basedir}/integration-tests/compression-grouped/pom.xml</destinationPom>
+                                        </mergePom>
```
aldettinger, line 912: By similarity, I've also added this. I wonder if it's really useful as I was able to execute the compression grouped tests without this `mergePom` previously ?

Issue #3790, opened 2022-05-16
Improve camel-quarkus-velocity test coverage


JiriOndrusek:
Add tests for:

- Custom Velocity context
- Dynamic templates

aldettinger:
I remember scenarios where velocity is using a lot of reflection. Then, I wonder if it would would be interesting to add more coverage based on camel tests above ?
https://github.com/apache/camel/blob/main/components/camel-velocity/src/test/java/org/apache/camel/component/velocity/VelocityMethodInvokationTest.java
https://github.com/apache/camel/blob/main/components/camel-velocity/src/test/java/org/apache/camel/component/velocity/VelocityBodyAsDomainObjectTest.java

JiriOndrusek:
@aldettinger I can try that. I made the tests according to the documentation. 

Just for your information, there is a lot of coverage already in the repository. 
For example https://github.com/apache/camel-quarkus/blob/main/integration-tests/velocity/src/test/java/org/apache/camel/quarkus/component/velocity/it/VelocityTest.java#L79 and https://github.com/apache/camel-quarkus/blob/main/integration-tests/velocity/src/test/java/org/apache/camel/quarkus/component/velocity/it/VelocityTest.java#L129

(which at first glance should cover both tests you mentioned) 

This PR only adds 2 missing configuration options which were not covered so far.

aldettinger:
@JiriOndrusek Good point. I've double-checked and actually the camel-quarkus `testBodyAsDomainObject` is already covering both. The involved classes just need to be marked as @RegisterForReflection.

I wonder if a cq-velocity specific doc statement would be needed to document that ?

JiriOndrusek:
There is no mention of registering for reflection in the doc - https://github.com/apache/camel-quarkus/blob/main/extensions/velocity/runtime/src/main/doc/configuration.adoc
I'll add a chapter about it and ammend it to this PR.

PR #3793, opened 2022-05-17
Improve camel-quarkus-velocity test coverage #3790

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/3790

### DIFF
--- a/integration-tests/velocity/src/main/java/org/apache/camel/quarkus/component/velocity/it/VelocityResource.java
+++ b/integration-tests/velocity/src/main/java/org/apache/camel/quarkus/component/velocity/it/VelocityResource.java
@@ -32,8 +32,11 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
 import org.apache.camel.Exchange;
+import org.apache.camel.Processor;
 import org.apache.camel.ProducerTemplate;
 import org.apache.camel.component.velocity.VelocityConstants;
+import org.apache.camel.util.CollectionHelper;
+import org.apache.velocity.VelocityContext;
 import org.jboss.logging.Logger;
 
 @Path("/velocity")
@@ -112,6 +115,34 @@ public class VelocityResource {
                 .build();
     }
 
+    @Path("/velocityContext")
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response velocityContext(String msg, @QueryParam("name") String name, @QueryParam("name2") String name2,
+            @QueryParam("item") String item) throws Exception {
+
+        final Exchange ex = producerTemplate.request(
+                "velocity://template/velocityContext.vm?allowTemplateFromHeader=true&allowContextMapAll=true",
+                (Processor) exchange -> {
+                    exchange.getIn().setBody("");
+                    exchange.getIn().setHeader("name", name2);
+                    Map<String, Object> variableMap = new HashMap<>();
+                    variableMap.put("headers", CollectionHelper.mapOf("name", name));
+                    variableMap.put("body", "Monday");
+                    variableMap.put("properties", exchange.getProperties());
+                    VelocityContext velocityContext1 = new VelocityContext(variableMap);
+                    exchange.getIn().setHeader(VelocityConstants.VELOCITY_CONTEXT, velocityContext1);
+                    exchange.setProperty("item", item);
+                });
+
+        return Response
+                .created(new URI("https://camel.apache.org/"))
+                .entity(CollectionHelper.mapOf("headers.name", ex.getMessage().getHeader("name"),
+                        "result", ex.getMessage().getBody(String.class)))
+                .build();
+    }
+
     @Path("/templateViaHeader")
     @POST
     @Consumes(MediaType.TEXT_PLAIN)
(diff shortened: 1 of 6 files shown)


oscerd (approved):

integration-tests/velocity/src/main/resources/application.properties
```diff
@@ -15,4 +15,6 @@
 ## limitations under the License.
 ## ---------------------------------------------------------------------------
 
-quarkus.native.resources.includes = template/*
\ No newline at end of file
+quarkus.native.resources.includes = template/*
+
+quarkus.camel.native.reflection.include-patterns=org.apache.camel.Exchange
```
jamesnetherton, line 20: Is the reflective config for `Exchange` definitely needed? Just curious how we don't need this with the other templating components.

zhfeng: yeah, I agree and is there any issue if this reflective config is not added? `velocityContext` test will be failing?

aldettinger (approved):
I agree with previous comments. Once clarified, I think it will be ok to merge this PR.

Further, I wonder if more tests would be needed ? More details [in the ticket](https://github.com/apache/camel-quarkus/issues/3790#issuecomment-1129984896).

integration-tests/velocity/src/main/resources/application.properties
JiriOndrusek: As you mentioned it, I agree that this should work automatically without registration. (At least from the PoV of user this is internal stuff)

Without registration of the Exchange, the execution fails in native mode, because I'm using exchange as an object in template (see `Dear ${headers.name}. You ordered item ${exchange.properties.item} on ${body}`).

```
VelocityTest.testVelocityContext:157 expected: <
Dear Sheldon. You ordered item Earth 1 on Monday.> but was: <
Dear Sheldon. You ordered item ${exchange.properties.item} on Monday.>
```

I was looking for an automatic registration for reflections of internal classes, but I was able to find only classes for serialization - https://github.com/apache/camel-quarkus/blob/main/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelSerializationProcessor.java#L86

I'd say that the similar mechanism would make sense also for registration for reflection and that the `org.apache.camel.Exchange` should be there. What do you think @jamesnetherton , @zhfeng ?
(I can create separate ticket and work on it)

JiriOndrusek: Other option is to register `org.apache.camel.Exchange` only in velocity extension (in deployment module)

zhfeng: Thanks @JiriOndrusek for explanations and I think it should only be in velocity extension.

jamesnetherton: Probably it's needed because the exchange properties are not handled here:

https://github.com/apache/camel/blob/985ea1061e9d2c1db81c9e6da751258700454978/core/camel-support/src/main/java/org/apache/camel/support/ExchangeHelper.java#L487-L505

In the template, it'd nicer if you could do:

```
Dear ${headers.name}. You ordered item ${properties.item} on ${body}.
```

I.e not have to reference the `Exchange` class. So we should open a Camel ticket and add that capability if possible.

I guess we could live with the workaround in the test until then. I don't really want us to promote reflective access to `Exchange` etc. We already [discourage](https://github.com/apache/camel-quarkus/blob/5127a3c6fa25d4788b2ffddca220545df08cd42e/docs/modules/ROOT/pages/reference/extensions/velocity.adoc#allowcontextmapall-option-in-native-mode) this.

JiriOndrusek: That makes sense, thanks for the suggestion. I'll create an improvement ticket on camel to allow ${properties.item}.
Not sure whether the following feature will make sense, but it would be nice, if camel can handle `${exchange.properties.item}` internally as `${properties.item}`. Because user could use `${exchange.properties.item}`

zhfeng: I think `${properties.item}` does not work. only `headers` and `body` can be used in templates.

JiriOndrusek: If I summarize the whole problem. We suggest to users to not use exchange or context as body  used in templates. Therefore the test should be probably only changed to use different property (to work without the registration of reflection).

Option for ${properties.item) would be a nice improvement on came side.

JiriOndrusek: @jamesnetherton would it make sense to register ${properties} only for camel-quarkus? (I think that making it accessible in that way in camel is does not make sense because there is no limitations with exchange, thus there is no need to implement it `${exchange.properties}` is OK from camel's PoV)

jamesnetherton: > would it make sense to register ${properties} only for camel-quarkus

I think we should avoid that. It should work the same across all of the runtimes. If you access headers without referencing the exchange, then to me it seems reasonable to do the same with properties.

JiriOndrusek: ok, then I'll change tests to not use this and propose an improvement on camel side -> we will see wheter it will be accepted.

zhfeng: is it possible to only register the custom properties and not internal properties associated with the exchange in camel?

JiriOndrusek: I just realized that the whole problem was caused on the test side. The test is putting exchange into context to be accessible, I switched it to properties map and all works without any registration required. I'm sorry for wasting your time for this.

zhfeng: I think if we expose all the properties (both custom and internal), it can not be accepted for security concerns.

integration-tests/velocity/src/main/java/org/apache/camel/quarkus/component/velocity/it/VelocityResource.java
```diff
@@ -112,6 +115,34 @@ public Response bodyAsDomainObject(Person person, @QueryParam("givenName") Strin
...
+
+        final Exchange ex = producerTemplate.request(
+                "velocity://template/velocityContext.vm?allowTemplateFromHeader=true&allowContextMapAll=true",
+                (Processor) exchange -> {
+                    exchange.getIn().setBody("");
+                    exchange.getIn().setHeader("name", name2);
+                    Map<String, Object> variableMap = new HashMap<>();
+                    variableMap.put("headers", CollectionHelper.mapOf("name", name));
+                    variableMap.put("body", "Monday");
+                    variableMap.put("properties", exchange.getProperties());
```
JiriOndrusek, line 133: (Just pointing to the place, where exchange was originally put into context, I switched it for the properties map, therefore now there is no problem)

zhfeng (approved):

Issue #3797, opened 2022-05-19
Registry lookup for overridden `DefaultBean` types does not work


jamesnetherton:
The Quarkus Qpid JMS Extension exposes a `ConnectionFactory` `DefaultBean` which can be overridden by users.

If you try to do this in conjunction with the JMS extension, autowiring of the `ConnectionFactory` fails with:

```
IllegalArgumentException: connectionFactory must be specified
```

And the root cause is that the JMS component cannot figure out which `ConnectionFactory` bean to use:

```
Cannot autowire ConnectionFactory as 2 instances found in registry.
```

It seems that the Camel registry [lookup](https://github.com/apache/camel-quarkus/blob/ddf4d9fd2d05ec302deb95354da4063f8138b50a/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java#L44) for the `ConnectionFactory` returns both the original `DefaultBean` instance and the overridden one.

Not sure if that's a bug or intended behavior from the `BeanManager`. I noticed some alternate lookup methods on `Arc.container` and  `BeanManager` correctly return only the overridden bean.

JiriOndrusek:
This is intended behavior. Quarkus returns `javax.enterprise.inject.spi.BeanManager`. We are calling method [getBeans](https://docs.oracle.com/javaee/7/api/javax/enterprise/inject/spi/BeanManager.html#getBeans-java.lang.String-) from our [lookup](https://github.com/apache/camel-quarkus/blob/ddf4d9fd2d05ec302deb95354da4063f8138b50a/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java#L44).
This method functionality is to return all beans which matches requirements.

```
Return the set of beans which have the given EL name and are available for injection in the module or library containing the class into which the BeanManager was injected or the Java EE component from whose JNDI environment namespace the BeanManager was obtained, according to the rules of EL name resolution.
```

To apply ambiguous dependency resolution (to remove @DefaultBeans), we need to call method [resolve](https://docs.oracle.com/javaee/7/api/javax/enterprise/inject/spi/BeanManager.html#resolve-java.util.Set-) afterwards.

```
Apply the ambiguous dependency resolution rules to a set of [beans](https://docs.oracle.com/javaee/7/api/javax/enterprise/inject/spi/Bean.html).
``` 

[...]

PR #3821, opened 2022-05-30
Registry lookup for overridden DefaultBean types does not work

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/3797

This is work in progress/draft to verify that this change does not affect any existing junit tests.

### DIFF
--- a/integration-test-groups/foundation/bean/src/main/java/org/apache/camel/quarkus/component/bean/BeanResource.java
+++ b/integration-test-groups/foundation/bean/src/main/java/org/apache/camel/quarkus/component/bean/BeanResource.java
@@ -18,6 +18,8 @@ package org.apache.camel.quarkus.component.bean;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
@@ -30,9 +32,11 @@ import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 
+import org.apache.camel.CamelContext;
 import org.apache.camel.Exchange;
 import org.apache.camel.Produce;
 import org.apache.camel.ProducerTemplate;
+import org.apache.camel.quarkus.component.bean.cdi.Producers;
 import org.apache.camel.quarkus.component.bean.model.Employee;
 
 @Path("/bean")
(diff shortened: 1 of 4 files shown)


JiriOndrusek:
Hi @jamesnetherton , can you pls look at this, whether it makes sense to you (it is nut urgent, I'm just tagging you here)

jamesnetherton:
I think it's probably ok. I've been trying to think of scenarios where the way we handle ambiguous beans could cause issues or lead to confusion, but I can't think of any.

Feel free to take the PR out of draft. I'd prefer to have the thoughts of others before merging though.

JiriOndrusek:
Hi @ppalaga, what do you think about this change?

integration-test-groups/foundation/bean/src/main/java/org/apache/camel/quarkus/component/bean/BeanResource.java
```diff
@@ -164,4 +169,18 @@ String awaitFirst(String key) {
...
+    public int withDefaultBeanCount() {
+        return camelContext.getRegistry().findByType(Producers.WithDefaultBeanInstance.class).size();
+    }
+
+    @Path("/withAlternativeBeanCount")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public int withAlternativeBeanCount() {
+        return camelContext.getRegistry().findByType(Producers.WithAlternateBeanInstance.class).size();
+    }
```
ppalaga, line 184: I see that this primarily covers the cause of the original bug report. But wouldn't it be better to make sure that the registry returns the right single bean returned by the right producer method?

ppalaga (commented):
The ambiguity resolution fix looks good. I am just suggesting to be more paranoid in the test.

integration-test-groups/foundation/bean/src/main/java/org/apache/camel/quarkus/component/bean/BeanResource.java
JiriOndrusek: That makes sense. thanks for the point, I'll change it/

JiriOndrusek: - I fixed tests to control also bean instances 
- I added test to cover scenario with 2 beans (without default and alternatives)
- I added test to cover scenario with more beans (from all tests covered by this PR) - 1 of them is default and 1 of them is alternative - to cover back-compatibility on ambiguous failure

ppalaga (approved):
Perfect, thanks!

aldettinger (approved):
The failing CI tests seems unrelated. Nice work @JiriOndrusek.

jamesnetherton (approved):

PR #3855, opened 2022-06-17
Next is 2.11.0

zhfeng:


aldettinger:
@zhfeng Many thanks for handling the release. The first time could seems a bit challenging.
I think this PR reports errors because you would need to run:
```
mvn clean install -DskipTests
```
If I'm right, it would update some file that could be committed.

jamesnetherton (approved):

PR #3867, opened 2022-06-21
Add extension DSL modeline

### LINKED ISSUE #3834: Add extension for camel-dsl-modeline
https://github.com/apache/camel/tree/main/dsl/camel-dsl-modeline

zbendhiba:
Fixes #3834


zbendhiba:
@davsclaus could you review please ?

zbendhiba:
@davsclaus @jamesnetherton I'm trying to get some examples in the project. The lookup of properties doesn't work out of the box.

This[ line of code](https://github.com/apache/camel/blob/ff571a78846e0f292753749a771cd808cea84b65/dsl/camel-dsl-modeline/src/test/java/org/apache/camel/dsl/modeline/ModelineParserTest.java#L108) doesn't get the expected result. 
The reason behind is that the property is not available in the `camel-microprofile-config` propertiesComponent.

Any idea how can I bind those camel-dsl-modeline properties?

davsclaus:
Yes you cannot test it as if that is added to the property - the modeline is in use during route creations and are available for these routes.

Can you try to get the property trait via

```
        PropertiesComponent pc = camelContext.getPropertiesComponent();
        PropertiesSource ps = pc.getPropertiesSource("property");
```

zbendhiba:
> Yes you cannot test it as if that is added to the property - the modeline is in use during route creations and are available for these routes.
> 
> Can you try to get the property trait via
> 
> ```
>         PropertiesComponent pc = camelContext.getPropertiesComponent();
>         PropertiesSource ps = pc.getPropertiesSource("property");
> ```

Nope, but I guess the way I test is not okay.
When you say `the modeline is in use during route creations`. Is having the dependency is enough, or do we need to configure the context before starting the routes ?

zbendhiba:
@davsclaus what makes [this propertiesSource](https://github.com/apache/camel/blob/2245f5c1eb1c63ec0beaaf598cc476bb6038a1ff/dsl/camel-dsl-modeline/src/main/java/org/apache/camel/dsl/modeline/DefaultModelineFactory.java#L64) different from null. I'm missing the part of what creates this one.

davsclaus:
Its loaded via classpath service detection
https://github.com/apache/camel/blob/2245f5c1eb1c63ec0beaaf598cc476bb6038a1ff/dsl/camel-dsl-modeline/src/generated/resources/META-INF/services/org/apache/camel/properties-source-factory

Which is done on startup at
https://github.com/apache/camel/blob/main/core/camel-base/src/main/java/org/apache/camel/component/properties/PropertiesComponent.java#L740

It may be that camel-quarkus need some "build step" code that detect that camel-modeline-dsl is on the classpath and automatic loads the properties source and enlist in the registry.

For what its worth is similar to how many of the other optional services in Camel are auto-detect from classpath, where quarkus has some build time optimizations.

zbendhiba:
@davsclaus thanks for all the explanations. I resolved it by adding the trait via a recorder on Runtime Init.
I'll add more tests, and try native mode, so I'm keeping this one as draft.

zbendhiba:
The PR is ready for reviews. I've added integrations tests and native support

extensions/dsl-modeline/deployment/src/main/java/org/apache/camel/quarkus/component/dsl/modeline/deployment/DslModelineProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+public class DslModelineProcessor {
+    private static final String FEATURE = "camel-dsl-modeline";
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    @BuildStep(onlyIf = NativeBuild.class)
+    void nativeUnsupported() {
```
jamesnetherton, line 38: PR comment states that native support was added but this suggests that there is no native support?

extensions/dsl-modeline/deployment/src/main/java/org/apache/camel/quarkus/component/dsl/modeline/deployment/DslModelineProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+        throw new RuntimeException("The " + FEATURE + " extension is not supported in native mode "
+                + "as loading DSL modeline code at runtime is not supported on GraalVM");
+    }
+
+    @Record(value = ExecutionTime.RUNTIME_INIT)
+    @BuildStep
+    void addPropertyTrait(CamelContextBuildItem camelContextBuildItem, PropertyTraitRecorder propertyTraitRecorder) {
+        RuntimeValue<CamelContext> context = camelContextBuildItem.getCamelContext();
+        propertyTraitRecorder.addPropertyTrait(context);
```
jamesnetherton, line 47: We don't have to do this now, but we can probably optimize this to be `STATIC_INIT`. E.g create the `PropertiesComponent` at build time, configure any additional stuff like traits and then pass it to the `CamelContext` we create.

integration-tests/dsl-modeline/src/test/java/org/apache/camel/quarkus/component/dsl/modeline/it/DslModelineTest.java
```diff
@@ -0,0 +1,195 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@QuarkusTest
+public class DslModelineTest {
+
+    @Test
+    void testModelineSingleDependency() {
+        String line = "// camel-k: dependency=mvn:org.my:application:1.0";
```
jamesnetherton, line 38: We test that the parsing of the modeline bits work correctly, which is great.

But, does stuff like declaring additional dependencies actually work in a real Camel Quarkus app?

extensions/dsl-modeline/deployment/src/main/java/org/apache/camel/quarkus/component/dsl/modeline/deployment/DslModelineProcessor.java
zbendhiba: Oh yes I need to delete this one.
Thanks

extensions/dsl-modeline/deployment/src/main/java/org/apache/camel/quarkus/component/dsl/modeline/deployment/DslModelineProcessor.java
jamesnetherton: Yes `@Record(value = ExecutionTime.STATIC_INIT)`.

But there is probably a bit more work involved to make it work that way, which I why I proposed we could consider doing a follow up issue for it.

zbendhiba: it seems to work for this one

integration-tests/dsl-modeline/src/test/java/org/apache/camel/quarkus/component/dsl/modeline/it/DslModelineTest.java
zbendhiba: I put it to JVM only for now

jamesnetherton: Might be best to tweak the extension metadata so that it's consistent with the other JVM only extensions. E.g add `unlisted: true`. Otherwise it shows up on code.quarkus.io.

zbendhiba: @jamesnetherton  oh yes nice catch. I regenerated this file. The status was also incorrect. Moved to "Preview", as other JVM extensions

PR #3872, opened 2022-06-24
Fix PDF encryption in native mode

jamesnetherton:


aldettinger:
Also, the failing test is not related.

jamesnetherton:
> Looks good to me.
> 
> Out of curiosity `.body(containsString("content to be included in the created pdf document"));` is used because some extra chars are added and `.body(is(...))` does not work ?

Yes, exactly. The existing tests are doing the same thing.

PR #3878, opened 2022-06-27
[2.7.x] Upgrade to Camel 3.14.4

ppalaga:
The last commit needs to be removed/reverted before merging

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -27,7 +27,7 @@ asciidoc:
 
     min-maven-version: 3.8.2 # replace ${min-maven-version}
     target-maven-version: 3.8.4 # replace ${target-maven-version}
-    camel-version: 3.14.3 # replace ${camel.version}
+    camel-version: 3.14.4 # replace ${camel.version}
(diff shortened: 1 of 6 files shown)


aldettinger:
The failure in group-09 seems unrelated.
Note sure about the failure in examples: `java.lang.ClassNotFoundException: com.jcraft.jsch.jce.HMACSHA256ETM`

zhfeng:
@ppalaga @aldettinger I think we need to upgrade `quarkiverse-jsch.version` to `2.0.0` which includes the fix https://github.com/quarkiverse/quarkus-jsch/commit/ee7b71beeadfb7a93ad1265f1b13d6012c3100ff

aldettinger (approved):
Ah, ok should be good now.

zhfeng:
@ppalaga we should still wait for camel `3.14.4` releasing ?

ppalaga:
> we should still wait for camel 3.14.4 releasing ?

We can release (create our staging repo) against the Camel 3.14.4 staging repo, open our vote and release our staging repo only after camel staging repo is released. We did something like that in the past .

PR #3882, opened 2022-06-29
Simplify support for Quartz clustering

jamesnetherton:


integration-tests/quartz-clustered/src/main/java/org/apache/camel/quarkus/component/quartz/it/QuartzClusteredResource.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Inject
+    CamelContext camelContext;
+
+    @javax.enterprise.inject.Produces
+    @Singleton
+    @Named("quartz")
+    public QuartzComponent quartzComponent(Scheduler scheduler) {
+        QuartzComponent component = new QuartzComponent();
+        component.setScheduler(scheduler);
```
zhfeng, line 41: `Scheduler` paramter is not `autowired` in camel-quartz? Maybe we can improve it in camel?

jamesnetherton: I thought about that. But the problem there is Camel Quartz will always be driven by Quarkus Quartz & I don't think we want that as it'd hurt the UX in camel-k / kamelets. Hence I've restricted it just to the clustering scenario as it's a special case.

zhfeng: yeah, that makes sense and is there any way to support `cluster` without quarkus-quartz?

zhfeng (approved):

zhfeng:
@jamesnetherton it's weird that CI does not be triggered?

jamesnetherton:
> @jamesnetherton it's weird that CI does not be triggered?

Yeah I noticed GitHub is not reporting the CI status. The build is running:

https://github.com/apache/camel-quarkus/actions/runs/2582916589

aldettinger (approved):

PR #3892, opened 2022-06-30
Fix #3809 add reflections for all transports

### LINKED ISSUE #3809: Configuration of ActiveMQ broker URL properties may require reflective access
Observation from a stackoverflow.com post:

https://stackoverflow.com/questions/72366070/unable-to-pass-activemq-configuration-in-quarkus-native-build-with-camel-setup

AFAIK we do not test with many (or any) properties set in the broker URL. We are probably missing registering various classes for reflection to make it work properly.

zhfeng:


aldettinger (approved):
Looks good to me.

zhfeng:
The failure tests look unrelated. I will merge it.

Issue #3910, opened 2022-07-08
Improve google-pubsub test coverage


JiriOndrusek:
Add tests for:

-  **PRODUCER**
- Add assert to cover Exchange.GROUPED_EXCHANGE property for grouped exchanges.
 - Cover ordering capabilities.
 - Add assert to cover existence of GooglePubsubConstants.MESSAGE_ID after succsesfull delivery.
 - **CONSUMER**
 - Cover redelivery if message is not acknowledged or the header GooglePubsubConstants.ACK_ID removed.
 - Assert that consumer returns byte[]

 
 - Test rollback/redelivery (ideally with use of GooglePubsubConstants.ACK_DEADLINE)

JiriOndrusek:
There is a problem with `orderingTest`. It fails on CI - reported as https://github.com/apache/camel-quarkus/issues/3944
Test is disabled on CI.

JiriOndrusek:
Another problem (seems like an issue in the test) reported as https://github.com/apache/camel-quarkus/issues/3957.
(there is a workaround in the test, so it does not block merging of the PR)

PR #3919, opened 2022-07-18
Improve google-pubsub test coverage #3910

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/3910

- `Ordering` test is disabled when the real account is used.  Problem has to be inspected in detail, I reported this issue for the Camel, see https://issues.apache.org/jira/browse/CAMEL-18277
- I created a new test-support module for `google-*` extensions, which is similar to `..-aws2-support`. In the future, more google extensions should use the test support module (which has to be extended for different google services) and therefore  should have the same testing capabilities/functionality as the other google extensions
- There is a `readme.adoc` file, which describe how to use the real service for the test execution.


oscerd (approved):

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudTestResource.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Set;
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+import org.apache.camel.quarkus.test.mock.backend.MockBackendUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.PubSubEmulatorContainer;
+import org.testcontainers.utility.DockerImageName;
+
+public class GoogleCloudTestResource implements QuarkusTestResourceLifecycleManager {
```
jamesnetherton, line 35: If the intention is that this module is supposed to be a 'support' utility that can be shared across other Google Cloud extensions, then we should remove any references to PubSub.

integration-tests/google-pubsub/pom.xml
```diff
@@ -39,7 +39,18 @@
...
             <artifactId>quarkus-resteasy</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jsonb</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-mock</artifactId>
```
jamesnetherton, line 48: Add the additional CQ dependencies to the `virtualDependencies` profile.

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubSubRoutes.java
```diff
@@ -0,0 +1,81 @@
...
+        from(ORDERING_DIRECT_IN)
+                .setHeader(ORDERING_KEY, constant("orderkey"))
+                .to("google-pubsub:{{project.id}}:{{google-pubsub.ordering-topic-name}}?messageOrderingEnabled=true&pubsubEndpoint=us-east1-pubsub.googleapis.com:443");
+
+        from(ACK_DIRECT_IN)
+                .to("google-pubsub:{{project.id}}:{{google-pubsub.ack-topic-name}}");
+
+        from("google-pubsub:{{project.id}}:{{google-pubsub.ack-subscription-name}}?synchronousPull=true")
+                .routeId("Fail_Receive").autoStartup(true).process(exchange -> {
+                    if (acKFailing.isFail()) {
```
jamesnetherton, line 57: I guess we don't need the empty `else` block. Also do we really need to `Thead.sleep`?

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudTestResource.java
JiriOndrusek: I understand your comment. But this support extension should take care of the containers related to pubsub (test scope), unfortunatyaly there is no general container to cover all, services (like in aws2 case)

integration-tests/google-pubsub/pom.xml
JiriOndrusek: ok, thanks

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubSubRoutes.java
JiriOndrusek: empty else block is a remnant of my previous code - I'll remove it.

I took the sleep from the camel IT, but it makes sense there. If the exchange fails, redelivery follows. If the fallback is too quick, redelivery may happen too soon and the assert for failure may fail.

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudTestResource.java
JiriOndrusek: ok, makes sense thanks for clarification

zhfeng (approved):

jamesnetherton:
Maybe a flaky test:

```
2022-07-19T11:00:15.5493101Z [ERROR]   GooglePubsubTest.testOrdering:132 1 expectation failed.
2022-07-19T11:00:15.5493460Z Expected status code <201> but was <500>.
```

JiriOndrusek:
I find some suspicious small issues - fixed
and added some logging + test to wait for consumer to stop.
I hope that these small changes will make tests successful.

JiriOndrusek:
I still don't know why, but `ordering test`, as was created in the beginning, fails in CLI.
The original test was sending **6** messages and receives them afterwards. In CLI there was a failure (without any obvious reason) with the 3rd message. Here is a log snippet:

```
022-07-27T13:37:14.9565463Z 2022-07-27 13:37:14,950 DEBUG [org.apa.cam.com.goo.pub.GooglePubsubProducer] (executor-thread-0) uploader thread/id: 126 / 64C35E7D9B618FD-0000000000000008. api call completed.
2022-07-27T13:37:14.9565962Z 2022-07-27 13:37:14,950 TRACE [org.apa.cam.pro.err.RedeliveryErrorHandler] (executor-thread-0) This exchange is not handled or continued so its marked as failed: Exchange[64C35E7D9B618FD-0000000000000008]
2022-07-27T13:37:14.9566895Z 2022-07-27 13:37:14,953 ERROR [org.apa.cam.pro.err.DefaultErrorHandler] (executor-thread-0) Failed delivery for (MessageId: 64C35E7D9B618FD-0000000000000008 on ExchangeId: 64C35E7D9B618FD-0000000000000008). Exhausted after delivery attempt: 1 caught: java.util.concurrent.ExecutionException: java.util.concurrent.CancellationException: Execution cancelled because executing previous runnable failed.
```

Ordering test in Camel Quarkus does not need to test functionality, but it should cover that CQ extension works with the requested configuration. In my POV it is enough to see that **2** messages are delivered during ordering test, because we can state, that it is possible to call ordering capability of the extension. Ordering itself is "executed" by the google pubsub provider.

I reported this "unknown" failure for further investigation as a [ticket](https://github.com/apache/camel-quarkus/issues/3944)

I think that this PR could be merged and the ordering test is covering the ordering feature even with this stripped scenario.

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubTest.java
```diff
@@ -56,6 +73,223 @@ public void jacksonSerializer() {
...
+
+        Assertions.assertTrue(results.contains("body1"));
+        Assertions.assertTrue(results.contains("body2"));
+
     }
+
+    //Disabled on real account because of https://issues.apache.org/jira/browse/CAMEL-18277
+    @DisabledIfEnvironmentVariable(named = "GOOGLE_APPLICATION_CREDENTIALS", matches = ".+")
+    // https://github.com/apache/camel-quarkus/issues/3944
+    @DisabledIfEnvironmentVariable(named = "CI", matches = "true")
```
JiriOndrusek, line 121: `OrderingTest` is disabled on CI. Issue to investoigate further is raised https://github.com/apache/camel-quarkus/issues/3944. PR could be merged.

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudContext.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.ListIterator;
+import java.util.Map;
+import java.util.Optional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GoogleCloudContext {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(GoogleCloudTestResource.class);
```
aldettinger, line 31: Shouldn't we use `GoogleCloudContext .class` to have clearer logs ?

integration-tests/google-pubsub/README.adoc
```diff
@@ -0,0 +1,21 @@
+== Google Pubsub integration tests
+
+=== Local container
+
+Google emulator differs a little bit from the real account.  For example there is a test which passes on local emulator but fails on real account (see an https://issues.apache.org/jira/browse/CAMEL-18277[issue] in Camel project).
+
+Tests are executed with local container, unless `GOOGLE_APPLICATION_CREDENTIALS` and `GOOGLE_PROJECT_ID` properties (both) are set.
```
aldettinger, line 7: Neat: avoiding parenthesis could make it much clearer ? Like `unless both A and B properties are set` ?

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
```diff
@@ -64,7 +82,95 @@ public Response sendPojoToTopic(String fruitName) {
...
+        return consumeEndpoint(subscriptionName, null);
+    }
+
+    @Path("receive/subscriptionOrdering/{subscriptionName}")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public String receiveFromSubscriptionOrdered(@PathParam("subscriptionName") String subscriptionName) throws Exception {
+
+        return consumeEndpoint(subscriptionName,
+                "?messageOrderingEnabled=true&pubsubEndpoint=us-east1-pubsub.googleapis.com:443");
```
aldettinger, line 118: So, this url will never change ?

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubSubCustomizer.java
```diff
@@ -0,0 +1,213 @@
...
+                subscriptionClient.deleteSubscription(subscriptionGrouped.getName());
+                topicClient.deleteTopic(topicGrouped.getName());
+                subscriptionClient.deleteSubscription(subscriptionAck.getName());
+                topicClient.deleteTopic(topicAck.getName());
+                subscriptionClient.deleteSubscription(subscriptionOrdering.getName());
+                topicClient.deleteTopic(topicOrdering.getName());
+                subscriptionClient.deleteSubscription(subscription.getName());
+                topicClient.deleteTopic(topic.getName());
+
+                topicClient.shutdown();
```
aldettinger, line 129: Do you think it would work to `shutdown()` both and then `awaitTermination()` both both at the same time ? Would it be faster anyway ?

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: Yes. See https://cloud.google.com/pubsub/docs/reference/service_apis_overview#service_endpoints

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubSubCustomizer.java
JiriOndrusek: You are right, I'll fix it.

integration-tests/google-pubsub/README.adoc
JiriOndrusek: done

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubSubCustomizer.java
JiriOndrusek: done

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudContext.java
JiriOndrusek: done

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: explained

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudContext.java
```diff
@@ -0,0 +1,97 @@
...
+
+public class GoogleCloudContext {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(GoogleCloudContext.class);
+
+    private final ArrayList<AutoCloseable> closeables = new ArrayList<>();
+    private final Map<String, String> properties = new LinkedHashMap<>();
+    private boolean usingMockBackend;
+
+    public <C> Optional<C> getResource(Class<C> type) {
```
aldettinger, line 37: It the intent of this method to do `getCloseable(Class<C> type)` ? Or maybe `Resource` means something specific in the context of google-test-support ?

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudContext.java
```diff
@@ -0,0 +1,97 @@
...
+            try {
+                c.close();
+            } catch (Exception e) {
+                LOGGER.warn(String.format("Could not close %s", c), e);
+            }
+        }
+    }
+
+    /**
+     * Add a key-value pair to the system properties seen by AWS 2 tests
```
aldettinger, line 68: Do you think we could remove `AWS 2` reference from the doc ?

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudContext.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * @return a read-only view of {@link #properties}
+     */
+    public Map<String, String> getProperties() {
+        return Collections.unmodifiableMap(properties);
+    }
+
+    public ArrayList<AutoCloseable> getCloseables() {
```
aldettinger, line 86: Is this method used somewhere ?

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
aldettinger: Thanks for the link @JiriOndrusek . Any reason why we specifically target us-east-1 ? Why not europe of global for instance ?

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubSubRoutes.java
```diff
@@ -0,0 +1,82 @@
...
+                        Thread.sleep(750);
+                        throw new Exception("fail");
+                    }
+                }).to(ACK_MOCK_RESULT);
+    }
+
+    static class AcKFailing {
+
+        boolean fail;
```
aldettinger, line 68: Could this boolean be changed from both vertx and camel threads ? Should we mark it as volatile ?

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
```diff
@@ -64,7 +82,95 @@ public Response sendPojoToTopic(String fruitName) {
...
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public String mockReceive(@QueryParam("reset") boolean reset, @PathParam("mockName") String mockName)
+            throws Exception {
+        MockEndpoint mock = context.getEndpoint(mockName, MockEndpoint.class);
+
+        List<String> results = mock.getExchanges().stream().map(e -> e.getIn().getBody(String.class))
+                .collect(Collectors.toList());
+        if (reset) {
+            mock.reset();
```
aldettinger, line 145: I vaguely remember that we had some flaky tests when using the `MockEndpoint.reset()` in the past. I wonder if we could use a distinct mock for each test scenario ? Not sure it's feasible in this case.

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubSubCustomizer.java
aldettinger: Great that we do this :+1:

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubTest.java
```diff
@@ -16,18 +16,34 @@
...
 
 @QuarkusTest
-@QuarkusTestResource(GooglePubsubTestResource.class)
+@QuarkusTestResource(GoogleCloudTestResource.class)
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 class GooglePubsubTest {
+    private static final Logger LOG = Logger.getLogger(GooglePubsubTest.class);
 
     @Test
+    @Order(1)
```
aldettinger, line 46: Is it needed to order tests ? What happen if they are executed in the stated order ?

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
```diff
@@ -64,7 +82,95 @@ public Response sendPojoToTopic(String fruitName) {
...
+    public Response mockReceive(boolean fail)
+            throws Exception {
+        acKFailing.setFail(fail);
+        return Response.created(URI.create("https://camel.apache.org")).build();
+    }
+
+    @Path("/stopConsumer")
+    @GET
+    public void stopConsumer() throws Exception {
+        ExecutorService executor = Executors.newSingleThreadExecutor();
```
aldettinger, line 162: Is it really needed to call `consumerTemplate.stop()` in another thread since we'll poll during 60s in the test ?

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: I chose `us-east1-pubsub` to stay consistent with camel integration test - see https://github.com/apache/camel/blob/main/components/camel-google/camel-google-pubsub/src/test/java/org/apache/camel/component/google/pubsub/integration/MessageOrderingIT.java#L46

The condition for pubsub ordering feature is, that all messages have to be published in the same region to ensure ordering. I don't see a problem using global or different endpoint in the test.  @aldettinger  Which one do you suggest?

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudContext.java
JiriOndrusek: This is the code I created in my first draft of the test. It was replaced by the similar technique which is used in aws. Therefore it should be removed, Thanks for noticing!

JiriOndrusek: This method was replaced with the automatic `AutoCloseable` registration  and execution from the `GoogleCloudContext.close`. It will be removed. Thanks for noticing!

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubSubRoutes.java
JiriOndrusek: Good point, thanks!

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: Each test which is using the mock endpoint, uses its own mock endpoint. So no problem here.

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubTest.java
JiriOndrusek: Ordering of the test was one of my attempts to find a reason of CI failures (in case that they are caused by different ordering). Unfortunately ordering didn't help, so I will remove it.

JiriOndrusek: I'm keeping the ordering of the tests, because it is a workaround of the https://github.com/apache/camel-quarkus/issues/3957. I mentioned it also as the comment in the test class.

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: Fixed by setting the timeout of the restAssured ` .config(RestAssuredConfig.config()
                        .httpClient(HttpClientConfig.httpClientConfig().setParam("http.connection.timeout", 60000)))`
Once te timeout is set, there is no need to keep new threadExecutor

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubSubRoutes.java
JiriOndrusek: fixed

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudContext.java
JiriOndrusek: removed

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudContext.java
JiriOndrusek: fixed

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudContext.java
JiriOndrusek: removed. (I refactored the only one place, this method was called from)

JiriOndrusek:
@aldettinger I fixed all your comments. I kept them `not-resolved`, to be able to find them (there are already some older resolved conversations)

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubTest.java
aldettinger: Okay it's good to disable to CI.

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
aldettinger: The documentation mention 3 reasons for using regional endpoints. In this PR, my understanding is that we deal with small volumes, producers/subscribers in same region, without constraints on data locality. If one agree with this, then it would be tempting to use global so that tests run on a nearby available region. Maybe tests would run quicker and fail less frequently.

That said, it may not be that simple. Maybe we could try global and integrate in the PR only of it works straight forward ? Maybe it's the subject of another ticket ? Or maybe requesting to the global endpoint is not a good idea for some reasons ?

What do you think ?

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudContext.java
aldettinger: Ok, good we have spotted this :+1:

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
```diff
@@ -64,7 +80,84 @@ public Response sendPojoToTopic(String fruitName) {
...
+        if (reset) {
+            mock.reset();
+        }
+        String s = results.stream().collect(Collectors.joining(","));
+        return s;
+    }
+
+    @Path("setFail/")
+    @POST
+    public Response mockReceive(boolean fail)
```
aldettinger, line 151: Maybe we could rename this medhod `public Response setFail(boolean fail)` ?

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
aldettinger: The issue I vaguely remember is more something like in `mockReceive`, `results` is empty, then the camel route thread put a message in the mock and then reset is called from the vertx thread. In such a case `Message 1` would not be received within 5 seconds (actually never received).

I "vaguely" remember. Would you think it's worth investigating more right now ? Or maybe put a note ? Or maybe no issue at all ?

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: @aldettinger  You are right about the global, that makes more sense. I tried it and the configuration works without an issue.So I changed the PR to ise global, if you agree please mark this conversation resolved.

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: I refactored the reset use in a way, that when reset is called, there is enforced, that no messages bould be received during that time.

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: done

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
aldettinger: I think the flaky issue is there and the probability of occurrence could be raised to observe it.
For instance, locally, I sleep 4s before resetting:
```
        if (reset) {
            Thread.sleep(4000L);
            mock.reset();
        }
```

And then, in the route, I delay the first message only:
```
static int i = 0;
...
...
...
                }).process(new Processor() {
                    @Override
                    public void process(Exchange exchange) throws Exception {
                        if (i == 0) {
                            Thread.sleep(1000L);
                        }
                        i++;
                    }
                }).to(ACK_MOCK_RESULT);
```

And then, I hit exception below:
```
Assertion condition defined as a org.apache.camel.quarkus.component.google.pubsub.it.GooglePubsubTest 1 expectation failed.
Response body doesn't match expectation.
Expected: is "1"
  Actual: 
 within 5 seconds.
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:985)
	at org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:769)
	at org.apache.camel.quarkus.component.google.pubsub.it.GooglePubsubTest.testAck(GooglePubsubTest.java:228)
...
Caused by: java.util.concurrent.TimeoutException
...
```

aldettinger: Ok, I see it was changed meanwhile. So the the synchronization is ensured at test logic level.

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
aldettinger: In the PR, we still use `us-east1-pubsub.googleapis.com` ? Isn't it regional ? Would public endpoint be `us-east1-pubsub.googleapis.com` ?

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudTestResource.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.ServiceLoader;
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+import org.apache.camel.quarkus.test.mock.backend.MockBackendUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+
+public class GoogleCloudTestResource implements QuarkusTestResourceLifecycleManager {
+    public static final String GOOGLE_EMULATOR_IMAGE = "gcr.io/google.com/cloudsdktool/cloud-sdk:emulators";
```
aldettinger, line 32: Do we have an idea whether 'emulators' is a constant version ? The link [here](https://cloud.google.com/sdk/docs/downloads-docker) would suggest that we could use a constant version by using `:VERSION-emulators` ?

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubTest.java
```diff
@@ -56,6 +74,219 @@ public void jacksonSerializer() {
...
+                .body("6")
+                .post("/google-pubsub/sendToEndpoint")
+                .then()
+                .statusCode(201);
+        LOG.info("All messages were sent");
+
+        await().atMost(5, TimeUnit.SECONDS).until(() -> RestAssured.given()
+                .get("/google-pubsub/receive/subscriptionOrdering/google-pubsub.ordering-subscription-name")
+                .then()
+                .statusCode(200)
```
aldettinger, line 168: Maybe this line is redundant ?

aldettinger: Same for other lines below.

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubTest.java
```diff
@@ -56,6 +74,219 @@ public void jacksonSerializer() {
...
+                Matchers.is("6"));
+        LOG.info("Message \"6\" received.");
+
+    }
+
+    @Test
+    @Order(5)
+    public void testAck() throws Exception {
+        LOG.info("Start of the acking test");
+        //enable ack
```
aldettinger, line 213: The comments in the test are really helpful :+1:

aldettinger:
Checking the feature coverage at `https://camel.apache.org/components/3.18.x/google-pubsub-component.html`:
 + Producers => OK
 + Consumers => OK
 + Message body => OK
 + AUTHENTICATION CONFIGURATION => I'm note sure ?
 + ROLLBACK AND REDELIVERY => I think it is covered by the ACK test ?

JiriOndrusek:
> Checking the feature [coverage](https://camel.apache.org/components/3.18.x/google-pubsub-component.html):
> 
> * Producers => OK
> * Consumers => OK
> * Message body => OK
> * AUTHENTICATION CONFIGURATION => I'm not sure ?
> * ROLLBACK AND REDELIVERY => I think it is covered by the ACK test ?
est ?
yes
> 
> Paying attention to the native test coverage of authentication might be a good investment. For instance, there are some reflections logic used in google's [DefaultCredentialsProvider](https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java#L254-L269). I wonder if users could hit that ?

I didn't check this google source, I agree that this code should be covered more carefully.
I'll revisit authentication case again, because it could be possible, that I removed some coe, which was covering part of this use case during refactoring. (without noticing it is important)

aldettinger:
> > Checking the feature [coverage](https://camel.apache.org/components/3.18.x/google-pubsub-component.html):
> > 
> > * Producers => OK
> > * Consumers => OK
> > * Message body => OK
> > * AUTHENTICATION CONFIGURATION => I'm not sure ?
> > * ROLLBACK AND REDELIVERY => I think it is covered by the ACK test ?
> >   est ?
> 
> yes
> 
> > Paying attention to the native test coverage of authentication might be a good investment. For instance, there are some reflections logic used in google's [DefaultCredentialsProvider](https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java#L254-L269). I wonder if users could hit that ?
> 
> I didn't check this google source, I agree that this code should be covered more carefully. I'll revisit authentication case again, because it could be possible, that I removed some coe, which was covering part of this use case during refactoring. (without noticing it is important)

@JiriOndrusek That could be the subject of another ticket ? What do you think ?

JiriOndrusek:
As @aldettinger  suggested in the above comment, authentication configuration should be covered more in detail to cover all possible scenarios (in native). Because the google authentication module is common in all google-*  extensions, it makes sense to create another ticket covering this functionality. (and to let this PR to be merged). What do you think @jamesnetherton

JiriOndrusek:
I looked into [DefaultCredenbtialProvider](https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java)

There are several ways of authentication for google clouds available, see the [documentation](https://cloud.google.com/docs/authentication#:~:text=You%20need%20to%20pass%20the,using%20the%20GOOGLE_APPLICATION_CREDENTIALS%20environment%20variable.).

There is no limitation listed nor in [camel component doc](https://camel.apache.org/components/3.17.x/google-pubsub-component.html#_authentication_configuration) neither in [camel quarkus extension doc](https://camel.apache.org/camel-quarkus/2.9.x/reference/extensions/google-pubsub.html).

-  Scenario with disabled authentication is covered via emulated tests. 
- Basic scenario with `GoogleCredentials.getApplicationDefault()` using service account (via export of  `GOOGLE_APPLICATION_CREDENTIALS`) is covered by using real account as is described in `RADME.adoc` file in the test module.

Which brings me to several findings:

1. It is not possible to cover all scenarios via emulated tests or via tests with real account. (which uses `Accessing private data on behalf of a service account outside Google Cloud environments`)
2. I think, that we can put into documentation, that only several ways of authentications are supported in native mode -> service account, by exporting `GOOGLE_APPLICATION_CREDENTIALS` 
3. Another option is to manually cover different authentication ways mainly for native mode.
4. Current coverage is acceptable and there is no need to add more sub-scenarios.

I prefer option 2 - to list limitations for native mode.

What do you think? @aldettinger @jamesnetherton @ppalaga

aldettinger:
> 

@JiriOndrusek That's great investigation :+1:  If we take it from camel-google-pusbsub component/endpoint options, then `authenticate` true/fase is covered ? However `serviceAccountKey` is not covered ?

If that's the case, then couldn't we document that `serviceAccountKey` is not yet supported in native mode. And then add support of `serviceAccountKey` in another ticket, maybe with a manual test.

JiriOndrusek:
> > 
> 
> @JiriOndrusek That's great investigation +1 If we take it from camel-google-pusbsub component/endpoint options, then `authenticate` true/fase is covered ? However `serviceAccountKey` is not covered ?
> 
> If that's the case, then couldn't we document that `serviceAccountKey` is not yet supported in native mode. And then add support of `serviceAccountKey` in another ticket, maybe with a manual test.

I reported a ticket for missing authentications - https://github.com/apache/camel-quarkus/issues/3970

JiriOndrusek:
Hi @aldettinger , do you see any remaining work on this PR or may it be approved/merged?

aldettinger:
@JiriOndrusek Yes, there are still 2 unresolved conversations in this PR.

JiriOndrusek:
> @JiriOndrusek Yes, there are still 2 unresolved conversations in this PR.

I'm sorry, I missed them. I'll fix them both, thanks for reminding me

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubTest.java
JiriOndrusek: Yes, duplicated status are there by mistake. Removed.

integration-tests-support/google/src/main/java/org/apache/camel/quarkus/test/support/google/GoogleCloudTestResource.java
JiriOndrusek: You are right. I put a version for the emulator image.

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: There were 2 occurrences of the `us-east1`, I fixed also the second one.

JiriOndrusek:
@aldettinger I went through the all conversations and all are resolved now. (Once again I'm sorry for missing the opened ones)

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
```diff
@@ -38,24 +46,32 @@ public class GooglePubsubResource {
     @Inject
     ConsumerTemplate consumerTemplate;
 
+    @Inject
+    CamelContext context;
+
+    @Inject
+    GooglePubSubRoutes.AcKFailing acKFailing;
+
+    private Future<Boolean> stopResult;
```
aldettinger, line 55: Is this one really needed ?

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
```diff
@@ -64,7 +80,88 @@ public Response sendPojoToTopic(String fruitName) {
...
+            @QueryParam("toEndpoint") String toEndpoint)
+            throws Exception {
+        producerTemplate.sendBody(toEndpoint, message);
+        return Response.created(URI.create("https://camel.apache.org")).build();
+    }
+
+    @Path("/getFromEndpoint")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public String getSchedulerResult(@QueryParam("fromEndpoint") String fromEndpoint) throws Exception {
```
aldettinger, line 99: Maybe `getFromEndpoint` is a better method name ?

integration-tests/google-pubsub/src/test/resources/META-INF/services/org.apache.camel.quarkus.test.support.google.GoogleTestEnvCustomizer
```diff
@@ -0,0 +1 @@
+org.apache.camel.quarkus.component.google.pubsub.it.GooglePubSubCustomizer
```
aldettinger, line 1: Could you please remind why we need this ?

aldettinger: ok, I recall now.

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: removed

integration-tests/google-pubsub/src/main/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubResource.java
JiriOndrusek: renamed

integration-tests/google-pubsub/src/test/resources/META-INF/services/org.apache.camel.quarkus.test.support.google.GoogleTestEnvCustomizer
JiriOndrusek: This is a registration of google-customizer through SPI, which is used to create/delete pubsub resources before/after the test. Implements `googleTestEnvCustomizer.java` from `google-tests-support` module (also part of this PR)

JiriOndrusek:
Work on this PR is done

aldettinger (changes requested):

integration-tests/google-pubsub/src/test/java/org/apache/camel/quarkus/component/google/pubsub/it/GooglePubsubTest.java
```diff
@@ -56,6 +74,220 @@ public void jacksonSerializer() {
...
+                .statusCode(200)
+                .body(Matchers.is("2")));
+
+        LOG.info("Message \"2\" was delivered after acking was enabled.");
+    }
+
+    @Test
+    @Order(6)
+    public void stopConsumerTest() throws Exception {
+        RestAssured.given()
```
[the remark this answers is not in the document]
JiriOndrusek: I added a comment to the follow-up issue to answer you (https://github.com/apache/camel-quarkus/issues/3957#issuecomment-1217698290)

>The problem seems to be somehow connected to the close of the consumer in the native mode.There is a test stopConsumerTest, which is closing the consumer after all the tests (and requires method order to achieve this)

I would keep the method here, because it forces the right closure of the consumer in the native mode. Once https://github.com/apache/camel-quarkus/issues/3957 is solved, this method should be removed.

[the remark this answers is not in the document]
JiriOndrusek: I commented the code of the test method and pointed to the issue https://github.com/apache/camel-quarkus/issues/3957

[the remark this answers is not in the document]
aldettinger: Ok, great many thanks for sharing :+1:

aldettinger (approved):
Looks good to me. I think all tricks and improvements are explained/listed.

Well done @JiriOndrusek, this was quite a PR :+1:

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/3957

JiriOndrusek:
This PR is ready to merge. I'm finishing another PR which depends on this one. Would it be possible to merge this PR? @jamesnetherton  @ppalaga 

(My last change which fixes https://github.com/apache/camel-quarkus/issues/3957 unfortunately contains also rebase to the current `main` branch, therefore the `compare` link shows a lot of affected files. You can see the changed files only in https://github.com/apache/camel-quarkus/commit/c77e6f2e4cb8a9917294fc83171b237ea9137d92

jamesnetherton:
One final question. Are these dependencies needed in the `virtualDependencies` profile?

```
<dependency>
    <artifactId>camel-quarkus-mock</artifactId>
</dependency>
<dependency>
</dependency>
```

PR #3928, opened 2022-07-20
[closes #3927] Improve camel quarkus master integration test

### LINKED ISSUE #3927: Improve camel-quarkus-master integration test
If the clustering wouldn't work, than it could happen that the [test](https://github.com/apache/camel-quarkus/blob/main/integration-tests/master/src/test/java/org/apache/camel/quarkus/component/master/it/MasterTest.java#L37) would write `follower` because both would write to that file all the time and it would be just about when the `follower` is catched by the `readLeaderFile`. 

 

I'm suggesting improvement in a way that both routes would write to different files. And we check the `follower` file is empty and after we kill the leader, checking the file is not empty.

llowinge:

### DIFF
--- a/integration-tests/master/src/test/java/org/apache/camel/quarkus/component/master/it/MasterTest.java
+++ b/integration-tests/master/src/test/java/org/apache/camel/quarkus/component/master/it/MasterTest.java
@@ -27,6 +27,7 @@ import io.quarkus.test.junit.QuarkusTest;
 import io.restassured.RestAssured;
 import org.apache.camel.quarkus.test.support.process.QuarkusProcessExecutor;
 import org.awaitility.Awaitility;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


integration-tests/master/src/test/java/org/apache/camel/quarkus/component/master/it/MasterTest.java
```diff
@@ -28,6 +28,7 @@
 import org.apache.camel.quarkus.test.support.process.QuarkusProcessExecutor;
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.Test;
+import org.wildfly.common.Assert;
```
jamesnetherton, line 31: Please use JUnit assertions.

llowinge: Thanks for great catch ! Fixed

jamesnetherton (approved):

Issue #764, opened 2020-02-26
CXF JAX-WS support


ppalaga:
A.k.a. SOAP client and SOAP server.

Related: 

* https://github.com/quarkusio/quarkus/issues/4005
* https://github.com/quarkusio/quarkus/issues/3479
* https://github.com/quarkusio/quarkus/pull/5538

dufoli:
Soap server is on good way with cxf. I need to check what is the possibilities to add client side too.

famod:
FWIW: `org.apache.cxf.ws.security.trust.STSClient` seems to work out of the box in JVM mode.
Next we are going to try `JaxWsProxyFactoryBean` in our Quarkus project. I hope we'll manage to get it working somehow until an official extension is available.
So a definite +1!

janstey:
Seems there is a CXF extension in progress in the quarkiverse https://github.com/quarkiverse/quarkiverse-cxf

grigoni:
it would really help me to implement SOAP services, I hope the extension comes soon
thanks to all who contributes to it

famod:
@grigoni If you don't need something that is tightly integrated into camel(-quarkus) then you can just use https://github.com/quarkiverse/quarkiverse-cxf

grigoni:
thx @famod but I'm really looking for camel integrations

aymenfurter:
Thanks to everyone contributing to this cause. I would love to use camel-cxf on quarkus aswell.

dufoli:
hello,

I can work on camel-cxf... I see that a soap folder already exist on camel. Someone can point me any documentation to help on that ?

ppalaga:
> I can work on camel-cxf... I see that a soap folder already exist on camel. Someone can point me any documentation to help on that ?

That would be highly welcome, here is is the guide: https://camel.apache.org/camel-quarkus/latest/contributor-guide/create-new-extension.html . You'd start with 

```
$ cd camel-quarkus
$ mvn cq:create -N -Dcq.artifactIdBase=cxf
```

And then go on to adding tests to `integration-tests/cxf` and fixing the issues until the tests pass in both JVM and native mode.

dufoli:
not working on windows.... and it will take too much time to make the build working on windows...

ppalaga:
Ups, sorry about that. Could you please report the issues your have? We'd be definitely interested in making the Windows builds work.

dufoli:
@ppalaga  
[...]

PR #3931, opened 2022-07-20
Camel Quarkus CXF Extension

javaduke:
This PR adds Camel Quarkus CXF Extension, addressing the issue #764 . Both SOAP client and SOAP server supported, in JVM and native modes.


extensions/qute/component/src/generated/resources/org/apache/camel/component/qute/qute.json
```diff
@@ -11,7 +11,7 @@
     "supportLevel": "Preview",
     "groupId": "org.apache.camel.quarkus",
     "artifactId": "camel-quarkus-qute-component",
-    "version": "2.10.0-SNAPSHOT",
+    "version": "2.11.0-SNAPSHOT",
```
zhfeng, line 14: Please rebase it with upstream.

integration-tests/google-bigquery/pom.xml
```diff
@@ -107,20 +107,6 @@
...
-                    <version>${project.version}</version>
-                    <type>pom</type>
-                    <scope>test</scope>
-                    <exclusions>
-                        <exclusion>
-                            <groupId>*</groupId>
-                            <artifactId>*</artifactId>
-                        </exclusion>
-                    </exclusions>
-                </dependency>
```
zhfeng, line 123: Please rebase

poms/bom/pom.xml
```diff
@@ -1844,6 +1930,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>org.apache.camel</groupId>
+                <artifactId>camel-directvm</artifactId>
+                <version>${camel.version}</version>
+            </dependency>
```
zhfeng, line 1937: Why you need this component?

extensions/qute/component/src/generated/resources/org/apache/camel/component/qute/qute.json
javaduke: done

poms/bom/pom.xml
zhfeng: @ppalaga can you take a look? maybe we can exclude it from `camel-cxf-soap`?

jamesnetherton: It's needed because of this:

https://github.com/apache/camel/blob/main/components/camel-cxf/camel-cxf-common/pom.xml#L104

It should be `test` scope.

extensions/cxf-soap/runtime/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-cxf-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-cxf-transport</artifactId>
+        </dependency>
+
+        <!-- CXF WS-* dependencies -->
```
jamesnetherton, line 58: Not sure if these should be direct dependencies of the extension or not.

I guess they're just here for convenience so that if somebody wants to use them, they are automatically available. If that's the case, I think it'd be better moving them to the BOM, so that folks can add them manually to their project if required.

extensions/cxf-soap/runtime/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+            <artifactId>cxf-rt-ws-security</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
+            <artifactId>cxf-rt-ws-policy</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
+            <artifactId>cxf-rt-transports-http-jetty</artifactId>
```
jamesnetherton, line 78: This is for the WS server part, right?

If so, similar to my previous comment, I don't think this should be a dependency of the extension by default. So it should be moved to the BOM.

However, I do not want us to be supporting HTTP transports like Jetty. We should be integrating with the Quarkus Vert.x HTTP server.

extensions/cxf-soap/runtime/pom.xml
javaduke: Well, you can't really use cxf without a transport, so I'm inclined to leave it here, but if you insist, I can move it back to the BOM, it's no big deal really.

extensions/cxf-soap/runtime/pom.xml
javaduke: Correct, I can move it out. I agree that ideally we want to support Vert.x, but CXF doesn't provide transport implementation for it (and I'm not sure if they will). I can try to implement one, but it may take a while. So as an interim solution let's support it and I'll see what I can do about supporting other transports.

extensions/cxf-soap/runtime/pom.xml
jamesnetherton: Not so much the transport. But this stuff:

```
        <dependency>
            <groupId>org.apache.cxf</groupId>
            <artifactId>cxf-rt-ws-rm</artifactId>
        </dependency>
        <dependency>
            <groupId>org.apache.cxf</groupId>
            <artifactId>cxf-rt-ws-addr</artifactId>
        </dependency>
        <dependency>
            <groupId>org.apache.cxf</groupId>
            <artifactId>cxf-rt-ws-security</artifactId>
        </dependency>
        <dependency>
            <groupId>org.apache.cxf</groupId>
            <artifactId>cxf-rt-ws-policy</artifactId>
        </dependency>
```

Presumably `ws-security` is not strictly required in every app?

javaduke: Oh, I see, sure, I can move it out as well.

extensions/cxf-soap/runtime/pom.xml
javaduke: I created a ticket here: https://issues.apache.org/jira/browse/CXF-8742
Please feel free to add your comments and/or corrections.

zhfeng (approved):

extensions/cxf-soap/runtime/pom.xml
jamesnetherton: To me it's mandatory that the endpoints are run on the Quarkus Vert.x HTTP server. Otherwise we're just making the app performance worse and exposing ourselves to more CVEs.

AFAIK `quarkiverse-cxf` has already solved this problem.

javaduke: Yes, I tried it, for some reason CXF does not recognize the VertxDestinationFactory that is provided by quarkiverse-cxf. I'll keep investigating.

zhfeng: So maybe we can consider to leverage on [quarkus-cxf](https://github.com/quarkiverse/quarkus-cxf).

javaduke: I'm trying to figure out how it's done there, but it seems like we'd have to rewrite Camel CXF completely, because there seems to be no way to pass the factories defined in the quarkus-cxf to Camel...

javaduke: OK, I think I know what's going on - currently Camel does not allow creating custom JaxWsServerFactroryBean: https://github.com/apache/camel/blob/main/components/camel-cxf/camel-cxf-soap/src/main/java/org/apache/camel/component/cxf/jaxws/CxfEndpoint.java#L683
This means that destination factories are initialized in the Camel CXF code, and apparently custom destination factories are ignored. Not sure if there's any way around it, apparently not...

zhfeng: @ffang can you take a look?

javaduke:
OK, I think we're done - this PR now uses Quarkus CXF for SOAP services and WS-Security, all tests pass correctly.

javaduke:
Apparently the Quarkus CXF uses a newer version of the xmlsec which causes test failures in the xmlsecurity extension, but I have a fix for that. See https://santuario.apache.org/faq.html **4. Secure Validation**

javaduke:
Any updates? Comments?

jamesnetherton:
> Any updates? Comments?

We should probably wait until Camel 3.18.1 is available (hopefully next week) because it fixes the dependency that was not properly test scoped:

https://github.com/apache/camel/blob/main/components/camel-cxf/camel-cxf-common/pom.xml#L104

So we wont need to pull in a bunch of redundant dependencies.

jamesnetherton:
Thinking also we should have a 'code first' test for the consumer. Right now we're just feeding the endpoint a WSDL file. Would be good to also test a JAX-WS annotated service interface.

javaduke:
Sure, I can work on code first test, although I see it a bit redundant, since the service interface is generated from the WSDL by the cxf-codegen-plugin during the build and then configured in the endpoint: `camel.beans.soapServiceEndpoint.serviceClass=com.helloworld.service.HelloPortType`

javaduke:
OK, code-first test added

ppalaga (approved):
Very nice @javaduke! Some nitpicks inline which I can try to fix myself tomorrow. Thank you very much for your endurance!

zbendhiba:
Many thanks @javaduke for this great work! Happy to find out this has been merged.

PR #3953, opened 2022-08-03
Increase test coverage (file/http) for camel-quarkus-validator extens…

svkcemk:
…ion.

### DIFF
--- a/integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorResource.java
+++ b/integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorResource.java
@@ -32,11 +32,28 @@ public class ValidatorResource {
     @Inject
     ProducerTemplate producerTemplate;
 
-    @Path("/xml")
+    @Path("/classpath")
     @POST
     @Consumes(MediaType.APPLICATION_XML)
     @Produces(MediaType.TEXT_PLAIN)
-    public String processOrder(String statement) {
-        return producerTemplate.requestBody("direct:start", statement, String.class);
+    public String processOrderInXmlFromClassPath(String statement) {
+        return producerTemplate.requestBody("direct:classpath", statement, String.class);
     }
+
+    @Path("/filesystem")
+    @POST
+    @Consumes(MediaType.APPLICATION_XML)
+    @Produces(MediaType.TEXT_PLAIN)
+    public String processOrderInXmlFromFileSystem(String statement) {
+        return producerTemplate.requestBody("direct:filesystem", statement, String.class);
+    }
+
+    @Path("/http")
+    @POST
+    @Consumes(MediaType.APPLICATION_XML)
+    @Produces(MediaType.TEXT_PLAIN)
+    public String processOrderInXmlFromHTTP(String statement) {
+        return producerTemplate.requestBody("direct:http", statement, String.class);
+    }
+
 }
--- a/integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorRouteBuilder.java
+++ b/integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorRouteBuilder.java
@@ -21,7 +21,14 @@ import org.apache.camel.builder.RouteBuilder;
 public class ValidatorRouteBuilder extends RouteBuilder {
     @Override
     public void configure() {
-        from("direct:start")
+        from("direct:classpath")
                 .to("validator:message.xsd");
+
+        from("direct:filesystem")
+                .to("validator:file:src/main/resources/message.xsd");
+
+        from("direct:http")
+                .toD("validator:https://raw.githubusercontent.com/apache/camel-quarkus/main/integration-tests/validator/src/main/resources/message.xsd");
+
     }
 }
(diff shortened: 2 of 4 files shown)


integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorResource.java
```diff
@@ -32,11 +32,28 @@ public class ValidatorResource {
...
+        return producerTemplate.requestBody("direct:filesystem", statement, String.class);
+    }
+
+    @Path("/http")
+    @POST
+    @Consumes(MediaType.APPLICATION_XML)
+    @Produces(MediaType.TEXT_PLAIN)
+    public String processOrderInXmlFromHTTP(String statement) {
+        return producerTemplate.requestBody("direct:http", statement, String.class);
+    }
```
zhfeng, line 57: This test is working in the native mode?

svkcemk: @zhfeng yes I tested with `mvn clean install -Pnative` as well and not seeing any failures on native mode. Thanks!

svkcemk: Sorry my mistake you are right the http one is failing I can see that now, thanks for pointing out.

svkcemk: Added `quarkus.ssl.native=true` for https endpoint, now tests are working in native mode as well, thanks @zhfeng

zhfeng (approved):

zhfeng:
Thanks @svkcemk for your contribution!

oscerd (approved):

integration-tests/validator/src/main/resources/application.properties
```diff
@@ -15,4 +15,5 @@
 ## limitations under the License.
 ## ---------------------------------------------------------------------------
 
-quarkus.native.additional-build-args =-H:ResourceConfigurationFiles=resources-config.json
\ No newline at end of file
+quarkus.native.additional-build-args =-H:ResourceConfigurationFiles=resources-config.json
```
ppalaga, line 18: I see that this is an old code, but we could perhaps use this occasion to change it towards Quarkus way of doing things:
```suggestion
quarkus.native.resources.includes = *.xsd
```

integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorResource.java
ppalaga, line 57: I'd vote for having just one parameterized endpoint instead of three non-parameterized. Something like

```
    @Path("/validate/{scheme}")
    @POST
    @Consumes(MediaType.APPLICATION_XML)
    @Produces(MediaType.TEXT_PLAIN)
    public String processOrderInXmlFromFileSystem(String statement, @PathParam("scheme") String scheme) {
        return producerTemplate.requestBody("direct:" + scheme, statement, String.class);
    }
```

integration-tests/validator/src/test/java/org/apache/camel/quarkus/component/validator/it/ValidatorTest.java
```diff
@@ -25,26 +25,75 @@
...
 
         RestAssured.given()
                 .contentType(ContentType.XML)
                 .body("<message><firstName>MyFirstname</firstName><lastName>MyLastname</lastName></message>")
-                .post("/validator/xml")
+                .post("/validator/classpath")
                 .then()
                 .statusCode(200);
 
     }
```
ppalaga, line 37: I'd vote for having one parameterized test for the valid case and one for invalid case. The params being `"classpath", "filesystem", "http"`. Just grep for `@ParameterizedTest` in the source tree to see some examples.
Less code is perhaps easier to maintain.

integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorRouteBuilder.java
```diff
@@ -21,7 +21,14 @@
...
     public void configure() {
-        from("direct:start")
+        from("direct:classpath")
                 .to("validator:message.xsd");
+
+        from("direct:filesystem")
+                .to("validator:file:src/main/resources/message.xsd");
+
+        from("direct:http")
+                .toD("validator:https://raw.githubusercontent.com/apache/camel-quarkus/main/integration-tests/validator/src/main/resources/message.xsd");
```
ppalaga, line 31: This might be risky for several reasons: 

* If we change message.xsd in the future, the tests in older branches may start failing
* GitHub might be inaccessible in some environments
* GitHub's availability will impact the reliability of this test

I'd vote for serving message.xsd from an ad hoc WireMock server. There are plenty of examples in the source tree.

integration-tests/validator/src/main/resources/application.properties
```diff
@@ -15,4 +15,5 @@
...
+quarkus.ssl.native=true
```
ppalaga, line 19: I see that this is needed for accessing the GitHub https URL. But I wonder how should we help end users to get around this. Maybe we should just document it in `usage.adoc`, something like 

```
If you are accessing XSD resources via HTTPS (e.g. `from(...).to("validator:https://acme.org/foo/schema.xsd")`) than you need to add `quarkus.ssl.native = true` to your `application.properties`.
```

Hardcoding SSL support (like we do in some other extensions) seems to be an overkill in this case, IMO. I guess users will mostly use classpath resources. WDYT @jamesnetherton?

ppalaga (commented):
Thanks @svkcemk , some suggestions inline

PR #3956, opened 2022-08-03
CAMEL-18341 - Upgrade from Codehaus Groovy 3.0.12 to Apache Groovy 4.0.4

apupier:
might worth merging at same time than https://github.com/apache/camel/pull/8102

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,7 +165,7 @@
         <formatter-maven-plugin.version>2.17.1</formatter-maven-plugin.version>
         <frontend-maven-plugin.version>1.6</frontend-maven-plugin.version>
         <groovy-maven-plugin.version>2.1.1</groovy-maven-plugin.version>
-        <groovy.version>3.0.8</groovy.version>
+        <groovy.version>4.0.4</groovy.version>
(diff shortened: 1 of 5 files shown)


apupier:
I do not understand why it complains about uncommited files.
```
	modified:   poms/bom/src/main/generated/flattened-full-pom.xml
	modified:   poms/bom/src/main/generated/flattened-reduced-pom.xml
	modified:   poms/bom/src/main/generated/flattened-reduced-verbose-pom.xml
```

the mentioned files have a lot of modifications. i tried to run the build locally again and I do not have other changes

jamesnetherton:
Attached a diff of the changes that I see when running the build.

Not sure what Maven version you are running, but try to run the build with the `mvnw` script if you have not already done so. That may help.

[diff.txt](https://github.com/apache/camel-quarkus/files/9257486/diff.txt)

apupier:
ah when i see the diff, i think it is because we need to merge first https://github.com/apache/camel/pull/8102

apupier:
outdated.
Note that Camel Core has migrated to Apache groovy 4 https://github.com/apache/camel/pull/9487

Issue #3905, opened 2022-07-07
Merge the performance regression prototype in camel-quarkus main


aldettinger:
I've been experimenting with a performance regression detection tool in [a separate repository](https://github.com/aldettinger/cq-perf-sandbox).

I think it's a valid proof of concept that we could keep attention toward performance early in the process against release candidate and maybe snapshot versions.

There are a lot of possible directions for the future, like including the possibility to:
+ test with/without atlasmap involved
+ test with/without platform-http involved (like pure camel-quarkus-core scenarios)
+ run the tool on MAC and Windows
+ have guard rail to ensure the performance regression tool will work when reaching release candidate time
+ have accurate and meaningful regression reports computed on CI, maybe only at important time like quarkus upgrade

Right now, the next step for me would be to merge it from the separate repository into the camel-quarkus main branch. And then complete the release process to record performance regression reports in a dedicated ticket at each release candidate.

jamesnetherton:
In addition to the list of future enhancements, we should aim to capture more metrics other than throughput. Like startup time, memory usage etc.

> Regression reports computed on CI

Definitely good to have. I wonder if GitHub CI is the best place to do that though. The nodes are quite underpowered.

aldettinger:
Later today, we had the opportunity to discuss this with @ppalaga. So, let's emphasize that the proposal would to run the performance regression tool on the release manager machine and then record the results inside a ticket.
We would not block a release for performance regression right now, the idea to more to learn whether this toll brings value.

At this stage, Peter and I would be in favor of merging the tool upstream and updating the release doc to collect the report/issues. Let's keep time to discuss a bit further in case someone come up with a good reason not to merge.

zbendhiba:
+1

aldettinger:
Ok, let's try to merge the tool in the main branch. I'm renaming this ticket accordingly.

PR #3959, opened 2022-08-04
Merge the performance regression detection prototype #3905

aldettinger:


oscerd (approved):

zhfeng (commented):
@aldettinger is it possible to reuse such `mvnw` in the root directory?

aldettinger:
> @aldettinger is it possible to reuse such `mvnw` in the root directory?

@zhfeng Thanks for feedback. It looks that the root mvnw has a far more recent version. So we should probably align on this one if feasible, and probably we can remove the .mvn folder after that.

I would take it as a dedicated task in another ticket. What do you think ?

zhfeng:
yeah, it looks good to me.

aldettinger:
> yeah, it looks good to me.

Ok, I've filed #3960 in this respect

aldettinger:
Another pair of eyes on this subject ? Feel welcome to ask questions if something is not clear enough.

jamesnetherton:
It looks mostly ok to me.

A nitpick on the Maven Wrapper. If we are going to (at least temporarily) have the sample project use its own wrapper, we should probably make sure the version is aligned with the one used by CQ.

Also, do you think we should add a dedicated docs page for this? I.e perhaps eventually  it'd be part of the release process to do a perf check, so we could link to the relevant bits from the release guide.

aldettinger:
@jamesnetherton The wrapper version is next on my list #3960. And yes, we could also add some improvements like integrating the tool in the release process #3967.

Issue #3966, opened 2022-08-08
CxfSoapClientIT.wsSecurityClient fails in native mode: wsse:Nonce not present in the request


ppalaga:
```
[INFO] Running org.apache.camel.quarkus.component.cxf.soap.it.CxfSoapClientIT
Aug 08, 2022 6:45:42 P.M. org.jboss.threads.Version <clinit>
INFO: JBoss Threads version 3.4.2.Final
Aug 08, 2022 6:45:42 P.M. org.apache.camel.quarkus.test.wiremock.WireMockTestResourceLifecycleManager createServer
INFO: Starting WireMockServer
Aug 08, 2022 6:45:42 P.M. org.apache.camel.quarkus.test.mock.backend.MockBackendUtils logMockBackendUsed
INFO: Mock backend will be used
2022-08-08 18:45:42,788 INFO  [org.ecl.jet.uti.log] (pool-2-thread-1) Logging initialized @2914ms to org.eclipse.jetty.util.log.Slf4jLog
2022-08-08 18:45:42,930 INFO  [org.ecl.jet.ser.Server] (pool-2-thread-1) jetty-9.4.48.v20220622; built: 2022-06-21T20:42:25.880Z; git: 6b67c5719d1f4371b33655ff2d047d24e171e49a; jvm 11.0.7+10
2022-08-08 18:45:42,950 INFO  [org.ecl.jet.ser.han.ContextHandler] (pool-2-thread-1) Started o.e.j.s.ServletContextHandler@452708ae{/__admin,null,AVAILABLE}
2022-08-08 18:45:42,952 INFO  [org.ecl.jet.ser.han.ContextHandler] (pool-2-thread-1) Started o.e.j.s.ServletContextHandler@1b6f967f{/,null,AVAILABLE}
2022-08-08 18:45:42,974 INFO  [org.ecl.jet.ser.AbstractConnector] (pool-2-thread-1) Started NetworkTrafficServerConnector@1a4e4007{HTTP/1.1, (http/1.1)}{0.0.0.0:43467}
2022-08-08 18:45:42,974 INFO  [org.ecl.jet.ser.Server] (pool-2-thread-1) Started @3102ms
Aug 08, 2022 6:45:42 P.M. org.apache.camel.quarkus.test.wiremock.WireMockTestResourceLifecycleManager start
INFO: WireMock started on http://localhost:43467
======================================================================================
  native.image.path was not set, making a guess for the correct path of native image
[...]

PR #3968, opened 2022-08-08
Improve the CXF extension and its tests:

ppalaga:
* Target release is 2.12.0
* Added native tests
* Fixed some issues in native mode
* Make one of the service tests to use the Java service client rather than rest-assured
* Still to do: #3966

@javaduke would you like to review?


integration-tests/cxf-soap/src/main/resources/application.properties
```diff
@@ -64,5 +56,4 @@ camel.beans.codeFirstServiceEndpoint.address=/codefirst
 camel.beans.codeFirstServiceEndpoint.features[0]=#bean:loggingFeature
```
jamesnetherton, line 56: Not something that we have to do for this PR, but I think we should avoid `camel.beans`. IIRC it works via reflection at runtime. It'd be more optimized to have CDI producers for those items.

ppalaga: Thanks, good point, let me do it.

ppalaga:
The OpenStack test passed for me locally. Thus merging.

PR #3972, opened 2022-08-09
Improve/clean-up camel-quarkus-validator extension tests.

svkcemk:

### DIFF
--- a/integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorResource.java
+++ b/integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorResource.java
@@ -21,8 +21,10 @@ import javax.inject.Inject;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
 
 import org.apache.camel.ProducerTemplate;
 
@@ -32,28 +34,15 @@ public class ValidatorResource {
     @Inject
     ProducerTemplate producerTemplate;
 
-    @Path("/classpath")
+    @Path("/validate/{scheme}")
     @POST
     @Consumes(MediaType.APPLICATION_XML)
     @Produces(MediaType.TEXT_PLAIN)
-    public String processOrderInXmlFromClassPath(String statement) {
-        return producerTemplate.requestBody("direct:classpath", statement, String.class);
+    public Response processOrderInXml(String statement, @PathParam("scheme") String scheme) {
+        try {
+            return Response.ok().entity(producerTemplate.requestBody("direct:" + scheme, statement, String.class)).build();
+        } catch (Exception e) {
+            return Response.serverError().entity(e.getMessage()).build();
+        }
     }
-
-    @Path("/filesystem")
-    @POST
-    @Consumes(MediaType.APPLICATION_XML)
-    @Produces(MediaType.TEXT_PLAIN)
-    public String processOrderInXmlFromFileSystem(String statement) {
-        return producerTemplate.requestBody("direct:filesystem", statement, String.class);
-    }
-
-    @Path("/http")
-    @POST
-    @Consumes(MediaType.APPLICATION_XML)
-    @Produces(MediaType.TEXT_PLAIN)
-    public String processOrderInXmlFromHTTP(String statement) {
-        return producerTemplate.requestBody("direct:http", statement, String.class);
-    }
-
 }
--- a/integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorRouteBuilder.java
+++ b/integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorRouteBuilder.java
@@ -16,19 +16,53 @@
  */
 package org.apache.camel.quarkus.component.validator.it;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+
 import org.apache.camel.builder.RouteBuilder;
+import org.eclipse.microprofile.config.ConfigProvider;
 
 public class ValidatorRouteBuilder extends RouteBuilder {
+
     @Override
     public void configure() {
+        // validator from the classpath resource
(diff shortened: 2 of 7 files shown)


integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorResource.java
```diff
@@ -32,28 +33,12 @@ public class ValidatorResource {
...
-        return producerTemplate.requestBody("direct:filesystem", statement, String.class);
-    }
-
-    @Path("/http")
-    @POST
-    @Consumes(MediaType.APPLICATION_XML)
-    @Produces(MediaType.TEXT_PLAIN)
-    public String processOrderInXmlFromHTTP(String statement) {
-        return producerTemplate.requestBody("direct:http", statement, String.class);
+    public String processOrderInXmlFromFileSystem(String statement, @PathParam("scheme") String scheme) {
```
aldettinger, line 40: `processOrderInXmlFromFileSystem` => Don't we do more than file system ?

svkcemk: Thanks just renamed to a generic one.

aldettinger:
@svkcemk It's look good to me. Well done :+1: 

Just one more thing. It's really nice to have separate commits during review. Before merging, it would be good to rebase all commits into a single one in order to help maintaining a clear git history.

svkcemk:
Thanks @aldettinger I was waiting for your review, now commits are squashed into one single commit.

aldettinger (approved):
Looks good to me :+1:  Let's merge after ci build for the sake of completeness.

integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorRouteBuilder.java
```diff
@@ -16,19 +16,47 @@
...
 
         from("direct:http")
-                .toD("validator:https://raw.githubusercontent.com/apache/camel-quarkus/main/integration-tests/validator/src/main/resources/message.xsd");
+                .toD("validator:" + serverURL + "/xsd");
 
     }
+
+    public String createTempXsd(String sourceXsd) {
+        try {
+            InputStream resourceAsStream = Thread.currentThread().getContextClassLoader().getResourceAsStream(sourceXsd);
```
ppalaga, line 53:
```suggestion
        try (InputStream resourceAsStream = Thread.currentThread().getContextClassLoader().getResourceAsStream(sourceXsd)) {
```

Otherwise the stream is not closed. It may cause problems especially on Windows.

integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorRouteBuilder.java
```diff
@@ -16,19 +16,47 @@
...
+            Path tempXsd = Files.createTempFile("temp", ".xsd");
+            Files.copy(resourceAsStream, tempXsd, StandardCopyOption.REPLACE_EXISTING);
+            return tempXsd.toAbsolutePath().toString();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
ppalaga, line 58:
```suggestion
            throw new RuntimeException("Could not read "+ sourceXsd + " from classpath or copy it to " + tempXsd, e);
```

The wrapped exception message does not say related to which file the operation failed. Adding the path to the wrapper exception will help future users to faster figure out where the problem is.

integration-tests/validator/src/test/java/org/apache/camel/quarkus/component/validator/it/ValidatorTest.java
```diff
@@ -16,84 +16,40 @@
...
+    @ValueSource(strings = { "classpath", "filesystem", "http" })
+    public void inValidXML(String scheme) {
 
         RestAssured.given()
                 .contentType(ContentType.XML)
                 .body("<message><firstName>MyFirstname</firstName></message>")
-                .post("/validator/classpath")
+                .post("/validator/validate/" + scheme)
                 .then()
                 .statusCode(500);
```
ppalaga, line 52: Here I wonder whether there is some body text (or any other way) from which we could see that what failed is really the XSD validation? 500 could mean a failure somewhere else in the stack, which we definitely would not want to accept as "test passed".

ppalaga (commented):
Very nice, thanks @svkcemk ! A few small nitpicks inline.

integration-tests/validator/src/main/java/org/apache/camel/quarkus/component/validator/it/ValidatorRouteBuilder.java
svkcemk: Thanks @ppalaga added some changes.

integration-tests/validator/src/test/java/org/apache/camel/quarkus/component/validator/it/ValidatorTest.java
svkcemk: @ppalaga I have added some checks based on the body, I think it will make more sense now, Thanks for pointing out !

ppalaga: The tests are failing. You might want to forward the exception message to the client in `ValidatorResource`. Something like 

```
    public javax.ws.rs.core.Response processOrderInXml(String statement, @PathParam("scheme") String scheme) {
        try {
            return javax.ws.rs.core.Response.ok().entity(producerTemplate.requestBody("direct:" + scheme, statement, String.class));
        } catch (Exception e) {
                return javax.ws.rs.core.Response.serverError().entity(e.getMessage());
        }
            
    }
```

svkcemk: > The tests are failing. You might want to forward the exception message to the client in `ValidatorResource`. Something like
> 
> ```
>     public javax.ws.rs.core.Response processOrderInXml(String statement, @PathParam("scheme") String scheme) {
>         try {
>             return javax.ws.rs.core.Response.ok().entity(producerTemplate.requestBody("direct:" + scheme, statement, String.class));
>         } catch (Exception e) {
>                 return javax.ws.rs.core.Response.serverError().entity(e.getMessage());
>         }
>             
>     }
> ```

ppalaga (approved):

ppalaga:
Thanks @svkcemk !

PR #3984, opened 2022-08-11
Upgrade Camel to 3.18.1

jamesnetherton:


aldettinger (approved):

jamesnetherton:
I'll do a temporary workaround for the failure in the `health` example (disabling the test). I need to understand the cause of the failure better before fixing properly.

PR #3991, opened 2022-08-12
WS_Addressing test added

javaduke:


javaduke:
@ppalaga Turns out, implementing WS-A is trivial but I added a test anyway

PR #3996, opened 2022-08-16
Fix route inclusion / exclusion filtering

jamesnetherton:


extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RegistryRoutesLoaders.java
```diff
@@ -88,6 +88,15 @@ public void configure() throws Exception {
                 boolean abs = Modifier.isAbstract(routesBuilder.getClass().getModifiers());
                 if (!abs) {
                     String name = routesBuilder.getClass().getName();
+
```
aldettinger, line 91: On the top of my head, the naming pattern will no vary a lot. Plus, we would catch any breaking change via tests. So, it should be ok if I get it  well.

aldettinger (approved):
Ok, make sense.

PR #4017, opened 2022-08-18
Restore ability to run js-dsl native tests

jamesnetherton:


aldettinger (approved):
Nice trick @jamesnetherton :+1:

zhfeng (approved):

zhfeng:
Is there any chance to add `gu install js` in https://github.com/quarkusio/quarkus-images/blob/main/modules/graalvm/22.2.0-java17/configure?

jamesnetherton:
> Is there any chance to add `gu install js`

The thing is, 99% of users will never use it. If we get reports from folks wanting JS native support, then we can ask if it could be added to the builder image.

Issue #1615, opened 2020-08-26
XSLT Saxon native support


ppalaga:

CynanX:
I hit this same issue when trying to build a native executable with `camel-quarkus-xslt-saxon`, though my use case is slightly different in that I exclude the home edition and import the personal edition.

Doing this I hit the same error as above. I then ran Quarkus along with the native agent to generate the required `reflect-config.json` file and can report that Saxon is now working fine for me in native mode.

zhfeng:
@CynanX it would be very useful if you can share with ```reflect-ocnfig.json``` here. And we can take a look to improve in the build steps. Thanks!

CynanX:
I'm currently using Quarkus 2.7.5.Final and building on GraalVM 21.3.1 Java 17 CE (Java Version 17.0.2+8-jvmci-21.3-b09).

The relevant sections of my pom are:

```
<dependency>
    <groupId>org.apache.camel.quarkus</groupId>
    <artifactId>camel-quarkus-xslt-saxon</artifactId>
    <exclusions>
        <exclusion>
            <groupId>net.sf.saxon</groupId>
            <artifactId>Saxon-HE</artifactId>
        </exclusion>
    </exclusions>
</dependency>

<dependency>
    <groupId>net.sf.saxon</groupId>
    <artifactId>saxon-pe</artifactId>
    <version>11.3</version>
</dependency>

<dependency>
    <groupId>com.sun.mail</groupId>
    <artifactId>javax.mail</artifactId>
    <version>1.6.2</version>
</dependency>
```

Here's is my `resource-config.json` file:

```
{
  "resources": {
    "includes": [
      {
        "pattern": "\\Qsaxon-license.lic\\E"
      }
    ]
  }
}
```

And finally here is my `reflect-config.json` file:

```
[
  {
    "name": "com.saxonica.config.ProfessionalConfiguration",
    "methods": [
      {
        "name": "<init>",
        "parameterTypes": []
      }
    ]
  },
  {
    "name": "com.saxonica.functions.extfn.Parse",
    "methods": [
      {
        "name": "<init>",
        "parameterTypes": []
      }
    ]
  },
  {
    "name": "net.sf.saxon.functions.StringJoin",
    "methods": [
      {
        "name": "<init>",
        "parameterTypes": []
      }
    ]
  },
  {
    "name": "net.sf.saxon.functions.String_1",
    "methods": [
      {
        "name": "<init>",
        "parameterTypes": []
      }
[...]

PR #4018, opened 2022-08-19
Fix #3904 increase xslt extension test coverage

### LINKED ISSUE #3904: Increase XSLT extension test coverage
- complete URL (classpath, file, http, ref, bean)
- different outputs (string, bytes, dom, file)
- using xsl:include to include files
- custom `XsltUriResolverFactory`
- custom `URIResolver`
- using `org.apache.camel.component.xslt.XsltAggregationStrategy`
- caching
- Being able to raise errors with xsl:message and messages getting into `Exchange.XSLT_ERROR`, `Exchange.XSLT_FATAL_ERROR`, or `Exchange.XSLT_WARNING`

zhfeng:
- [x] Add `ref`, `bean`, `http`, `file` schemas which could only work in jvm mode since generating xslt template classes dynamically does not be supported in native mode.
- [x] Different outputs (string, bytes, dom, file)
- [x] Using xsl:include to include files
- [x] Add aggergate test
- [x] Add custom URIResolver 
- [ ] Caching
- [x] Being able to raise errors with xsl:message and messages getting into `Exchange.XSLT_ERROR`, `Exchange.XSLT_FATAL_ERROR`, or `Exchange.XSLT_WARNING`
- [x] Update the documents

### DIFF
--- a/extensions/xslt/runtime/src/main/doc/configuration.adoc
+++ b/extensions/xslt/runtime/src/main/doc/configuration.adoc
@@ -9,7 +9,22 @@ quarkus.camel.xslt.sources = transform.xsl, classpath:path/to/my/file.xsl
 
 Scheme-less URIs are interpreted as `classpath:` URIs.
 
-Only `classpath:` URIs are supported on Quarkus. `file:`, `http:` and other kinds of URIs do not work by design.
+Only `classpath:` URIs are supported on Quarkus native mode. `file:`, `http:` and other kinds of URIs can be used on JVM mode only.
+
+`<xsl:include>` and `<xsl:messaging>` XSLT elements are also supported in JVM mode only right now.
+
+If `aggregate` DSL is used, `XsltSaxonAggregationStrategy` has to be used such as
+[source,java]
+----
+from("file:src/test/resources?noop=true&sortBy=file:name&antInclude=*.xml")
+   .routeId("aggregate").noAutoStartup()
+   .aggregate(new XsltSaxonAggregationStrategy("xslt/aggregate.xsl"))
+   .constant(true)
+   .completionFromBatchConsumer()
+   .log("after aggregate body: ${body}")
+   .to("mock:transformed");
+----
+Also, it's only supported on JVM mode.
 
 === Configuration
 TransformerFactory features can be configured using following property:
@@ -35,9 +50,4 @@ public class FunctionsConfiguration {
 ====
 The content of the XSLT source URIs is parsed and compiled into Java classes at build time. These Java classes are the
 only source of XSLT information at runtime. The XSLT source files may not be included in the application archive at all.
-====
-
-[WARNING]
-====
-the extension does not yet support Java 11.
 ====
--- a/extensions/xslt/runtime/src/main/java/org/apache/camel/quarkus/component/xslt/CamelXsltRecorder.java
+++ b/extensions/xslt/runtime/src/main/java/org/apache/camel/quarkus/component/xslt/CamelXsltRecorder.java
@@ -20,9 +20,12 @@ import java.util.Map;
 
 import javax.xml.transform.TransformerException;
 import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.URIResolver;
 
 import io.quarkus.runtime.RuntimeValue;
 import io.quarkus.runtime.annotations.Recorder;
+import org.apache.camel.CamelContext;
+import org.apache.camel.component.xslt.DefaultXsltUriResolverFactory;
 import org.apache.camel.component.xslt.TransformerFactoryConfigurationStrategy;
 import org.apache.camel.component.xslt.XsltComponent;
 import org.apache.camel.component.xslt.XsltEndpoint;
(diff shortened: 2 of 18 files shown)


integration-tests/xml/pom.xml
```diff
@@ -35,14 +35,30 @@
...
             <artifactId>camel-quarkus-direct</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-file</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-mock</artifactId>
+        </dependency>
```
ppalaga, line 61: @zhfeng I have some concerns about adding this many new dependencies to the test. I mean it is generally better to test with as little dependencies as possible, so that it does not happen that an extension that is there just for testing some niche feature brings some build steps that become vital for the execution of some basic functionality.

Wouldn't it perhaps be worthwhile to introduce a new XML testing group like we already have with [Azure, AWS an others](https://github.com/apache/camel-quarkus/tree/main/integration-test-groups)? We could have several XML modules there that would test the use cases coverable with their minimal set of deps and ideally still be able to run them all together on the CI (in the grouped module), so that the exec time does not explode. See also https://camel.apache.org/camel-quarkus/2.11.x/contributor-guide/extension-testing.html#_grouping
WDYT?

ppalaga: > But how to group the sub modules?

The grouping is given by the required dependencies. I.e. if the `bean:` test protocol requires an additional `camel-quarkus-bean` dependency, then it should be a separate test module with that dependency under the new group.

zhfeng: OK, let's raise a follow up issue to group the xml tests.

zhfeng:
@ppalaga please can you take a review of these changes?

extensions/xslt/runtime/src/main/java/org/apache/camel/quarkus/component/xslt/CamelXsltRecorder.java
```diff
@@ -52,6 +55,23 @@ public void addRuntimeUriResolverEntry(RuntimeValue<RuntimeUriResolver.Builder>
...
+            this.uriResolver = uriResolver;
+        }
+
+        @Override
+        public URIResolver createUriResolver(CamelContext camelContext, String resourceUri) {
+            if (uriResolver.getTransletClassName(resourceUri) != null) {
+                return uriResolver;
+            } else {
+                return super.createUriResolver(camelContext, resourceUri);
+            }
```
ppalaga, line 71: A comment would be nice about what these two options are supposed to do. The first one is supposed to catch the cases where we compile the translet at build time, right? Which is for `classpath:` XSLT resources in both JVM and native mode?
The `else` branch is for all other cases, right?

extensions/xslt/runtime/src/main/java/org/apache/camel/quarkus/component/xslt/RuntimeUriResolver.java
```diff
@@ -60,11 +60,7 @@ public Source resolve(String href, String base) throws TransformerException {
      * @return     the unqualified translet name associated with the given {@code uri}
```
ppalaga, line 60: Just guessing - please correct me if needed:

```suggestion
     * @return     the unqualified translet name associated with the given {@code uri} or {@code null} if the given XSLT resource was not compiled to a translet at build time.
```

integration-tests/xml/src/main/java/org/apache/camel/quarkus/component/xml/it/XmlRouteBuilder.java
```diff
@@ -18,9 +18,17 @@
...
 import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.xslt.saxon.XsltSaxonAggregationStrategy;
 import org.apache.camel.support.builder.Namespaces;
 
+@RegisterForReflection(classNames = {
+        "net.sf.saxon.Configuration",
+        "net.sf.saxon.functions.String_1",
+        "net.sf.saxon.functions.Tokenize_1",
+        "net.sf.saxon.functions.StringJoin",
+        "org.apache.camel.component.xslt.saxon.XsltSaxonBuilder" })
```
ppalaga, line 31: I see that camel-quarkus-xslt-saxon is JVM only. 

If porting camel-quarkus-xslt-saxon to native would be too much work, then it would be worthwhile to put a comment here saying that `@RegisterForReflection` should be removed with fixing https://github.com/apache/camel-quarkus/issues/1615

In https://github.com/apache/camel-quarkus/issues/1615, we should note that XmlRouteBuilder.java needs some care

ppalaga (commented):
Looks great! Inline, I am asking for adding some comments here and there.

extensions/xslt/runtime/src/main/doc/configuration.adoc
```diff
@@ -9,7 +9,22 @@ quarkus.camel.xslt.sources = transform.xsl, classpath:path/to/my/file.xsl
 
 Scheme-less URIs are interpreted as `classpath:` URIs.
 
-Only `classpath:` URIs are supported on Quarkus. `file:`, `http:` and other kinds of URIs do not work by design.
+Only `classpath:` URIs are supported on Quarkus native mode. `file:`, `http:` and other kinds of URIs can be used on JVM mode only.
+
+`xsl:include` and `xsl:messaging` are aslo supported in JVM mode only right now.
```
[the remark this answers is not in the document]
zhfeng: It looks good to me.

extensions/xslt/runtime/src/main/java/org/apache/camel/quarkus/component/xslt/CamelXsltRecorder.java
zhfeng: yeah, it is right.

integration-tests/xml/src/main/java/org/apache/camel/quarkus/component/xml/it/XmlRouteBuilder.java
zhfeng: yeah, I will leave a comment here.

PR #4019, opened 2022-08-19
Register HttpOperationFailedException for reflection

jamesnetherton:


aldettinger (approved):
Nice catch :+1:

ppalaga (approved):
I wondered whether we need to add seda for this, but it is actually OK. Seda does not bring any additional transitives or build steps.

Issue #4014, opened 2022-08-18
perf-regression: java.lang.NumberFormatException: For input string: "782,36"


aldettinger:
On some machine, the hyperfoil-maven-plugin could report the throughput using the french locale. Thus generating the below stack trace:

```
--\___\_\____/_/ |_/_/|_/_/|_|\____/___/   
2022-08-18 16:08:09,549 INFO  [io.quarkus] (main) perf-regression 2.12.0-SNAPSHOT on JVM (powered by Quarkus 2.12.0.CR1) started in 0.828s. 
2022-08-18 16:08:09,607 INFO  [io.quarkus] (main) Profile prod activated. 
2022-08-18 16:08:09,607 INFO  [io.quarkus] (main) Installed features: [cdi, picocli]
java.lang.NumberFormatException: For input string: "782,36"
    at java.base/jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2054)
    at java.base/jdk.internal.math.FloatingDecimal.parseDouble(FloatingDecimal.java:110)
    at java.base/java.lang.Double.parseDouble(Double.java:543)
    at org.apache.camel.quarkus.performance.regression.PerfRegressionCommand.runPerfRegression(PerfRegressionCommand.java:117)
    at org.apache.camel.quarkus.performance.regression.PerfRegressionCommand.runPerfRegressionForCqVersion(PerfRegressionCommand.java:88)
    at org.apache.camel.quarkus.performance.regression.PerfRegressionCommand.run(PerfRegressionCommand.java:64)
    at picocli.CommandLine.executeUserObject(CommandLine.java:1939)
    at picocli.CommandLine.access$1300(CommandLine.java:145)
    at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2358)
    at picocli.CommandLine$RunLast.handle(CommandLine.java:2352)
    at picocli.CommandLine$RunLast.handle(CommandLine.java:2314)
    at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
    at picocli.CommandLine$RunLast.execute(CommandLine.java:2316)
    at io.quarkus.picocli.runtime.PicocliRunner$EventExecutionStrategy.execute(PicocliRunner.java:26)
    at picocli.CommandLine.execute(CommandLine.java:2078)
    at io.quarkus.picocli.runtime.PicocliRunner.run(PicocliRunner.java:40)
[...]

PR #4024, opened 2022-08-22
perf-regresssion: fix number format exception when java and mvnw don'…

aldettinger:
…t have same default locale #4014

### DIFF
--- a/tooling/perf-regression/src/main/java/org/apache/camel/quarkus/performance/regression/MvnwCmdHelper.java
+++ b/tooling/perf-regression/src/main/java/org/apache/camel/quarkus/performance/regression/MvnwCmdHelper.java
@@ -68,14 +68,14 @@ public class MvnwCmdHelper {
 
             Map<String, String> environment = EnvironmentUtils.getProcEnvironment();
 
-            String newMavenOpts = "MAVEN_OPTS=--add-opens java.base/java.lang=ALL-UNNAMED";
+            String newMavenOpts = "-Duser.language=en -Duser.country=US --add-opens java.base/java.lang=ALL-UNNAMED";
             if (environment.containsKey("MAVEN_OPTS")) {
                 String currentMavenOpts = environment.get("MAVEN_OPTS");
                 LOGGER.debugf("MAVEN_OPTS is already set up in the main process with value: %s", currentMavenOpts);
-                newMavenOpts = "MAVEN_OPTS=" + currentMavenOpts + " --add-opens java.base/java.lang=ALL-UNNAMED";
+                newMavenOpts = currentMavenOpts + " " + newMavenOpts;
             }
             LOGGER.debugf("Setting MAVEN_OPTS in child process with value: %s", newMavenOpts);
-            EnvironmentUtils.addVariableToEnvironment(environment, newMavenOpts);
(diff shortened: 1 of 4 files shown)


aldettinger:
@zbendhiba Time permitting, it would be awesome if this PR could be tested in an environment where the bug is truly reproduceable. More info in #4014

zbendhiba (changes requested):
@aldettinger   This works fine for Locale = fr_FR.UTF-8, but not for some other Locales. For instance when I try Locale = de_DE.UTF-8, the format of the Double is like that "#,##0.00", and we get a different error, as the String is not compatible with the Regex anymore. This is the kind of error we get with Locale =  de_DE.UTF-8 : ```Expected: a string containing " req/s [+0.00%] "
     but: was "       2.12.0-SNAPSHOT  |        1s  |  166,00 req/s [+0,00%]  |      OK  |"```. IMHO, we need to set the Locale to en_US, to get the right sentence, and avoid any Double/String conversion

aldettinger:
@zbendhiba The reported error message seems to indicate that the expected for the `nominalShouldPrintReport` integration-test is locale dependent. I can't reproduce this though, steps to reproduce would help here.

Overriding the default locale in sub-processes with MAVEN_OPTS would put us in a more secure/repeatable situation to the price of overriding some environment settings made by users for good reasons. For instance, maybe there is a good reason that the default locale is different between java and maven on your machine ? Or do you think it would always be safe to override the default locale ?

zbendhiba:
I was thinking of setting in the code with Java. If this meant to be executed also for native mode. Add this to the configuration: 
``` 
quarkus.native.user-country=US
quarkus.native.user-language=en 
```
Just to prevent it from failing locally, as Double is different regarding to our native languages. My laptop is setup with French language by default.

To reproduce on Linux, you can change the Locale this way : `export LANG=de_DE.UTF-8` in a terminal, before executing the test

aldettinger:
The code formatting the double is from the hyperfoil-maven-plugin, not sure they would like to override the default locale with java code. Plus it would need a release. Anyway, it should be possible to do it on our side [with MAVEN_OPTS](https://github.com/apache/camel-quarkus/blob/main/tooling/perf-regression/src/main/java/org/apache/camel/quarkus/performance/regression/MvnwCmdHelper.java#L71).

Concerning native, there is no need for native support in the tool itself. The sample application under test could also be tested in native with the `-an` option. This is already working without tweaking locales.

The current PR should work as soon as java and maven have the same default locale. However itests are not working.
For instance, below would be expected to work:
```
mvn clean install -Dquickly
java -jar target/quarkus-app/quarkus-run.jar -d 1s 2.11.0
```

Thanks for sharing the way to reproduce. Experimenting with that should help to identify the best solution :+1:

aldettinger:
Actually, both ways could work. I've opted to use a fixed en_US locale:
 + for interactions between the tool and underlying maven builds
 + when the tool format the report that will be printed to the console

This should work as soon as:
 + using `-Duser.language=...` has precedence over other ways of overriding the default locale
 + when passing mutliple `-user.language=` statements to the JVM, only the last takes effect
 + the en_US locale is available on all JREs

If overriding user environments/configs/variables generate other issues. Then we could still fallback to the initial solution with a regex.

zbendhiba (approved):

PR #4030, opened 2022-08-23
Pass encrypt property to Debezium MS SQL Server JDBC URL

jamesnetherton:
I took the liberty of adding `acceptLicense()` to the MS SQL Server container setup so that we can always have the SQL Sever tests run by default. Hopefully there's no issues with us doing that.

### DIFF
--- a/integration-tests/debezium/src/main/java/org/apache/camel/quarkus/component/debezium/common/it/AbstractDebeziumResource.java
+++ b/integration-tests/debezium/src/main/java/org/apache/camel/quarkus/component/debezium/common/it/AbstractDebeziumResource.java
@@ -61,13 +61,13 @@ public abstract class AbstractDebeziumResource {
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     public Map<String, String> getAdditionalProperties() {
-        DebeziumEndpoint endpoint = (DebeziumEndpoint) camelContext.getEndpoint(getEndpointUrl()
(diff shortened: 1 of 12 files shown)


aldettinger:
@jamesnetherton Great to have multiple commits. It looks easier to review actually. Maybe, that could become a general practise to keep distinct commits during reviews. Maybe sometimes, squashing at the end of review would be needed...

Reading the EULA, It looks ok for to use in itests. However, couldn't we have issues when someone clone the repository and run itests on its own machine.

For instance, the EULA mention US export restriction law. When calling `acceptLicense()`, it could mean that explicit consent would be given that the MS SQL database is run in a country that is not subject to US export restrictions. However, anybody can clone this code and use it in another country subject to US export restrictions...

That said, it's a tricky question. Maybe we should ask Apache legal ? Or maybe we should keep a kind of express consent in a simpler form, like an environment variable `QUARKUS_***_ACCEPT_MS_SQL_EULA=True` ? It could be true on CI and one would need to set it to true when running locally ?

jamesnetherton:
> It could be true on CI and one would need to set it to true when running locally 

On reflection, I think, I'll drop that commit. I thought I had seen in the past that Quarkus dev services was doing auto acceptance and figured maybe we could do the same. But I just checked and they have the same requirement of manually adding `src/test/resources/container-license-acceptance.txt`.

aldettinger (approved):
Good trick to check how Quarkus handles it. Looks good to me.

PR #4045, opened 2022-08-25
Log warning instead of throw exception for Debezium MongoDB container script execution failure

jamesnetherton:
Otherwise it seems to make all of the following Debezium tests fail. It's safe to do this as there's a check in `DebeziumMongodbTest` to see if there's a MongoDB connection available, and if not, fails the tests.


oscerd (approved):

aldettinger:
So the `DebeziumMongodbTestResource` is initialized only once in `DebeziumMongodbTest` ? 

Tests are executed in below order ?
 + testReceiveInit
 + testInsert
 + testUpdate
 + testDelete

And showing a warning instead of throwing a RuntimeException would make it possible to run some of latter tests ?

jamesnetherton:
> And showing a warning instead of throwing a RuntimeException would make it possible to run some of latter tests ?

By 'latter tests` I mean the ones for the other Debezium components. MySQL, Postgres, SQL Server etc. Having `RuntimeException` thrown seems to prevent that. We only want the Mongo tests to fail.

aldettinger (approved):
Ok, make sense now :+1:

jamesnetherton:
Hmmm - looks likes there's a different failure. I wonder if previously we have been masking issues with Mongo. I'll dig into it a bit...

jamesnetherton:
> Hmmm - looks likes there's a different failure.

It was because there was no tag specified for the MongoDB container image.

PR #4049, opened 2022-08-25
Springless JPA extension

zhfeng:
- It replaces the spring transaction template with Quarkus TX API
- It excludes `spring-orm`
- I have to add some tricks to make it work in native mode

### DIFF
--- a/extensions/jpa/runtime/pom.xml
+++ b/extensions/jpa/runtime/pom.xml
@@ -53,10 +53,6 @@
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-jpa</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-tx</artifactId>
-        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/jpa/runtime/pom.xml
+++ b/extensions/jpa/runtime/pom.xml
@@ -36,6 +36,11 @@
     </properties>
 
     <dependencies>
+	<dependency>
+            <groupId>org.graalvm.nativeimage</groupId>
+            <artifactId>svm</artifactId>
+            <scope>provided</scope>
+        </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-hibernate-orm</artifactId>
@@ -45,12 +50,12 @@
             <artifactId>camel-quarkus-core</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.apache.camel.quarkus</groupId>
-            <artifactId>camel-quarkus-support-spring</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-jpa</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-tx</artifactId>
         </dependency>
     </dependencies>
 
--- a/extensions/jpa/runtime/src/main/java/org/apache/camel/quarkus/component/jpa/DefaultTransactionStrategySubstitution.java
+++ b/extensions/jpa/runtime/src/main/java/org/apache/camel/quarkus/component/jpa/DefaultTransactionStrategySubstitution.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.camel.quarkus.component.jpa;
+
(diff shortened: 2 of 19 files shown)


oscerd (approved):

zhfeng:
@jamesnetherton @ppalaga please review these changes and I think it might be experimental feature currently.

extensions/jpa/runtime/src/main/java/org/apache/camel/quarkus/component/jpa/DefaultTransactionStrategySubstitution.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.persistence.EntityManagerFactory;
+
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+import org.apache.camel.component.jpa.DefaultTransactionStrategy;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.support.TransactionTemplate;
+
+@TargetClass(DefaultTransactionStrategy.class)
+final public class DefaultTransactionStrategySubstitution {
```
jamesnetherton, line 29: I wonder if we could annotate this class with `@Delete` instead of doing the method substitutions? If we are forcing the transaction strategy in the extension, then in theory there should be nothing that needs access to it.

extensions/jpa/runtime/pom.xml
```diff
@@ -44,14 +49,14 @@
...
-            <groupId>org.apache.camel.quarkus</groupId>
-            <artifactId>camel-quarkus-support-spring</artifactId>
-        </dependency>
         <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-jpa</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-tx</artifactId>
```
jamesnetherton, line 58: Could maybe scope as `provided`?

Not sure if it'd be possible eventually to remove this entirely. It'd be nice if we could.

extensions/jpa/runtime/src/main/java/org/apache/camel/quarkus/component/jpa/DefaultTransactionStrategySubstitution.java
zhfeng: No, graalvm has to find `DefaultTransactionStrategy` during the static analyst phase.

extensions/jpa/runtime/pom.xml
zhfeng: `java.lang.ClassNotFoundException: org.springframework.transaction.support.TransactionTemplate` is throws during static analyst if we add `provided` scope.

zhfeng:
It needs https://github.com/apache/camel/pull/8264 to use TransactionStrategy in `JpaMessageIdRepository`

integration-tests/jpa/src/main/java/org/apache/camel/quarkus/component/jpa/it/JpaRoute.java
```diff
@@ -73,7 +77,12 @@ public void configure() throws Exception {
         from("direct:idempotent")
                 .idempotentConsumer(
                         header("messageId"),
-                        jpaMessageIdRepository(entityManagerFactory, "idempotentProcessor"))
+                        new JpaMessageIdRepository(entityManagerFactory, new TransactionStrategy() {
+                            @Override
+                            public void executeInTransaction(Runnable runnable) {
+                                QuarkusTransaction.run(runOptions().semantic(RunOptions.Semantic.JOIN_EXISTING), runnable);
+                            }
+                        }, "idempotentProcessor"))
```
ppalaga, line 85: Will end users always have to create the `JpaMessageIdRepository` like this? If so, can't we automate it for them somehow? Or at least document that they need to do this?

ppalaga (commented):
Nice work, @zhfeng ! One comment inline.

Besides, I wonder how much Spring-less we are now? It looks like the extension still has Spring dependencies? Are they all left there only to make the native compiler happy? If so, can't we maybe add some fake java stubs for the required classes (like @jamesnetherton did IIRC elsewhere for other Spring stuff)?

integration-tests/jpa/src/main/java/org/apache/camel/quarkus/component/jpa/it/JpaRoute.java
zhfeng: yeah, I think we can add `QuarkusTransactionStrategy`.

zhfeng:
Thanks @ppalaga only `spring-tx` is used right now and yeah, most of them needed in native to make compiler happy are `PlatformTransactionManager` and `TransactionTemplate`. I'm not sure if it is possible to add fake classes but will try it.

extensions/jpa/runtime/pom.xml
zhfeng: Finnaly we remove it! and add fake `PlatformTransactionManager` and `TransactionTemplate` to make it work with graalvm native bulding.

extensions/jpa/runtime/src/main/java/org/apache/camel/quarkus/component/jpa/DefaultTransactionStrategySubstitution.java
zhfeng: It is deleted now.

integration-tests/jpa/src/main/java/org/apache/camel/quarkus/component/jpa/it/JpaRoute.java
zhfeng: Done

jamesnetherton:
Do we need any additional documentation? E.g in the route you inject `TransactionStrategy` and use it in the `JpaMessageIdRepository`. If the same thing is expected by users, then we should add a note into the extension docs.

zhfeng:
yeah, sure. I'm working the doc.

jamesnetherton (approved):

PR #4060, opened 2022-08-31
Ban camel-directvm 

### LINKED ISSUE #4059: Ban camel-directvm
Follows up https://github.com/apache/camel-quarkus/pull/3847#discussion_r938639850

`camel-directvm` is a transitive dependency of camel-test-junit5. I wonder whether it is really needed on Camel Quarkus?
IIRC, we agreed recently not to port camel-directvm to CQ. But when it soaks to end user tests in this way, it might awake a false impression that it is supported on CQ. Would it perhaps be possible to exclude camel-directvm from camel-test-junit5 and stop managing camel-directvm?

ppalaga:
Fix #4059


aldettinger (approved):
Indeed and further directvm interest may even be questionable with mono context microservices.

PR #4067, opened 2022-09-02
pg-replication-slot: fix warning

aldettinger:


aldettinger:
Need to check group_07 itests on CI

Issue #4063, opened 2022-09-01
[Quarkus 2.13.0] JTA integration test fails due to undiscoverable `DataSource`


jamesnetherton:
Not sure what's changed, but the JTA integration test fails due to:

```
2022-09-01 08:58:13,084 ERROR [org.apa.cam.qua.mai.CamelMainRuntime] (main) Failed to start application: org.apache.camel.FailedToCreateRouteException: Failed to create route route11 at: >>> Transacted <<< in route: Route(route11)[From[direct:sqltx] -> [OnException[[java.lang... because of Failed to resolve endpoint: sql://insert%20into%20example(message,%20origin)%20values%20(:%23message,%20'sqltx') due to: DataSource must be configured
        at org.apache.camel.reifier.RouteReifier.doCreateRoute(RouteReifier.java:240)
        at org.apache.camel.reifier.RouteReifier.createRoute(RouteReifier.java:74)
        at org.apache.camel.impl.DefaultModelReifierFactory.createRoute(DefaultModelReifierFactory.java:49)
        at org.apache.camel.impl.DefaultCamelContext.startRouteDefinitions(DefaultCamelContext.java:862)
        at org.apache.camel.impl.DefaultCamelContext.startRouteDefinitions(DefaultCamelContext.java:750)
        at org.apache.camel.impl.engine.AbstractCamelContext.doInit(AbstractCamelContext.java:2947)
        at org.apache.camel.quarkus.core.FastCamelContext.doInit(FastCamelContext.java:174)
        at org.apache.camel.support.service.BaseService.init(BaseService.java:83)
        at org.apache.camel.impl.engine.AbstractCamelContext.init(AbstractCamelContext.java:2630)
        at org.apache.camel.support.service.BaseService.start(BaseService.java:111)
        at org.apache.camel.impl.engine.AbstractCamelContext.start(AbstractCamelContext.java:2649)
        at org.apache.camel.impl.DefaultCamelContext.start(DefaultCamelContext.java:262)
        at org.apache.camel.quarkus.main.CamelMain.doStart(CamelMain.java:94)
        at org.apache.camel.support.service.BaseService.start(BaseService.java:119)
        at org.apache.camel.quarkus.main.CamelMain.startEngine(CamelMain.java:140)
        at org.apache.camel.quarkus.main.CamelMainRuntime.start(CamelMainRuntime.java:49)
[...]

PR #4094, opened 2022-09-15
Upgrade Quarkus to 2.13.0.CR1

jamesnetherton:
I could possibly drop commit be839b3 and cover the workaround with documentation. I added it to restore the capability to do autowiring of `DataSource` when only 1 named `DataSource` exists. It relates to [this Quarkus change](https://github.com/quarkusio/quarkus/commit/60f62f7283a62fe28d9282be78cec97ff2866120). Not sure what others think?


zhfeng:
@jamesnetherton what is the workaround if there is only one named `DataSource`, users have to use it explictly and can not be Autowired?

jamesnetherton:
> What is the workaround if there is only one named DataSource, users have to use it explictly and can not be Autowired?

Yes, exactly.

However, the JDBC example project test is failing. So it looks like I would need to rethink the fix for #4063 anyway.

I'm going to work around the JTA itest failure for now so we can merge the 2.13 upgrade.

zhfeng:
yeah, I recall the issue https://github.com/apache/camel-quarkus/issues/4063. For the failing of jdbc example, is it possible to add some similar codes in jdbc extension https://github.com/apache/camel-quarkus/pull/4094/commits/be839b34d1260987bcf8fd0b43be70d5ef4b9bbf#diff-15efe9e17b3cf8bebde051bb494ba73ed69148b908162e1191492f6ad141cda2

zhfeng (approved):

jamesnetherton:
I will fix up the failing JDBC example project test post merging this.

Issue #4097, opened 2022-09-15
Deprecated parameters in several annotations are ignored - inconsistent with other annotations.


JiriOndrusek:
Camel currently supports annotations (like `@Produce`, `@Consume`, ..) with either parameter called `value` or depricated parameter `uri`.
In camel-quarkus there is an inconsistency.  I.E `@Consume` supports both parameters, `@EndointInject` supports only `value` parameter.

Camel-quarkus should be consistent and support both parameters (as is in Camel) in similar annotations.

JiriOndrusek:
Here is the code for `@Consume`, which already exists - https://github.com/apache/camel-quarkus/blob/main/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/ConsumeProcessor.java#L172-L175

PR #4099, opened 2022-09-15
Deprecated parameters in several annotations are ignored - inconsisten

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4097

### DIFF
--- a/extensions-core/core/deployment/src/test/java/org/apache/camel/quarkus/core/deployment/main/CamelMainDeprecatedConsumeTest.java
+++ b/extensions-core/core/deployment/src/test/java/org/apache/camel/quarkus/core/deployment/main/CamelMainDeprecatedConsumeTest.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.core.deployment.main;
+
+import io.quarkus.test.QuarkusUnitTest;
+import org.apache.camel.Consume;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+public class CamelMainDeprecatedConsumeTest {
+    @RegisterExtension
+    static final QuarkusUnitTest CONFIG = new QuarkusUnitTest().setExpectedException(IllegalArgumentException.class)
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
+
+    @Consume(uri = "direct:consumeAnnotation")
+    public String consumeAnnotation(String name) {
+        return "Consumed " + name;
+    }
+
+    @Test
+    public void consumeAnnotationWithDeprecatedParamsThrowsIllegalArgumentException() {
+        // Noop - we expect IllegalArgumentException to be thrown on application startup
+    }
+}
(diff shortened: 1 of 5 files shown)


ppalaga:
I am perhaps missing something, but why should we add support for deprecated stuff?
The supported cases are documented in https://camel.apache.org/camel-quarkus/2.12.x/user-guide/cdi.html#_endpointinject_and_produce

llowinge:
@ppalaga Problem is as follows. Using

```
@EndpointInject(id = "mock:result")
    protected MockEndpoint resultEndpoint;

    @Produce(id = "direct:start")
    protected ProducerTemplate template;
 ``` fails with `java.lang.IllegalArgumentException: uri must be specified and not empty.`. I'm not sure if it is bug in Camel or CQ. But even using deprecated stuff shouldn't break it IMO. I think if some feature is deprecated + it doesn't work, it should be removed.

**Comment by JiriOndrusek** [comment]
From my PoV support of deprecated parameters is not worth of a fix - I agree. But the stat of the code is, that user may use `@Consume` with deprecated parameter, but if `@Produce` or `@EndpountInject` is used with deprecated parameter, execution fails. The doc from the only says, that the same approach as in plain Camel can used -but plain Caml allows deprecated parameters.

**Comment by ppalaga** [comment]
> fails with java.lang.IllegalArgumentException: uri must be specified and not empty.. I'm not sure if it is bug in Camel or CQ. But even using deprecated stuff shouldn't break it IMO. I think if some feature is deprecated + it doesn't work, it should be removed.

The evaluation of the annotation params happens (or is missing in) in Camel Quarkus. I'd vote for throwing a better error message (something like `@org.apache.camel.EndpointInject(uri = "my-component:...") is not supported on Camel Quarkus. Please replace it with just @org.apache.camel.EndpointInject("my-component:...")`. We could output a similar warning for `@Consume`. WDYT?

**Comment by JiriOndrusek** [comment]
I agree with that approach and will change the PR accordingly. 

@llowinge Do you agree?

**Comment by llowinge** [comment]
@JiriOndrusek I agree with @ppalaga.

**Comment by JiriOndrusek** [comment]
PR is changed to throw an IllegalArgumentException with appropriate message + tests covering these use cases are added as well.

File: extensions-core/core/deployment/src/test/java/org/apache/camel/quarkus/core/deployment/main/CamelMainDeprecatedConsumeTest.java [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+    static final QuarkusUnitTest CONFIG = new QuarkusUnitTest().setExpectedException(IllegalArgumentException.class)
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
+
+    @Consume(uri = "direct:consumeAnnotation")
+    public String consumeAnnotation(String name) {
+        return "Consumed " + name;
+    }
+
+    @Test
+    public void camelMainEventBridgeNotConfigured() {
```
jamesnetherton, line 37: If the empty test is required to make the tests run, please can we name it something relevant to the thing under test. E.g:

```
    @Test
    public void consumeAnnotationWithDeprecatedParamsThrowsIllegalArgumentException() {
        // Noop - we expect IllegalArgumentException to be thrown on application startup
    }
```

extensions-core/core/deployment/src/test/java/org/apache/camel/quarkus/core/deployment/main/CamelMainDeprecatedEndpointInjectTest.java
```diff
@@ -0,0 +1,47 @@
...
+public class CamelMainDeprecatedEndpointInjectTest {
+    @RegisterExtension
+    static final QuarkusUnitTest CONFIG = new QuarkusUnitTest().setExpectedException(IllegalArgumentException.class)
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
+
+    @EndpointInject(uri = "direct:start")
+    DirectEndpoint start;
+
+    @Test
+    public void camelMainEventBridgeNotConfigured() {
```
jamesnetherton, line 37: See my previous comment about the empty test method.

extensions-core/core/deployment/src/test/java/org/apache/camel/quarkus/core/deployment/main/CamelMainDeprecatedProduceTest.java
```diff
@@ -0,0 +1,47 @@
...
+public class CamelMainDeprecatedProduceTest {
+    @RegisterExtension
+    static final QuarkusUnitTest CONFIG = new QuarkusUnitTest().setExpectedException(IllegalArgumentException.class)
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
+
+    @Produce(uri = "direct:start")
+    FluentProducerTemplate produceProducerFluent;
+
+    @Test
+    public void camelMainEventBridgeNotConfigured() {
```
jamesnetherton, line 37: See my previous comment about the empty test method.

extensions-core/core/deployment/src/test/java/org/apache/camel/quarkus/core/deployment/main/CamelMainDeprecatedConsumeTest.java
JiriOndrusek: fixed.

extensions-core/core/deployment/src/test/java/org/apache/camel/quarkus/core/deployment/main/CamelMainDeprecatedEndpointInjectTest.java
JiriOndrusek: fixed

JiriOndrusek:
@jamesnetherton Would it be possible to merge this fix also to `2.13.x`. To make sure it makes into any patch which may be released?

jamesnetherton (approved):

Issue #4076, opened 2022-09-07
[Camel 3.19.0] Quartz: solve autowired scheduler properly and remove quick workaround


JiriOndrusek:
Change in Camel  (https://issues.apache.org/jira/browse/CAMEL-18143) is not compatible with the Quarkus Quartz extension.

Once the scheduler is autowired, scheduler from the Quarkus is instantiated and detects a problem (see [code](https://github.com/quarkusio/quarkus/blob/main/extensions/quartz/runtime/src/main/java/io/quarkus/quartz/runtime/QuartzScheduler.java#L340)).

Ideal solution would be to not produce a scheduler if there is no configuration requiring it.

If the right solution is done, please remove quick workaround from the [application.properties](https://github.com/apache/camel-quarkus/blob/camel-main/integration-tests/quartz/src/main/resources/application.properties#L23) of the Quartz test module.

```
#quick workaround caused by https://issues.apache.org/jira/browse/CAMEL-18143
camel.component.quartz.autowired-enabled = false
```

JiriOndrusek:
The same workaround is appliad alsi in file tests. It has to be removed once problem is solved properly - https://github.com/apache/camel-quarkus/blob/camel-main/integration-tests/file/src/main/resources/application.properties

JiriOndrusek:
Enhancement proposed on Quarkus as https://github.com/quarkusio/quarkus/issues/27929

PR #4119, opened 2022-09-21
Quartz: solve autowired scheduler properly and remove quick workaroun…

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4076

I changed the default value of [quarkus.quartz.start-mode](https://quarkus.io/guides/quartz#quarkus-quartz_quarkus.quartz.start-mode) to `forced`

```
If set to "forced", scheduler will be started even if no scheduled business methods are found. This is necessary for "pure" programmatic scheduling.
````

This mode seems to be the right one for the camel component. If user needs to use different mode, it is possible.
I mentioned this change in Quartz doc.

With this change, there is no need to change Quarkus extension (see https://github.com/quarkusio/quarkus/issues/27929)


oscerd (approved):

JiriOndrusek:
@jamesnetherton I think that this solution is the right one and does not need any fix on Quarkus extension.

jamesnetherton:
I think we should consider this a temporary fix. It forces the Quarkus quartz scheduler to always be autowired into the Camel component, which I don't think is what we want as the default if no `quarkus.quartz` properties were defined (I.e Camel should create the scheduler itself in that scenario).

JiriOndrusek:
> I think we should consider this a temporary fix. It forces the Quarkus quartz scheduler to always be autowired into the Camel component, which I don't think is what we want as the default if no `quarkus.quartz` properties were defined (I.e Camel should create the scheduler itself in that scenario).

Hi @jamesnetherton  I proposed to the quarkus an "enhancment" to not create Quartz scheduler in case tat there is no `quarkus.quartz.*` property (or @Scheduled), but it seems, that this approach is not possible (see https://github.com/quarkusio/quarkus/pull/28168)

I'll try to find a solution which does not require above change in quarkus.

Issue #4085, opened 2022-09-13
[Camel 3.19.0] Splunk: tests are failing after the upgrade of Splunk in Camel


JiriOndrusek:
Splunk was upgraded in https://github.com/apache/camel/commit/d991f82b50cc3c536cffec48156f9169dc448ec3
(currently the issue is happening on `camel-main` branch)

Tests in Splunk extension have to be fixed.
- version of docker image should be bumped to `splunk/splunk:9.0.1`
- even with that change there is a problem probably with login (version in splunk.Service is not initialized)

Enable tests once fixed.

PR #4124, opened 2022-09-22
Splunk: tests are failing after the upgrade of Splunk in Camel #4085

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4085

### DIFF
--- a/integration-tests/splunk/src/main/java/org/apache/camel/quarkus/component/splunk/it/SplunkResource.java
+++ b/integration-tests/splunk/src/main/java/org/apache/camel/quarkus/component/splunk/it/SplunkResource.java
@@ -82,7 +82,7 @@ public class SplunkResource {
     @Produces(MediaType.APPLICATION_JSON)
     public List normal(String search) throws Exception {
         String url = String.format(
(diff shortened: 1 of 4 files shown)


oscerd (approved):

aldettinger:
It looks the upgrade in camel is linked to TLS version. Are we covering TLS in those splunk tests ?

JiriOndrusek:
This fix is meant to fix the failure int the current tests, which was caused by the upgrade.
I looked into the test and ssl is turned off  (see [this line](https://github.com/apache/camel-quarkus/blob/main/integration-tests/splunk/src/test/java/org/apache/camel/quarkus/component/splunk/it/SplunkTestResource.java#L54).

I'll create another ticket to extend the coverage of the splunk extension by covering TLD.

aldettinger (approved):
Ok, many thanks for double checking @JiriOndrusek :+1: . So, I see the coverage is the same than before and that's good.

Concerning the need to create another ticket for TLS, maybe it makes sense as TLS is handled  a bit special in native mode if I recall correctly.

jamesnetherton:
@JiriOndrusek can we redo this one with just the Splunk test changes on `main`?

JiriOndrusek:
Can be merged.

PR #4133, opened 2022-09-26
Improve test coverage for scheduler component.

svkcemk:

### DIFF
--- a/integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerResource.java
+++ b/integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerResource.java
@@ -26,14 +26,24 @@ import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 
-import org.apache.camel.ProducerTemplate;
-
 @Path("/scheduler")
 @ApplicationScoped
 public class SchedulerResource {
 
     @Inject
-    ProducerTemplate producerTemplate;
+    @Named("withDelayCounter")
+    AtomicInteger withDelayCounter;
+
+    @Inject
+    @Named("useFixedDelayCounter")
+    AtomicInteger useFixedDelayCounter;
+
+    @Named("withDelayRepeatCounter")
+    AtomicInteger withDelayRepeatCounter;
+
+    @Inject
+    @Named("greedyCounter")
+    AtomicInteger greedyCounter;
 
     @Inject
     @Named("schedulerCounter")
@@ -46,10 +56,66 @@ public class SchedulerResource {
         return schedulerCounter.get();
     }
 
+    @Path("/get-delay-count")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public int getCountDelay() {
+        return withDelayCounter.get();
+    }
+
+    @Path("/get-fixed-delay-count")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public int getCountFixedDelay() {
+        return useFixedDelayCounter.get();
+    }
+
+    @Path("/get-repeat-count")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public int getRepeatCount() {
+        return withDelayRepeatCounter.get();
+    }
+
+    @Path("/get-greedy-count")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public int getGreedyCount() {
+        return greedyCounter.get();
+    }
+
     @javax.enterprise.inject.Produces
     @ApplicationScoped
     @Named("schedulerCounter")
     AtomicInteger schedulerCounter() {
+        return new AtomicInteger();
+    }
+
+    @javax.enterprise.inject.Produces
+    @ApplicationScoped
+    @Named("withDelayRepeatCounter")
+    AtomicInteger withDelayRepeatCounter() {
+        return new AtomicInteger();
+    }
+
+    @javax.enterprise.inject.Produces
+    @ApplicationScoped
+    @Named("withDelayCounter")
+    AtomicInteger withDelayCounter() {
+        return new AtomicInteger();
+    }
+
+    @javax.enterprise.inject.Produces
+    @ApplicationScoped
+    @Named("useFixedDelayCounter")
+    AtomicInteger useFixedDelayCounter() {
+        return new AtomicInteger();
+    }
+
+    @javax.enterprise.inject.Produces
+    @ApplicationScoped
+    @Named("greedyCounter")
+    AtomicInteger greedyCounter() {
         return new AtomicInteger();
     }
 
--- a/integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerRoute.java
+++ b/integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerRoute.java
@@ -31,10 +31,37 @@ public class SchedulerRoute extends RouteBuilder {
     @Named("schedulerCounter")
     AtomicInteger schedulerCounter;
 
+    @Inject
+    @Named("withDelayCounter")
+    AtomicInteger withDelayCounter;
+
+    @Inject
+    @Named("useFixedDelayCounter")
+    AtomicInteger useFixedDelayCounter;
+
+    @Inject
+    @Named("withDelayRepeatCounter")
+    AtomicInteger withDelayRepeatCounter;
+
+    @Inject
+    @Named("greedyCounter")
+    AtomicInteger greedyCounter;
+
     @Override
     public void configure() throws Exception {
-        from("scheduler:start?initialDelay=1")
+        from("scheduler:withInitialDelay?initialDelay=1")
                 .process(e -> schedulerCounter.incrementAndGet());
 
+        from("scheduler:withDelay?delay=100")
+                .process(e -> withDelayCounter.incrementAndGet());
+
+        from("scheduler:useFixedDelay?initialDelay=200&useFixedDelay=true")
+                .process(e -> useFixedDelayCounter.incrementAndGet());
+
+        from("scheduler:withDelayRepeat?delay=1&repeatCount=5")
+                .process(e -> withDelayRepeatCounter.incrementAndGet());
+
+        from("scheduler:greedy?delay=100&greedy=true")
+                .process(e -> greedyCounter.incrementAndGet());
     }
 }
--- a/integration-test-groups/foundation/scheduler/src/test/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerTest.java
+++ b/integration-test-groups/foundation/scheduler/src/test/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerTest.java
@@ -28,11 +28,47 @@ import static org.awaitility.Awaitility.await;
 class SchedulerTest {
 
     @Test
-    public void test() throws Exception {
-        // wait until the scheduler has run and return a counter that is > 0
+    public void testInitialDelay() throws Exception {
         await().atMost(5, TimeUnit.SECONDS).until(() -> {
             String body = RestAssured.get("/scheduler/get").then().statusCode(200).extract().body().asString();
             return !body.equals("0");
+        });
+    }
+
+    @Test
+    public void testDelay() throws Exception {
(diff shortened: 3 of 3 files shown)


oscerd (approved):

integration-test-groups/foundation/scheduler/src/test/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerTest.java
```diff
@@ -28,12 +28,43 @@
...
+        });
+    }
+
+    @Test
+    public void testDelayWithFixed() throws Exception {
+        await().atMost(4, TimeUnit.SECONDS).until(() -> {
+            String body = RestAssured.get("/scheduler/get-count").then().statusCode(200).extract().body().asString();
+            return Integer.parseInt(body) > 1;
+        });
+    }
```
ppalaga, line 52: It looks like you assume `testDelay()` to be executed before `testDelayWithFixed()` ? Maybe one test method would serve better?

```suggestion
    @Test
    public void testDelay() throws Exception {
        await().atMost(2, TimeUnit.SECONDS).until(() -> {
            String body = RestAssured.get("/scheduler/get-count").then().statusCode(200).extract().body().asString();
            return Integer.parseInt(body) > 1;
        });
        await().atMost(2, TimeUnit.SECONDS).until(() -> {
            String body = RestAssured.get("/scheduler/get-count").then().statusCode(200).extract().body().asString();
            return Integer.parseInt(body) > 2;
        });
    }
```

integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/ShoppingList.java
```diff
@@ -0,0 +1,43 @@
...
+    public void addItems() {
+        int item = new Random().nextInt(100);
+        lists.add(item);
+    }
+
+    public void removeItems() {
+        if (lists != null && lists.size() > 0) {
+            lists.remove(lists.size() - 1);
+        }
+    }
```
ppalaga, line 38: This does not seem to be used. So we may remove it, no?

integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/ShoppingList.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public List<Integer> getItems() {
+        return lists;
+    }
```
ppalaga, line 42: There is a potential race between `getItems()` and `addItems()` - you may get concurrent modification exception when accessing the list via `getItems()`. You may want to switch from ArrayList to java.util.concurrent.CopyOnWriteArrayList.

integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/ShoppingList.java
```diff
@@ -0,0 +1,43 @@
...
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Named;
+
+@ApplicationScoped
+@Named("shoppingList")
+public class ShoppingList {
+    List<Integer> lists = new ArrayList<>();
+
+    public void addItems() {
```
ppalaga, line 29: You are adding just a single item.
```suggestion
    public void addItem() {
```

integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/GreedyCounter.java
```diff
@@ -0,0 +1,34 @@
...
+public class GreedyCounter {
+    private int count = 0;
+
+    public int getCount() {
+        return count;
+    }
+
+    public void increment() {
+        count++;
+    }
```
ppalaga, line 33: Note that getCount and increment() can be called from differen threads concurrently. While I do not think it matters much in how we call them from tests currently, it is still potentially dangerous (end users may copy this and use differently, we may extend the test in the future). So producing a named AtomicInteger is perhaps a better choice.

integration-test-groups/foundation/scheduler/pom.xml
```diff
@@ -34,6 +34,10 @@
         <dependency>
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-scheduler</artifactId>
+        </dependency>
+            <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-bean</artifactId>
         </dependency>
```
ppalaga, line 41: I'd vote for doing without this additional dependency. We could e.g. have a an ad hoc processor to increment the counter, something like:

```
.process(e -> myCounter.incrementAndGet());
```

integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerRoute.java
```diff
@@ -30,11 +31,24 @@ public class SchedulerRoute extends RouteBuilder {
     @Inject
     @Named("schedulerCounter")
     AtomicInteger schedulerCounter;
+    @Inject
+    CamelContext context;
```
ppalaga, line 35: The context field does not seem to be used. Can we perhaps remove it?

integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerRoute.java
```diff
@@ -30,11 +31,24 @@ public class SchedulerRoute extends RouteBuilder {
...
     public void configure() throws Exception {
-        from("scheduler:start?initialDelay=1")
+        from("scheduler:withInitialDelay?initialDelay=1")
                 .process(e -> schedulerCounter.incrementAndGet());
 
+        from("scheduler:withDelay?delay=1000")
+                .to("bean:myCounter?method=increment");
+
+        from("scheduler:useFixedDelay?initialDelay=2000&useFixedDelay=true")
+                .to("bean:myCounter?method=increment");
```
ppalaga, line 46: I'd vote for using shorter delays, to save some testing time. 200ms should be fine. At the same time, we should not check using `==` in the tests, only `<`, for the case that the test is too slow to observe the exact value.

ppalaga (commented):
Thanks @svkcemk , some comments inline.

aldettinger:
The Camel quarkus documentation is talking about [forcing scheduler to be idle](https://camel.apache.org/components/3.18.x/scheduler-component.html#_forcing_the_scheduler_to_be_idle).

Would it make sense to test this one ? Maybe that's another PR.

integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/GreedyCounter.java
aldettinger: And maybe the proposal is even to produce one `AtomicInteger` named `greedyCounter` and a second one named `mycounter`

aldettinger:
That's great start @svkcemk , keep up the good work.
Also note that CI issues might not be related to this PR.

svkcemk:
Thanks @ppalaga @aldettinger for review, sorry for late update ( I was on PTO last week ) updated the PR based o the review. @aldettinger I am creating a separate PR for the case : Forcing the scheduler to be idle.

ppalaga:
Sorry, no time to review, because I am on a conference. If @aldettinger finds this good, I find it good either :)

integration-test-groups/foundation/scheduler/src/test/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerTest.java
```diff
@@ -28,12 +28,42 @@
...
         await().atMost(5, TimeUnit.SECONDS).until(() -> {
             String body = RestAssured.get("/scheduler/get").then().statusCode(200).extract().body().asString();
             return !body.equals("0");
         });
     }
 
+    @Test
+    public void testDelay() throws Exception {
+
+        await().atMost(2, TimeUnit.SECONDS).until(() -> {
```
aldettinger, line 41: Could the body could be > 1 even if withDelay is never called ?
At the end of the day, we could maybe better to separate `withDelay` and `useFixedDelay`.
So, 2 distinct tests and 2 distinct atomic integers.
What do you think ?

integration-test-groups/foundation/scheduler/src/test/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerTest.java
```diff
@@ -28,12 +28,42 @@
...
+            String body = RestAssured.get("/scheduler/get-count").then().statusCode(200).extract().body().asString();
+            return Integer.parseInt(body) > 2;
+        });
+    }
+
+    @Test
+    public void testDelayWithRepeat() throws Exception {
+        await().atMost(4, TimeUnit.SECONDS).until(() -> {
+            String body = RestAssured.get("/scheduler/get-items").then().statusCode(200).extract().body().asString();
+            return body.split(",").length >= 4;
```
aldettinger, line 56: Here, we expect a list of items >= 4. Can't we do that with a simple AtomicInteger ?

integration-test-groups/foundation/scheduler/src/test/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerTest.java
svkcemk: @aldettinger do you mean two separate Atomic int like myCounter or its just during testing two separate int one for withDelay and one for withFixedDelay ?

aldettinger: The main idea is that, from my point of view, the current implementation of this test could pass even when the `withDelay` route is never called.
From there, one solution could be to have two distinct AtomicInteger, let's say `withDelayCounter` and `useFixedDelayCounter`. And then, it's possible to check that each route has been invoked separately.

integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerResource.java
```diff
@@ -33,11 +31,15 @@
...
+    AtomicInteger withDelayCounter;
 
-    @Named("shoppingList")
-    CopyOnWriteArrayList<Integer> shoppingList;
+    @Inject
+    @Named("useFixedDelayCounter")
+    AtomicInteger useFixedDelayCounter;
+
+    @Named("withDelayRepeat")
+    AtomicInteger withDelayRepeat;
```
aldettinger, line 42: Maybe `withDelayRepeatCounter` would align with other names ? What do you think ?

integration-test-groups/foundation/scheduler/src/test/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerTest.java
```diff
@@ -28,12 +28,43 @@
...
 
+    @Test
+    public void testDelay() throws Exception {
+        await().atMost(2, TimeUnit.SECONDS).until(() -> {
+            String body = RestAssured.get("/scheduler/get-delay-count").then().statusCode(200).extract().body()
+                    .asString();
+            return Integer.parseInt(body) > 2;
+        });
+
+        await().atMost(2, TimeUnit.SECONDS).until(() -> {
```
aldettinger, line 46: And so, at the end of the day. Those 2 tests are independent ? So we could better separate them back in tests ? Like `testDelay` and `testFixedDelay` ?

integration-test-groups/foundation/scheduler/src/main/java/org/apache/camel/quarkus/component/scheduler/it/SchedulerResource.java
svkcemk: Yes sounds good!

aldettinger (approved):
Cool, it's already rebased to a single commit.
Let's check the build and we can merge then.
Well done @svkcemk :clap:

aldettinger:
The failing test is not related. Merging then.

PR #4144, opened 2022-09-28
Increase test coverage of ref extension

jamesnetherton:


zhfeng (approved):

aldettinger:
@jamesnetherton This PR should work. I have few questions below if you have time.

This PR introduce a second route in tests: Are the 2 routes testing exactly the same scenario ?

Also, we switch from requestBody+direct to sendBody+direct+seda: Is the purpose to test a kind of subtlety of the ref component ?

jamesnetherton:
> This PR introduce a second route in tests: Are the 2 routes testing exactly the same scenario ?

Sort of. One case uses the `from("endpont-bean-id")` / `to(endpont-bean-id")` idiom. The other case explicitly specifies the `ref` scheme like `from("ref:endpont-bean-id")` / `to(ref:endpont-bean-id")`.

> Also, we switch from requestBody+direct to sendBody+direct+seda: Is the purpose to test a kind of subtlety of the ref component?

Not really - we just need to test routing with `from` & `to` so figured it was a simple enough way of doing that.

aldettinger (approved):
Ok, thanks for sharing explanation and well done :+1:

PR #4183, opened 2022-10-14
Exclude com.github.stephenc.jcip:jcip-annotations from Azure extensions

jamesnetherton:


oscerd (approved):

aldettinger (approved):

ppalaga:
@jamesnetherton could you please explain, why is the exclusion needed? Don't we perhaps want to ban jcip-annotations altogether?

jamesnetherton:
> Why is the exclusion needed

To avoid some dependency convergence check errors. I don't think the dependency has much value on the runtime classpath so figured it was ok if it were excluded.

>  Don't we perhaps want to ban jcip-annotations altogether?

Yes we can try to do that. I can open an issue.

PR #4204, opened 2022-10-24
[closes #4203] Add tests for CXF SOAP MTOM

llowinge:

PR #4206, opened 2022-10-24
CXF cleanup

ppalaga:
* camel-quarkus-cxf-soap should depend on quarkus-cxf-rt-features-logging because camel-cxf-soap depends on cxf-rt-features-logging
* Rename ws-security test to-ws-security-client test, better separate package names, test names and property names among the individual cxf test modules so that they do not clash when run together


ppalaga:
Replaced by #4204

PR #4207, opened 2022-10-24
More CXF tests and cleanup

### LINKED ISSUE #4203: Create test coverage for CXF SOAP MTOM

ppalaga:
fix #4203


oscerd (approved):

ppalaga:
@llowinge I have added some fixes for MTOM in native mode, most notably https://github.com/apache/camel-quarkus/pull/4207/files#diff-7c81397449fd40a8260c63c1f182320f8aab02b7e56879589ffce1d46b36df6fR35-R38 and https://github.com/apache/camel-quarkus/pull/4207/files#diff-a6565d3563e7493df31d39fe68eba12b8d5892af636f96541f78d22e385cac8fR72-R74 but it still fails. 
So I filed https://github.com/apache/camel-quarkus/issues/4208 and disabled the test in native mode https://github.com/apache/camel-quarkus/pull/4207/files#diff-e9e596e55a8e034e9d04f6b83ba0a06cc87694d14824540234ece03404e73082R37 
Feel free to try to fix it.
I think it would be worth to try to see an MTOM test with java.awt.Image working on plain quarkus-cxf. If it won't work there, then we know that it is an AWT problem rather than CXF problem. Otherwise it will point in direction CQ. Note that a DataHandler MTOM test is already there in quarkus-cxf: https://github.com/quarkiverse/quarkus-cxf/tree/main/integration-tests/mtom

llowinge:
I've fixed it with not using the awt.* classes in the test app (it is needed for the test scenario anyway). I was unable to work on that with macOS because of https://github.com/oracle/graal/issues/4124. Fix is in commit https://github.com/llowinge/camel-quarkus/commit/2fae5ab2019b7278bc84ffdab8e6ccfdf7332953.

Issue #4084, opened 2022-09-13
Springless JPA extension


zhfeng:

zhfeng:
We have to revert the current fix due to https://github.com/apache/camel-quarkus/issues/4279. It is because we introduce a fake `PlatformTransactionManager` to make graal native building happy. But it is incompatiable with the other extensions such as `camel-quarkus-jms` which depends on `spring-tx`.

Possible changes with came-jpa:

- Deprecated PlatformTransactionManger on camel-jpa component level in 3.x and remove it in camel 4.x
- All related set/get method should be moved to DefaultTransactionStrategy which depends on spring-tx to do the transaction operations (commit / rollback)
- Discuss these changes with the upstream community at first

ppalaga:
@zhfeng is there any chance to get the required changes to Camel 4?

zhfeng:
Yeah, I already deprecated `transactionManager` in [CAMEL-18742](https://issues.apache.org/jira/browse/CAMEL-18742) but have not got a chance to remove it in Camel 4. I will create a ticket on Camel to track it.

zhfeng:
https://issues.apache.org/jira/browse/CAMEL-19225

zhfeng:
I open an upstream PR to remve `transactionManager`
- https://github.com/apache/camel/pull/10954

zhfeng:
This should be fixed.

PR #4280, opened 2022-11-17
Fix #4279 to import PlatformTransactionManager completely in camel-qu…

### LINKED ISSUE #4279: JMS consumer in a XA router can not work with JPA
I'm trying to incorporate JMS consumer in a XA router such lik:
```
from("xajms:outbound")
                .transacted()
                .log("Message out: ${body}")
                .to("bean:auditLog?method=createAuditLog(${body}-ok)")
                .to("jpa:org.acme.AuditLog");
``` 

And a XA JMS Component could be 

```
 @Produces
    @Named("xajms")
    public JmsComponent xaJms() {
        JmsComponent component = new JmsComponent();
        component.setConnectionFactory(connectionFactory);
        component.setTransacted(false);

        JtaTransactionManager jtaTransactionManager = new JtaTransactionManager(userTransaction, transactionManager);
        jtaTransactionManager.setTransactionSynchronizationRegistry(transactionSynchronizationRegistry);
        component.setTransactionManager(jtaTransactionManager);
        return component;
    }
```
But it can not receive any message from the queue and the warn messaging is
```
2022-11-17 10:21:31,014 WARN  [org.apa.cam.com.jms.DefaultJmsMessageListenerContainer] (Camel (camel-1) thread #1 - JmsConsumer[outbound]) Setup of JMS message listener invoker failed for destination 'outbound' - trying to recover. Cause: 'org.springframework.transaction.TransactionStatus org.springframework.transaction.PlatformTransactionManager.getTransaction(org.springframework.transaction.TransactionDefinition)': java.lang.NoSuchMethodError: 'org.springframework.transaction.TransactionStatus org.springframework.transaction.PlatformTransactionManager.getTransaction(org.springframework.transaction.TransactionDefinition)'
	at org.springframework.jms.listener.AbstractPollingMessageListenerContainer.receiveAndExecute(AbstractPollingMessageListenerContainer.java:243)
	at org.springframework.jms.listener.DefaultMessageListenerContainer$AsyncMessageListenerInvoker.invokeListener(DefaultMessageListenerContainer.java:1237)
	at org.springframework.jms.listener.DefaultMessageListenerContainer$AsyncMessageListenerInvoker.executeOngoingLoop(DefaultMessageListenerContainer.java:1227)
	at org.springframework.jms.listener.DefaultMessageListenerContainer$AsyncMessageListenerInvoker.run(DefaultMessageListenerContainer.java:1120)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)

```

I think the root cause is that we introduced [PlatformTransactionManager](https://github.com/apache/camel-quarkus/blob/main/extensions/jpa/runtime/src/main/java/org/springframework/transaction/PlatformTransactionManager.java) in `camel-quarkus-jpa` which is just to make the graal native building happy to do the static analyst.  So it needs to import `PlatformTransactionManager` interface completely since the `camel-jms` has to depend on `spring-tx`.

It might be an other better way to do such things with native building. @jamesnetherton @ppalaga WDYT?

zhfeng:
…arkus-jpa

### DIFF
--- a/extensions/jpa/runtime/src/main/java/org/springframework/transaction/SavepointManager.java
+++ b/extensions/jpa/runtime/src/main/java/org/springframework/transaction/SavepointManager.java
@@ -14,12 +14,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.springframework.transaction;
 
-public interface PlatformTransactionManager {
-    TransactionStatus getTransaction(TransactionDefinition definition) throws TransactionException;
+public interface SavepointManager {
+    Object createSavepoint() throws TransactionException;
 
-    void commit(TransactionStatus status) throws TransactionException;
+    void rollbackToSavepoint(Object savepoint) throws TransactionException;
 
-    void rollback(TransactionStatus status) throws TransactionException;
+    void releaseSavepoint(Object savepoint) throws TransactionException;
 }
--- a/extensions/jpa/runtime/src/main/java/org/springframework/transaction/StaticTransactionDefinition.java
+++ b/extensions/jpa/runtime/src/main/java/org/springframework/transaction/StaticTransactionDefinition.java
@@ -14,12 +14,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.springframework.transaction;
 
-public interface PlatformTransactionManager {
-    TransactionStatus getTransaction(TransactionDefinition definition) throws TransactionException;
+final class StaticTransactionDefinition implements TransactionDefinition {
+    static final StaticTransactionDefinition INSTANCE = new StaticTransactionDefinition();
 
-    void commit(TransactionStatus status) throws TransactionException;
-
-    void rollback(TransactionStatus status) throws TransactionException;
+    private StaticTransactionDefinition() {
+    }
 }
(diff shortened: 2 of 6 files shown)


oscerd (approved):

jamesnetherton:
We should probably open an issue to rethink how we're handling the removal of Spring from the JPA extension.

With the current way, I wonder if we're creating a potential maintenance headache? E.g if the Spring APIs change in the future, do we need to update these placeholder classes and interfaces?

zhfeng:
Yeah, it is exactly what I'm worrying about. In case of APIs changes, I think we need to update these classes and interfaces otherwise `camel-jms` or othere components which depend on `spring-tx` will be failing when combining with `camel-jpa`.

jamesnetherton:
Wondering whether an alternative 'fix' might be to revert the Spring removal from the JPA extension? That would give us some more time to think through a better solution for the future. WDYT?

zhfeng:
It suppressed by https://github.com/apache/camel-quarkus/pull/4286

zhfeng:
@jamesnetherton I've reopened https://github.com/apache/camel-quarkus/issues/4084 to think about removing spring from JPA.

PR #4295, opened 2022-11-24
docs: Fix some quarkus.camel property names

christophd:

### DIFF
--- a/docs/modules/ROOT/pages/reference/extensions/core.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/core.adoc
@@ -135,7 +135,7 @@ When set to true, the `CamelRuntime` will be started automatically.
 | `boolean`
 | `true`
 
-|icon:lock[title=Fixed at build time] [[quarkus.camel.service.discovery.exclude-patterns]]`link:#quarkus.camel.service.discovery.exclude-patterns[quarkus.camel.service.discovery.exclude-patterns]`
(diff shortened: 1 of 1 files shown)


jamesnetherton:
Does `quarkus.camel.service.discovery.exclude-patterns` not work?

The config property doc sections are auto generated by some Quarkus tooling. So I think this change would get overwritten when the build runs.

ppalaga:
I think the proposal of @christophd is indeed correct. There is a separate `.service.` node in the config: https://github.com/apache/camel-quarkus/blob/main/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java#L128-L140

... and @jamesnetherton is also right that the docs is generated. So we need to fix the generator code.

ppalaga:
I am looking into it.

The first step is done by annotation processor `io.quarkus:quarkus-extension-processor` plugged to the compiler plugin. 
The result is stored in `target/asciidoc/generated/config/all-configuration-roots-generated-doc/org.apache.camel.quarkus.core.CamelConfig` of the source tree root dir. The prop names seem to be correct there: 

```
...
 {
      "configDocKey":{
         "type":"string",
         "key":"quarkus.camel.service.discovery.exclude-patterns",
...
```

Those files are then picked by `update-extension-doc-page` of `camel-quarkus-maven-plugin`. Let my try to debug it.

christophd:
sorry for the confusion. I was getting confused by `quarkus.camel.routes-discovery` and another property that was provided to me with another `-`. Reviewing the config structure I think `quarkus.camel.service.discovery.*` is correct as you guys have suggested.

I think we can safely close this one then

PR #4307, opened 2022-11-30
Test Java first CXF server endpoint with multiple SEI methods 

### LINKED ISSUE #4306: Test Java first CXF server endpoint with multiple SEI methods
Citing @llowinge 

>I don't have the wsdl (will be autogenerated by cxf), but I have the interface and the implementation which is returned via `to("bean:implementor")`.
>Ideally i would like the interface to have atleast two methods, so we verify the mapping from `from(cxf)` and its soap request will be handled correctly to desired method.

ppalaga:
Fix #4306


llowinge:
LGTM

oscerd (approved):

PR #4312, opened 2022-12-02
Manage jakarta.xml.soap-api and sync jakarta.ws properties with CXF

jamesnetherton:
@ppalaga Not sure if it'd be better to move these to the `quarkus-cxf` BOM? They were added originally to support `camel-quarkus-soap`.


aldettinger (approved):

ppalaga:
> @ppalaga Not sure if it'd be better to move these to the quarkus-cxf BOM? They were added originally to support camel-quarkus-soap.

+1, see https://github.com/apache/camel-quarkus/issues/4321

Issue #4196, opened 2022-10-19
Expand AWS CW test coverage


JiriOndrusek:
The following scenarios from the [documentation](https://camel.apache.org/components/3.18.x/aws2-cw-component.html) are not covered in Camel Quarkus:

(Producer only)
- different ways of configuring credentials (may not be possible to cover) - [link](https://camel.apache.org/components/3.18.x/aws2-cw-component.html#_static_credentials_vs_default_credential_provider)
- message headers for sending information - [link](https://camel.apache.org/components/3.18.x/aws2-cw-component.html#_static_credentials_vs_default_credential_provider)
- custom client - [link](https://camel.apache.org/components/3.18.x/aws2-cw-component.html#_advanced_cloudwatchclient_configuration)

PR #4345, opened 2022-12-16
Expand AWS CW test coverage #4196

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4196

Different ways of configuring credentials are not covered. (It is covered by https://github.com/apache/camel-quarkus/issues/4346)

### DIFF
--- a/integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwResource.java
+++ b/integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwResource.java
@@ -17,19 +17,26 @@
 package org.apache.camel.quarkus.component.aws2.cw.it;
 
 import java.net.URI;
+import java.time.Instant;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 import javax.ws.rs.Consumes;
+import javax.ws.rs.HeaderParam;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.Response;
 
 import io.quarkus.scheduler.Scheduled;
+import org.apache.camel.CamelExecutionException;
 import org.apache.camel.ProducerTemplate;
+import org.apache.camel.component.aws2.cw.Cw2Constants;
 
 @Path("/aws2-cw")
 @ApplicationScoped
@@ -57,6 +64,49 @@ public class Aws2CwResource {
             @PathParam("metric-name") String name,
             @PathParam("metric-unit") String unit) throws Exception {
         endpointUri = "aws2-cw://" + namespace + "?name=" + name + "&value=" + value + "&unit=" + unit;
+        return Response
+                .created(new URI("https://camel.apache.org/"))
+                .build();
+    }
+
+    @Path("/send-metric-map/{namespace}")
+    @POST
+    @Consumes("application/x-www-form-urlencoded")
+    @Produces(MediaType.TEXT_PLAIN)
+    public Response postMap(
+            @PathParam("namespace") String namespace,
+            @HeaderParam("customClientName") String customClientName,
+            MultivaluedMap<String, String> formParams) throws Exception {
+
+        String uri = "aws2-cw://" + namespace;
+        uri = customClientName != null && !customClientName.isEmpty()
+                ? uri + "?autowiredEnabled=false&amazonCwClient=#" + customClientName : uri;
+
+        Map<String, Object> typedHeaders = formParams.entrySet().stream().collect(Collectors.toMap(
+                e -> e.getKey(),
+                e -> {
+                    final String val = e.getValue().get(0);
+                    if (Cw2Constants.METRIC_TIMESTAMP.equals(e.getKey())) {
+                        return Instant.ofEpochMilli(Long.parseLong(val));
+                    } else if (Cw2Constants.METRIC_VALUE.equals(e.getKey())) {
+                        return Long.parseLong(val);
+                    } else if (Cw2Constants.METRIC_DIMENSIONS.equals(e.getKey())) {
+                        String[] keyVal = val.split("=");
+                        return Map.of(keyVal[0], keyVal[1]);
+                    }
+                    return val;
+                }));
+        try {
+            producerTemplate.requestBodyAndHeaders(uri, null, typedHeaders, String.class);
+        } catch (Exception e) {
+            if (e instanceof CamelExecutionException && e.getCause() != null) {
+                return Response
+                        .ok(e.getCause().getMessage())
+                        .build();
+            }
+            throw e;
+        }
+
         return Response
                 .created(new URI("https://camel.apache.org/"))
                 .build();
--- a/integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/ClientBuilder.java
+++ b/integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/ClientBuilder.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.aws2.cw.it;
+
+import javax.enterprise.inject.Produces;
+import javax.inject.Named;
+
+import org.apache.camel.component.aws2.cw.CloudWatchClientMock;
+import software.amazon.awssdk.services.cloudwatch.CloudWatchClient;
+
+public class ClientBuilder {
+
+    @Named("customClient")
+    CloudWatchClient produceClient() {
+        return new CloudWatchClientMock();
+    }
+
+    @Produces
+    @Named("customClient2")
+    CloudWatchClient produceClient2() {
+        //second client is required to avoid autowiring of the first client if there is only one client
+        //workaround because of https://github.com/apache/camel-quarkus/issues/4400
+        return new CloudWatchClientMock();
+    }
+}
--- a/integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/CloudWatchClientMock.java
+++ b/integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/CloudWatchClientMock.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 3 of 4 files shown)


oscerd (approved):

integration-test-groups/aws2/aws2-cw/src/test/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwTest.java
```diff
@@ -90,4 +97,148 @@ public void metric() {
...
+                        return false;
+                    }
+
+                    Datapoint oddDp = oddDatapoints.get(0);
+                    Datapoint evenDp = evenDatapoints.get(0);
+
+                    if (!(oddDp.sampleCount().intValue() == 2
+                            && oddDp.minimum().intValue() == 2 * value + 1
+                            && oddDp.maximum().intValue() == 2 * value + 3)) {
+                        throw new RuntimeException("Unexpected odd datapoint " + oddDp
```
aldettinger, line 204: Is there any specific reason to throw RuntimeException instead of using assert* methods ?

integration-test-groups/aws2/aws2-cw/pom.xml
```diff
@@ -39,6 +39,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-resteasy</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jackson</artifactId>
```
aldettinger, line 44: @ppalaga If I remember well, we try our best not to use `quarkus-resteasy-jackson` in tests because it registers some native reflections that could potentially hide bug in our extensions ? If yes, what could we advise in the test at hand ?

JiriOndrusek: @aldettinger  I can refactor the test to use a "custom" serialization. I tried that in this case, but it seems like a potential cause of errors. Maybe we could add some kind of "util", "helper" class to serialize and deserialize maps into some test extension, what do you think?

integration-test-groups/aws2/aws2-cw/src/test/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwTest.java
JiriOndrusek: I'm asserting the returned message in the test class. It doesn't feel good to call assert from runtime (not test) scope.

integration-test-groups/aws2/aws2-cw/src/test/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwTest.java
```diff
@@ -90,4 +97,148 @@ public void metric() {
...
+                Cw2Constants.METRIC_UNIT, "Count"));
+
+        RestAssured.given()
+                .contentType(ContentType.JSON)
+                .queryParam("customClientName", "customClient")
+                .body(data)
+                .post("/aws2-cw/send-metric-maps/" + namespace)
+                .then()
+                .statusCode(200)
+                .body(is("MockedClient"));
```
JiriOndrusek, line 241: @aldettinger Here is assertion for the result of the mocked client exception from previous comment.

integration-test-groups/aws2/aws2-cw/pom.xml
aldettinger: Yes, we should come with an idea that improve the situation, if feasible. On top of my mind, I don't have much. Would passing a Map<String, String> works in plain text ? Maybe we could then add long/int as extra parameters... Not sure it would make it.

integration-test-groups/aws2/aws2-cw/src/test/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwTest.java
aldettinger: Ok, think I get that. It's actually a `pollInterval()` block.

integration-test-groups/aws2/aws2-cw/src/test/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwTest.java
aldettinger: ok, caught. Many thanks.

JiriOndrusek:
@aldettinger I created a helper class `SerializationUtil`, which is able to serialize/deserialize maps, list and some primitive types (String, integer, long).
I didn't find a right location for such code, so I created a simple module for it (without dependencies - to not affect native build). Helper method can be used by different modules.

@ppalaga , @jamesnetherton , @aldettinger Once we agreed that this kind of solution is acceptable, I'll add support to more primitive types and extend the test coverage. I'd like personally  to create a separated issue for such feature and a separated PR for it. This is only a "working preview" to discuss whether we want to use this. I'm changing this PR to draft for above reasons.

JiriOndrusek:
Here is a commit with the serialization feature and its use in aws2-cw - 6c42c54aec56d7bbf9abc8488a02baa9aab5b0ca

ppalaga:
I'd vote for keeping this as simple as possible. The serialization task gets a bit easier when we pass a single map per HTTP request instead of sending a list of maps at once. We can do that because we loop over the list-of-maps entries in the REST handler anyway. A map can be smuggled to plain RESTeasy in many ways: via QueryParams, Headers or URL-encoded body, which I tried to implement here: https://github.com/ppalaga/camel-quarkus/commits/pr4345 What do others think?

aldettinger:
The proposal based on plain resteasy seems to me more readable and easier to maintain. I wonder what other think.

I wonder what key words we could search to find an existing library that would do this. Maybe a non reflective java serialization library.

JiriOndrusek:
I agree with the simple solution with plain RESTeasy. Solution is straightforward and easily understandable. 

I'll amend the change from https://github.com/ppalaga/camel-quarkus/commits/pr4345 into this PR. Do you agree, @aldettinger @ppalaga .

JiriOndrusek:
Change is amended, thanks @ppalaga for the simple solution.

integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/ClientBuilder.java
```diff
@@ -0,0 +1,875 @@
...
+    @Named("customClient")
+    CloudWatchClient produceClient() {
+
+        return new MockedClient();
+    }
+
+    @Produces
+    @Named("customClient2")
+    CloudWatchClient produceClient2() {
+        //second client is required to avoid autowiring of the first client if there is only one client
```
jamesnetherton, line 143: Bit of a nitpick - but can we not just set `camel.component.aws2-cw.autowired-enabled=false` in `application.properties`?

ppalaga (approved):

integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwResource.java
```diff
@@ -61,4 +68,48 @@ public Response post(
...
+    public Response postMap(
+            @PathParam("namespace") String namespace,
+            @HeaderParam("customClientName") String customClientName,
+            MultivaluedMap<String, String> formParams) throws Exception {
+
+        String uri = "aws2-cw://" + namespace;
+        uri = customClientName != null && !customClientName.isEmpty()
+                ? uri + "?autowiredEnabled=false&amazonCwClient=#" + customClientName : uri;
+
+        //use Instant as timestamp header
```
aldettinger, line 85: Does this comment apply to the whole set of instructions below ?

integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/ClientBuilder.java
```diff
@@ -0,0 +1,875 @@
...
+        return new MockedClient();
+    }
+
+    class MockedClient implements CloudWatchClient {
```
aldettinger, line 147: Do we need to re-implement the default methods ?

In Camel, it looks possible not do to so:
https://github.com/apache/camel/blob/main/components/camel-aws/camel-aws2-cw/src/test/java/org/apache/camel/component/aws2/cw/CloudWatchClientMock.java

integration-test-groups/aws2/aws2-cw/pom.xml
JiriOndrusek: fixed

integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/ClientBuilder.java
JiriOndrusek: Problem in camel prevents that -> fix was already proposed  - https://issues.apache.org/jira/browse/CAMEL-18878
I reported an issue to remove this workaround once change in camel is consumed by cq. - https://github.com/apache/camel-quarkus/issues/4400

integration-test-groups/aws2/aws2-cw/src/test/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwTest.java
```diff
@@ -90,4 +97,148 @@ public void metric() {
...
+                Cw2Constants.METRIC_NAMESPACE, namespace, Cw2Constants.METRIC_NAME, metricName,
+                Cw2Constants.METRIC_VALUE, 2 * value + 3,
+                Cw2Constants.METRIC_UNIT, "Count",
+                Cw2Constants.METRIC_DIMENSION_NAME, "type",
+                Cw2Constants.METRIC_DIMENSION_VALUE, "odd"));
+        // ignored because of timestamp
+        data.add(CollectionHelper.mapOf(
+                Cw2Constants.METRIC_NAMESPACE, namespace, Cw2Constants.METRIC_NAME, metricName,
+                Cw2Constants.METRIC_VALUE, 2 * value + 5,
+                Cw2Constants.METRIC_TIMESTAMP, System.currentTimeMillis() - 24 * 60 * 60 * 10000,
```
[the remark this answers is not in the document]
JiriOndrusek: I fixed the test to use 24h * 60 * 60s * 1000.

I'm not sure whether using fixed value is the right approach here. What if real service somehow limits the timestamp-metric? (for example performance wise - to not save old data) I tried simple run with fixed time of 1970-01-06T19:04:15.080Z (long - 24 * 60 * 60 * 1000 * 1000) and execution with real service failed (local execution was successful). Therefore I think that using metrics as is supposed to be - using current time +/- 1 day is a correct way. What do you think, @aldettinger ?

integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwResource.java
JiriOndrusek: removed

integration-test-groups/aws2/aws2-cw/src/test/java/org/apache/camel/quarkus/component/aws2/cw/it/Aws2CwTest.java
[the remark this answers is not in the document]
aldettinger: Looks it make sense. The real service is in need of real data.

integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/CloudWatchClientMock.java
```diff
@@ -0,0 +1,43 @@
...
+package org.apache.camel.component.aws2.cw;
+
+import software.amazon.awssdk.services.cloudwatch.CloudWatchClient;
+import software.amazon.awssdk.services.cloudwatch.model.PutMetricDataRequest;
+import software.amazon.awssdk.services.cloudwatch.model.PutMetricDataResponse;
+
+public class CloudWatchClientMock implements CloudWatchClient {
+
+    public CloudWatchClientMock() {
+        System.out.println(1);
```
aldettinger, line 26: Maybe this println could be removed ?

aldettinger (approved):
Looks good. Many thanks to all for collaborative improvement of this pr :clap:

integration-test-groups/aws2/aws2-cw/src/main/java/org/apache/camel/quarkus/component/aws2/cw/it/CloudWatchClientMock.java
JiriOndrusek: oh, I missed that, will be removed, thanls!

JiriOndrusek: removed

PR #4359, opened 2022-12-23
Ref #4358: Java jOOR DSL native support

### LINKED ISSUE #4358: Java jOOR DSL native support
[To support routes written in Java language in Camel-K](https://github.com/apache/camel-k/issues/3023), we have to make the extension `java-joor-dsl` compatible with the native mode.

essobedo:
fixes #4358

## Motivation

[To support routes written in Java language in Camel-K](https://github.com/apache/camel-k/issues/3023), we have to make the extension `java-joor-dsl` compatible with the native mode.

## Modifications

The main idea is to compile the java resources using jOOR at build time and ignore runtime compilation requests.

* Compile java resources using jOOR and declare the result as generated
* Register the default constructor of the generated classes for reflection
* Substitute the body of the method `JavaRoutesBuilderLoader#doLoadRoutesBuilders` to do nothing at runtime
* Substitute the body of the method `ClassRoutesBuilderLoader#doLoadRoutesBuilders` to throw an exception at runtime 
* Simplify code allowing iterating over matching resources

## Result

Routes written in Java resources can be compiled using jOOR and included in the native image.

**Note:** I had to use an ugly workaround to provide the classpath, I intend to change the code in Camel to clean it up later.


essobedo:
Unless I miss something, the build failure is not related to those changes since the extension `java-joor-dsl` is part of the group 13 while the group 03 is failing

integration-tests/foundation-grouped/pom.xml
```diff
@@ -96,12 +96,10 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-log</artifactId>
         </dependency>
-        <!-- https://github.com/apache/camel-quarkus/issues/4225
```
ppalaga, line 99: camel-quarkus-microprofile-fault-tolerance should be kept out otherwise you'll keep getting the failure in foundation-grouped/pom.xml

extensions/java-joor-dsl/deployment/src/main/java/org/apache/camel/quarkus/dsl/java/joor/deployment/JavaJoorDslProcessor.java
```diff
@@ -0,0 +1,181 @@
...
+                        unit.addClass(name, content);
+                        nameToResource.put(name, resource);
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
+                    }
+                });
+        if (nameToResource.isEmpty()) {
+            return;
+        }
+        // Temporary workaround until being able to provide the classpath somehow to the method MultiCompile.compileUnit
```
ppalaga, line 110: Could you please paste an issue or PR URL here so that future readers know what is this blocked by?

extensions/java-joor-dsl/deployment/src/main/java/org/apache/camel/quarkus/dsl/java/joor/deployment/JavaJoorDslProcessor.java
```diff
@@ -0,0 +1,181 @@
...
+                    }
+                    try (InputStream is = resource.getInputStream()) {
+                        generatedClass
+                                .produce(new JavaJoorGeneratedClassBuildItem(asClassName(resource), resource.getLocation(),
+                                        is.readAllBytes()));
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
+                    }
+                });
+    }
```
ppalaga, line 85: Why would anybody want to pass .class files as resources? Is this also supported by plain Camel?

ppalaga: Isn't Java actually loading those classes without our extension needing to do anything?

ppalaga:
@essobedo very nice! I have left some minor questions and comments inline.

In addition to that, I'd vote for 

* Removing the dependency on quarkus-resteasy-jsonb in the test. We generally prefer a minimal set of dependencies in integration tests to avoid the risk that the additional dependency fixes some native issue in the test, but otherwise the extension under test would not work without it. We have seen this happening with quarkus-resteasy-jackson, maybe it would not happen with quarkus-resteasy-jsonb, but anyway, it is easy to do without by splitting to context lookups into multiple REST endpoints as I tried to implement here: https://github.com/ppalaga/camel-quarkus/commit/ef0361fe28c261c5d621fc2ab07433c58a4aa7a3#diff-050e5976f7d86cda20a440f5326cc7fd3f6dbe5ae9f52700c53329a4da1033f7R46-R73
* Adding a test that checks that the jOOR route really works - something like this: https://github.com/ppalaga/camel-quarkus/commit/ef0361fe28c261c5d621fc2ab07433c58a4aa7a3#diff-050e5976f7d86cda20a440f5326cc7fd3f6dbe5ae9f52700c53329a4da1033f7R75-R81

If you like the above proposal feel free to amend the commit https://github.com/ppalaga/camel-quarkus/commit/ef0361fe28c261c5d621fc2ab07433c58a4aa7a3 to this PR.

extensions/java-joor-dsl/deployment/src/main/java/org/apache/camel/quarkus/dsl/java/joor/deployment/JavaJoorDslProcessor.java
essobedo: > Why would anybody want to pass .class files as resources? Is this also supported by plain Camel?

Yes it also supported by Camel as you can see [here](https://github.com/apache/camel/blob/main/dsl/camel-java-joor-dsl/src/main/java/org/apache/camel/dsl/java/joor/ClassRoutesBuilderLoader.java)

ppalaga: Hm... https://issues.apache.org/jira/browse/CAMEL-18014 says 

> When compiling with camel-java-dsl we can have it capture the compiled byte code, which allows us to write to disk, which we can make possible as a build step, that allows to load the routes from disk and avoid the compilation step at runtime.

@davsclaus is that use case described somewhere in more details? I mean what is build time and runtime for Camel JBang? I mostly wonder why would the build time part need to pass the .class files as resources to the runtime? Why can runtime simply pick those via stock java class path? Does this use case make any sense on Quarkus?

ppalaga:
Thanks @essobedo, the last open thing for me is the `.class` resource loading. IMO, we should either

A. not support it at all (throw an exception if somebody tries to use it) or

B. If the `.class` resource loading makes sense on Quarkus, then there should be a test covering it.

extensions/java-joor-dsl/runtime/src/main/java/org/apache/camel/quarkus/dsl/java/joor/runtime/JavaJoorDslRecorder.java
```diff
@@ -0,0 +1,54 @@
...
+                    // inject context and resource
+                    CamelContextAware.trySetCamelContext(obj, context.getValue());
+                    ResourceAware.trySetResource(obj, ResourceHelper.fromString(location, ""));
+                    context.getValue().addRoutes((RoutesBuilder) obj);
+                }
+            } catch (Exception e) {
+                throw new RuntimeCamelException("Cannot create instance of class: " + className, e);
+            }
+        }
+    }
```
aldettinger, line 53: Would there be value in logging/warning about non matched items ?

essobedo:
@ppalaga ok let's see the answer of @davsclaus and I will make the corresponding changes

extensions/java-joor-dsl/deployment/src/main/java/org/apache/camel/quarkus/dsl/java/joor/deployment/JavaJoorDslProcessor.java
essobedo: I guess, I can remove the support of class files to keep it simple

davsclaus:
The .class is from https://issues.apache.org/jira/browse/CAMEL-18014 back at the time when camel-jbang and (indirectly also camel-k) could benefit from pre-loading java routes from .class that was compiled via java dsl (joor).

However since we have less of that need as camel-jbang now exports to standard quarkus/SB and allow it to compile java dsl as regular java.

The current Camel K however still has a limitation as all the DSL are runtime loaded and "compiled", also .java so there is a big cost to run java routes in camel-k as on startup. 

Even if there is a special Camel Quarkus build-step then Camel K cannot benefit from that, as it bootup a "empty" Camel.

I think we can remove the .class also in camel core for java-joor dsl as its not really needed anymore.

extensions/java-joor-dsl/runtime/src/main/java/org/apache/camel/quarkus/dsl/java/joor/runtime/JavaJoorDslRecorder.java
essobedo: Done, thx for the feedback

extensions/java-joor-dsl/deployment/src/main/java/org/apache/camel/quarkus/dsl/java/joor/deployment/JavaJoorDslProcessor.java
essobedo: class files support removed

essobedo:
@davsclaus thx for your answer

> Even if there is a special Camel Quarkus build-step then Camel K cannot benefit from that, as it bootup a "empty" Camel.

I know, I have already something working to make it possible at Camel-K level but first, I need it to be released and a new version of camel-k-runtime depending on this new version of camel-quarkus

> 
> I think we can remove the .class also in camel core for java-joor dsl as its not really needed anymore. I think (I am getting older so its maybe no longer there) there is an option in camel-main to configure a "compile directory" so removing would also need to remove that.

I removed it from Camel Quarkus for now

essobedo:
> I think we can remove the .class also in camel core for java-joor dsl as its not really needed anymore. I think (I am getting older so its maybe no longer there) there is an option in camel-main to configure a "compile directory" so removing would also need to remove that.

The related ticket https://issues.apache.org/jira/browse/CAMEL-18864, I can do it asap

ppalaga (approved):
Thanks for the explanation, @davsclaus!

Thanks again for a great contribution, @essobedo!

jamesnetherton (approved):
Nice work @essobedo :sunglasses:

aldettinger (approved):

essobedo:
As it is my first contribution to the project, I'm not really aware of the validation process, I assume that someone from the team will merge it once the process is complete. Let me know if it is not the case.

Issue #4362, opened 2022-12-31
Multiple test methods do not work in a class extending CamelQuarkusTestSupport 


ppalaga:
Originally reported by @guhilling via chat.

A reproducer project is here: https://github.com/guhilling/camel-quarkus-demos

ppalaga:
@JiriOndrusek would you like to have a look?

JiriOndrusek:
Yes, I'll look into it.

JiriOndrusek:
The problem is caused by the usage of the `javax.enterprise.inject.Produces` in the test (line https://github.com/guhilling/camel-quarkus-demos/blob/main/src/test/java/de/hilling/camel/SampleRouteBuilderTest.java#L35). 

The `CamelQuarkusTestSupport` aims to help replace Camel's CamelTestSupport tests. `@Produce` can not be used with them and CamrlQuarkusTestSupport does not work with them easily. Using @Produce to create routeBuilders is not recommanded.

Here are some problems caused by the differences of lifecycle in Camel and CamelQuarkus:

• Application is started once for all the tests together. Therefore all @Produces methods are applied together - in case there were several Test classes each producing its own route builder - this will force application to create all routes at the start (even if they are not necessary for the first test class)
• Support of `CamelTestSupport` like tests forces camel-quarkus to emulate several lifecycle actions without stopping the application. Which leads for example to stopping and removing routes and creating them again for another test method/classes.
• . Fully supported style of registering RouteBuilders is to use overriding method. 


Please replace following code (lines https://github.com/guhilling/camel-quarkus-demos/blob/main/src/test/java/de/hilling/camel/SampleRouteBuilderTest.java#L33-L39)

```
    @Configuration
    public static class TestConfig {
        @Produces
        RoutesBuilder route() {
            return new SampleRouteBuilder();
        }
    }
```

with 

```
    @Override
    protected RouteBuilder createRouteBuilder() {
        return new SampleRouteBuilder();
    }
```

and the tests will finish successfully.

I've added the suggestion to not use @Produce with RouteBuilders int the documentation.

guhilling:
Thanks for the hint! Unfortunately now I need to put the "advising" (AdviceWith.adviceWith(context() ...) into the test-methods directly as it won't work in "doBeforeEach()" ... Any idea for that?

JiriOndrusek:
Hi @guhilling  thanks for spotting this. I refactored my fork obviously a little bit more to make the test work (and didn;t notice this last change).

[...]

PR #4365, opened 2023-01-03
Test coverage for produced RouteBuilder + doc suggestion

JiriOndrusek:
Test coverage and explanation for https://github.com/apache/camel-quarkus/issues/4362

More information in the [comment](https://github.com/apache/camel-quarkus/issues/4362#issuecomment-1369885091).

PR #4370, opened 2023-01-04
DoBeforeEach does not work with Advice #4362

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4362

Originally reported issue problem is not a bug -> this is intentional (forced) behaviour.
Second problem -> Advice not working in @BeforeEach method is na issue, which is fixed in this PR.

Tests covering both cases  and a small change in the documentation are added.

### DIFF
--- a/test-framework/junit5-extension-tests/src/test/java/org/apache/camel/quarkus/test/extensions/producedRouteBuilder/ProducedRouteBuilderET.java
+++ b/test-framework/junit5-extension-tests/src/test/java/org/apache/camel/quarkus/test/extensions/producedRouteBuilder/ProducedRouteBuilderET.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.test.extensions.producedRouteBuilder;
+
+import javax.enterprise.inject.Produces;
+
+import io.quarkus.test.junit.QuarkusTest;
+import org.apache.camel.builder.AdviceWith;
+import org.apache.camel.builder.AdviceWithRouteBuilder;
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.quarkus.test.CamelQuarkusTestSupport;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+public class ProducedRouteBuilderET extends CamelQuarkusTestSupport {
+
+    @BeforeEach
+    public void doSomethingBefore() throws Exception {
+        AdviceWith.adviceWith(this.context, "sampleRoute",
+                ProducedRouteBuilderET::enhanceRoute);
+    }
+
+    @Produces
+    public RouteBuilder routes() {
+        return new RouteBuilder() {
+            @Override
+            public void configure() {
+                from("direct:in").routeId("sampleRoute").to("file:target/data/RouteBuilderET?filename=hello_true.txt");
+            }
+        };
+    }
+
+    @Test
+    public void firstTest() throws Exception {
+        Assertions.assertTrue(true);
+    }
+
+    @Test
+    public void secondTest() throws Exception {
+        Assertions.assertTrue(true);
+    }
+
+    private static void enhanceRoute(AdviceWithRouteBuilder route) {
+        route.replaceFromWith("direct:ftp");
+        route.interceptSendToEndpoint("file:.*samples.*")
+                .skipSendToOriginalEndpoint()
+                .to("mock:file:sample_requests");
+    }
+
+}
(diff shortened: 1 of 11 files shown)


JiriOndrusek:
Replaces https://github.com/apache/camel-quarkus/pull/4365

test-framework/junit5-extension-tests/src/test/java/org/apache/camel/quarkus/test/extensions/producedRouteBuilder/ProducedRouteBuilderET.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.camel.builder.AdviceWith;
+import org.apache.camel.builder.AdviceWithRouteBuilder;
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.quarkus.test.CamelQuarkusTestSupport;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+public class ProducedRouteBuilderET extends CamelQuarkusTestSupport {
```
aldettinger, line 31: What's the meaning of ET ? it is IT ?

test-framework/junit5-extension-tests/src/test/java/org/apache/camel/quarkus/test/extensions/producedRouteBuilder/ProducedRouteBuilderTest.java
```diff
@@ -0,0 +1,61 @@
...
+import io.quarkus.test.QuarkusDevModeTest;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+/**
+ * Test for https://github.com/apache/camel-quarkus/issues/4362
```
aldettinger, line 31: 4362 seems to be linked to "multiple test methods" bug. Is it the case here ?

test-framework/junit5-extension-tests/src/test/java/org/apache/camel/quarkus/test/extensions/producedRouteBuilder/ProducedRouteBuilderTest.java
```diff
@@ -0,0 +1,61 @@
...
+                    return ShrinkWrap.create(JavaArchive.class).addClasses(ProducedRouteBuilderET.class);
+                }
+            });
+
+    @Test
+    public void checkTests() {
+        ContinuousTestingTestUtils utils = new ContinuousTestingTestUtils();
+        ContinuousTestingTestUtils.TestStatus ts = utils.waitForNextCompletion();
+
+        Assertions.assertEquals(1L, ts.getTestsFailed());
```
aldettinger, line 58: Could you please explain why we expect 1 failing test and the other passing ? I would expect both `firstTest` and `secondTest` to pass as the instruction is just `Assertions.assertTrue(true);` for both of them.

JiriOndrusek: The reason of the failure is the issue reported here https://github.com/apache/camel-quarkus/issues/4362
As you can see, RouteBuilder is registered with @Produces (which is not recommended). In a lot of cases with multiple test methods (e.g. his test) , test framework removes the routes after the first test but can not call RouteBuilder. Which means that the second test will fail with the reason, can not advice non existing route. You can see more detailed description in this [comment](https://github.com/apache/camel-quarkus/issues/4362#issuecomment-1369885091).

test-framework/junit5-extension-tests/src/test/java/org/apache/camel/quarkus/test/extensions/producedRouteBuilder/ProducedRouteBuilderET.java
JiriOndrusek: ET means extension test. (or at least I suppose it is this abbreviation).
You can see several cases in Quarkus [codebase](https://github.com/quarkusio/quarkus/tree/main/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/testrunner)

These *ET classes are used in the test classes to create a tested application (for example this [line](https://github.com/quarkusio/quarkus/blob/main/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/testrunner/TestRunnerSmokeTestCase.java#L35))

The *ET java file can not be run directly, so it is not *Test or *IT. *ET seems like a nice solution, as the test executes another tests using devmode (new QuarkusDevModeTest()).

test-framework/junit5-extension-tests/src/test/java/org/apache/camel/quarkus/test/extensions/producedRouteBuilder/ProducedRouteBuilderTest.java
JiriOndrusek: I think that it is better to cover also this negative scenario. In case that something is changed (because lifecycle of Quarkus is different then Camel's one) and this test fails -> we should carefully review whether there is no other impact.

aldettinger: Ok, I see. This works as expected and so enforced by a test.

aldettinger (approved):
Thanks for explanation.

I wonder whether it would be possible to detect/warn/block cases where `@Produces` is used with `RouteBuilder` in a subclass of `CamelQuarkusTestSupport`. That would be another ticket anyway.

JiriOndrusek:
Good idea, I'll think about that, that could make the user experience much better.

ppalaga:
@JiriOndrusek thanks for taking care for this!

In the light of current changes, is there anything to add/improve in https://camel.apache.org/camel-quarkus/2.15.x/user-guide/testing.html#_cameltestsupport_style_of_testing ?

JiriOndrusek:
@ppalaga I made 2 changes into the doc 

https://github.com/apache/camel-quarkus/pull/4370/files#diff-b0192eeed7b3641688fc3b73fa7c01e066a6038d0565eed34a679fa171798694L261
https://github.com/apache/camel-quarkus/pull/4370/files#diff-b0192eeed7b3641688fc3b73fa7c01e066a6038d0565eed34a679fa171798694L263

PR #4376, opened 2023-01-05
Remove dependency on camel-health from camel-quarkus-core

jamesnetherton:
@JiriOndrusek do you recall why the override for `createHealthCheckRegistry` was added to `FastCamelContext` in the 3.20 upgrade? I think it's not needed and enough to inherit the default implementation in `SimpleCamelContext`:

https://github.com/apache/camel/blame/3bcc047dcb54b064df5de3beacc4af57971a47f6/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/SimpleCamelContext.java#L131-L140

### DIFF
--- a/extensions-core/core/runtime/pom.xml
+++ b/extensions-core/core/runtime/pom.xml
@@ -85,10 +85,6 @@
             <artifactId>camel-microprofile-config</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.apache.camel</groupId>
-            <artifactId>camel-health</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.jboss.spec.javax.xml.bind</groupId>
             <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
         </dependency>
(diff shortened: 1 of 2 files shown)


JiriOndrusek:
I vaguely remember that without the method, some kind of aws test was failing.  HealthCheckRegistry could not be found therefore some data was missing. I was related to  https://issues.apache.org/jira/browse/CAMEL-18131 and this [line](https://github.com/apache/camel/blame/main/components/camel-aws/camel-aws2-sns/src/main/java/org/apache/camel/component/aws2/sns/Sns2Endpoint.java#L117) if I'm not mistaken

jamesnetherton:
> some kind of aws test was failing

Thanks for clarifying.

I think it was enough to add [this](https://github.com/apache/camel-quarkus/blob/main/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelProcessor.java#L181) and enable the health check discovery to work. Lets see what the CI result is.

In any case, the logic in `Sns2Endpoint` is flawed because it proceeds to try and register a health check when `healthCheckRepository` is potentially `null`.

aldettinger (approved):

PR #4385, opened 2023-01-06
Ref #4379: Groovy DSL native support

### LINKED ISSUE #4379: Groovy DSL native support
To support routes written in Groovy language in Camel-K, we have to make the extension `groovy-dsl` compatible with the native mode.

essobedo:
fixes #4379 

## Motivation

To support routes written in Groovy language in Camel-K, we have to make the extension `groovy-dsl` compatible with the native mode.

## Modifications

The main idea is to compile the groovy resources at build time and ignore runtime compilation requests.

* Convert the groovy scripts into groovy classes to be able to compile them as normal classes
* Compile groovy resources and declare the result as generated
* Substitute the body of the method `GroovyRoutesBuilderLoader#doLoadRoutesBuilders` to do nothing at runtime
* Add a set of substitutions to get rid of or replace unneeded code.
* Add a note to indicate that Groovy extensions are not yet supported

## Result

Routes written in Groovy resources can be compiled and included in the native image.


integration-tests/groovy-dsl/pom.xml
```diff
@@ -39,6 +39,26 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-direct</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-seda</artifactId>
```
jamesnetherton, line 44: Please can you make sure these `org.apache.camel-quarkus` dependencies are also in the `virtualDependencies` profile (you can add them automatically with `mvn process-resources -Pformat -N`).

integration-tests/groovy-dsl/pom.xml
```diff
@@ -100,6 +120,36 @@
...
+            <activation>
+                <property>
+                    <name>native</name>
+                </property>
+            </activation>
+            <properties>
+                <quarkus.package.type>native</quarkus.package.type>
+                <!-- To prevent getting DeletedElementException: Unsupported method java.lang.Class.registerNatives()
+                is reachable for some use cases -->
+                <quarkus.native.additional-build-args>--report-unsupported-elements-at-runtime</quarkus.native.additional-build-args>
```
jamesnetherton, line 134: Is this suppressing potential issues that users may see?

zhfeng: I think we have to handle this carefully. So users need to have `--report-unsupported-elements-at-runtime` in the applications in native build? Is there any chance to remove it in the future?

integration-tests/groovy-dsl/pom.xml
essobedo: It should be good now, thx

aldettinger:
Nice job @essobedo :+1:
There are few nitpicks comments if it make sense.

`--report-unsupported-elements-at-runtime` might be the only showstopper, let's see how the discussion goes

integration-tests/groovy-dsl/pom.xml
essobedo: Yeah, I know, fortunately, this flag is only needed for some use cases not all. I may be wrong but I'm afraid that it can only be fixed at Graal VM level, indeed the method `Class#registerNatives` is private/native and it is called only on static initialization of the class `Class` and I don't know any way to prevent that, IMHO it should never occur.

essobedo: If it is really a showstopper for you, the best I can do is to remove the use cases that need this flag and add a note in the limitations or simply keep it like this and add a note in the doc indicating in which cases it could be needed. WDYT?

essobedo: Please also note that since it should never occur, no exception will ever be thrown at runtime.

zhfeng: It should be better to remove this flag and add a note in the limitations for such use cases which need it. Also if

- https://github.com/quarkusio/quarkus/issues/2720 get resolved and can we have a chance to leverage on it?

essobedo: I've just pushed some changes that remove the flag and add a note in the limitations

essobedo: Supporting Groovy directly in Quarkus won't be easy, I believe that they will need to limit the support to Groovy classes that can be statically compiled which is a big limitation. Here, as we are already limited to route definitions, I did not have to add this limitation.

zhfeng: Thanks @essobedo and it looks good to me.

zhfeng (approved):

jamesnetherton (approved):

Issue #4023, opened 2022-08-22
perf-regression: PerfRegressionIT should use `java.home` system property for the java executable location


jamesnetherton:
PerfRegressionIT currently makes the assumption that whatever `java` binary is found on the `PATH`, is the one that should be used in the test. That may not always be the case.

It'd probably be safer to look up the value of the `java.home` system property and use the same JDK that is running the test.

PR #4390, opened 2023-01-09
perf-regression: try resolving 'java.home' system property first in itest #4023

aldettinger:


aldettinger:
@jamesnetherton FYI

oscerd (approved):

jamesnetherton:
My original intention was to use the `java.home` System property as it'll point to the JDK that the test is running under and is always guaranteed to be present.

aldettinger:
Ok, the pr has been updated and now uses `java.home` system property.

jamesnetherton (approved):

PR #4403, opened 2023-01-12
Ref #4392: Kotlin DSL native support

### LINKED ISSUE #4392: Kotlin DSL native support
To support routes written in Kotlin language in Camel-K, we have to make the extension `kotlin-dsl` compatible with the native mode.

essobedo:
fixes #4392 

## Motivation

To support routes written in Kotlin language in Camel-K, we have to make the extension `kotlin-dsl` compatible with the native mode.

## Modifications

The main idea is to compile the Kotlin resources at build time and ignore runtime compilation requests.

* Convert the Kotlin scripts into Kotlin classes to be able to compile them as normal classes
* Compile Kotlin resources thanks to `KotlinCompilationProvider` and declare the result as generated
* Substitute the body of the method `KotlinRoutesBuilderLoader#doLoadRoutesBuilders` to do nothing at runtime

## Result

Routes written in Kotlin resources can be compiled and included in the native image.


essobedo:
The failing test is not related to those changes

oscerd (approved):

integration-tests/kotlin-dsl/pom.xml
```diff
@@ -100,6 +124,44 @@
...
+                <property>
+                    <name>native</name>
+                </property>
+            </activation>
+            <properties>
+                <quarkus.package.type>native</quarkus.package.type>
+            </properties>
+            <build>
+                <plugins>
+                    <!-- remove me -->
```
jamesnetherton, line 139: I guess this can be removed to reinstate the sanity-checks execution.

integration-tests/kotlin-dsl/pom.xml
```diff
@@ -39,6 +39,30 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-direct</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-seda</artifactId>
```
jamesnetherton, line 44: Please regenerate the `virtualDependencies` profile if the corresponding `-deployment` dependencies are not present.

integration-tests/kotlin-dsl/pom.xml
essobedo: Oops nice catch 👍

integration-tests/kotlin-dsl/pom.xml
essobedo: You make me realize that I launched the command from the wrong module, thx for the feedback

jamesnetherton (approved):

extensions/kotlin-dsl/deployment/src/main/java/org/apache/camel/quarkus/dsl/kotlin/deployment/KotlinDslProcessor.java
```diff
@@ -0,0 +1,224 @@
...
+
+    @BuildStep(onlyIf = NativeBuild.class)
+    void compileScriptsAOT(BuildProducer<GeneratedClassBuildItem> generatedClass,
+            BuildProducer<GeneratedResourceBuildItem> generatedResource,
+            BuildProducer<ReflectiveClassBuildItem> reflectiveClass,
+            BuildProducer<KotlinGeneratedClassBuildItem> generatedKotlinClass,
+            CurateOutcomeBuildItem curateOutcomeBuildItem) throws Exception {
+        LOG.debug("Loading .kts resources");
+        final Path projectDir = Paths.get(".").toAbsolutePath().normalize();
+        final Path generatedSourceDir = projectDir.resolve("target/kotlin-dsl/generated-sources");
```
ppalaga, line 94: This might fail on Gradle or even with some non-standard Maven setup. 
`io.quarkus.deployment.pkg.builditem.BuildSystemTargetBuildItem` or something similar provided by Quarkus Core could perhaps deliver safer results.

essobedo: Nice one, thx for sharing

ppalaga (commented):
Great work, @essobedo, I especially like the test coverage!

I wonder what's the idea behind the timer to log test routes? They do not seem to be producing any testable effect, no? Would any test fail if any of those routes fails during processing? 

Moreover, given that there are so many of them, I guess the log is quite hard to follow because each of those routes is concurrently outputting its messages.

I wonder whether those could be transformed to some sort of http request -> direct -> (do something) -> http response. That would produce less messages in the log and the tests could do some assertions on the HTTP responses.

I also wonder whether we could reduce the number of maven dependencies? We could perhaps do without timer. If we intend to test Camel REST then we could perhaps use it is a testing bridge in each test, and get rid of RestEasy? 

The rest of deps is perhaps hard to reduce without sacrificing the coverage. If so, we could perhaps make use of [Grouping](https://camel.apache.org/camel-quarkus/2.15.x/contributor-guide/extension-testing.html#_grouping) that IMO suits well this situation when we'd like to test one extension in combination with some others. A follow up PR for this would be nice.

essobedo:
> I wonder what's the idea behind the timer to log test routes? They do not seem to be producing any testable effect, no? Would any test fail if any of those routes fails during processing?
> 
> Moreover, given that there are so many of them, I guess the log is quite hard to follow because each of those routes is concurrently outputting its messages.

The goal here is more to test that the route can be started and executed properly, what is actually tested is the code before the route definition.

> I wonder whether those could be transformed to some sort of http request -> direct -> (do something) -> http response. That would produce less messages in the log and the tests could do some assertions on the HTTP responses.

OK let me do this

zbendhiba (approved):

essobedo:
@ppalaga ok I removed 2 deps (seda and timer)

ppalaga (approved):
Thanks @essobedo, looks good to merge from my PoV. The substitution would deserve a followup issue.

essobedo:
Does it need to be reviewed by someone else? Or is it good enough to be merged?

davsclaus (approved):

PR #4409, opened 2023-01-16
Ref #4407: js-dsl - Improve the interoperability with Java code

### LINKED ISSUE #4407: js-dsl - Improve the interoperability with Java code
The JavaScript DSL already supports the native mode but it is really limited in terms of interoperability with Java code, the goal of this task is to improve the code to at least support the basic use cases available [in the Came code base](https://github.com/apache/camel/tree/camel-3.x/dsl/camel-js-dsl/src/test/resources/routes).

essobedo:
fixes #4407 

## Motivation

The JavaScript DSL already supports the native mode but it is really limited in terms of interoperability with Java code, the goal of this task is to improve the code to at least support the basic use cases available [in the Came code base](https://github.com/apache/camel/tree/camel-3.x/dsl/camel-js-dsl/src/test/resources/routes).

## Modifications:

* Add a filter to the build step `registerReflectiveClasses` to call it only in native mode
* Register the missing methods and fields for reflection allowing to be able to configure Camel components from a JavaScript resource
* Add missing Camel reflective classes
* Add implementation of the main functional interfaces to have a way to implement them in native mode
* Add a limitation note about the function `Java.extend` and how to work around it
* Add `mjs` to the mapping of the `license-maven-plugin` to prevent check style issue


davsclaus (approved):

essobedo:
@jamesnetherton good question, for me yes as it is still useable on GraalVM and personally I cannot imagine the Quarkus team enforcing the community to use Mandrel definitively just because of a bug that will be fixed one day or another. AFAIU, they only switch as long as the bug is open.

For me, the question is more, do you want to keep on proposing the js-DSL in native mode? If not, you can close it otherwise IMHO it is worth keeping it but it is only my point of view.

essobedo:
As you probably have understood my goal at the end is to make Camel-K evolves to be able to write an integration in the language of my choice and still be able to compile it natively in order to save as much resources as possible to reduce the costs when deploying it on a public cloud.

jamesnetherton:
At the moment the native tests aren't being executed on the CI build. So we should open an issue to get that figured out (for the future). Otherwise we end up in our previous situation where stuff gets broken and nobody notices.

ppalaga:
>  Otherwise we end up in our previous situation where stuff gets broken and nobody notices.

+100

jamesnetherton:
I opened https://github.com/apache/camel-quarkus/issues/4418.

PR #4417, opened 2023-01-17
Ref #4416: Make the csimple extension agnostic to build systems

### LINKED ISSUE #4416: Make the csimple extension agnostic to build systems
The csimple extension has hard-coded references to the directory `target` which makes only sense in the case of maven. It should be reviewed to be agnostic to build systems.

essobedo:
fixes #4416 

## Motivation

The csimple extension has hard-coded references to the directory `target` which makes only sense in the case of maven. It should be reviewed to be agnostic to build systems.

## Modifications:

* Replace `target` with the output directory of the `BuildSystemTargetBuildItem`


oscerd (approved):

jamesnetherton (approved):

ppalaga:
Nice catch, thanks @essobedo !

PR #4433, opened 2023-01-20
Fix #4128 add a dependabot to upgrade quarkiverse versions

### LINKED ISSUE #4128: Dependabot to upgrade quarkiverse versions
Have a way to automatic to upgrade quarkiverse versions, if possible ?

zhfeng:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,25 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


oscerd (approved):

aldettinger:
Good time to experiment with new tools right after a release. Thus few questions.

Will dependabot upgrade versions only linked to CVEs ?
Or upgrade to any new release quarkiverse.* version available ?

Also, in case a new quarkiverse.* version is available but we would skip it in camel-quarkus, because of a bug for instance. Then, how could we skip that version ?

zhfeng:
It could open a PR for any new release of  quarkiverse artifacts we are using. More details on https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

Of casue we can skip a version if it is not feasible for merging.

aldettinger:
So, when we want to skip a version. Then we just modify the yaml file ?
That would be a nice way to document what quarkiverse updates are dangling.

zhfeng:
yeah. I think we can use `ignore` in yaml file to skip some artifacts or some version we don't want to update.

aldettinger (approved):
Ok, thanks for sharing @zhfeng. I would vote to experiment with this.

I wonder what other think ?

jamesnetherton:
> I would vote to experiment with this.
> 
> I wonder what other think ?

Yeah let's see how it works for us.

For `quarkus-amazon-services` we maybe need to be a bit careful. I don't think we always want to merge version upgrades, given that project is a Quarkus Platform member and we should not use a version higher than what is specified there.

aldettinger:
Indeed, well spotted, not all updates are to be consumed.

Also, I'm thinking the dependabot approach would encourage to increase each quarkiverse.* version independently. I see this as a benefit at this stage.

ppalaga (approved):

zbendhiba:
Many thanks! This is awesome and helpful.
Just quick question, will it open a PR every time a new quarkiverse dependency is released ?

zhfeng:
@zbendhiba yeah, it checks daily and will open a PR for new release of quarkiverse.

PR #4453, opened 2023-01-27
Ref #4452: Allow to disable build time compilation of jOOR expressions

### LINKED ISSUE #4452: Allow to disable build time compilation of jOOR expressions
In order to give the opportunity to the end user to decide whether the compilation of jOOR expressions in JVM should be done at runtime like before or at build time, a new dedicated property configuration should be added.

essobedo:
fixes #4452 

## Motivation

In order to give the opportunity to the end user to decide whether the compilation of jOOR expressions in JVM should be done at runtime like before or at build time, a new dedicated property configuration should be added.

## Modifications:

* Add a new configuration property called `compileAtBuildTime` for this purpose, set to `false` by default to keep the same behavior as the previous version 
* Add a specific `BooleanSupplier` to enable the compilation in native mode or when `compileAtBuildTime` is `true`
* Rename the test resource `bean.joor` to `bean.txt` to avoid having to take care of the license header.


docs/modules/ROOT/pages/reference/extensions/joor.adoc
```diff
@@ -74,6 +74,12 @@ The specific location of the configuration of the jOOR language.
 The specific default result type of an expression expressed in jOOR language.
 | `string`
 | 
+
+|icon:lock[title=Fixed at build time] [[quarkus.camel.joor.compile-ahead-of-time]]`link:#quarkus.camel.joor.compile-ahead-of-time[quarkus.camel.joor.compile-ahead-of-time]`
```
jamesnetherton, line 78: Wonder if `quarkus.camel.joor.compile-at-build-time` would be better? I think it's more consistent with the nomenclature that Quarkus uses (I.e AOT is rarely mentioned in docs. It's usually 'build time') .

I'm nitpicking a bit here, so happy to be ignored if you disagree.

essobedo: 👍 It makes sense indeed, let me rename it

aldettinger (approved):

essobedo:
The failure is not related to this PR since it is an mail integration test that fails and it fails because of an `QuarkusBindException: Port(s) already bound: 32775: Address already in use`

oscerd (approved):

PR #4473, opened 2023-02-02
Draft: Refactor Gson Json Dataformat tests

ldrozdo:


ldrozdo:
@aldettinger @ppalaga Hi, can I go this way with splitting the json dataformats? This draft is only for gson, I will do it similarly with other formats. I've created modules in the `integration-test-groups`. I didn't create any base test class in `integration-tests-support`, it seemed to me that it is more readable this way. But I can create it if you want. It's possible that while doing other dataformats, I will see and implement more options for refactoring.

integration-test-groups/dataformats-json/json-gson/pom.xml
```diff
@@ -0,0 +1,289 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-xml-jaxb</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-jackson</artifactId>
+        </dependency>
+        <dependency>
```
aldettinger, line 48: After the spiit we would need only required dependency per dataformats. For instance, maybe camel-quarkus-jsonb is not needed to test json-gson.

aldettinger:
@ldrozdo The direction looks good :+1: 

Then a grouped test might be needed as `integration-tests/aws2-grouped`.
After that, the grouped test could be assigned to a group where ci run is shorter than others... like `tooling/scripts/test-categories.yaml` where `aws2-grouped` has been assigned `group-12`.

Concerning base test, it should be fine. Some common pieces behave the same but on the other hand, most data formats have their own specificities. And we can always refine later.

ppalaga:
@ldrozdo, could you please rebase (plz take care to replace `<version>2.17.0-SNAPSHOT</version>` with `<version>3.0.0-SNAPSHOT</version>`  in `pom.xml` files ) and add the `-grouped` module as sketched in https://camel.apache.org/camel-quarkus/2.16.x/contributor-guide/extension-testing.html#_grouping ?

ldrozdo:
> @ppalaga yes I will, there were just too many conflicts and I had to move on to a more priority task. I will finish hopefully this week and I will let you know.

Issue #4442, opened 2023-01-26
Aws2: Add testing of defaultCredentialProvider to each extension


JiriOndrusek:
Follows https://github.com/apache/camel-quarkus/issues/4401

Add test coverage of defaultCredentialProvider into each aws extension. Try to keep majority of the code in the support extension.

JiriOndrusek:
Add comment about reading properties via ConfigProvider and why they are set as system.properties

Issue #4401, opened 2023-01-11
Aws2: investigate possibilty of testing defaultXCredentialProvider with local stack


JiriOndrusek:
Follows https://github.com/apache/camel-quarkus/issues/4346.

If local stack is used for testing. Aws2 credentials + override url is passed to the aws client (as static credentials)
Perhaps it could be possible to use default credentials as well. (In case that default credentials are not configured in the system)

PR #4474, opened 2023-02-02
Aws2: Add testing of defaultCredentialProvider to each extension #4442

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4442
(follows https://github.com/apache/camel-quarkus/issues/4401)

PR extends `camel-quarkus-integration-tests-support-aws2` to cover tests with default credentials provider without a lot of changes in each aws2 test module.
To fully use the support there are a few requirements for each aws2 test module:

1. Add  dependencies of both flavors (of `camel-quarkus-integration-tests-support-aws2`) into the pom:

```
        <dependency>
            <artifactId>camel-quarkus-integration-tests-support-aws2</artifactId>
        </dependency>
        <dependency>
            <artifactId>camel-quarkus-integration-tests-support-aws2</artifactId>
            <type>test-jar</type>
            <scope>test</scope>
        </dependency>
```

2. Extend `BaseAws2Resource` by the resource class.
3. Extend `BaseAWs2TestSupport` by the test class.
4. Add parameter `"?useDefaultCredentialsProvider=" + useDefaultCredentials` into endpoint url called from the resource.
5. Implement simple test method (overriding) `testMethodForDefaultCredentialsProvider()`

If all items are satisfied, 2 new tests are executed for the test module (`failingDefaultCredentialsProviderTest` and `successfulDefaultCredentialsProviderTest`).
One covers positive scenario, the other negative scenario.

Be aware that the tests are executed only with **mocked backend**, **without quarkus client**, and only in case that **awsDefaultCredentialsProvider is not defined on the system**. Otherwise tests are skipped. See `Aws2DefaultCredentialsProviderAvailabilityCondition`.

 

- Cw: Tests were refactored to use helper methods from aws-support.
- Kinesis: Test class was split into kinesis & firehose test classes to cover default credentials testing for  both extensions.
- S3: Name of the rest resource used for testing was renamed to the same pattern as other aws2 testing extensions.


oscerd (approved):

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/Aws2Helper.java
```diff
@@ -0,0 +1,35 @@
...
+            //if message starts with "Unable to load credentials", allow testing
+            if (e instanceof SdkClientException && e.getMessage() != null
+                    && e.getMessage().startsWith(UNABLE_TO_LOAD_CREDENTIALS_MSG)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public static void setAwsSysteCredentials(String accessKey, String secretKey) {
```
aldettinger, line 23: small typo here ?

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/Aws2Helper.java
```diff
@@ -0,0 +1,35 @@
...
+        System.setProperty("aws.accessKeyId", accessKey);
+        System.setProperty("aws.secretAccessKey", secretKey);
+
+    }
+
+    public static void clearAwsSysteCredentials() {
```
aldettinger, line 29: Same way: System ?

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/BaseAws2Resource.java
```diff
@@ -0,0 +1,96 @@
...
+import javax.ws.rs.core.Response;
+
+import io.quarkus.runtime.ShutdownEvent;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.jboss.logging.Logger;
+
+/**
+ * Parent for testing resources for aws2 extensions.
+ *
+ * This class adds endpoints to force default credentials to be used + provides protected varialbe useDefaultCredentials
```
aldettinger, line 33: protected variable ?

integration-tests-support/aws2/src/test/java/org/apache/camel/quarkus/test/support/aws2/BaseAWs2TestSupport.java
```diff
@@ -0,0 +1,108 @@
...
+ * Parent for aws test classes which should cover default credentials test case.
+ * Parent adds two test methods {@link #successfulDefaultCredentialsProviderTest()} and
+ * {@link #failingDefaultCredentialsProviderTest()}
+ *
+ * Test expects that test resource extends {@link BaseAws2Resource}.
+ *
+ */
+public abstract class BaseAWs2TestSupport {
+
+    //Rest path to connect to rest resource. Path differs for different aws2 extensions (like "/aws2-ddb" or "/aws2-s3"
```
aldettinger, line 34: We may close the ")" ?

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/Aws2Helper.java
JiriOndrusek: I'll replace it cross the PR:)

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/BaseAws2Resource.java
```diff
@@ -0,0 +1,96 @@
...
+            clearAwsredentials = false;
+        }
+
+        return Response.ok().build();
+    }
+
+    /**
+     * Listeners ensures, that system credentials are cleared at the end of the lifecycle.
+     * Tests are clearing them by itself, this is just a precaution.
+     */
```
aldettinger, line 86: Are we also clearing the properties when an unexpected exception is thrown during a test ?

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/Aws2Helper.java
JiriOndrusek: fixed

aldettinger:
The comments in the pr and the code are very much welcome @JiriOndrusek :clap: It helps with review from my perspective.

It looks good pr, only few nitpicks and 1 question.

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/BaseAws2Resource.java
JiriOndrusek: I thought that this listener should be called also with unexpected shutdown. Therefore it should clear properties for all cases. Am I missing here something? (https://quarkus.io/guides/lifecycle#listening-for-startup-and-shutdown-events)

JiriOndrusek: I can hook another safety clear into [Aws2TestResource.java](https://github.com/apache/camel-quarkus/blob/main/integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/Aws2TestResource.java#L109). That could help, what do you think, @aldettinger ?

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/BaseAws2Resource.java
JiriOndrusek: I kept it as protected, so the inherited classes can call it directly. I will  use getter (`is*`) and keep variable private and remove the word `protected` from the comment

integration-tests-support/aws2/src/test/java/org/apache/camel/quarkus/test/support/aws2/BaseAWs2TestSupport.java
```diff
@@ -0,0 +1,108 @@
...
+                    .body(true)
+                    .post(restPath + "/initializeDefaultCredentials")
+                    .then()
+                    .statusCode(200);
+
+            //should succeed
+            testMethodForDefaultCredentialsProvider();
+
+        } finally {
+            RestAssured.given()
```
JiriOndrusek, line 74: @aldettinger  the test is trying to clear the system.properties in final block. (so it covers unexpected exception)

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/BaseAws2Resource.java
aldettinger: Oh, it was just to state another typo `varialbe`

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/BaseAws2Resource.java
aldettinger: As far as I get, the `onStop` listener is called when quarkus shutdown.

The question would be more like, do we have situations where we set properties, then exception is thrown without cleanup. And then, a next test would fail because the properties was not cleaned up.

integration-tests-support/aws2/src/test/java/org/apache/camel/quarkus/test/support/aws2/BaseAWs2TestSupport.java
aldettinger: ok, that was the missing part :+1 Thanks

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/BaseAws2Resource.java
aldettinger: Ok, it was solved in a comment below. There is a finally block in the test that would restore the properties.

aldettinger (approved):

integration-tests-support/aws2/src/test/java/org/apache/camel/quarkus/test/support/aws2/BaseAWs2TestSupport.java
JiriOndrusek: fixed

integration-tests-support/aws2/src/main/java/org/apache/camel/quarkus/test/support/aws2/BaseAws2Resource.java
JiriOndrusek: fixed

jamesnetherton:
@JiriOndrusek can you rebase this with the latest work from `main`. If CI is good we can merge.

JiriOndrusek:
sure, I'll do that, thanks.

aldettinger:
I would say this one can be merged ?

ppalaga (approved):
Thanks @JiriOndrusek!

PR #4529, opened 2023-02-08
Mail test fails in native mode #4519

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4519


jamesnetherton:
We should probably be modifying the mail support extension. IIRC it already has logic to register the providers.

https://github.com/apache/camel-quarkus/blob/main/extensions-support/mail/deployment/src/main/java/org/apache/camel/quarkus/support/mail/deployment/SupportMailProcessor.java

JiriOndrusek:
I didn't notice that support module. I  will refactor my PR

Issue #4528, opened 2023-02-08
PerfRegressionIT failing with Camel 4 and Quarkus 3


ppalaga:
```
[INFO] Running org.apache.camel.quarkus.performance.regression.it.PerfRegressionIT
[ERROR] Tests run: 3, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 10.775 s <<< FAILURE! - in org.apache.camel.quarkus.performance.regression.it.PerfRegressionIT
[ERROR] nominalShouldPrintReport{String}[1]  Time elapsed: 10.098 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: The perf-regression process has finished with an unexpected exit value
        at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:42)
        at org.junit.jupiter.api.Assertions.fail(Assertions.java:147)
        at org.apache.camel.quarkus.performance.regression.it.PerfRegressionIT.nominalShouldPrintReport(PerfRegressionIT.java:116)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
        at java.base/java.lang.reflect.Method.invoke(Method.java:578)
        at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:727)
        at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
        at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
        at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:156)
        at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:147)
        at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestTemplateMethod(TimeoutExtension.java:94)
        at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(InterceptingExecutableInvoker.java:103)
        at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.lambda$invoke$0(InterceptingExecutableInvoker.java:93)
[...]

PR #4535, opened 2023-02-08
perf-regression: restore PerfRegressionIT to test against 3.0.0-SNAPS…

aldettinger:
…HOT #4528


ppalaga (approved):
THanks @aldettinger

ppalaga:
Group 12 is stuck for some reason, but it can be seen in its log that the tests passed already. Thus merging.

PR #4539, opened 2023-02-09
Fix JAXB and Salesforce native compilation issues

jamesnetherton:


ppalaga (approved):
Thanks @jamesnetherton !

PR #4567, opened 2023-02-14
Use XML DSL to define templated route in integration test

osmman:


aldettinger:
The failing test is unrelated, so merging.

Issue #4560, opened 2023-02-12
Multiple test methods do not work when extending CamelQuarkusTestSupport


salagator:
I am trying to write camel quarkus tests using the CamelQuarkusTestSupport facility.

From the documentation (here: https://camel.apache.org/camel-quarkus/2.16.x/user-guide/testing.html), I gather that the camel context is created once, for all the tests.

However, when I try writing tests the first one succeeds and any subsequent tests fail. If I apply the "TestInstance(TestInstance.Lifecycle.PER_CLASS)" annotation, the tests in one class succeed and the tests in any subsequently executed class fail.

Something in the test scaffolding seems to be stoping all routes, but not starting them back up (either at the test method boundary, or the class boundary respectively).

I must be doing something wrong, but I cannot understand what. I tried googling myself out of the problem, but to no avail. Please help.

I am attaching a "reproducer" project that shows what I am doing (wrong): [camel-testing.zip](https://github.com/apache/camel-quarkus/files/10715736/camel-testing.zip)

ppalaga:
Thanks for the reproducer, @salagator! 
@JiriOndrusek could you please have a look?

JiriOndrusek:
I'll check that, thanks for the information.

JiriOndrusek:
@salagator Thanks for finding those issues. There are probably 2 isolated issues affecting this case

1.  Scenario  without `TestInstance(TestInstance.Lifecycle.PER_CLASS)` should succeed. I've almost finished fix for this sub-issue and I'll provide a PR soon.
2. Usage of `TestInstance(TestInstance.Lifecycle.PER_CLASS)` should work as a workaround for the first sub-issue. Unfortunately there is probably a small issue with propagating camel context. I'll investigate the second scenario after the first one and will create a separate ticket for it.

PR #4568, opened 2023-02-14
Multiple test methods do not work when extending CamelQuarkusTestSupport

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4560
fixes https://github.com/apache/camel-quarkus/issues/4569

Brief summary of changes:

Problem was caused by the difference between lifecycle of camel and camel-quarkus. As the camel context is created once, there are several callbacks, which executes context "resetting" so it behave likes it starts for each test again. During the resetting, all routes were stopped and removed. Unfortunately in a scenario with global routeBuilder (like the reproducer from the issue) it meant that  even routes from tat global routeBuilder were removed ,  and the next test classes (or methods) missed routes from that routeBuilder. This PR changes the behavior in a way, that only routes added during setup of the test are removed. Therefore there is a local variable `createdRoutes` which , holds routes created during the test setup.


oscerd (approved):

test-framework/junit5/src/main/java/org/apache/camel/quarkus/test/CallbackUtil.java
```diff
@@ -46,15 +46,22 @@ static Optional<TestInstance.Lifecycle> getLifecycle(CamelQuarkusTestSupport tes
...
+        //if routeBuilder (from the test) was used, all routes from that builder has to be stopped and removed
         //because routes will be created again (in case of TestInstance.Lifecycle.PER_CLASS, this method is not executed)
-        if (testInstance.isUseRouteBuilder()) {
-            try {
-                testInstance.context().getRouteController().stopAllRoutes();
-                testInstance.context().getRouteController().removeAllRoutes();
+        if (testInstance.isUseRouteBuilder() && testInstance.createdRoutes != null) {
 
+            try {
+                if (keepOriginalRoutes) {
```
aldettinger, line 56: Could you please details what `keepOriginal` means in this context ?

It's like we keep the state of route ? Or we want a fresh route ?

JiriOndrusek: I'll add comment here. that makes sense, it didn't occur to me that it is my own definition of routes.

JiriOndrusek: I removed this if. It should not be there. There is a following ticket, which I'm starting to work on, which takes care of the lifecycle with Lifecycle.PER_CLASS - https://github.com/apache/camel-quarkus/issues/4569

aldettinger (approved):
Ok, many thanks for adjustement :+1:

PR #4580, opened 2023-02-15
create itest project for xml-io + bean #4579

aldettinger:


oscerd (approved):

jamesnetherton (approved):

aldettinger:
It was not clear whether this is xml-io or bean tests, so a wrapping project solve this.
Plus, it could avoid cases where bean would hide issues from xml-io (without any particular cases in head).

That's a tie, we could do both. Shall we create a new project as in this pr ?

jamesnetherton:
> That's a tie, we could do both. Shall we create a new project as in this pr ?

I'll leave it to your choice. I have no strong opinion on it.

zhfeng (approved):

aldettinger:
OK, let's merge what's ready. We can always revert later if this prove to be useless. Many thanks all for stepping in.

PR #4589, opened 2023-02-20
Ref #4393: Groovy DSL - Get rid of --report-unsupported-elements-at-runtime

### LINKED ISSUE #4393: Get rid of `--report-unsupported-elements-at-runtime` for Groovy DSL
Follows up https://github.com/apache/camel-quarkus/pull/4385/files#diff-fa25b87249559ee321efea1b839659150f97ca18cf5718cab8826bde025d7a84R52-R67

While this is fine for this initial version of the Groovy DSL support, we should not leave it as is in the long term. We generally do not want end users to add any GraalVM extra flags. Our extensions should be able to configure everything needed. However, `--allow-incomplete-classpath` and `--report-unsupported-elements-at-runtime` are special: we should try to always avoid them. The reason is that they hide potential problems that may make the app fail at runtime.

The root cause of why `--report-unsupported-elements-at-runtime` is required in some situations needs to be understood in order to be able to tell whether this is an issue of how Camel Quarkus or Camel is using Groovy or whether this is an issue fixable in Groovy (how about Groovy 4?) or whether we should seek for a fix in GraalVM. Once a tracker issue is created in any external project, its reference should be pasted to `docs/modules/ROOT/pages/reference/extensions/groovy-dsl.adoc`

essobedo:
fixes #4393 

## Motivation

The flag `--report-unsupported-elements-at-runtime` can be needed in specific use cases to prevent errors during a native compilation, the goal is to see how it can be avoided.

## Modifications:

I could not reproduce the problem with the latest versions of Graal VM / Quarkus / Camel Quarkus so this PR simply adds new integration tests to ensure that the problem won't appear anymore.

* Add more integration tests to cover the use cases that used to fail
* Align the code of the integration tests and their dependencies with the integration tests of other DSLs
* Adapt the code of the processor to Java 17
* Remove the limitation from the documentation


oscerd (approved):

jamesnetherton (approved):

aldettinger:
@essobedo Really nice to investigate those issues :+1: 

> I could not reproduce the problem with the latest versions of Graal VM

Is it native-image 22.3.1 ?

On main, users are currently bound to [native-image 22.3.0](https://github.com/apache/camel-quarkus/blob/main/pom.xml#L102).

jamesnetherton:
> users are currently bound to native-image 22.3.0.

If you do a container build with `ubi-quarkus-mandrel-builder-image:22.3-java17`, then you'll get Mandrel `22.3.1.0`.

essobedo:
> @essobedo Really nice to investigate those issues 👍
> 
> > I could not reproduce the problem with the latest versions of Graal VM
> 
> Is it native-image 22.3.1 ?
> 
> On main, users are currently bound to [native-image 22.3.0](https://github.com/apache/camel-quarkus/blob/main/pom.xml#L102).

It also passes with GaalVM `22.3.0` and `22.2.0`

aldettinger (approved):

Issue #4514, opened 2023-02-07
zendesk test cannot be compiled to native with Camel 4 and Quarkus 3


ppalaga:
```
[INFO] [stdout] GraalVM Native Image: Generating 'camel-quarkus-integration-test-zendesk-3.0.0-SNAPSHOT-runner' (executable)...
[INFO] [stdout] ========================================================================================================================
[INFO] [stdout] [1/7] Initializing...                                                                                    (8.7s @ 0.29GB)
[INFO] [stdout]  Version info: 'GraalVM 22.3.1.0-Final Java 17 Mandrel Distribution'
[INFO] [stdout]  Java version info: '17.0.6+10'
[INFO] [stdout]  C compiler: gcc (redhat, x86_64, 8.5.0)
[INFO] [stdout]  Garbage collector: Serial GC
[INFO] [stdout]  3 user-specific feature(s)
[INFO] [stdout]  - io.quarkus.runner.Feature: Auto-generated class by Quarkus from the existing extensions
[INFO] [stdout]  - io.quarkus.runtime.graal.DisableLoggingFeature: Disables INFO logging during the analysis phase for the [org.jboss.threads] categories
[INFO] [stdout]  - io.quarkus.runtime.graal.ResourcesFeature: Register each line in META-INF/quarkus-native-resources.txt as a resource on Substrate VM
[INFO] [stdout] [2/7] Performing analysis...  []                                                                        (11.3s @ 2.42GB)
[INFO] [stdout]   10,219 (91.11%) of 11,216 classes reachable
[INFO] [stdout]   14,247 (58.82%) of 24,221 fields reachable
[INFO] [stdout]   43,156 (80.14%) of 53,851 methods reachable
[INFO] [stdout]      740 classes,     0 fields, and     0 methods registered for reflection
[INFO] [stdout] 
[WARN] [stderr] Fatal error: com.oracle.graal.pointsto.util.AnalysisError$ParsingError: Error encountered while parsing org.asynchttpclient.request.body.multipart.FilePart.<init>(java.lang.String, java.io.File, java.lang.String, java.nio.charset.Charset, java.lang.String, java.lang.String, java.lang.String) 
[WARN] [stderr] Parsing context:
[WARN] [stderr]    at org.asynchttpclient.request.body.multipart.FilePart.<init>(FilePart.java:45)
[WARN] [stderr]    at org.asynchttpclient.request.body.multipart.FilePart.<init>(FilePart.java:41)
[...]

Issue #4591, opened 2023-02-21
Zendesk: update zendesk-java-client once it depends on asyncclient 3.x


JiriOndrusek:
Follows https://github.com/apache/camel-quarkus/issues/4514. Async-httpclient is upgraded to 3.x. This dependency should be brought by `com.cloudbees.thirdparty:zendesk-java-client`. If such java client exists, enable @sync task for `ahc.version` https://github.com/apache/camel-quarkus/blob/main/pom.xml#L68

jamesnetherton:
No longer relevant since https://github.com/apache/camel-quarkus/pull/5202.

PR #4592, opened 2023-02-21
Zendesk test cannot be compiled to native with Camel 4 and Quarkus 3 

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4514

Async-httpclient us upgraded to 3.0.0.Beta1 to support jakarta.

Synchronization between `com.cloudbees.thirdparty:zendesk-java-client` and `ahc.version` is disabled until a new client exist. Ticket to enable it back was created https://github.com/apache/camel-quarkus/issues/4591.

### DIFF
--- a/extensions-support/ahc/deployment/src/main/java/org/apache/camel/quarkus/component/support/ahc/deployment/SupportAhcProcessor.java
+++ b/extensions-support/ahc/deployment/src/main/java/org/apache/camel/quarkus/component/support/ahc/deployment/SupportAhcProcessor.java
@@ -18,6 +18,7 @@ package org.apache.camel.quarkus.component.support.ahc.deployment;
 
 import java.util.stream.Stream;
 
+import io.netty.incubator.channel.uring.IOUringEventLoopGroup;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.ExtensionSslNativeSupportBuildItem;
@@ -38,6 +39,11 @@ class SupportAhcProcessor {
     @BuildStep
     ExtensionSslNativeSupportBuildItem activateSslNativeSupport() {
         return new ExtensionSslNativeSupportBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    RuntimeInitializedClassBuildItem runtimeInitializedClasses() {
+        return new RuntimeInitializedClassBuildItem(IOUringEventLoopGroup.class.getName());
     }
 
     @BuildStep
--- a/extensions-support/ahc/runtime/pom.xml
+++ b/extensions-support/ahc/runtime/pom.xml
@@ -47,6 +47,23 @@
             <groupId>org.asynchttpclient</groupId>
             <artifactId>async-http-client</artifactId>
         </dependency>
+        <dependency>
+            <artifactId>netty-transport-classes-epoll</artifactId>
+            <groupId>io.netty</groupId>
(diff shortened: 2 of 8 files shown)


oscerd (approved):

aldettinger (approved):
LGTM

extensions/zendesk/deployment/src/main/java/org/apache/camel/quarkus/component/zendesk/deployment/ZendeskProcessor.java
```diff
@@ -38,6 +40,11 @@ AdditionalApplicationArchiveMarkerBuildItem boxArchiveMarker() {
         return new AdditionalApplicationArchiveMarkerBuildItem("org/zendesk");
     }
 
+    @BuildStep
+    RuntimeInitializedClassBuildItem test() {
```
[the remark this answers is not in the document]
JiriOndrusek: I moved code zendesk extension to `ahc-support`, where it belongs

extensions/zendesk/runtime/pom.xml
```diff
@@ -51,6 +51,12 @@
         <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-zendesk</artifactId>
+            <exclusions>
+                <exclusion>
+                    <artifactId>async-http-client</artifactId>
+                    <groupId>org.asynchttpclient</groupId>
+                </exclusion>
+            </exclusions>
```
ppalaga, line 59: Why is this exclusion needed? If it there to solve dependency convergence issues, then I'd rather vote for managing `async-http-client` in our BOM. In that way, we can make sure that all our extensions (and tests) using it will use the same version which is better for our and end users' reproducibility.

ppalaga: Hm... we already manage it in the BOM. Why is the exclusion needed then?

JiriOndrusek: I'll check, it is possible that it is not needed anymore - as it is in the BOM

JiriOndrusek: I forgot to remove this exclude during my last update of the PR - I'll fix it in a moment

extensions-support/ahc/deployment/pom.xml
```diff
@@ -42,6 +42,11 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-support-ahc</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.graalvm.nativeimage</groupId>
+            <artifactId>svm</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
ppalaga, line 49: I guess this should rather go to the runtime module, where we do substitutions?

Possible followup: we should slowly Migrate from `org.graalvm.nativeimage:svm` to `org.graalvm.sdk:graal-sdk` https://github.com/apache/camel-quarkus/issues/4602

ppalaga (approved):
One little nitpick inline, otherwise looks good. Thanks @JiriOndrusek !

extensions-support/ahc/deployment/pom.xml
JiriOndrusek: I see, I added dependency to the wrong module. I'll fix it in a few minutes.

JiriOndrusek: fixed

zbendhiba:
Nice @JiriOndrusek !

PR #4593, opened 2023-02-21
Minimize exclusions in the BOM, remove superfluous exclusions in extensions

ppalaga:
This is an attempt to 
* Put the exclusions in the BOM only on the closest managed dependent. Because the BOMs we import do not do the same, we sometimes still need to exclude also on higher levels
* Remove superfluous exclusions in extension modules - those accumulated over the years. They are not needed anymore for various reasons, most notably: 
  * The given exclusion is available in some BOM
  * The transitives have changed


oscerd (approved):

ppalaga:
615ac95: fixed the failure in azure tests

Issue #4598, opened 2023-02-22
Micrometer test coverage - possible issue with @Counted annotation and metrics mandatory


JiriOndrusek:
[Comment](https://github.com/apache/camel-quarkus/issues/4582#issuecomment-1433125354) shows possible error and therefore a shift of optional to mandatory for microprofile metrics API. 

It makes sense to investigate this potential issue separately.

PR #4600, opened 2023-02-23
Micrometer test coverage - @Counted

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4598

### DIFF
--- a/integration-tests/micrometer/src/main/java/org/apache/camel/quarkus/component/micrometer/it/TestMetric.java
+++ b/integration-tests/micrometer/src/main/java/org/apache/camel/quarkus/component/micrometer/it/TestMetric.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.micrometer.it;
+
+import io.micrometer.core.annotation.Counted;
+import io.micrometer.core.annotation.Timed;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.config.MeterFilter;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Produces;
+import jakarta.inject.Named;
+import jakarta.inject.Singleton;
+
+@ApplicationScoped
+@Named("testMetric")
+public class TestMetric {
+
+    @Counted(value = "TestMetric.counted1")
+    @Timed(value = "TestMetric.timed1")
+    public void call1() {
+        try {
+            //wait 1 second
+            Thread.sleep(1000);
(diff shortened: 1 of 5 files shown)


integration-tests/micrometer/src/main/java/org/apache/camel/quarkus/component/micrometer/it/TestMetric.java
```diff
@@ -0,0 +1,64 @@
...
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.config.MeterFilter;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Produces;
+import jakarta.inject.Singleton;
+
+@ApplicationScoped
+public class TestMetric {
+
+    private int count;
```
aldettinger, line 30: Maybe w ecould remove count ?

aldettinger (approved):

oscerd (approved):

integration-tests/micrometer/src/main/java/org/apache/camel/quarkus/component/micrometer/it/TestMetric.java
JiriOndrusek: I need count because of the test `testAnnotations`, I didn't force an order to the test methods, therefore I don't know which test goes first. I'll add another comment linked to the code I talking about to show you

integration-tests/micrometer/src/test/java/org/apache/camel/quarkus/component/micrometer/it/MicrometerTest.java
```diff
@@ -83,20 +84,58 @@ public void testMicrometerExchangeEventNotifier() {
...
+    @Test
+    public void testAnnotations() {
+        RestAssured.get("/micrometer/annotations/call")
+                .then()
+                .statusCode(200);
+        RestAssured.get("/micrometer/annotations/call")
+                .then()
+                .statusCode(200);
+
+        int count = RestAssured.get("/micrometer/annotations/get")
```
JiriOndrusek, line 96: @aldettinger I have to get count from the object, because it can have value 2 or 3. Depends whether testMetricsApiRenamed was executed or not.

integration-tests/micrometer/src/main/java/org/apache/camel/quarkus/component/micrometer/it/TestMetric.java
aldettinger: On a local execution with `TestMetric.count` attribute removed, the value of count seems still to be propagated to the test.

Is the count value returned from `MicrometerResource` handled internally by micrometer ?

jamesnetherton: Do we need the `testMetricsApiRenamed` test? Removing it would avoid the need to maintain an internal counter of method invocations. We can assume the general MicroProfile functionality is already tested in Quarkus.

The Objective here is to test MicroProfile metric annotations in conjunction with Camel.  So instead of just manually invoking a CDI mean from a JAX-RS method, we should do it via a Camel route. E.g like...

```java
from("direct:annotatedBean")
   .bean("someBeanWithMethodHaving@Counted")
```

JiriOndrusek: I understand you point of view, but let me explain the whole motivation:
1- Test `testMetricsApiRenamed` covers Quarkus's use of metrics API ([doc](https://quarkus.io/guides/micrometer#support-for-the-microprofile-metrics-api)) therefore I think it should stay. I renamed the test to `testQuarkusMetricsApi` and separated it from the other annotations test, so no `count` is needed anymore.
2 - Nice catch. I refactored the test to use route.

JiriOndrusek:
@jamesnetherton , @aldettinger thanks for the review! I refactored tests to fulfill your hints. I think thet the PR could be merged.

aldettinger (approved):
LGTM. Maybe another review would be great before merging.

jamesnetherton (approved):

PR #4611, opened 2023-03-01
Upgrade to cq-maven-plugin 3.5.3: flatten faster, sync-versions in

ppalaga:


oscerd (approved):

aldettinger (approved):
Maybe it could be needed to re-flatten depending on merge order.

PR #4645, opened 2023-03-13
Auto synchronize dependabot branches with generated changes

jamesnetherton:
Automatically syncs up dependabot branches if there are any changes in generated files.

One slight quirk of how this works, is that the PR CI workflow is re-triggered, but the status is never reported in the GitHub UI on the PR. So I add a comment with a link to the build so that it can be checked.

I tested on my CQ fork and it seemed to be working ok. We might want to hold merging until the next dependabot PR is opened. That way we could test and monitor the workflow to make sure nothing bad happens.


jamesnetherton:
Well... that failed quickly :sweat_smile:. I'll follow up on the failure.

oscerd (approved):

.github/workflows/ci-build.yaml
```diff
@@ -57,9 +57,74 @@ env:
...
+            ./mvnw clean install -N ${CQ_MAVEN_ARGS}
+
+            cd poms
+            ../mvnw clean install -N ${CQ_MAVEN_ARGS}
+
+            cd bom
+            ../../mvnw process-resources -Dcq.flatten-bom.format --fail-never ${CQ_MAVEN_ARGS}
+            ../../mvnw clean install ${CQ_MAVEN_ARGS}
+
+            cd ../../
```
ppalaga, line 101: A similar result could perhaps be achieved with less lines:
```suggestion
            ./mvnw clean install -f poms/bom -am -Dcq.flatten-bom.format --fail-never ${CQ_MAVEN_ARGS}
            # the above may change the BOM in the source tree, but does not install it, so let's install it
            ./mvnw clean install ${CQ_MAVEN_ARGS} -f poms/bom
```

zhfeng (approved):

ppalaga:
Impressive @jamesnetherton!

.github/workflows/ci-build.yaml
jamesnetherton: Thanks! I'll give that a test.

zbendhiba:
Awesome !

.github/workflows/ci-build.yaml
aldettinger: The flatten-bom default to process-resources phase and is not installed in the first "mvn clean install" line ? If someone has light to share, that would be welcome.

Also, if second line is needed, maybe `mvnw install` without `clean` could be enough ?

aldettinger (approved):
As far as I get looks good. It would be awesome if it work.

.github/workflows/ci-build.yaml
jamesnetherton: > The flatten-bom default to process-resources phase and is not installed in the first "mvn clean install" line ? If someone has light to share, that would be welcome.

I think the suggestion made [here](https://github.com/jamesnetherton/camel-quarkus/commit/281c5789e68733dfcbf8e16da9c24adfc15b8b1e#r100230924), was that we want the relevant POM hierarchy installed before `cq.flatten-bom.format` runs. Thus `mvnw clean install -pl poms/bom -am -Dcq.flatten-bom.format` would be enough to do everything we need in one step.

jamesnetherton: I have reduced the number of required `mvnw` steps as suggested.

aldettinger: Ok, so in such a situation. It's really needed to do it in 2 steps. Many thanks for sharing :+1:

zhfeng:
@jamesnetherton is it good to merge? I can't wait to use it!

jamesnetherton:
I commented above:

> We might want to hold merging until the next dependabot PR is opened. That way we could test and monitor the workflow to make sure nothing bad happens.

If folks feel ok about merging now, I have no objections.

zhfeng:
Nice, I will do a new release of  `quarkus-mybatis` for trigger a dependabot PR. Let's see what could happen!
Thanks again for your amazing works!

.github/workflows/ci-build.yaml
ppalaga: > The flatten-bom default to process-resources phase and is not installed in the first "mvn clean install" line ?

Yes, and I know no way how to do it better. The problem is that `-Dcq.flatten-bom.format` may add some exclusions and other edits to the source `pom.xml` that the further flattening does not see, because it uses Maven APIs to get the entries of the flattened POM. Maven APIs read the source pom.xml file only once before I do the edits. 
So currently, if `-Dcq.flatten-bom.format` does some edits, it throws an exception that recommends the user to invoke `mnv install` again. `--fail-never` is thus need so that it does not fail the whole CI job.

aldettinger: Thanks for sharing the details @ppalaga. Indeed, this is a tricky situation and no better idea than a 2 pass treatment from my side. That's great to have those flatten boms.

PR #4668, opened 2023-03-17
Fix #4667 to use quay.io/jbosstm/lra-coordinator

### LINKED ISSUE #4667: jbosstm/lra-coordinator is not available
I think we need to change to use `quay.io/jbosstm/lra-coordinator`

zhfeng:

### DIFF
--- a/integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
+++ b/integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
@@ -28,7 +28,7 @@ import org.testcontainers.containers.wait.strategy.Wait;
 public class LraTestResource implements QuarkusTestResourceLifecycleManager {
 
     private static final Integer LRA_PORT = AvailablePortFinder.getNextAvailable();
-    private static final String LRA_IMAGE = "jbosstm/lra-coordinator:5.12.0.Final";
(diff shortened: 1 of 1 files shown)


integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
```diff
@@ -28,7 +28,7 @@
 public class LraTestResource implements QuarkusTestResourceLifecycleManager {
 
     private static final Integer LRA_PORT = AvailablePortFinder.getNextAvailable();
-    private static final String LRA_IMAGE = "jbosstm/lra-coordinator:5.12.0.Final";
+    private static final String LRA_IMAGE = "quay.io/jbosstm/lra-coordinator";
```
jamesnetherton, line 31: Is there a tag that we can pin to?

aldettinger (approved):

integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
zhfeng: Good point! I think it could be `5.13.1.Final-2.16.4.Final` which is composed by `narayana-version` and `quarkus-version`. Anyway, I can check it with narayana team.

zbendhiba (approved):
Other than adding a tag, LGTM !

Issue #4676, opened 2023-03-20
jira extension throws ClassNotFoundException: com.fasterxml.jackson.datatype.joda.JodaModule


claudio4j:
When using camel-quarkus 2.13.2 and quarkus 2.13.5.Final

Given the following route
```
fromF("kamelet:jira-source?jql=RAW(project=MYP)&delay=30000&%s", params)
    .log("JIRA issue: ${body}");
```

And building in native mode with `mvn -Pnative -Dquarkus.native.builder-image=mandrel`
The execution throws:
```
ClassNotFoundException: com.fasterxml.jackson.datatype.joda.JodaModule
```

There is no error in jvm mode.

claudio4j:
While investigating, it led me to add most of jira model classes from package `com.atlassian.jira.rest.client.api.domain`
and `JSONObject, JSONArray` from `org.codehaus.jettison.json` to camel-quarkus-jira-deployment.
Also add `com.fasterxml.jackson.datatype.joda.JodaModule` and `org.joda.time.DateTime` to camel-quarkus-jackson-deployment, because jira model classes directly depends on jackson joda module.

Issue #4524, opened 2023-02-07
jira test fails with Camel 4 and Quarkus 3


ppalaga:
```
[ERROR] org.apache.camel.quarkus.component.jira.it.JiraTest  Time elapsed: 5.383 s  <<< ERROR!
java.lang.RuntimeException: java.lang.RuntimeException: Failed to start quarkus
        at io.quarkus.test.junit.QuarkusTestExtension.throwBootFailureException(QuarkusTestExtension.java:625)
        at io.quarkus.test.junit.QuarkusTestExtension.interceptBeforeAllMethod(QuarkusTestExtension.java:680)
        at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(InterceptingExecutableInvoker.java:103)
        at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.lambda$invoke$0(InterceptingExecutableInvoker.java:93)
        at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
        at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:156)
        at org.junit.jupiter.engine.extension.TimeoutExtension.interceptLifecycleMethod(TimeoutExtension.java:128)
        at org.junit.jupiter.engine.extension.TimeoutExtension.interceptBeforeAllMethod(TimeoutExtension.java:70)
        at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(InterceptingExecutableInvoker.java:103)
        at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.lambda$invoke$0(InterceptingExecutableInvoker.java:93)
        at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
        at org.junit.jupiter.engine.execution.InvocationInterceptorChain.proceed(InvocationInterceptorChain.java:64)
        at org.junit.jupiter.engine.execution.InvocationInterceptorChain.chainAndInvoke(InvocationInterceptorChain.java:45)
        at org.junit.jupiter.engine.execution.InvocationInterceptorChain.invoke(InvocationInterceptorChain.java:37)
[...]

PR #4677, opened 2023-03-20
Fix (jira): Add jira model classes and jackson joda datatype to @BuildStep (2.13.x)

claudio4j:
Fix https://github.com/apache/camel-quarkus/issues/4676

### DIFF
--- a/extensions/jackson/deployment/src/main/java/org/apache/camel/quarkus/component/jackson/JacksonProcessor.java
+++ b/extensions/jackson/deployment/src/main/java/org/apache/camel/quarkus/component/jackson/JacksonProcessor.java
@@ -19,7 +19,7 @@ package org.apache.camel.quarkus.component.jackson;
 import java.util.ArrayList;
 import java.util.List;
 
-import com.fasterxml.jackson.databind.JsonNode;
+import io.quarkus.bootstrap.classloading.QuarkusClassLoader;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
@@ -37,7 +37,11 @@ public class JacksonProcessor {
     List<ReflectiveClassBuildItem> registerReflectiveClasses() {
         List<ReflectiveClassBuildItem> items = new ArrayList<>();
         items.add(new ReflectiveClassBuildItem(false, true, "com.fasterxml.jackson.module.jaxb.JaxbAnnotationModule"));
-        items.add(new ReflectiveClassBuildItem(false, false, JsonNode.class));
+        items.add(new ReflectiveClassBuildItem(false, false, "com.fasterxml.jackson.databind.JsonNode"));
+        if (QuarkusClassLoader.isClassPresentAtRuntime("com.fasterxml.jackson.datatype.joda.JodaModule")) {
+            items.add(new ReflectiveClassBuildItem(true, true, "com.fasterxml.jackson.datatype.joda.JodaModule"));
+            items.add(new ReflectiveClassBuildItem(true, true, "org.joda.time.DateTime"));
+        }
         return items;
     }
 }
--- a/extensions/jira/deployment/src/main/java/org/apache/camel/quarkus/component/jira/deployment/JiraProcessor.java
+++ b/extensions/jira/deployment/src/main/java/org/apache/camel/quarkus/component/jira/deployment/JiraProcessor.java
@@ -19,17 +19,20 @@ package org.apache.camel.quarkus.component.jira.deployment;
 import java.util.ArrayList;
 import java.util.List;
 
-import com.atlassian.jira.rest.client.api.domain.Issue;
 import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
 import io.quarkus.deployment.builditem.ExtensionSslNativeSupportBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.deployment.builditem.IndexDependencyBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.NativeImageResourceBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+import org.jboss.jandex.IndexView;
 import org.joda.time.DateTimeZone;
 
 class JiraProcessor {
 
     private static final String FEATURE = "camel-jira";
+    private static final String JIRA_MODEL_PACKAGE = "com.atlassian.jira.rest.client.api.domain";
 
     @BuildStep
     ExtensionSslNativeSupportBuildItem activateSslNativeSupport() {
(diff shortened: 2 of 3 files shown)


extensions/jackson/deployment/src/main/java/org/apache/camel/quarkus/component/jackson/JacksonProcessor.java
```diff
@@ -37,7 +36,9 @@ FeatureBuildItem feature() {
     List<ReflectiveClassBuildItem> registerReflectiveClasses() {
         List<ReflectiveClassBuildItem> items = new ArrayList<>();
         items.add(new ReflectiveClassBuildItem(false, true, "com.fasterxml.jackson.module.jaxb.JaxbAnnotationModule"));
-        items.add(new ReflectiveClassBuildItem(false, false, JsonNode.class));
+        items.add(new ReflectiveClassBuildItem(false, false, com.fasterxml.jackson.databind.JsonNode.class));
```
zhfeng, line 39: It could use class name here.
```suggestion
        items.add(new ReflectiveClassBuildItem(false, false, "com.fasterxml.jackson.databind.JsonNode"));
```

extensions/jackson/deployment/src/main/java/org/apache/camel/quarkus/component/jackson/JacksonProcessor.java
```diff
@@ -37,7 +36,9 @@ FeatureBuildItem feature() {
...
+        items.add(new ReflectiveClassBuildItem(true, true, com.fasterxml.jackson.datatype.joda.JodaModule.class));
```
zhfeng, line 40: Same here.

extensions/jira/deployment/src/main/java/org/apache/camel/quarkus/component/jira/deployment/JiraProcessor.java
```diff
@@ -55,8 +54,121 @@ NativeImageResourceBuildItem nativeImageResources() {
...
+        items.add(new ReflectiveClassBuildItem(true, false, com.atlassian.jira.rest.client.api.domain.TimeTracking.class));
+        items.add(new ReflectiveClassBuildItem(true, false, com.atlassian.jira.rest.client.api.domain.Transition.class));
+        items.add(new ReflectiveClassBuildItem(true, false, com.atlassian.jira.rest.client.api.domain.User.class));
+        items.add(new ReflectiveClassBuildItem(true, false, com.atlassian.jira.rest.client.api.domain.Version.class));
+        items.add(new ReflectiveClassBuildItem(true, false,
+                com.atlassian.jira.rest.client.api.domain.VersionRelatedIssuesCount.class));
+        items.add(new ReflectiveClassBuildItem(true, false, com.atlassian.jira.rest.client.api.domain.Visibility.class));
+        items.add(new ReflectiveClassBuildItem(true, false, com.atlassian.jira.rest.client.api.domain.Votes.class));
+        items.add(new ReflectiveClassBuildItem(true, false, com.atlassian.jira.rest.client.api.domain.Watchers.class));
+        items.add(new ReflectiveClassBuildItem(true, false, com.atlassian.jira.rest.client.api.domain.Worklog.class));
```
zhfeng, line 169: It seems that we have to add all classes in `com.atlassian.jira.rest.client.api.domain` package?

zhfeng (changes requested):

extensions/jackson/deployment/pom.xml
```diff
@@ -45,6 +45,10 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-jackson</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-joda</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
jamesnetherton: We probably should not be adding `jackson-datatype-joda` as a dependency to `camel-quarkus-jackson` at all. Instead, we could have some logic to detect if its on the classpath and register `JodaModule` for reflection if it is.

Or if we must add it, add it as scope 'provided'.

extensions/jira/deployment/src/main/java/org/apache/camel/quarkus/component/jira/deployment/JiraProcessor.java
jamesnetherton: I am curious as to why this has not been a requirement before. I am guessing it's because you (or the Kamelet) is doing JSON serialization on these types?

extensions/jackson/deployment/src/main/java/org/apache/camel/quarkus/component/jackson/JacksonProcessor.java
claudio4j: Out of curiosity, the advantage to use the class name as string, is to not have the hard dependency ?

extensions/jira/deployment/src/main/java/org/apache/camel/quarkus/component/jira/deployment/JiraProcessor.java
claudio4j: Given the example route to retrieve jira issues, the message object is [`Issue`](https://bitbucket.org/atlassian/jira-rest-java-client/src/master/api/src/main/java/com/atlassian/jira/rest/client/api/domain/Issue.java) class, which has a hard dependency on many other jira classes. And it would be imprecise to say which classes are really required, as the issue itself could contain any of the attributes. 
I will try to reduce the number of classes.

zbendhiba: Maybe we can try to catch all those classe in this package, instead of naming each of them?

extensions/jackson/deployment/src/main/java/org/apache/camel/quarkus/component/jackson/JacksonProcessor.java
ppalaga: In the first place, strings allow registering also non-public classes that might be inaccessible from here. Second, a string class name allows to avoid loading the class at build time which might sometimes come in handy, e.g. if the given class initializer does something nasty. Not loading the class may make the build time processing generally faster, but I never collected any numbers. Referencing the classes is of course better for maintenance, because if the class is renamed or removed, the compilation fails.

claudio4j:
Should this change go to 2.15.x and 2.16.x branches as well ?

extensions/jira/deployment/src/main/java/org/apache/camel/quarkus/component/jira/deployment/JiraProcessor.java
```diff
@@ -55,8 +58,27 @@ NativeImageResourceBuildItem nativeImageResources() {
...
     @BuildStep
-    ReflectiveClassBuildItem registerForReflection() {
-        // Required by org.apache.camel.component.jira.consumer.WatchUpdatesConsumer
-        return new ReflectiveClassBuildItem(true, false, Issue.class.getName());
+    ReflectiveClassBuildItem registerJiraClassesForReflection(CombinedIndexBuildItem combinedIndex) {
+        IndexView index = combinedIndex.getIndex();
+        String[] modelClasses = index.getKnownClasses().stream()
+                .map(ci -> ci.name().toString())
+                .filter(n -> n.startsWith(JIRA_MODEL_PACKAGE))
+                .sorted()
```
aldettinger, line 66: Out of curiosity, is there any specific need to sort them ?

claudio4j: not really.

zbendhiba (approved):

zbendhiba:
> Should this change go to 2.15.x and 2.16.x branches as well ?

When this is merged, please open a PR on the master branch. Jira integration testing is temporarily disabled, but we wouldn't want to lose this change when this [issue](https://github.com/apache/camel-quarkus/issues/4524) is resolved.

extensions/jira/deployment/src/main/java/org/apache/camel/quarkus/component/jira/deployment/JiraProcessor.java
aldettinger: So maybe it could be removed if we consider it's better for build time and readibility.

aldettinger:
> > Should this change go to 2.15.x and 2.16.x branches as well ?
> 
> When this is merged, please open a PR on the main branch. Jira integration testing is temporarily disabled, but we wouldn't want to lose this change when this [issue](https://github.com/apache/camel-quarkus/issues/4524) is resolved.

Indeed, we would expect to have this one in main, 2.16.x and 2.13.x from my point of view.

claudio4j:
I tried to run it main  branch, but got some errors. I will try again with the recent changes and open the PR accordingly.
Thanks for the review.

jamesnetherton:
Thinking also that we should have a test for the use case that this change is being added for. Doesn't strictly need to be done in this PR, but we should at least follow up with another issue / PR if possible.

zhfeng (approved):

Issue #4723, opened 2023-03-30
Intermittent failure in Jdbc native tests


zhfeng:
```
2023-03-30 11:32:18,113 WARN  [io.agr.pool] (Camel (camel-1) thread #1 - timer://interval-polling) Datasource 'camel-ds': JDBC resources leaked: 0 ResultSet(s) and 1 Statement(s)
2023-03-30 11:32:18,116 ERROR [org.apa.cam.pro.err.DefaultErrorHandler] (Camel (camel-1) thread #1 - timer://interval-polling) Failed delivery for (MessageId: 4C4B23931637794-0000000000000000 on ExchangeId: 4C4B23931637794-0000000000000000). Exhausted after delivery attempt: 1 caught: org.h2.jdbc.JdbcSQLSyntaxErrorException: Table "CAMELSGENERATED" not found (this database is empty); SQL statement:
select * from camelsGenerated order by id desc [42104-214]
```

It looks like there is no data in the h2 database. I think  we could use [quarkus-flyway](https://quarkus.io/guides/flyway) to setup the database at the application startup time.

ldrozdo:
Hi @zhfeng, looks like @jamesnetherton already solved it, thanks for that :) I'm working on testing with multiple JDBC providers, I can check the quarkus-flyway and try to set up the database with it, thanks for the tip.

zhfeng:
Yeah, it looks good to me.

jamesnetherton:
https://github.com/apache/camel-quarkus/issues/4730

PR #4728, opened 2023-03-31
Start JDBC polling after database initialization #4723

jamesnetherton:
Should fix intermittent JDBC test failures.

I can leave #4723 open. If we want to start testing with multiple JDBC providers, then it might be better to have flyway...


oscerd (approved):

aldettinger (approved):
Ah interesting. Maybe this one would end up in 2.13.x too.

zbendhiba (approved):

zhfeng (approved):

zhfeng:
Thanks @jamesnetherton and nice work!

Issue #4470, opened 2023-02-01
Quarkiverse Tika incompatible with Camel 4


ppalaga:
https://github.com/quarkiverse/quarkus-tika/issues/22
https://github.com/quarkiverse/quarkus-tika/issues/79

zhfeng:
@ppalaga we already get a `2.0.0.CR` release of quarkus-tika.

Issue #3599, opened 2022-03-08
Tika 2.x incompatible with Quarkus Tika


jamesnetherton:
Camel has moved on to Tika 2.x, Quarkus Tika is currently aligned with 1.x.

Seems most functionality actually works when I exclude the `org.apache.tika` dependencies from `org.apache.camel:camel-tika`, except for the `textMain` content type:

```
Caused by: java.lang.NoClassDefFoundError: org/apache/tika/sax/boilerpipe/BoilerpipeContentHandler
	at org.apache.camel.component.tika.TikaProducer.getContentHandler(TikaProducer.java:154)
	at org.apache.camel.component.tika.TikaProducer.doParse(TikaProducer.java:117)
	at org.apache.camel.component.tika.TikaProducer.process(TikaProducer.java:91)
	at org.apache.camel.support.AsyncProcessorConverterHelper$ProcessorToAsyncProcessorBridge.process(AsyncProcessorConverterHelper.java:66)
	... 40 more
```

For native mode it's possible to fix and work around this via a substitution. For JVM mode it's a bit trickier to fix.

jamesnetherton:
I have a workaround in place until things are aligned better.

jamesnetherton:
See also https://github.com/quarkiverse/quarkus-tika/issues/22.

zhfeng:
[TikaProducerSubstitutions.java](https://github.com/apache/camel-quarkus/blob/camel-main/extensions%2Ftika%2Fruntime%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fcamel%2Fquarkus%2Fcomponent%2Ftika%2Fgraalvm%2FTikaProducerSubstitutions.java) is missing license header.

jamesnetherton:
> [TikaProducerSubstitutions.java](https://github.com/apache/camel-quarkus/blob/camel-main/extensions%2Ftika%2Fruntime%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fcamel%2Fquarkus%2Fcomponent%2Ftika%2Fgraalvm%2FTikaProducerSubstitutions.java) is missing license header.

I've added it now.

zhfeng:
@jamesnetherton shoud it be closed?

jamesnetherton:
> @jamesnetherton shoud it be closed?

I was leaving it open because the proper fix is to upgrade to a Quarkus Tika release that is aligned with Tika 2.x.

[...]

PR #4739, opened 2023-04-04
Add back Tika component

zhfeng:
It could close
- #4470 
- #3599


zbendhiba (approved):

oscerd (approved):

zhfeng:
@jamesnetherton It seems the maven connection timeout happens again and again. I just find
- https://github.com/actions/runner-images/issues/1499
and mentioned a work around to add 
`-Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120`.

How do you think about it?

jamesnetherton:
> How do you think about it?

We can add it to `CQ_MAVEN_ARGS` in each of our workflow YAML files.

poms/bom/pom.xml
```diff
@@ -6266,6 +6266,17 @@
                 <groupId>io.quarkiverse.tika</groupId>
                 <artifactId>quarkus-tika</artifactId>
                 <version>${quarkiverse-tika.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>commons-logging</groupId>
+                        <artifactId>commons-logging</artifactId>
+                    </exclusion>
+                    <!-- It could be removed when https://github.com/quarkiverse/quarkus-tika/issues/109 gets resolved -->
```
jamesnetherton, line 6274: Unless we can get a quick release of `quarkus-tika` containing the fix for this, we should probably open a follow-up issue here as a reminder to remove this exclusion.

jamesnetherton (approved):

poms/bom/pom.xml
zhfeng: I will take a new releas of `quarkus-tika` ASAP.

zhfeng: It should be fine with `2.0.0.CR2`.

jamesnetherton (approved):

aldettinger (approved):

PR #4740, opened 2023-04-04
Fix maven connection time out

zhfeng:

### DIFF
--- a/.github/workflows/camel-master-cron.yaml
+++ b/.github/workflows/camel-master-cron.yaml
@@ -25,7 +25,7 @@ on:
 env:
   LANG: en_US.UTF-8
   MAVEN_OPTS: -Xmx3000m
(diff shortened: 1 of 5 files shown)


zbendhiba (approved):

jamesnetherton (approved):

oscerd (approved):

aldettinger (approved):

zhfeng:
@ppalaga so it needs to change these configurations in `maven.config`, or it will be override?

jamesnetherton:
> @ppalaga so it needs to change these configurations in `maven.config`, or it will be override?

I think Peter was hinting at removing `maven.config` (or just removing the file contents). I don't think we want to rely on it for the CI build because we sometimes build other projects like Quarkus & Camel.

zhfeng:
OK, I see and +1 to remove it.

zhfeng:
The CI looks good and I will merge it.

PR #4755, opened 2023-04-07
Ref #4749 - java-joor-dsl - Add templated route support to native mode

### LINKED ISSUE #4749: java-joor-dsl - Add templated route support to native mode
If we try to add a `RouteBuilder` with a `templatedRoute` in native mode, no route will be created at runtime.

essobedo:
fixes #4749 

## Motivation

If we try to add a `RouteBuilder` with a `templatedRoute` in native mode, no route will be created at runtime.

## Modifications:

* Add the templated routes that are defined by the `RouteBuilder`s
* Add a unit test to ensure that it works as expected


ppalaga (approved):
Thanks @essobedo!

PR #4758, opened 2023-04-11
#4291 avoid In/Out Message soap headers conflict

### LINKED ISSUE #4291: [CXF] SoapFault: BSP:R3227: A SECURITY_HEADER MUST NOT contain more than one TIMESTAMP
This happens with WS-Trust and WS-SecurityPolicy tests ported from Quarkiverse CXF, the only difference being the Camel way (through `from("cxf:...")` clause) of deploying the SOAP endpoint.

I have tests reproducing this issue. Let me prepare the PRs.

ffang:
fix #4291

### DIFF
--- a/integration-test-groups/cxf-soap/cxf-soap-ws-security-server/src/main/java/org/apache/camel/quarkus/component/cxf/soap/securitypolicy/server/it/WsSecurityPolicyServerRoutes.java
+++ b/integration-test-groups/cxf-soap/cxf-soap-ws-security-server/src/main/java/org/apache/camel/quarkus/component/cxf/soap/securitypolicy/server/it/WsSecurityPolicyServerRoutes.java
@@ -19,9 +19,12 @@ package org.apache.camel.quarkus.component.cxf.soap.securitypolicy.server.it;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.inject.Produces;
 import jakarta.inject.Named;
+import org.apache.camel.Exchange;
+import org.apache.camel.Processor;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.component.cxf.jaxws.CxfEndpoint;
 import org.apache.cxf.ext.logging.LoggingFeature;
+import org.apache.cxf.headers.Header;
 
 @ApplicationScoped
 public class WsSecurityPolicyServerRoutes extends RouteBuilder {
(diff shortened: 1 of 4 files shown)


ppalaga:
I have restarted the failing CI jobs.

integration-test-groups/cxf-soap/cxf-soap-ws-security-server/src/main/java/org/apache/camel/quarkus/component/cxf/soap/securitypolicy/server/it/WsSecurityPolicyServerRoutes.java
```diff
@@ -19,20 +19,25 @@
...
     @Override
     public void configure() {
 
-        from("cxf:bean:wsSecurityPolicyHelloService?dataFormat=POJO")
-                .log("exchange: ${exchange}")
-                .setBody(exchange -> "Secure good morning " + exchange.getMessage().getBody(String.class));
-
+        from("cxf:bean:wsSecurityPolicyHelloService?dataFormat=POJO").process(new Processor() {
+            public void process(final Exchange exchange) throws Exception {
+                exchange.getIn().removeHeader(Header.HEADER_LIST);
```
ppalaga, line 37: @ffang could you please shed some light on where this HEADER_LIST is coming from and why we need to remove it to fix #4291

ppalaga (approved):
@ffang you cannot imagine how thankful I am for this fix! I am asking for some details inline.

aldettinger:
Maybe this one will need to be backported to 2.13.x before release too.

ppalaga:
> Maybe this one will need to be backported to 2.13.x before release too.

Yes

ppalaga:
I am working on a fix for the failure in CxfWsTrustIT.programmaticSts()

zbendhiba:
Thanks @ffang

integration-test-groups/cxf-soap/cxf-soap-ws-security-server/src/main/java/org/apache/camel/quarkus/component/cxf/soap/securitypolicy/server/it/WsSecurityPolicyServerRoutes.java
ppalaga: Great, thanks for the clarification @ffang! 

So similarly written route would also fail on CSB and there is nothing Camel Quarkus specifix about the failure. Nice to know, thanks a lot!

It sounds like this could happen to any user writing a new route using WS-Security/WS-SecurityPolicy. I wonder whether there is way to improve the default behavior? Shouldn't perhaps the security headers be filtered by default? Or wouldn't it be better not to propagate the headers by default unless the user sets the CxfHeaderFilterStrategy (thus giving a sign that he knows what he does)? If changing the code is not feasible, should we not warn about this in the documentation?

PR #4808, opened 2023-04-20
[2.13.x] Remove javax.servlet-api exclusion from jetty-server

jamesnetherton:
Relates to https://github.com/quarkusio/quarkus-platform/pull/816#issuecomment-1514539665.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -124,7 +124,7 @@
         <jnr-ffi.version>2.1.2</jnr-ffi.version><!-- Mess in web3j transitive deps -->
         <jsch.version>0.2.4</jsch.version><!-- @sync io.quarkiverse.jsch:quarkus-jsch-parent:${quarkiverse-jsch.version} prop:jsch.version -->
         <json-path.version>${json-path-version}</json-path.version>
-        <json-smart.version>2.4.7</json-smart.version>
+        <json-smart.version>2.4.7</json-smart.version><!-- @sync com.jayway.jsonpath:json-path:${json-path.version} dep:net.minidev:json-smart -->
         <kafka.version>3.2.3</kafka.version><!-- @sync io.quarkus:quarkus-bom:${quarkus.version} dep:org.apache.kafka:kafka-clients -->
         <kudu.version>${kudu-version}</kudu.version>
         <kotlin.version>1.7.20</kotlin.version><!-- @sync io.quarkus:quarkus-bom:${quarkus.version} dep:org.jetbrains.kotlin:kotlin-stdlib -->


oscerd (approved):

zhfeng (approved):

poms/bom/pom.xml
```diff
@@ -7249,10 +7249,6 @@
                                     <gavPattern>org.apache.kafka:connect-api</gavPattern>
                                     <addExclusions>javax.ws.rs:javax.ws.rs-api</addExclusions>
                                 </autogeneratedBomEntryTransformation>
-                                <autogeneratedBomEntryTransformation>
-                                    <gavPattern>org.eclipse.jetty:jetty-server</gavPattern>
-                                    <addExclusions>javax.servlet:javax.servlet-api</addExclusions>
-                                </autogeneratedBomEntryTransformation>
```
ppalaga, line 7255: Unfortunately, this will come back with the next `mvn process-resources -Dcq.flatten-bom.format`. These autogenerated exclusions are driven by banned dependencies defined by us in [tooling/enforcer-rules/camel-quarkus-banned-dependencies.xml](https://github.com/apache/camel-quarkus/blob/main/tooling/enforcer-rules/camel-quarkus-banned-dependencies.xml) or by [Quarkus](https://github.com/quarkusio/quarkus/blob/main/independent-projects/enforcer-rules/src/main/resources/enforcer-rules/quarkus-banned-dependencies.xml). javax.servlet:javax.servlet-api comes from Quarkus. There is currently no way to tell cq-maven-plugin to ignore some specific banned artifact only for some specific dependent, like we'd need here. It's now only possible to remove javax.servlet:javax.servlet-api from quarkus banns via XSLT in [tooling/enforcer-rules/quarkus-banned-dependencies.xsl](https://github.com/apache/camel-quarkus/blob/main/tooling/enforcer-rules/quarkus-banned-dependencies.xsl). I do not think we should do that. 
We could enhance cq plugin to be able to do these fine grained un-exclusions but I wonder if there are easier ways? 

Maybe we could stop managing jetty in the application bom and perhaps not manage it at all or manage it only in the test bom? 

Or we could perhaps upgrade to WireMock 3.x that runs on top of Jetty 11? [Here](https://github.com/wiremock/wiremock/issues/1760) they say they have -standalone artifacts, which perhaps shade jetty so the would be no conflict with those two JVM extensions using Jetty 9?

jamesnetherton: > Unfortunately, this will come back with the next mvn process-resources -Dcq.flatten-bom.format

That's what I assumed, but it does not seem to be the case when I run that command.

> upgrade to WireMock 3.x

I can try that. Alternatively, we could perhaps leave things as-is for 2.13.x, given we have a workaround for the platform side.

ppalaga: Ah, I am bind! I have not noticed, this is for 2.13. Sorry! Having javax.servlet:javax.servlet-api in 2.13 would not be that bad. Is it banned at all in 2.13? Let me check.

ppalaga: > Is it banned at all in 2.13

Yes, it is, by Quarkus. It would normally have to be replaced with jakarta servlet API. But we cannot expect end users (and Kogito team) to do that whenever they use WireMock. 
Can't we stop managing jetty-server in camel-quarkus-bom? It could work. 
Upgrading to wiremock 3.x would be only for main.

jamesnetherton: > Can't we stop managing jetty-server in camel-quarkus-bom

What happens if the extensions that need it pull different versions? Won't we get convergence check errors?

jamesnetherton:
I'll revist this in another PR.

poms/bom/pom.xml
ppalaga: > Won't we get convergence check errors?

Hard to say without trying

PR #4810, opened 2023-04-20
Upgrade quarkus-qpid-jms to 2.0.0

zhfeng:


pom.xml
```diff
@@ -81,6 +81,7 @@
         <azure-core-http-vertx.version>1.0.0-beta.3</azure-core-http-vertx.version> <!-- TODO: https://github.com/apache/camel-quarkus/issues/4181 -->
         <cassandra-driver-test.version>3.7.1</cassandra-driver-test.version><!-- Keep in sync with testcontainers instead of Debezium bom -->
         <bouncycastle.version>1.72</bouncycastle.version><!-- @sync io.quarkus:quarkus-bom:${quarkus.version} dep:org.bouncycastle:bcprov-jdk18on -->
+        <bouncycastle-jdk15on.version>1.70</bouncycastle-jdk15on.version>
```
zbendhiba, line 84: Do we need the JDK 15 version ?

zhfeng: The `camel-quarkus-pulsar` depends on it.

jamesnetherton: Can we exclude `jdk15` from pulsar and replace with the `jdk18` alternatives?

zhfeng: Let't me check it.

zhfeng: @jamesnetherton It looks good but there is no any test for crypto in `camel-quarkus-pulsar`.

jamesnetherton: > @jamesnetherton It looks good but there is no any test for crypto in `camel-quarkus-pulsar`.

I think for now we'll just have to live with that. I'd prefer we avoid the alternative of maintaining multiple BC versions if possible.

jamesnetherton (approved):

Issue #4707, opened 2023-03-28
MinIO: Extend test coverage


JiriOndrusek:
Following minio features should be covered by the tests:

- Producer operation `GetObjectRange
- Automatic detection of client from registry
- Bucket autocreation
- Moving data between buckets
- Move/Delete after read option
- Cover some potentially risky options (risky in the native mode) from CamelMinioContentMD5 , CamelMinioETag, CamelMinioStorageClass, CamelMinioVersionId, CamelMinioCannedAcl

PR #4828, opened 2023-04-25
MinIO: Extend test coverage #4707

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4707

Last item was not covered fully (`Cover some potentially risky options (risky in the native mode) from CamelMinioContentMD5 , CamelMinioETag, CamelMinioStorageClass, CamelMinioVersionId, CamelMinioCannedAcl`),
because the code responsible for this feature in Camel component is just very simple setter for MinIO.

### DIFF
--- a/integration-tests/minio/src/main/java/org/apache/camel/quarkus/component/minio/it/MinioResource.java
+++ b/integration-tests/minio/src/main/java/org/apache/camel/quarkus/component/minio/it/MinioResource.java
@@ -16,11 +16,14 @@
  */
 package org.apache.camel.quarkus.component.minio.it;
 
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import io.minio.GetObjectArgs;
+import io.minio.GetObjectResponse;
 import io.minio.Result;
 import io.minio.messages.Bucket;
 import io.minio.messages.Item;
@@ -30,13 +33,18 @@ import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import org.apache.camel.CamelContext;
 import org.apache.camel.ConsumerTemplate;
+import org.apache.camel.Exchange;
 import org.apache.camel.ProducerTemplate;
 import org.apache.camel.component.minio.MinioConstants;
 import org.apache.camel.component.minio.MinioOperations;
+import org.apache.camel.util.Pair;
 
 @Path("/minio")
 @ApplicationScoped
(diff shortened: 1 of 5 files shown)


integration-tests/minio/src/main/java/org/apache/camel/quarkus/component/minio/it/MinioResource.java
```diff
@@ -30,76 +34,169 @@
...
+    @Inject
+    CamelContext camelContext;
+
+    @Path("/consumerWithClientCreation/{endpoint}")
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public Response consumerWithClientCreation(@PathParam("endpoint") String endpoint) {
+        MinioComponent minioComponent = camelContext.getComponent("minio", MinioComponent.class);
+        MinioClient client = minioComponent.getConfiguration().getMinioClient();
+        minioComponent.getConfiguration().setMinioClient(null);
```
jamesnetherton, line 74: May be better to have a dedicated component instance for this, then you don't need to mess with the original `minio` instance.

E.g

```
@Produces
@Named("minioCamel")
MinioComponent camelMinioComponent() {
    // Code to set null client, autowired disabled etc
}
```

integration-tests/minio/src/test/java/org/apache/camel/quarkus/component/minio/it/MinioTestResource.java
```diff
@@ -30,6 +30,7 @@ public class MinioTestResource implements QuarkusTestResourceLifecycleManager {
     public static final String CONTAINER_SECRET_KEY = "MINIO_SECRET_KEY";
     private final String CONTAINER_IMAGE = "minio/minio:RELEASE.2020-12-03T05-49-24Z";
```
jamesnetherton, line 31: Might be good to update the container image. Maybe to this:

https://github.com/quarkiverse/quarkus-minio/blob/main/minio-client/deployment/src/main/java/io/quarkiverse/minio/client/deployment/devservices/MinioDevServicesBuildTimeConfig.java#L29

zbendhiba:
@JiriOndrusek Since you are working on Minio, maybe you could take a look at test failing while upgrading the dependency. cf https://github.com/apache/camel-quarkus/pull/4831

integration-tests/minio/src/main/java/org/apache/camel/quarkus/component/minio/it/MinioResource.java
JiriOndrusek: you are right, that would be "cleaner" solution, thanks for the hint, I'll change that

JiriOndrusek: done

integration-tests/minio/src/test/java/org/apache/camel/quarkus/component/minio/it/MinioTestResource.java
JiriOndrusek: done

jamesnetherton:
Is this good to merge? I guess we have to delay the Minio upgrade until the okhttp issue is fixed.

JiriOndrusek:
It is good to merge. This PR extends coverage on current version.

Upgrade of minio might not happen because of https://github.com/quarkiverse/quarkus-minio/issues/220

PR #4861, opened 2023-04-29
simple("${exchange.getMessage().getBody()}") causes a MethodNotFoundException in native mode

### LINKED ISSUE #4860: simple("${exchange.getMessage().getBody()}") causes a MethodNotFoundException in native mode
# Bug description
Using `simple("${exchange.getMessage().getBody()}"` throws an exception at runtime when compiling natively:

```
Caused by: org.apache.camel.component.bean.MethodNotFoundException: Method with name: getMessage() not found on bean: Exchange[ED6D02F79FBB419-0000000000000000] of type: org.apache.camel.support.DefaultExchange on the exchange: Exchange[ED6D02F79FBB419-0000000000000000]
...
```

# Affected versions:
Quarkus,  at least from `2.13.x.Final` up to the current version (`3.0.1.Final`). 

# Reproducer:
- clone [this repository (`github.com`)](https://github.com/turing85/camel-simple-language-bug):
    ```bash
    git clone https://github.com/turing85/camel-simple-language-bug.git
    cd camel-simple-language-bug
    ```
- Build the application:
    ```bash
    ./mvnw clean verify
    ```
- Observe that the tests pass:
    ```bash
    ...
    [INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
    ...
    ```
- Build the application natively:
    ```bash
    ./mvn --define native clean verify
    ```
- Observe that the tests fail with the above mentioned error in the application log, and an unexpected status code on the tests, reporting a status of `500` instead of the expected `200`:
    ```bash
    ...
    Caused by: org.apache.camel.component.bean.MethodNotFoundException: Method with name: getMessage() not found on bean: Exchange[FD7D7D11C0B4B98-0000000000000000] of type: org.apache.camel.support.DefaultExchange on the exchange: Exchange[FD7D7D11C0B4B98-0000000000000000]
    ...
    Expected status code <200> but was <500>.
    ...
    ```

# Workaround
Add the following to `application.yml`

```yaml
quarkus:
  camel:
    native:
      reflection:
        include-patterns:
          - org.apache.camel.support.AbstractExchange
          - org.apache.camel.support.MessageSupport
```

# Possible solution
The corresponding extension should register those classes for reflection by default. This, however, begs the question: what else should be registered? We should analyze all relevant classes used by the `simple(...)` language and ensure that they all work as expected.

ppalaga:
Related #4860

### DIFF
--- a/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelNativeImageProcessor.java
+++ b/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelNativeImageProcessor.java
@@ -139,6 +139,11 @@ public class CamelNativeImageProcessor {
                 .map(ai -> ai.target().asMethod())
                 .forEach(mi -> reflectiveMethod.produce(new ReflectiveMethodBuildItem(mi)));
 
+        reflectiveClass.produce(
+                ReflectiveClassBuildItem.builder(
+                        "org.apache.camel.support.AbstractExchange",
+                        org.apache.camel.support.MessageSupport.class.getName())
+                        .methods().build());
     }
 
     @BuildStep
--- a/integration-test-groups/foundation/core-languages/src/main/java/org/apache/camel/quarkus/core/languages/it/SimpleRoutes.java
+++ b/integration-test-groups/foundation/core-languages/src/main/java/org/apache/camel/quarkus/core/languages/it/SimpleRoutes.java
@@ -36,6 +36,9 @@ public class SimpleRoutes extends RouteBuilder {
         from("direct:languageSimple")
                 .transform().language("simple", "Hello ${body} from language().simple()");
 
+        from("direct:simpleExchangeMethods")
+                .transform().simple("exchange.getMessage().getBody() evals to ${exchange.getMessage().getBody()}");
+
     }
 
 }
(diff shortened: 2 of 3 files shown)


extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelNativeImageProcessor.java
```diff
@@ -139,6 +139,11 @@ void reflectiveItems(
                 .map(ai -> ai.target().asMethod())
                 .forEach(mi -> reflectiveMethod.produce(new ReflectiveMethodBuildItem(mi)));
 
+        reflectiveClass.produce(
+                ReflectiveClassBuildItem.builder(
+                        "org.apache.camel.support.AbstractExchange",
+                        org.apache.camel.support.MessageSupport.class.getName())
```
turing85, line 145: Have you checked whether there are possibly other classes that should be registered in order for the simple language to work as expected on the `exchange`?

ppalaga: No, I have not. Do you have any other scenarios in mind this is not covering?

turing85: None I found. As I wrote in #4860 :

> ... This, however, begs the question: what else should be registered? We should analyze all relevant classes used by the `simple(...)` language and ensure that they all work as expected.

ppalaga: Thanks for explaining. Yeah, it would be nice to audit whether we need more. While I do not feel like I have free capacity do that these days, we can still keep the issue open.

PR #4864, opened 2023-04-30
Limit aether.connector.http.connectionMaxTtl to 120 seconds

ppalaga:
... and some workarounds to get aether.connector.http.connectionMaxTtl honored


ppalaga:
I am not going to wait for the examples. They use their own mvnw anyway.

PR #4865, opened 2023-04-30
[2.13.x] Backports

ppalaga:


ppalaga:
I have filed https://github.com/apache/camel-quarkus/issues/4866 for the failure in openstack test. It is passing for me locally on Java 17, thus merging.

PR #4869, opened 2023-05-02
Update documentation references for javax packages to jakarta

jamesnetherton:


ppalaga:
I introduced an AsciiDoc attribute for javax/jakarta in QCXF, so that docs improvements stay portable to older branches: 
https://github.com/quarkiverse/quarkus-cxf/blob/13f4192322e60ed8ddccb15ed8612adb23504ff1/docs/modules/ROOT/pages/reference/extensions/quarkus-cxf-rt-features-logging.adoc?plain=1#L96-L97
I wonder how much sense it would make also here? Maybe an overkill, because we do not add so much new docs in Camel Quarkus these days. End of thinking loud :)

jamesnetherton:
> because we do not add so much new docs in Camel Quarkus these days

Yes, true. I have no strong opinion on whether we add a new attribute or not.

If you think it'd be better to have it, I'm happy to add it.

jamesnetherton:
I'll merge as-is for now. If we find backports start to become an issue, I'll introduce the AsciiDoc attribute.

zbendhiba:
thanks @jamesnetherton

PR #4867, opened 2023-05-02
Snmp: reworked tests to avoid flaky failures

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4850

Attempt to remove possible `ConcurrentModificationException` on `Vector`

PR #4871, opened 2023-05-04
Upgraded snmp4j to another provider to avoid flaky tests

JiriOndrusek:
follows https://github.com/apache/camel-quarkus/pull/4870
follows https://github.com/apache/camel-quarkus/pull/4867#issuecomment-1534290658

Camel in `2.13.x`  brings `org.apache.servicemix.bundles.snmp4j` which causes random errors in snmp tests (with low resources). For example:
- java.nio.BufferOverflowException
- java.util.ConcurrentModificationException
- The actual length of the SEQUENCE object org.snmp4j.PDUv1 is 11, but 14 was expected

Solution is to switch to `org.snmp4j.snmp4j` (by forcing in CQ BOM - which is purpose of this PR).
There is different method signature (during call of ` SecurityModels.getInstance().addSecurityModel(this.usm);`, therefore both producers have to be recompiled in CQ. Because both classes contains a lot of private fileds, it was not possible to override only 1 method. Therefore both classes are copied from the Camel (without any change). If Camel 3.x in future contains `org.snmp4j.snmp4j`. This PR could be reverted.

### DIFF
--- a/extensions-jvm/snmp/deployment/src/main/java/org/apache/camel/quarkus/component/snmp/deployment/SnmpProcessor.java
+++ b/extensions-jvm/snmp/deployment/src/main/java/org/apache/camel/quarkus/component/snmp/deployment/SnmpProcessor.java
@@ -21,7 +21,10 @@ import io.quarkus.deployment.annotations.ExecutionTime;
 import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.pkg.steps.NativeBuild;
+import org.apache.camel.component.snmp.SnmpComponent;
+import org.apache.camel.quarkus.component.snm.graal.SnmpRecorder;
 import org.apache.camel.quarkus.core.JvmOnlyRecorder;
+import org.apache.camel.quarkus.core.deployment.spi.CamelBeanBuildItem;
 import org.jboss.logging.Logger;
 
 class SnmpProcessor {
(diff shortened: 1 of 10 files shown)


JiriOndrusek:
@ppalaga @jamesnetherton FYI

zhfeng (approved):

jamesnetherton (approved):

ppalaga:
Great work, @JiriOndrusek! I am totally happy that we can make this work also on CQ 2.13.

I have a couple of questions/suggestions:

1. Could you please check what is the impact on the generated catalog in `catalog/target/classes/org/apache/camel/catalog/quarkus` after building the whole source tree before and after the change? Ideally, there should be no changes. Esp. snmp.json should still be there and the metadata in it should mirror the real state. 

2. We should inform users in the docs that Camel Quarkus uses a different protocol implementation than plain Camel 3.18. Explain reasons, possible gotchas (if any) and explain that `org.snmp4j.snmp4j` will be used in Camel 4 and CQ 3 anyway.

3. Practical: Could please the code copied from Camel keep the original java source file structure, so that it is easier to compare what changes have been done between the two and so that it is easier to port any future changes from Camel 3.18?

4. Question: so now the SNMP tests in CQ 2.13.x and main are 100% identical, right?

JiriOndrusek:
@ppalaga thanks for the suggestions!

1 - There is no difference (I used `diff` with the generated stuff before this PR and after PR)

2,3- Following PR (https://github.com/apache/camel-quarkus/pull/4873) adds doc, refactors class files from Camel (I had to change the name of the files to avoid duplicate java class error) 

4 - PR for 2 and 3 with a small upstream change (https://github.com/apache/camel-quarkus/pull/4872), makes tests in CQ 2.13.x and main 100% identical (of course there is a difference javax vs jakarta)

ppalaga:
@JiriOndrusek thanks for adressing the suggestions!

I dared to fixup some small glitches in this PR by creating a new PR to avoid another review round: https://github.com/apache/camel-quarkus/pull/4876

First, `snm` in the `org.apache.camel.quarkus.component.snm.graal` package name looked like a typo. I changed it to simply `org.apache.camel.quarkus.component.snmp` because we tend to use graal packages for substitutions. I also moved the camel classes to `org.apache.camel.quarkus.component.snmp` because I think there is no reason to introduce a duplicate package. 
Other than that, i rephrased the docs text a bit. I hope I have not broken anything.

PR #4918, opened 2023-05-19
[IBM-MQ] Add tests for IBM MQ client

avano:
- add ibm mq test resource + tests with ibm mq client
- changed messaging tests to use "predictable" destination names (named after the test method), as IBM MQ requires destinations to exist before one can use them

### DIFF
--- a/integration-tests/jms-ibmmq-client/pom.xml
+++ b/integration-tests/jms-ibmmq-client/pom.xml
@@ -1,0 +1,168 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one or more
+    contributor license agreements.  See the NOTICE file distributed with
+    this work for additional information regarding copyright ownership.
+    The ASF licenses this file to You under the Apache License, Version 2.0
+    (the "License"); you may not use this file except in compliance with
+    the License.  You may obtain a copy of the License at
+
+         http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-build-parent-it</artifactId>
+        <version>3.0.0-SNAPSHOT</version>
+        <relativePath>../../poms/build-parent-it/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-integration-test-jms-ibmmq-client</artifactId>
+    <name>Camel Quarkus :: Integration Tests :: JMS IBM MQ Client</name>
+    <description>Integration tests for Camel Quarkus JMS extension with the IBM MQ Client</description>
+
+    <properties>
+    </properties>
+
+    <dependencies>
+        <!-- Messaging extension to test -->
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-jms</artifactId>
+        </dependency>
+
+        <!-- Inherit base messaging routes -->
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-integration-test-messaging-jms</artifactId>
+        </dependency>
+
+        <!-- The JMS client library to test with -->
+        <dependency>
+            <groupId>com.ibm.mq</groupId>
+            <artifactId>com.ibm.mq.jakarta.client</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- test dependencies -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- IBM MQ Test Resource -->
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>junit</groupId>
+                    <artifactId>junit</artifactId>
+                </exclusion>
+            </exclusions>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit4-mock</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <!-- Inherit base messaging tests -->
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-integration-test-messaging-common</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-integration-test-messaging-jms</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+    </dependencies>
+
+    <profiles>
+        <!-- IBM MQ client does not work in native: https://github.com/apache/camel-quarkus/issues/4924
+        <profile>
+            <id>native</id>
+            <activation>
+                <property>
+                    <name>native</name>
+                </property>
+            </activation>
+            <properties>
+                <quarkus.package.type>native</quarkus.package.type>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>integration-test</goal>
+                                    <goal>verify</goal>
+                                </goals>
+                            </execution>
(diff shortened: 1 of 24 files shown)


integration-tests/jms-ibmmq-client/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-integration-test-messaging-jms</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+    </dependencies>
+
+    <profiles>
+        <!-- IBM MQ client does not work in native
```
ppalaga, line 92: Could you please file a Camel Quarkus issue summarizing why it does not work and paste the issue URL here?

integration-tests/jms-ibmmq-client/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-integration-test-messaging-jms</artifactId>
+        </dependency>
+
+        <!-- The JMS client library to test with -->
+        <dependency>
+            <groupId>com.ibm.mq</groupId>
+            <artifactId>com.ibm.mq.jakarta.client</artifactId>
+            <scope>provided</scope>
```
ppalaga, line 54: Could you please explain why is the client jar provided?
It makes sense when writing a web app for a traditional EE container where the client jar is indeed provided by the container. But there is no Quarkus container that would provide stuff like this.

integration-tests/jms-ibmmq-client/src/test/java/org/apache/camel/quarkus/component/jms/ibmmq/it/IBMMQTest.java
```diff
@@ -0,0 +1,91 @@
...
+        for (Method method : test.getTestClass().get().getMethods()) {
+            destinations.createQueue(method.getName());
+            // Some tests use two queues
+            destinations.createQueue(method.getName() + "2");
+            destinations.createTopic(method.getName());
+        }
+
+        // Sleep for a while to allow the routes to reconnect
+        try {
+            Thread.sleep(5000L);
```
ppalaga, line 63: Can't we perhaps create the queues and topics in the `IBMMQTestResource`? That one is started before the application under test, so we'd perhaps could avoid this reconnection interval?

ppalaga:
Very nice, thanks a lot, @avano!

Besides the comments inline, I wonder why we need the separate camel-quarkus-integration-tests-support-ibm-mq module? It would make sense when it would be re-used at least twice, but there is no re-use here. Maybe there are other reasons for this structure?

integration-tests/jms-ibmmq-client/pom.xml
avano: that's an info I got from James (and Andrea), that due to a different license in the ibm client (and apache's policy), such dependencies can be only defined with `provided` or `system` scope

integration-tests/jms-ibmmq-client/src/test/java/org/apache/camel/quarkus/component/jms/ibmmq/it/IBMMQTest.java
avano: :thinking: is there a way how to access some test info in a test resource? that was my initial attempt, but I couldn't find a way how to do that, other than just hardcoding all the destinations it should create (which isn't very maintainable)

avano:
> Very nice, thanks a lot, @avano!
> 
> Besides the comments inline, I wonder why we need the separate camel-quarkus-integration-tests-support-ibm-mq module? It would make sense when it would be re-used at least twice, but there is no re-use here. Maybe there are other reasons for this structure?

you're right, currently a separate module is not needed. I though for future-proofing it would be easier to just add a dependency to a new module, rather than refactoring it, but maybe there isn't such case where another module would need it. So if you prefer I can squash it together

integration-tests/jms-ibmmq-client/pom.xml
zhfeng: If it does not support native mode right now, it should be moved to `integration-tests-jvm`.

integration-tests/jms-ibmmq-client/src/test/java/org/apache/camel/quarkus/component/jms/ibmmq/it/IBMMQTest.java
zhfeng: Is there any particular reason to use method name as queue and topic?

integration-tests/jms-ibmmq-client/pom.xml
jamesnetherton: > If it does not support native mode right now, it should be moved to `integration-tests-jvm`.

Initially I suggested to not do that as the messaging test 'framework' all lives under `integration-tests` currently.

poms/bom-test/pom.xml
```diff
@@ -399,6 +403,11 @@
                 <version>${unboundid-ldapsdk.version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
+                <groupId>com.ibm.mq</groupId>
+                <artifactId>com.ibm.mq.jakarta.client</artifactId>
+                <version>${ibm.mq.client.version}</version>
```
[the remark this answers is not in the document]
zhfeng: I think it should be enough to managed in `bom`.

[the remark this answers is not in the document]
avano: that didn't work for me as the `bom-test` does not inherit from the main `bom`

integration-tests/jms-ibmmq-client/pom.xml
ppalaga: Ok, thanks for explaining!

integration-tests/jms-ibmmq-client/src/test/java/org/apache/camel/quarkus/component/jms/ibmmq/it/IBMMQTest.java
ppalaga: I'd vote for using names having some random suffix. You could create those in the `IBMMQTestResource` and pass them to the application under test via config properties returned from IBMMQTestResource.start().

ppalaga:
> if you prefer I can squash it together

Yes, please :)

integration-tests/jms-ibmmq-client/src/test/java/org/apache/camel/quarkus/component/jms/ibmmq/it/IBMMQTest.java
avano: @ppalaga could you please elaborate? Wouldn't that require maintaining a list of destination names that should be created inside `IBMMQTestResource` when a new test is added to, for example, [AbstractMessagingTest](https://github.com/avano/camel-quarkus/blob/ibm-mq/integration-tests/messaging/common/src/test/java/org/apache/camel/quarkus/component/messaging/it/AbstractMessagingTest.java)?

Or did you mean to re-use one queue/topic for all tests?

poms/bom-test/pom.xml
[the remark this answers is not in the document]
avano: ok I see it now, the tests actually use the `bom` through `camel-quarkus-build-parent-it`. My issue is, when I add a dependency to the `poms/bom/pom.xml` [here](https://github.com/avano/camel-quarkus/blob/ibm-mq/poms/bom/pom.xml#L6058-L6062), after I build the bom with `mvn clean install`, the artifact is not present in the installed bom in the local repository. Am I doing it wrong?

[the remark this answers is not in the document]
ppalaga: Good point! We are flattening the BOM based on the set of artifacts that are actually required by Camel Quarkus. The set is defined by all our extensions plus some artifacts on top of it. Because the IBM MQ connector is not a transitive dep of any extensions, it needs to be added manually around here: https://github.com/apache/camel-quarkus/blob/main/poms/bom/pom.xml#L6739C30-L6750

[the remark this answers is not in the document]
avano: that did the trick, thanks!

integration-tests/jms-ibmmq-client/src/test/java/org/apache/camel/quarkus/component/jms/ibmmq/it/IBMMQTest.java
ppalaga: Yes, IBMMQTestResource would have to decide about the actual (random) names of the resources. The property names can be fixed in some common enum. I'd personally be for going that way rather than creating the resources in the test classes and having to wait for reconnecting. I guess the maintenance overhead will be smaller than the overall time lost due to waiting for reconnects. 

But maybe the problem with waiting for reconnect can be solved somehow. E.g. you do not auto-start the routes and instead you start them via REST from the test after you have created the required queues?

avano: > But maybe the problem with waiting for reconnect can be solved somehow. E.g. you do not auto-start the routes and instead you start them via REST from the test after you have created the required queues?

:+1: that should work, I'll look into it

integration-tests/messaging/common/src/test/java/org/apache/camel/quarkus/component/messaging/it/AbstractMessagingTest.java
```diff
@@ -25,15 +25,42 @@
...
+    protected String queue2;
+    protected String topic;
+
+    @BeforeAll
+    public void startRoutes(TestInfo info) {
+        // At this point RestAssured is not configured to use the test port
+        // see https://github.com/quarkusio/quarkus/issues/7690#issuecomment-596543310
+        // The comment states it does not work in native, however it seems to work fine
+        RestAssured.given()
+                .port(ConfigProvider.getConfig().getValue("quarkus.http.test-port", Integer.class))
```
avano, line 53: it we want to avoid this, it can be moved to `BeforeEach` where it should work without the property, either adding a boolean to the resource to start the routes only once, or just call `startAllRoutes` everytime

ppalaga: I do not mind leaving it like this as long as it works in both JVM and native modes.

ppalaga (approved):

zhfeng (approved):

jamesnetherton (approved):

zhfeng:
Thanks @avano and it is really great work! The failing test with netty seems unrelated. I will re-run the task to see if CI can pass.

ppalaga:
Thanks a lot @avano!

PR #4921, opened 2023-05-19
Remove "none" WSDLs as they are not needed

llowinge:


ppalaga:
The failure in KNative tests is unrelated.

PR #4936, opened 2023-05-26
Restore master itest platform exclusion

jamesnetherton:


oscerd (approved):

aldettinger (approved):
I reached the same conclusion. No idea how it was introduced.

@jamesnetherton I may update the platform pull request if noone is already on it

Issue #4582, opened 2023-02-16
Expand micrometer test coverage


JiriOndrusek:
The following scenarios from the [documentation](https://camel.apache.org/components/3.20.x/micrometer-component.html) are not covered in Camel Quarkus:

- message headers (just increment/decrement for counter and start/stop for timer)
- event notification - cover dump to json
- if possible cover instrumenting thread pool
-  exposing to JMX (JVM only)

Because micrometer extension leverages quarkus extension (see the [documentaion](https://quarkus.io/guides/micrometer#does-micrometer-support-annotations)) it would be nice to cover major features - it is **optional**.

- try MeterFilter
- annotations
- customized meterRegistry
- microprofile metrics API
-  review gauge metrics coverage

JiriOndrusek:
Thanks @llowinge for the information. I reported those problems in its own ticket https://github.com/apache/camel-quarkus/issues/4598 and will look at it.

JiriOndrusek:
@llowinge I found a reason of different behavior in Quarkus and reported a ticket there - https://github.com/quarkusio/quarkus/issues/31433

JiriOndrusek:
Test coverage revealed missing feature `instrumenting core thread pool`. Reported as https://github.com/apache/camel-quarkus/issues/4934.

ppalaga:
Is there anyhing more to do here @JiriOndrusek. Can't we perhaps close this issue?

JiriOndrusek:
There will be another PR covering the missing parts. Here is my WIP https://github.com/apache/camel-quarkus/compare/main...JiriOndrusek:CQ4582-expand-micrometer-coverage?expand=1
I hope to finish it this week or the next one.

PR #4951, opened 2023-06-01
Expand micrometer test coverage #4582

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4582

1. I added a test for jmx access to metrics (JVM only). For that porpose I had to add a dependency `io.micrometer:micrometer-registry-jmx` into CQ BOM.
2. Prometheus metrics ignores negative values for counter. I put that information into `limitations.adoc`, because there is Camel header `decrement` which is meant to be used with counter metrics. Possibly better place for such information is Camel itself.

### DIFF
--- a/extensions/micrometer/runtime/src/main/doc/limitations.adoc
+++ b/extensions/micrometer/runtime/src/main/doc/limitations.adoc
@@ -1,0 +1,8 @@
+=== Exposing Micrometer statisctics in JMX
+
+The `Exposing Micrometer statisctics in JMX` action is not available in native mode as JMX is not supported on GraalVM.
+
+=== Decrement header for Counter is ignored by Prometheus
+
+Prometheus backend ignores negative values during increment of Counter metrics.
+
--- a/docs/modules/ROOT/pages/reference/extensions/micrometer.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/micrometer.adoc
@@ -48,41 +48,30 @@ endif::[]
 == Usage
 This extension leverages https://quarkus.io/guides/micrometer[Quarkus Micrometer]. Quarkus supports a variety of Micrometer metric registry implementations.
 
-Your application should declare one of the following dependencies, depending on the monitoring solution you want to work with.
+Your application should declare the following dependency or  one of the dependencies listed in the https://quarkiverse.github.io/quarkiverse-docs/quarkus-micrometer-registry/dev/index.html[quarkiverse documentation], depending on the monitoring solution you want to work with.
 
-[cols="50,.^50]
-|===
-|Monitoring System | Dependency
+[source,xml]
+----
+<dependency>
+    <groupId>io.micrometer</groupId>
+    <artifactId>micrometer-registry-prometheus</artifactId>
+</dependency>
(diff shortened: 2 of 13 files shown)


ppalaga:
> So maybe a better way would be to add `io.quarkiverse.micrometer.registry:quarkus-micrometer-registry-jmx` only to the test BOM.

I see it managed in quarkus-bom: https://repo1.maven.org/maven2/io/quarkus/quarkus-bom/3.1.0.Final/quarkus-bom-3.1.0.Final.pom

> It's a bit of an open question whether we even need to test the JMX bits at all, given it's already tested elsewhere.

JMX support is underway for "GraalVM for JDK 17 / JDK 20 (June 13, 2023)", so these tests may come in handy for native testing, no?

jamesnetherton:
> I see it managed in quarkus-bom

That's just the plain `micrometer-registry-jmx` dependency. Not the Quarkus extension that lives here:

https://github.com/quarkiverse/quarkus-micrometer-registry/

ppalaga:
> > I see it managed in quarkus-bom
> 
> That's just the plain `micrometer-registry-jmx` dependency. Not the Quarkus extension that lives here:
> 
> https://github.com/quarkiverse/quarkus-micrometer-registry/

I am blind, sorry!

JiriOndrusek:
> I am not sure if we want to be supporting `micrometer-registry-jmx` directly in this project. Otherwise, we have to maintain the dependencies, tests and docs explaining how to configure it.
> 
> Support for the different Micrometer registry implementations lives here:
> 
> https://github.com/quarkiverse/quarkus-micrometer-registry/
> 
> So maybe a better way would be to add `io.quarkiverse.micrometer.registry:quarkus-micrometer-registry-jmx` only to the test BOM. And just replace the table in our [usage docs](https://camel.apache.org/camel-quarkus/next/reference/extensions/micrometer.html#extensions-micrometer-usage) with a link to [here](https://quarkiverse.github.io/quarkiverse-docs/quarkus-micrometer-registry/dev/index.html) .
> 
> It's a bit of an open question whether we even need to test the JMX bits at all, given it's already tested elsewhere.
> 
> WDYT @ppalaga?

The dependency is changed and moved.
Table in the doc is replaced but I had to keep onel dependency for prometheus which is not part of the quarkiverse doc.

ppalaga (approved):

ppalaga:
Thanks @JiriOndrusek!

extensions/micrometer/runtime/src/main/doc/limitations.adoc
```diff
@@ -0,0 +1,8 @@
+=== Exposing Micrometer statisctis in Jmx
+
+The `Exposing Micrometer statisctis in Jmx` action is not available in native mode as JMX is not supported on GraalVM.
```
jamesnetherton, line 3:
```suggestion
The `Exposing Micrometer statistics in JMX` action is not available in native mode as JMX is not supported on GraalVM.
```

JiriOndrusek: fixed

PR #4980, opened 2023-06-12
camel-quarkus-crypto: Added test to sign/verify raw keys

### LINKED ISSUE #4979: camel-quarkus-crypto: Add test coverage for raw keys
The camel docs for the crypto component describe a use case where raw keys are used to sign/verify an exchange. Here, raw keys are those that have been created by a key generator and not loaded from a keystore.

Add a test for this scenario to ensure it works in native mode.

djcoleman:
Also removed unused dependency on javassist from bouncycastle-support

Fixes #4979

[x] An issue should be filed for the change unless this is a trivial change (fixing a typo or similar). One issue should ideally be fixed by not more than one commit and the other way round, each commit should fix just one issue, without pulling in other changes.
[x] Each commit in the pull request should have a meaningful and properly spelled subject line and body. Copying the title of the associated issue is typically enough. Please include the issue number in the commit message prefixed by #.
[x] The pull request description should explain what the pull request does, how, and why. If the info is available in the associated issue or some other external document, a link is enough.
[x] Phrases like Fix #<issueNumber> or Fixes #<issueNumber> will auto-close the named issue upon merging the pull request. Using them is typically a good idea.
[x] Please run mvn process-resources -Pformat (and amend the changes if necessary) before sending the pull request.
[x] Contributor guide is your good friend: https://camel.apache.org/camel-quarkus/latest/contributor-guide.html

### DIFF
--- a/extensions-support/bouncycastle/runtime/pom.xml
+++ b/extensions-support/bouncycastle/runtime/pom.xml
@@ -55,10 +55,6 @@
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcpkix-jdk18on</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.javassist</groupId>
-            <artifactId>javassist</artifactId>
-        </dependency>
     </dependencies>
 
     <build>
--- a/integration-tests/crypto/src/main/java/org/apache/camel/quarkus/component/crypto/it/CryptoResource.java
+++ b/integration-tests/crypto/src/main/java/org/apache/camel/quarkus/component/crypto/it/CryptoResource.java
@@ -29,6 +29,7 @@ import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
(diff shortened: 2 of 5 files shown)


jamesnetherton (approved):

extensions-support/bouncycastle/runtime/pom.xml
```diff
@@ -55,10 +55,6 @@
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcpkix-jdk18on</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.javassist</groupId>
-            <artifactId>javassist</artifactId>
-        </dependency>
```
ppalaga, line 61: It might be that this was the only javaassist reference in the whole source tree because of which we were modifying quarkus enforcer bans around here: https://github.com/apache/camel-quarkus/blob/main/tooling/enforcer-rules/quarkus-banned-dependencies.xsl#LL33C64-L33C102
Could you please remove `text() = 'org.javassist:javassist' or ` from there and see whether enforcer will complain about javassist somewhere else?

djcoleman: I ran `mvn dependency:tree` and there are no other references to javassist. Also running the enforcer with that constraint removed didn't produce any errors.

ppalaga: Perfect, thanks @djcoleman!

Issue #5023, opened 2023-06-28
Micrometer: SimpleMeterRegistry is not created if no other registry is defined


JiriOndrusek:
### Bug description

Current extension does not create a SimpleMeterRegistry registry in case no other registries is defined. This feature is suggested to be used for testing purposes only. See Camel doc page - https://camel.apache.org/components/3.20.x/micrometer-component.html#MicrometerComponent-registry

PR #5024, opened 2023-06-28
Micrometer: SimpleMeterRegistry is not created if no other registry is defined #5023

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/5023

Quarkus gathers all registerie and puts them together as CompositeRegistry (created by [code](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/Metrics.java#L36))
This happens even if there is no metric implementation dependency (which causes thet CompositeRepository is empty and can not be used). For such purposes the Camel component creates a SimpleMeterRegistry for the testing purposes  ([code](https://github.com/apache/camel/blob/main/components/camel-micrometer/src/main/java/org/apache/camel/component/micrometer/MicrometerUtils.java#L67)).
This PR adds code, which detects cases where CompositeRegistry is empty and in such cases  creates `SimpleMeterRegistry`.

### DIFF
--- a/extensions/micrometer/deployment/src/main/java/org/apache/camel/quarkus/component/micrometer/deployment/MicrometerProcessor.java
+++ b/extensions/micrometer/deployment/src/main/java/org/apache/camel/quarkus/component/micrometer/deployment/MicrometerProcessor.java
@@ -16,12 +16,16 @@
  */
 package org.apache.camel.quarkus.component.micrometer.deployment;
 
+import java.util.List;
+import java.util.Optional;
+
 import io.micrometer.core.instrument.MeterRegistry;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.ExecutionTime;
 import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.micrometer.deployment.MicrometerProcessor.MicrometerEnabled;
+import io.quarkus.micrometer.deployment.MicrometerRegistryProviderBuildItem;
 import io.quarkus.micrometer.deployment.RootMeterRegistryBuildItem;
 import org.apache.camel.component.micrometer.MicrometerConstants;
 import org.apache.camel.quarkus.component.micrometer.CamelMicrometerConfig;
@@ -64,5 +68,17 @@ class MicrometerProcessor {
             CamelMicrometerConfig config) {
         return new RuntimeCamelContextCustomizerBuildItem(
                 recorder.createRuntimeContextCustomizer(config, rootMeterRegistryBuildItem.getValue()));
+    }
+
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep
+    void configureDefaultRegistry(
+            List<MicrometerRegistryProviderBuildItem> providers,
+            Optional<RootMeterRegistryBuildItem> registry,
+            CamelMicrometerRecorder recorder) {
+        // Register SimpleMeterRegistry to the CompositeMeterRegistry (created by Quarkus) if there is no MicrometerRegistryProviderBuildItem
+        if (registry.isPresent() && providers.isEmpty()) {
+            recorder.configureDefaultRegistry(registry.get().getValue());
+        }
     }
 }
--- a/extensions/micrometer/deployment/src/test/java/org/apache/camel/quarkus/component/micrometer/deployment/MicrometerSimpleMeterRegistryTest.java
+++ b/extensions/micrometer/deployment/src/test/java/org/apache/camel/quarkus/component/micrometer/deployment/MicrometerSimpleMeterRegistryTest.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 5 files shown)


extensions/micrometer/deployment/src/main/java/org/apache/camel/quarkus/component/micrometer/deployment/MicrometerProcessor.java
```diff
@@ -65,4 +67,13 @@ RuntimeCamelContextCustomizerBuildItem runtimeContextCustomizer(
         return new RuntimeCamelContextCustomizerBuildItem(
                 recorder.createRuntimeContextCustomizer(config, rootMeterRegistryBuildItem.getValue()));
     }
+
+    @Record(ExecutionTime.STATIC_INIT)
+    @BuildStep
+    CamelRuntimeBeanBuildItem tikaComponent(CamelMicrometerRecorder recorder) {
```
jamesnetherton, line 73:
```suggestion
    CamelRuntimeBeanBuildItem micrometerComponent(CamelMicrometerRecorder recorder) {
```

extensions/micrometer/deployment/src/test/java/org/apache/camel/quarkus/component/micrometer/deployment/MicrometerSimpleMeterRegistryTest.java
```diff
@@ -0,0 +1,44 @@
...
+public class MicrometerSimpleMeterRegistryTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest CONFIG = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
+    @Inject
+    CamelContext context;
+
+    @Test
+    public void testQuarkusSchedulerAutowired() throws Exception {
```
jamesnetherton, line 40:
```suggestion
    public void testDefaultSimpleMeterRegistry() throws Exception {
```

extensions/micrometer/deployment/src/test/java/org/apache/camel/quarkus/component/micrometer/deployment/MicrometerSimpleMeterRegistryTest.java
```diff
@@ -0,0 +1,44 @@
...
+        MicrometerComponent component = context.getComponent("micrometer", MicrometerComponent.class);
+        assertTrue(component.getMetricsRegistry() instanceof SimpleMeterRegistry, "SimpleMeterRegistry");
```
jamesnetherton, line 42:
```suggestion
        assertInstanceOf(SimpleMeterRegistry.class, component.getMetricsRegistry() );
```

jamesnetherton:
I wonder if we should avoid extending the component? 

And instead consume `List<MicrometerRegistryProviderBuildItem>`. And if it's empty, then we can just register `SimpleMeterRegistry` with the existing Quarkus composite registry (which has its own `BuildItem` where the value can be passed to a recorder).

I'm just thinking out loud, not sure if that would work or not.

JiriOndrusek:
> I wonder if we should avoid extending the component?
> 
> And instead consume `List<MicrometerRegistryProviderBuildItem>`. And if it's empty, then we can just register `SimpleMeterRegistry` with the existing Quarkus composite registry (which has its own `BuildItem` where the value can be passed to a recorder).
> 
> I'm just thinking out loud, not sure if that would work or not.

This is a nice idea. Registering simpleRegistry and keeping the Quarks mechanism working as expected should be nice and it would allow to share SimpleRegistry among more services/apps/beans. I'll try that solution.

and thanks for the founded issues in the code , I'll rework the PR with this new approach to see if it works as expected.

JiriOndrusek:
@jamesnetherton 
my first attempt was not successful. Unfortunately it is not possible to consume  `List<MicrometerRegistryProviderBuildItem>` and create a new `MicrometerRegistryProviderBuildItem` at the same time. I'm getting `io.quarkus.builder.ChainBuildException: Cycle detected`.

jamesnetherton:
I don't think we need to create `MicrometerRegistryProviderBuildItem`, we only need to consume them.

So I was thinking something like:

```java
@BuildStep
void configureDefaultRegistry(
        List<MicrometerRegistryProviderBuildItem> providers, 
        RootMeterRegistryBuildItem registry,
        CamelMicrometerRecorder recorder) {
    if (providers.isEmpty()) {
        // Adds SimpleMeterRegistry to the Quarkus composite one
        recorder.configureDefaultRegistry(registry.getValue());
    }
}
```

extensions/micrometer/deployment/src/main/java/org/apache/camel/quarkus/component/micrometer/deployment/MicrometerProcessor.java
JiriOndrusek: N/A

extensions/micrometer/deployment/src/test/java/org/apache/camel/quarkus/component/micrometer/deployment/MicrometerSimpleMeterRegistryTest.java
JiriOndrusek: N/A

extensions/micrometer/deployment/src/test/java/org/apache/camel/quarkus/component/micrometer/deployment/MicrometerSimpleMeterRegistryTest.java
JiriOndrusek: N/A

JiriOndrusek:
@jamesnetherton your suggestion work as expected. The PR now contains a different implementation of the fix. I added 1 line into the doc, describing autocreation of `SimpleMeterRegistry`.

Issue #5010, opened 2023-06-23
[Mandrel 23.0.0] google-storage integration test native compilation fails


jamesnetherton:
### Bug description

What's odd is that this test was successful a few times on CI and then started failing (see [here](https://github.com/apache/camel-quarkus/actions/runs/5350809287/jobs/9704366917))

```
Failed generating 'camel-quarkus-integration-test-google-storage-3.0.0-SNAPSHOT-runner' after 42.4s.

The build process encountered an unexpected error:

java.lang.NullPointerException: Cannot invoke "java.util.Set.iterator()" because "triggers" is null
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.ReachabilityHandlerFeature.processReachable(ReachabilityHandlerFeature.java:173)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.ReachabilityHandlerFeature.duringAnalysis(ReachabilityHandlerFeature.java:128)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.lambda$runPointsToAnalysis$10(NativeImageGenerator.java:770)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.FeatureHandler.forEachFeature(FeatureHandler.java:86)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.lambda$runPointsToAnalysis$11(NativeImageGenerator.java:770)
        at org.graalvm.nativeimage.pointsto/com.oracle.graal.pointsto.AbstractAnalysisEngine.runAnalysis(AbstractAnalysisEngine.java:179)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.runPointsToAnalysis(NativeImageGenerator.java:767)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:582)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:539)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:408)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:612)
[...]

Issue #5007, opened 2023-06-22
[Mandrel 23.0.0] google-pubsub integration test native compilation fails


jamesnetherton:
### Bug description

Oddly, the test passed ok on the first run of the PR to upgrade Quarkus to 3.2.0.CR1. Now it fails with:

```
Failed generating 'camel-quarkus-integration-test-google-pubsub-3.0.0-SNAPSHOT-runner' after 45.4s.

The build process encountered an unexpected error:

java.lang.NullPointerException: Cannot invoke "java.util.Set.iterator()" because "triggers" is null
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.ReachabilityHandlerFeature.isTriggered(ReachabilityHandlerFeature.java:142)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.ReachabilityHandlerFeature.duringAnalysis(ReachabilityHandlerFeature.java:121)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.lambda$runPointsToAnalysis$10(NativeImageGenerator.java:770)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.FeatureHandler.forEachFeature(FeatureHandler.java:86)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.lambda$runPointsToAnalysis$11(NativeImageGenerator.java:770)
        at org.graalvm.nativeimage.pointsto/com.oracle.graal.pointsto.AbstractAnalysisEngine.runAnalysis(AbstractAnalysisEngine.java:179)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.runPointsToAnalysis(NativeImageGenerator.java:767)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:582)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:539)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:408)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:612)
[...]

PR #5033, opened 2023-06-30
Re-enable google-pubsub and google-storage and add a workaround

zhfeng:
It should fix 
- #5007 
- #5010


jamesnetherton:
Looks fine.

But, I think we should drop a note into the extension docs about this so folks are aware that the workaround is required. I don't think there's any way for us to apply it for them automatically.

zhfeng:
Oh, yeah, so put it in e.g. `./extensions/google-pubsub/runtime/src/main/doc/limitations.adoc`?

jamesnetherton:
> put it in e.g. `./extensions/google-pubsub/runtime/src/main/doc/limitations.adoc`?

Yeah, that's fine to add for the `pubsub` & `storage` extensions :+1:

zhfeng:
OK, I add them all. `google-bigquery` has another issue with `apache-arrow` due to the netty upgrading. I hope it can be fixed with some new releases in the future.

jamesnetherton (approved):

oscerd (approved):

ppalaga:
Would please somebody explain, why the `RunReachabilityHandlersConcurrently` option cannot be passed via `quarkus.native.additional-build-args`?

PR #5035, opened 2023-06-30
Ref #4894: Make Groovy DSL ITs platform compliant

### LINKED ISSUE #4894: Groovy DSL test hard to make work on Quarkus Platform
The test can currently be characterized as follows:

1. The execution is moved to integration test phase 
2. It relies on the existence of the runner artifact under target directory
3. The path to the runner is passed from pom.xml to the test via property

All of these are quite against how the tests in Quarkus Platform are set up:

1. Native tests are not executed and thus their CI is only reaching Maven test phase. This was discussed several times and the general agreement was that this should change, but it has not changed yet. Hence point 1. may get solved at some point by the Platform team.
2. Running `mvn package` in the generated test module directory `generated-platform-project/quarkus-camel/integration-tests/camel-quarkus-integration-test-groovy-dsl` does not produce any jars under target. Maybe quarkus-maven-plugin is not set up. It needs to be investigated.
3. Setting the `quarkus.runner` property (and moving surefire to integration-test) can be done by this piece of XSL added to `src/main/resources/xslt/camel/test-pom.xsl` in the Platform source tree: 

```
    <!-- camel-quarkus-integration-test-groovy-dsl requires to move surefire:test to integration-test phase -->
    <!-- to be able to run java -jar target/*runner.jar from a test -->
    <xsl:template match="/pom:project/pom:build/pom:plugins/pom:plugin[(pom:artifactId/text() = 'maven-surefire-plugin') and (/pom:project/pom:artifactId/text() = 'camel-quarkus-integration-test-groovy-dsl')]">
        <xsl:copy>
            <xsl:apply-templates select="@* | node()" />
            <executions>
                <execution>
                    <id>default-test</id>
                    <goals>
                        <goal>test</goal>
                    </goals>
                    <phase>integration-test</phase>
                    <configuration>
                        <systemProperties>
                            <quarkus.runner>&#36;{project.build.directory}/quarkus-app/quarkus-run.jar</quarkus.runner>
                        </systemProperties>
                    </configuration>
                </execution>
            </executions>
        </xsl:copy>
    </xsl:template>
```

Instead on pushing on any of the above, I'll remove camel-quarkus-integration-test-groovy-dsl from the test list we publish for the platform for now.

@essobedo I wonder whether the classloading issue because of which we have this special setup cannot be fixed in Quarkus or elsewhere? If it can be fixed, we would not need to bother with any special measures to make the test work in the Platform.

essobedo:
fixes #4894 

## Motivation

The Groovy DSL tests are hard to make work on the Quarkus Platform.

## Modifications

* Revert changes that were required to workaround the ClassLoading issue

### DIFF
--- a/integration-tests/groovy-dsl/pom.xml
+++ b/integration-tests/groovy-dsl/pom.xml
@@ -30,9 +30,6 @@
     <name>Camel Quarkus :: Integration Tests :: Groovy DSL</name>
     <description>Integration tests for Camel Groovy DSL extension</description>
 
-    <properties>
-        <quarkus.runner>${project.build.directory}/quarkus-app/quarkus-run.jar</quarkus.runner>
-    </properties>
     <dependencies>
         <dependency>
             <groupId>org.apache.camel.quarkus</groupId>
@@ -70,18 +67,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.camel.quarkus</groupId>
-            <artifactId>camel-quarkus-integration-tests-process-executor-support</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.awaitility</groupId>
-            <artifactId>awaitility</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
@@ -185,7 +176,6 @@
             </activation>
             <properties>
                 <quarkus.package.type>native</quarkus.package.type>
-                <quarkus.runner>${project.build.directory}/${project.artifactId}-${project.version}-runner</quarkus.runner>
             </properties>
             <build>
                 <plugins>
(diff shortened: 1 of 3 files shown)


essobedo:
@ppalaga I've just checked and refreshed my mind a bit about it, the problem here is very specific and complex, it only occurs when using the Quarkus JUnit5 extension with Groovy extensions. The Groovy extensions are loaded with the System ClassLoader instead of the `Quarkus Runtime ClassLoader` because Groovy [is declared as a parent first artifact in Quarkus core](https://github.com/quarkusio/quarkus/blob/main/core/runtime/pom.xml#L195-L197) that is why the Camel extensions are not seen. Groovy has been declared as a parent's first artifact to fix a ClassValue leak in Quarkus described in [this ticket](https://github.com/quarkusio/quarkus/issues/12498). In theory, the way to fix the leak is more or less what is done in the [`GroovyCacheCleaner`](https://github.com/quarkusio/quarkus/blob/main/test-framework/common/src/main/java/io/quarkus/test/common/GroovyCacheCleaner.java) in case Groovy is in the System CL but if Groovy is in a custom CL like `Quarkus Runtime ClassLoader` it gets even more complex to clean up the `ClassValue` so I'm testing another approach that simply disables the `ClassValue` in Groovy if the tests pass I will propose the ticket and the corresponding PR hoping that they will accept them otherwise I'm afraid that the only way is to keep the code as it is in camel-quarkus and apply your proposal in Quarkus Plarform. I don't think it is worth too much effort since once again it is a very specific use case and having Groovy classes loaded in the `Quarkus Runtime ClassLoader` may have an impact on the duration of the tests since it will load potentially several times the Groovy classes per module instead of only once.

essobedo:
The related ticket in Quarkus https://github.com/quarkusio/quarkus/issues/34446

ppalaga:
Thanks for investigating @essobedo!

PR #5038, opened 2023-06-30
Replace org.graalvm.nativeimage:svm with org.graalvm.sdk:graal-sdk which has a status of public API

ppalaga:


ppalaga:
@zakkak replacing org.graalvm.nativeimage:svm with org.graalvm.sdk:graal-sdk is not that straightforward as I hoped for. We use com.oracle.svm.core.jdk.Resources.registerResource() to override the service provider file `META-INF/services/javax.xml.transform.TransformerFactory` from xalan.jar with our custom implementation. Do you have any idea how we could get rid of com.oracle.svm.core.jdk.Resources.registerResource()? Either replace it with some similar method from graal-sdk, or figure out a way how to force-pass our custom service file resource to GraalVM native-image? Do you perhaps know whether Quarkus/GraalVM considers the resources from the application before it looks into dependencies? If so, we could generate the service file in the app and use the NativeImageResourceBuildItem or ServiceProviderBuildItem.

ppalaga:
Thanks @zakkak! Let me try it.

PR #2008, opened 2020-11-13
Kudu: unshade and remove embedded netty and use quarkus-netty instead

ppalaga:
related #1907

Fix #1669

Issue #1907, opened 2020-10-13
Figure out how to port shaded netty to native


ppalaga:
Camel components known to pull an embedded Netty:

* Kudu https://github.com/apache/camel-quarkus/issues/1669
* Couchbase https://github.com/apache/camel-quarkus/issues/1188
* HBase https://github.com/apache/camel-quarkus/issues/1648

In case of Kudu, I simply copied the substitutions and deployment code from quarkus-netty, see https://github.com/apache/camel-quarkus/pull/1667 but this approach is hard to maintain. It would be even more hard to do it for n separate embedded Netties. 

As seen in case of Kudu asking the third party lib to stop shading netty may get refused for valid reasons.

Ideally both the substitutions and deployment BuildSteps should be made to accept the shaded package prefix somehow. I think it is rather easy in case of the BuildSteps (we'd prodoce some sort of ShadedPackagePrefixBuildItem and NettyProcessor would consume it). It is a bit harder in case of the Substitutions. They'd have to be generated either on source code or bytecode level.

Any other ideas?

ppalaga:
Hm... actually, if they shade a compatible netty version, we could revert the netty package relocations using Shade plugin and then simply strip the netty classes away. That could be both easier to maintain and the resulting apps could be slimmer.

ppalaga:
> we could revert the netty package relocations using Shade plugin and then simply strip the netty classes away, replacing it with Netty from Quarkus.

This seems to work, see #2008

PR #5041, opened 2023-07-03
Re-enable google-bigquery and add a work around

zhfeng:
Exclude `org.apache.arrow:arrow-memory-netty` because it is not compatible with the lastest netty `4.1.94.Final`.

### DIFF
--- a/extensions/google-bigquery/runtime/src/main/java/io/netty/buffer/LargeBuffer.java
+++ b/extensions/google-bigquery/runtime/src/main/java/io/netty/buffer/LargeBuffer.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.netty.buffer;
+
+/**
+ * A MutableWrappedByteBuf that also maintains a metric of the number of huge buffer bytes and
+ * counts.
+ */
+public class LargeBuffer extends MutableWrappedByteBuf {
+
+    public LargeBuffer(ByteBuf buffer) {
+        super(buffer);
+    }
+
+    @Override
+    public ByteBuf copy(int index, int length) {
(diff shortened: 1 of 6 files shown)


oscerd (approved):

jamesnetherton:
We should add the same doc notes for the workaround like we did for pubsub & storage.

Also, not sure how safe the Arrow exclusion is. It is referenced in a few places:

https://github.com/search?q=repo%3Agoogleapis%2Fjava-bigquery+arrow+language%3AJava&type=code&l=Java

I guess the fact that the native compilation + tests passed means that maybe we don't hit those code paths in Camel...

zhfeng:
Yeah, it is exactly what I was concerned. It seems that `google-bigquery` default uses `DataFormat.ARROW` when read/write a large data table result.

We don't hit those codes because  we lack of a test for the large table. So how can write a sentence for this limitation?

jamesnetherton:
Thinking a bit more, I'm not sure I like excluding Arrow because it potentially impacts negatively on JVM mode, and the issue is purely a native mode one.

jamesnetherton:
@zhfeng is it just the shaded Netty of Arrow that is causing the problem? If Arrow does not modify any of the Netty bits, maybe we could try something similar to what we did for Kudu and shaded Netty? See here: https://github.com/apache/camel-quarkus/pull/2008 & https://github.com/apache/camel-quarkus/issues/1907.

zhfeng:
@jamesnetherton I think `apache-arrow` does not shade Netty. The problem is Quarkus 3.2.0.Final update Netty to `4.1.94.Final` and there is an un-compatible change in `PooledByteBufAllocator`. So I think it could also impact the JVM mode if we test with a large table.

The fix in `apache-arrow` is https://github.com/apache/arrow/commit/ea4f03ac166e3961b59f29f60dcd160fbed29894#diff-d967dd684919ce097b81e439fafc96217f1263985f5864faa4b78eb96e7a4f16L164

So is it possible to have this file in our `camel-quarkus-google-bigquery`?

jamesnetherton:
Ah, yeah. Sorry for the confusion!

> So is it possible to have this file in our camel-quarkus-google-bigquery?

We could temporarily have it in our extension. Quarkus has `RemovedResourceBuildItem`, not sure if that could also be used to remove the bad impl from Arrow?

If it becomes too much work. Maybe we just stick with the original plan of removing Arrow until we get the fix.

zhfeng:
Sine `org.apache.arrow:arrow-memory-netty` is a runtime dependency, `RemovedResourceBuildItem` seems not working. I just try to copy 4 files from `apache-arrow`
- LargeBuffer.java
- MutableWrappedByteBuf.java
- PooledByteBufAllocatorL.java
- UnsafeDirectLittleEndian.java

and apply the fix for `PooledByteBufAllocatorL.java`. All of the `google-bigquery` tests look good both in JVM and native mode.

Also I think we need a test for a large table and enable the result with arrow data format. Can it be supported by wiremock?

ppalaga:
> storing huge files for the stubbed data

We could perhaps generate those at build time?

zhfeng:
I raise 
- https://github.com/apache/camel-quarkus/issues/5048

jamesnetherton (approved):

PR #5061, opened 2023-07-06
Build Camel on the nightly build instead of relying on the SNAPSHOT repo

jamesnetherton:
We can revert this if it takes too long or does not work out...


ppalaga:
Rebased to make the atlassian repo happy.

ppalaga:
335069e: Remove `BRANCH_OPTIONS: -Papache-snapshots` from camel-master-cron.yaml

PR #5065, opened 2023-07-06
Restrict downloading of com.atlassian dependencies to packages.atlassian.com

jamesnetherton:
Test to see if we can avoid slurping down the entire internet from packages.atlassian.com.


zbendhiba (approved):

oscerd (approved):

ppalaga:
Nice trick, I did not know this was even possible!

PR #5076, opened 2023-07-07
chore: Add git diff when there are uncommitted changes

essobedo:
## Motivation

There are builds failing due to uncommitted changes for which providing only the list of uncommitted files is not enough to understand the problem especially when we cannot reproduce.

## Modifications

* Add `git diff` to each command to have more info


jamesnetherton (approved):

essobedo:
Merging it as the build failure is not related to these changes

PR #5106, opened 2023-07-19
Fix hamcrest GAV in bom-test

claudio4j:


oscerd (approved):

aldettinger (approved):
AFAIR it indeed does not exist.

@jamesnetherton Do we need to remove it also from `.github/dependabot.yml` ?

jamesnetherton:
> @jamesnetherton Do we need to remove it also from .github/dependabot.yml ?

Yeah, we should update it rather than remove:

https://github.com/apache/camel-quarkus/blob/main/.github/dependabot.yml#L35C54-L35C54

claudio4j:
Updated in dependabot.yml

PR #5116, opened 2023-07-24
Ref#4772 Introduce group testing for different jdbc db types

### LINKED ISSUE #4772: Introduce group testing for different jdbc db types
As discussed in #4751, we could implement [group testing ](https://camel.apache.org/camel-quarkus/3.0.x/contributor-guide/extension-testing.html#_grouping) for db types in jdbc integration tests.

Supported DBs:
- Postgresql
- MySQL 
- MariaDB 
- H2 
- DB2
- MSSQL 
- Oracle

ldrozdo:
Fix #4772


oscerd (approved):

integration-test-groups/jdbc/db2/src/main/resources/reflection-config.json
zhfeng, line 1: I think we could avoid using `reflection-config.json` and how about adding `@RegisterForReflection` on these three classes?

integration-test-groups/jdbc/db2/src/main/resources/application.properties
```diff
@@ -0,0 +1,44 @@
...
+
+quarkus.datasource.h2.db-kind=h2
+quarkus.datasource.h2.jdbc.max-size=8
+
+quarkus.datasource.db2.db-kind=db2
+quarkus.datasource.db2.jdbc.max-size=8
+
+quarkus.datasource.mssql.db-kind=mssql
+quarkus.datasource.mssql.jdbc.max-size=8
+
```
zhfeng, line 41: Can we only keep `quarkus.datasource.db2` configurations sine it is for `db2` testing?

zhfeng:
Thanks @ldrozdo for working on this issue. I just add some comments.

integration-test-groups/jdbc/db2/src/main/resources/reflection-config.json
ldrozdo: @zhfeng thanks, I added the annotations (note: there is no `BaseCamelJdbcTest` class, I forgot to delete it, so the annotation should be only on two classes)

integration-test-groups/jdbc/db2/src/main/resources/application.properties
ldrozdo: @zhfeng I thought that the `application.properties` file should be the same for all modules, otherwise group tests don't work properly.

ldrozdo:
>  @jamesnetherton honestly, I'm not sure what is the right process. But our QE job for database certification already uses `integration-tests/jdbc`, so I would leave it like this at least for a while.

jamesnetherton:
> Is there something I can do to fix it?

There's no Docker on the Windows build. You can add a profile like this one so the test gets skipped on that platform:

https://github.com/apache/camel-quarkus/blob/main/integration-test-groups/azure/azure-storage-blob/pom.xml#L165-L175

integration-test-groups/jdbc/h2/pom.xml
```diff
@@ -0,0 +1,284 @@
...
+            <artifactId>quarkus-jdbc-mysql</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jdbc-oracle</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jdbc-postgresql</artifactId>
+        </dependency>
```
ppalaga, line 88: Are all these JDBC artifacts needed in this specific test module? We generally prefer keeping the [minimal set of dependencies](https://camel.apache.org/camel-quarkus/latest/contributor-guide/extension-testing.html#_minimal_set_of_dependencies). The same issue seems to exist in other submodules of `integration-test-groups/jdbc`

ppalaga:
> @jamesnetherton honestly, I'm not sure what is the right process. But our QE job for database certification already uses `integration-tests/jdbc`, so I would leave it like this at least for a while.

I guess, after merging this change, you want to run tests in all submodules of `integration-test-groups/jdbc`.

ldrozdo:
@jamesnetherton thank you, the profile helped. 
@ppalaga  thanks, I've fixed the dependencies. 
About the deleting  `integration-tests/jdbc`, I will create a task for it and I will make sure the deletion won't affect our jobs. 

I have one (hopefully last) problem. The native tests were already passing, but now they've started to fail with:
`Caused by: com.github.dockerjava.api.exception.DockerClientException: Could not pull image: failed to register layer: ApplyLayer exit status 1 stdout:  stderr: write /usr/lib/x86_64-linux-gnu/libLLVM-14.so.1: no space left on device`
Please, is there something I can do about it?

jamesnetherton:
> is there something I can do about it

I guess we're eating up a lot of disk space with all of the container images we pull....

I can create a separate PR to try and reclaim some space before the tests run.

jamesnetherton:
@ldrozdo if you rebase with the latest work from `main`, the disk space issue should hopefully be resolved.

jamesnetherton:
Follow up issue to remove the old itest module https://github.com/apache/camel-quarkus/issues/5164.

PR #5152, opened 2023-08-07
Manage test container versions in the root project pom.xml

jamesnetherton:
Also tested with the Quarkus Platform and this worked fine.


aldettinger (approved):
Having a single prop file generated for each test should be ok.
Fingers crossed that all those container version upgrade will pass in ci build :crossed_fingers: 

Also, if one day there is a valid reason to test against multiple release streams of a container, we might be able to use properties like:
```
kafka-7_x.container.image
kafka-8_x.container.image
```

tooling/scripts/generate-test-containers-config-properties.groovy
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/**
+ * Writes all Maven properties suffixed with 'container.image' to target/test-classes/META-INF/microprofile-config.properties
+ */
+def fileContent = project.properties.findAll { key, value ->
+    key.endsWith('container.image')
+}.collect { key, value ->
+    "${key}=${value}"
+}.join(System.lineSeparator())
```
ppalaga, line 25:
```suggestion
}.join('\n')
```

I'd vote for making this reproducible across all platforms.

ppalaga (commented):
Very nice! Thanks a lot @jamesnetherton!

tooling/scripts/generate-test-containers-config-properties.groovy
```diff
@@ -0,0 +1,34 @@
...
+
+File testClasses = new File("${project.build.testOutputDirectory}")
+if (testClasses.exists()) {
+    File metaInfDir = new File("${testClasses.absolutePath}/META-INF")
+    metaInfDir.mkdir()
+
+    File file = new File("${metaInfDir.absolutePath}/microprofile-config.properties")
+    file.write(fileContent)
```
ppalaga, line 33:
```suggestion
    file.write(fileContent)
```
Or the like. Also for the sake of reproducibility.

jamesnetherton: Suggestion code seems to be unaltered to the original.

Was there a specific change you'd like to be made here?

ppalaga: Ups, sorry, I meant something like `file.write(fileContent, 'UTF-8')` IIRC, Groovy has this kind of override.

jamesnetherton: https://github.com/apache/camel-quarkus/pull/5157

PR #5166, opened 2023-08-10
Fix compareVersion in sanity-checks.groovy #5165

### LINKED ISSUE #5165: InvocationTargetException when running sanity-checks on 3.0.0-RC2
### Bug description

The steps to reproduce

1. Checkout on `3.0.0-RC2`
2. Run `./mvnw -f extension/ldap/pom.xml`
3. It will be failling with
```
[ERROR] Failed to execute goal org.codehaus.gmavenplus:gmavenplus-plugin:3.0.0:execute (sanity-checks) on project camel-quarkus-ldap: Error occurred while calling a method on a Groovy class from classpath.: InvocationTargetException: For input string: "0-RC2" -> [Help 1]
```

It needs to strip anything after "-" in `compareVersion` method in `tooling/scripts/sanity-checks.groovy`

zhfeng:
Fix #5165

### DIFF
--- a/tooling/scripts/sanity-checks.groovy
+++ b/tooling/scripts/sanity-checks.groovy
@@ -51,7 +51,8 @@ if (!badDeps.isEmpty()) {
 }
 
 // check jvmSince and nativeCheck not newer than current SNAPSHOT version
-final String currentVersion = project.version - '-SNAPSHOT'
+def versionRegex = ~/\-.*/
+final String currentVersion = project.version - versionRegex
 final String jvmSince    = project.properties['camel.quarkus.jvmSince']
 final String nativeSince = project.properties['camel.quarkus.nativeSince']
 


aldettinger (approved):
Oh, String - regex in groovy. Looks good.

tooling/scripts/sanity-checks.groovy
```diff
@@ -68,10 +68,11 @@ int compareVersion(String a, String b) {
     List verB = b.tokenize('.')
 
     def commonIndices = Math.min(verA.size(), verB.size())
+    def versionRegex = ~/\-.*/
 
     for (int i = 0; i < commonIndices; ++i) {
        def numA = verA[i].toInteger()
-       def numB = verB[i].toInteger()
+       def numB = (verB[i] - versionRegex).toInteger()
```
ppalaga, line 75: I am bit lost what is this doing. The minus operator removes the matching part from the string, right?
I this way we are effectively disregard the `-RC` qualifiers, right? 
That's definitely OK for this current versions check, but I wonder whether it would not be safer for future copy-and-paste visitors, to move the suffix stripping before the compareVersion() method. After all we strip `-SNAPSHOT` from project.version before passing it to compareVersion() too.

ppalaga, line 75: But wait, which version actually contains the -RC qualifier? Is it the current version or some of the since versions? Both of those sounds a bit strange.

zhfeng: Only `currentVersion` contains `-RC`

zhfeng: Yeah, it intends to strip `-RC`. So Maybe it could move before `compareVersion()` like
```java
final String currentVersion = project.version - '~/\-.*/'
```

ppalaga: Yeah, I think that would be safer for any future reuse of `compareVersion()`

zhfeng: OK, I fix it.

ppalaga (approved):

ppalaga:
Thanks for taking care, @zhfeng

PR #5168, opened 2023-08-10
Extend test coverage of gRPC extension

jamesnetherton:


jamesnetherton:
I'll look into the flakiness on Windows.

docs/modules/ROOT/pages/reference/extensions/grpc.adoc
```diff
@@ -72,3 +118,34 @@ To avoid compilation issues, the `grpc-java` project recommendation is to add a
...
+The gRPC component has various options where resources are resolved from the classpath:
+
+* `keyCertChainResource`
+* `keyResource`
+* `serviceAccountResource`
+* `trustCertCollectionResource`
+
+When using these options in native mode, you must ensure that any such resources are included in the native image.
+
+This can be accomplished by adding the configuration property `quarkus.native.resources.includes` to `application.properties`.
```
aldettinger, line 133: Could we remove lines 133-140 to ease long term maintenance ? Everything is already described in embedding-`resource-in-native-executable[native mode guide]` at line 141.

jamesnetherton: I figured it was better to have some instructions inline, because it's nicer to be able to directly copy a code snippet, rather than click away to another page and loose the context of the thing you're interested in (gRPC).

If you think not having it is better, then I can change it. Although, there are other extensions where the docs are like this, so please open an issue so we clean up the rest of the codebase.

aldettinger: There are always tradeof in such questions. The justification looks fair, let's keep it.

aldettinger (approved):
LGTM, well done @jamesnetherton :+1:

PR #5170, opened 2023-08-10
[5162] Add elasticsearch tests

llowinge:
Covers:
* Index data (CQ)
* Search (CQ)
* Multisearch (CQ)
* GetById (CQ)
* Update (CQ)
* Delete (CQ)
* DeleteIndex (CQ)
* Bulk (CQ)
* Exists (CQ)
* Ping (CQ)

### DIFF
--- a/integration-tests-jvm/elasticsearch/src/test/java/org/apache/camel/quarkus/component/elasticsearch/it/ElasticsearchTest.java
+++ b/integration-tests-jvm/elasticsearch/src/test/java/org/apache/camel/quarkus/component/elasticsearch/it/ElasticsearchTest.java
@@ -1,0 +1,365 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.elasticsearch.it;
+
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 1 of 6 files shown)


llowinge:
@jamesnetherton Can you please review ?

integration-tests-jvm/elasticsearch/src/test/java/org/apache/camel/quarkus/component/elasticsearch/it/ElasticsearchTest.java
```diff
@@ -0,0 +1,359 @@
...
+                    .extract()
+                    .body()
+                    .asString();
+            return hits.equals("2");
+        });
+    }
+
+    @SuppressWarnings("unused")
+    private static String[] componentNames() {
+        return new String[] { "elasticsearch" };
```
jamesnetherton, line 357: In theory this is not required and we could just hard code the component name (using `to` instead of `toD`).

Or is it worth keeping in case we do eventually want to test the component with the Quarkus managed ES client?

jamesnetherton (commented):
I left a comment / question / potential improvement. Otherwise everything else LGTM.

integration-tests-jvm/elasticsearch/src/test/java/org/apache/camel/quarkus/component/elasticsearch/it/ElasticsearchTest.java
llowinge: I left it there for the Quarkus managed ES client. But i agree it is not useful now at all. Maybe i should add comment there why it is done like that to that method ?

llowinge: I left comment, i hope it is OK.

aldettinger:
LGTM. Do we need tests to cover things like `certificatePath` or `enableSSL` ? Not sure as native is not tested.

jamesnetherton:
> Do we need tests to cover things like `certificatePath` or `enableSSL`

We could maybe follow up with that. These initial tests are mostly a reproduction of what we had for the previous `elasticsearch-rest` extension.

aldettinger:
Yes, that's ok in this state. Plus if we promote it to native one day, then we could update the test coverage at this stage YAGNI style.

aldettinger (approved):

llowinge:
@aldettinger That is great idea, but given it was only for JVM, i think we are safe there. Basically all the tests are probably overkill, but i've ported all of them for future benefits. I've created followup issue for the SSL https://github.com/apache/camel-quarkus/issues/5174.

aldettinger:
Not sure, it's overkill. Those features are present in the camel component doc, so this is good to have them covered :+1:

jamesnetherton (approved):

Issue #5201, opened 2023-08-21
ZenDesk native integration test failure


jamesnetherton:
### Bug description

Quarkus is now managing `async-http-client` (for Pulsar) which effectively downgrades it to a non JakartaEE 10 friendly impl.

https://github.com/quarkusio/quarkus/issues/35128

llowinge:
I'm just adding note here about https://github.com/apache/camel-quarkus/blob/main/extensions-support/ahc/deployment/src/main/java/org/apache/camel/quarkus/component/support/ahc/deployment/JakartaEnablement.java#L40 - CQ still uses async-client which is not Jakarta 10 compatible, so we cannot remove the workaround.

PR #5202, opened 2023-08-21
Downgrade async-http-client to 2.12.3. Relates #5201

jamesnetherton:
In the absence of a Quarkus fix for #5201, it's maybe better to downgrade the `async-http-client` and just bytecode transform the single problem class.


jamesnetherton:
> @jamesnetherton should it be for `quarkus-main` branch?

I didn't see the point in releasing 3.2.0 aligned with one version of `async-http-client` and then downgrading it in the next. So thought I'd just make the change now on `main`.

PR #5204, opened 2023-08-21
Fix quarkus-maven-plugin groupId & version for gRPC itest module

jamesnetherton:
Make it consistent with how we do things elsewhere in the project (Avro, CXF, Protobuf etc).

### DIFF
--- a/integration-tests/grpc/pom.xml
+++ b/integration-tests/grpc/pom.xml
@@ -129,8 +129,9 @@
                 </executions>
             </plugin>
             <plugin>
-                <groupId>io.quarkus</groupId>
+                <groupId>${quarkus.platform.group-id}</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


aldettinger (approved):
LGTM

There is the same pattern in perf-regression, is it an issue ?

jamesnetherton:
> There is the same pattern in perf-regression, is it an issue ?

Probably not a big issue for the perf module. It's more important to have the plugin overridable for the core itests. Guess it can't hurt to change it if we think there's value in doing it.

aldettinger:
> > There is the same pattern in perf-regression, is it an issue ?
> 
> Probably not a big issue for the perf module. It's more important to have the plugin overridable for the core itests. Guess it can't hurt to change it if we think there's value in doing it.

Ok, opened https://github.com/apache/camel-quarkus/issues/5205 not to forget.

zhfeng (approved):

PR #5051, opened 2023-07-04
Micrometer: Message History factory and JMX MicrometerMessageHistory is not covered

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/5050

Issue #5209, opened 2023-08-22
Test Micrometer JMX in native mode


jamesnetherton:
Native support was added recently:

https://github.com/quarkiverse/quarkus-micrometer-registry/issues/306

It'd be nice to have some tests when a new release of `quarkus-micrometer-registry` is available.

PR #5210, opened 2023-08-22
Micrometer: Message History factory and JMX MicrometerMessageHistory is not covered

jamesnetherton:
Clone of https://github.com/apache/camel-quarkus/pull/5051, rebased from latest `main` + some minor tweaks.

### DIFF
--- a/integration-tests/micrometer/src/test/java/org/apache/camel/quarkus/component/micrometer/it/MicrometerTest.java
+++ b/integration-tests/micrometer/src/test/java/org/apache/camel/quarkus/component/micrometer/it/MicrometerTest.java
@@ -78,7 +78,7 @@ class MicrometerTest extends AbstractMicrometerTest {
                 .then()
                 .statusCode(200);
         //component with Quarkus's registry should still see only 1 from the first call
-        assertEquals(1, getMetricValue(Integer.class, "counter", "camel-quarkus-custom-counter", null, 200, null));
+        assertEquals(1, getMetricValue(Integer.class, "counter", "camel-quarkus-custom-counter"));
         //component with custom registry should be on 11 see only 1 from the first call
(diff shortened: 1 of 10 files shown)


oscerd (approved):

integration-tests/micrometer/src/test/java/org/apache/camel/quarkus/component/micrometer/it/MicrometerTest.java
```diff
@@ -207,4 +207,46 @@ public void testJMXQuarkusDomain(String domain) throws Exception {
...
+
+        ObjectName on = ObjectName.getInstance(name);
+
+        //return json result
+        String json = (String) mBeanServer.invoke(on, "dumpStatisticsAsJson", null, null);
+
+        assertTrue(json.contains("jmxHistory"));
+
+    }
+
```
aldettinger, line 238: Is it correct reading that `testJMXHistory` is checking that the service could be accessed through JMX ?
And that `testHistory` is ensuring that the service could be accessed through the camel context ?

If yes, what about renaming both tests ala:
 + micrometerHistoryShouldBeAvailableThroughJMX
 + micrometerHistoryShouldBeAvailableThroughCamelContext

If make sense.

jamesnetherton: Yeah, if I understand the intent correctly, there's a check to ensure the relevant MBeans are published and metrics are being recorded. And another test to get it directly from the `CamelContext`. I can rename the test methods.

Maybe after #5209 these could be combined into a single test. Checking that the bits are available in JMX is probably enough for the native / non-native case.

jamesnetherton: Test renamed as requested.

aldettinger (approved):
LGTM.

PR #5215, opened 2023-08-23
Populate FastComponentNameResolver component names from included component service paths

jamesnetherton:


aldettinger (approved):
LGTM. This add service exclude pattern which is good :+1: 

And actually, maybe it also help with custom component. Like custom component would not be present in the catalog and would be missed before this pr. Does is sound right ?

jamesnetherton:
> And actually, maybe it also help with custom component. Like custom component would not be present in the catalog and would be missed before this pr. Does is sound right ?

Yes, exactly. I did not consider custom components when I wrote the initial implementation of `FastComponentNameResolver`.

zhfeng (approved):

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelProcessor.java
```diff
@@ -399,15 +395,19 @@ public List<CamelRoutesBuilderClassBuildItem> discoverRoutesBuilderClassNames(
     @Record(ExecutionTime.STATIC_INIT)
     @BuildStep
     CamelComponentNameResolverBuildItem componentNameResolver(
-            BuildTimeCamelCatalogBuildItem camelCatalog,
```
zhfeng, line 402: It seems that `BuildTimeCamelCatalog` no other usage if we remove here. Just curious, what's the original idea about it?

jamesnetherton: It is used elsewhere. To figure out which types can be made unremovable, just in case folks want to use CDI to wire up components etc.

zhfeng: OK, I get it. Thanks!

jamesnetherton:
The Hazelcast failure is a known flake.

PR #5225, opened 2023-08-25
[CEQ-5762 test] Implement test for custom exception mappers

AnetaCadova:
https://issues.redhat.com/browse/CEQ-5762


oscerd (approved):

aldettinger:
Great to see contributions @AnetaCadova :+1: 

+1 to not using downstream info, ticket number...

Thinking more, the test looks good. However, is it a test for Camel Quarkus ? Like are we testing only a Quarkus feature there ? Should we expect that it's already covered in Quarkus tests ?

AnetaCadova:
@aldettinger sorry for using the downstream issue number, I removed it. And yes, the issue is covered in Quarkus tests: https://github.com/quarkusio/quarkus/pull/32619. The Quarkus test also focuses on the problem with exception mappers occurring in DEV mode. My tests are focused only on the precedence of given types of exceptions, which seems to me as the main exception mappers issue. So I covered it here too.

aldettinger:
Many thanks for moving the subject forward @AnetaCadova 

> My tests are focused only on the precedence of given types of exceptions, which seems to me as the main exception mappers issue. So I covered it here too.

So, this specific case is not covered by Quarkus test yet ? Are we testing something specific to camel ? Or could other Quarkus Platform participants be affected ?

Another subject, we generally open prs against main first, tag the pr with backport-2.13x and then do the actual backporting. Could you please change the target branch ?

vkasala:
@aldettinger @AnetaCadova The idea was to test the custome-execption together with camel as that was the primary use case where the original issue in Quarkus appeared.

AnetaCadova:
@vkasala thanks for the clarification, I will add camel context to the test. @aldettinger sorry for the confusion

ppalaga:
>  @ppalaga Is camel involved at all in the context of this pr where custom Quarkus exception mappers are at play ? And so, is it a good test for Quarkus or Camel Quarkus ?

I do not know, sorry.

aldettinger:
yeah, non intrusive would be difficult.

@AnetaCadova Maybe it could be experimented to create `integration-tests/platform-http-custom-exception-mapping` from `integration-tests/platform-http` and see where it leads.

AnetaCadova:
After the discussion in Camel Quarkus chat, I'm closing this PR. I moved the test to the camel-q test suit

Issue #5028, opened 2023-06-29
Micrometer: Router policy should be disabled by default (to be consistent with Camel)


JiriOndrusek:
### Bug description

As you can see in Camel [documentation](https://camel.apache.org/components/3.20.x/micrometer-component.html#_using_micrometer_route_policy_factory), RoutePolicyFactory is not enabled by default. CQ should be consistent with this behavior and change default value for property `camel.metrics.enableRoutePolicy` to `false` ([code](https://github.com/apache/camel-quarkus/blob/main/extensions/micrometer/runtime/src/main/java/org/apache/camel/quarkus/component/micrometer/CamelMicrometerConfig.java#L30))

jamesnetherton:
The docs state that the default is `true`:

https://camel.apache.org/components/3.20.x/micrometer-component.html#_sb_option_camel_metrics_enable-route-policy

JiriOndrusek:
Default is true for CSB, unfirtunately for **plain Camel default is `false**`. Because users have to set routePolicy like in the [test](https://github.com/apache/camel/blob/main/components/camel-micrometer/src/test/java/org/apache/camel/component/micrometer/routepolicy/AbstractMicrometerRoutePolicyTest.java#L59). (Explanation in the [doc](https://camel.apache.org/components/3.20.x/micrometer-component.html#_using_micrometer_route_policy_factory) is `From Java you just add the factory to the CamelContext as shown below: context.addRoutePolicyFactory(new MicrometerRoutePolicyFactory());`

Therefore CQ  should default to false and ticket should be reported for CSB.

zhfeng:
@JiriOndrusek it can be closed? or is there anything about document we need to do?

PR #5247, opened 2023-09-01
 Micrometer: Router policy should be disabled by default  #5028

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/5028

### DIFF
--- a/docs/modules/ROOT/pages/migration-guide/3.0.0.adoc
+++ b/docs/modules/ROOT/pages/migration-guide/3.0.0.adoc
@@ -36,3 +36,19 @@ The following extensions have been removed.
 | camel-quarkus-vm                          | camel-quarkus-seda
 | camel-quarkus-xstream                     | camel-quarkus-jacksonxml
 |===
+
+
+== Change to default values of `camel.metrics.enableRoutePolicy` configuration
+
+The Micrometer extension's route policy was enabled by default in previous releases.
+
+Current behaviour has changed to reflect the default behavior of Camel.
+`MicrometerRoutePolicyFactory` is now disabled by default.
+
+You can enable `MicrometerRoutePolicyFactory` using configuration property.
+
+```
+quarkus.camel.metrics.enable-route-policy = true
+```
+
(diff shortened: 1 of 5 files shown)


oscerd (approved):

jamesnetherton:
We need to document this change of behavior in the migration guide.

FWIW I'm not convinced this is a good from the perspective of UX or migration from the old MicroProfile Metrics extension where the same config property defaults to `true`.

JiriOndrusek:
Change in migration doc is part of the PR. 
I proposed the PR mainly to see reaction, whether we want this change.
If the change is approved, the same behavior should be proposed to CSB.

JiriOndrusek:
> FWIW I'm not convinced this is a good from the perspective of UX or migration from the old MicroProfile Metrics extension where the same config property defaults to `true`.

default **true** is only in CSB. See https://camel.apache.org/components/3.20.x/micrometer-component.html#_sb_option_camel_metrics_enable-route-policy
plan Camel default value is **false** (`context.addRoutePolicyFactory` has to be called). See https://camel.apache.org/components/3.20.x/micrometer-component.html#_using_micrometer_route_policy_factory

jamesnetherton:
@JiriOndrusek sorry, I introduced some conflicts because I had to rename the migration guide.

zbendhiba:
Thanks @JiriOndrusek 
I'm curious why this should be enabled or disabled by default for Camel-Quarkus ?

JiriOndrusek:
> default **true** is only in CSB. See https://camel.apache.org/components/3.20.x/micrometer-component.html#_sb_option_camel_metrics_enable-route-policy
> plain Camel default value is **false** (`context.addRoutePolicyFactory` has to be called). See https://camel.apache.org/components/3.20.x/micrometer-component.html#_using_micrometer_route_policy_factory

Just to be consistent with plain camel (where user has to set policy in the code to enable it)   ->  See https://camel.apache.org/components/3.20.x/micrometer-component.html#_using_micrometer_route_policy_factory

aldettinger:
No strong opinion. Not sure we need to align with Camel. Aligning Camel Quarkus and Camel Spring Boot looks good.

jamesnetherton:
Out of date.

PR #5270, opened 2023-09-06
Upgrade Quarkus to 3.4.0.CR1

jamesnetherton:
I am expecting some test failures. Lets see how it goes...

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -30,7 +30,7 @@ asciidoc:
     # Project versions
     camel-version: 4.0.0 # replace ${camel.version}
     camel-docs-version: 4.0.x # replace ${camel.docs.components.version}
-    quarkus-version: 3.3.1 # replace ${quarkus.version}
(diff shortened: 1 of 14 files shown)


oscerd (approved):

jamesnetherton:
For reasons that I don't quite understand, I have to add `quarkus-virtual-threads` explicitly to the `knative` extension. Some more chat about it here:

https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/3.2E4.2E0.2ECR1.20extension-descriptor.20build.20error

aldettinger:
Shouldn't be that bad anyway: knative is more an internal thing if I remember well ?
Also not sure about the why... maybe using provided scope could workaround and be less intrusive, not sure.

jamesnetherton:
> maybe using provided scope could workaround and be less intrusive

It's a transitive of `quarkus-vertx-http`, so I think that would override the original `compile` scope, which is probably not what we want.

The weird thing is that the error only occurs when the build is run from the project root. And not when run directly from the `knative` extension directory.

PR #5271, opened 2023-09-06
Test REST to SOAP bridge scenario

ppalaga:
This is related to https://issues.apache.org/jira/browse/CAMEL-19843

### DIFF
--- a/integration-test-groups/cxf-soap/cxf-soap-rest/src/main/java/org/apache/camel/quarkus/component/cxf/soap/rest/it/CxfSoapRestRoutes.java
+++ b/integration-test-groups/cxf-soap/cxf-soap-rest/src/main/java/org/apache/camel/quarkus/component/cxf/soap/rest/it/CxfSoapRestRoutes.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


aldettinger:
Shouldn't be that bad anyway: knative is more an internal thing if I remember well ?
Also not sure about the why... maybe using provided scope could workaround and be less intrusive, not sure.

oscerd (approved):

integration-test-groups/cxf-soap/cxf-soap-rest/src/main/java/org/apache/camel/quarkus/component/cxf/soap/rest/it/CxfSoapRestRoutes.java
```diff
@@ -0,0 +1,88 @@
...
+    @ApplicationScoped
+    @Named("loggingFeatureRest")
+    public LoggingFeature loggingFeatureRest() {
+        final LoggingFeature result = new LoggingFeature();
+        result.setPrettyLogging(true);
+        return result;
+    }
+
+    @Produces
+    @SessionScoped
```
aldettinger, line 69: Not used to `@SessionScoped`, there is doc mentioning it's only available for quarkus-undertow:
https://quarkus.io/guides/cdi#bean-scope-available

Maybe the doc is wrong.

aldettinger (approved):
As far as I get, looks good.

integration-test-groups/cxf-soap/cxf-soap-rest/src/main/java/org/apache/camel/quarkus/component/cxf/soap/rest/it/CxfSoapRestRoutes.java
ppalaga: It could be `@ApplicationScoped` here. I copied it from the client test where there are other reasons to have it `@SessionScoped` (and it works there for sure). Should I change it?

aldettinger: Many thanks for explanation @ppalaga. No strong advice there, if it's working then let it be.

PR #5339, opened 2023-09-22
Add xslt prefix to grouped XML test sub-modules so that the intent is clear

jamesnetherton:


oscerd (approved):

zhfeng:
@jamesnetherton does it need backport to `3.2.x`?

jamesnetherton:
> @jamesnetherton does it need backport to `3.2.x`?

Yeah probably should backport as I'm planning to follow up with some additional tests in `integration-test-groups/xml`.

PR #5346, opened 2023-09-25
Add basic test coverage for xml-jaxp type converters

jamesnetherton:


oscerd (approved):

integration-tests/xml-jaxp/src/main/java/apache/camel/quarkus/component/xml/jaxp/XmlJaxpRoutes.java
```diff
@@ -0,0 +1,146 @@
...
+                .convertBodyTo(StreamSource.class)
+                .convertBodyTo(String.class);
+
+        from("direct:streamSourceReaderConvert")
+                .convertBodyTo(StreamSource.class)
+                .convertBodyTo(Reader.class)
+                .convertBodyTo(String.class);
+
+        from("direct:xmlStreamReaderCharsetConvert")
+                .convertBodyTo(XMLStreamReader.class, StandardCharsets.ISO_8859_1.name())
```
aldettinger, line 99: Changing that charset does not seem to break the test. Does the scenario really test with charset ?

Like the incoming body a String ? The expected type is XMLStreamReader, so the chosen converter would be this [one](https://github.com/apache/camel/blob/main/core/camel-xml-jaxp/src/main/java/org/apache/camel/converter/jaxp/StaxConverter.java#L186) ? So it doesn't support charset ?

If right, maybe we would need to find a way to get ISO-8859-1 encoded byte array as input.

jamesnetherton: I pretty much stole the original code from Camel. I can double check it.

jamesnetherton: Seems I did too much tidying before the PR and removed the critical part of the route.

The issue should be fixed now.

PR #5365, opened 2023-09-26
Remove guide link metadata from camel-k extension

jamesnetherton:
FYI @holly-cummins.

### DIFF
--- a/extensions/camel-k/runtime/src/main/resources/META-INF/quarkus-extension.yaml
+++ b/extensions/camel-k/runtime/src/main/resources/META-INF/quarkus-extension.yaml
@@ -26,7 +26,6 @@ description: "Camel runtime specialization for Camel K"
 icon-url: "https://camel.apache.org/_/img/logo-d-f21b25ba38.svg"
 metadata:
   unlisted: true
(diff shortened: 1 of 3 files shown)


aldettinger (approved):

jamesnetherton:
Ok - didn't know we had a script that validates the guide link is present. Guess I'll fix that up too.

holly-cummins:
> Ok - didn't know we had a script that validates the guide link is present. Guess I'll fix that up too.

What would be *awesome* is if the script validated the guide link was valid. 

Although, actually, it's non-trivial since the codebase and the website are not in sync, so sometimes extensions are in code but nothing is on the web ... and sometimes the docs are removed from the web but the old extensions are still there, so the link needs to mapped from a `latest` link to an older version. I try and do all that before publishing a guide link on http://quarkus.io/extensions, but it would be a lot for the script to do. If the script *did* want to try and do that deeper validation, happy to share my logic. :)

jamesnetherton:
Latest commit tweaks things to just target the `camel-k` extension.

zhfeng (approved):

Issue #5366, opened 2023-09-26
Kafka: Unresolved type o.quarkus.kafka.client.serialization.JsonArraySerializer in basic app


JiriOndrusek:
### Bug description

Go to https://code.quarkus.io/ and generated app based only on Camel Kafka.

The native build fails with:

```
Discovered unresolved type during parsing: io.vertx.core.json.JsonArray. This error is reported at image build time because class io.quarkus.kafka.client.serialization.JsonArraySerializer is registered for linking at image build time by command line
```

Native build should succeed.

PR #5367, opened 2023-09-26
Kafka: native build fails because of missing vertx

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/5366


jamesnetherton:
For some of these missing dependency issues, are we checking that the same issue is not present for the plain Quarkus extension? (E.g `io.quarkus:quarkus-kafka-clients`).

JiriOndrusek:
The only relevant extensins with such issues are mongodb and kafka (there are quarkus only extensions for both of them )
Both of those quarkus extensions can be built in the native without an error.

(the rest cq extensions - xslt-saxon, salesforce aws2-kinesis - doesn't have quarkus only extension

zhfeng (approved):

PR #5380, opened 2023-09-29
feat(native): support Native Sources

squakez:
With this PR we enable Native Sources compilation - https://quarkus.io/guides/building-native-image#separating-java-and-native-image-compilation

This is going to be fully compatible with canonical native and it uses the Quarkus suggested way of checking native build step: https://github.com/quarkusio/quarkus/blob/main/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeOrNativeSourcesBuild.java

### DIFF
--- a/extensions-jvm/asn1/deployment/src/main/java/org/apache/camel/quarkus/component/asn1/deployment/Asn1Processor.java
+++ b/extensions-jvm/asn1/deployment/src/main/java/org/apache/camel/quarkus/component/asn1/deployment/Asn1Processor.java
@@ -20,7 +20,7 @@ import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.ExecutionTime;
 import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
-import io.quarkus.deployment.pkg.steps.NativeBuild;
+import io.quarkus.deployment.pkg.steps.NativeOrNativeSourcesBuild;
 import org.apache.camel.quarkus.core.JvmOnlyRecorder;
 import org.jboss.logging.Logger;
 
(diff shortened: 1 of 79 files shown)


essobedo (approved):

extensions-jvm/asn1/deployment/src/main/java/org/apache/camel/quarkus/component/asn1/deployment/Asn1Processor.java
```diff
@@ -37,7 +37,7 @@ FeatureBuildItem feature() {
     /**
      * Remove this once this extension starts supporting the native mode.
      */
-    @BuildStep(onlyIf = NativeBuild.class)
+    @BuildStep(onlyIf = NativeOrNativeSourcesBuild.class)
     @Record(value = ExecutionTime.RUNTIME_INIT)
     void warnJvmInNative(JvmOnlyRecorder recorder) {
```
essobedo, line 42: In case of warnings like here, it could remain `NativeBuild` but for the sake of uniformity, I understand that it is changed for `NativeOrNativeSourcesBuild`

squakez: IMO we should stick to it in order to warn also when native-sources is used.

aldettinger (approved):
Reviewing the involved build steps, it should work with native or native sources.
Worst case, this pr make it possible to experiment and refine.
Plus it is a good moment from release stream point of view to include such a change.

squakez:
Thanks folks. Are we good to merge it then? I'd like to have this in the next release so I can follow up the work on Camel K side.

aldettinger:
@squakez So with less coupling between Camel K and the runtime, this pr targets Camel Quarkus 3.5.x only ?

@jamesnetherton @ppalaga Please raise a hand in case there is any blocker to merge this.

squakez:
@aldettinger I'd be good with any version you're willing to release next.

aldettinger:
Syncing on this from side discussion, let's better merge in main to target 3.5.x.

extensions-jvm/asn1/deployment/src/main/java/org/apache/camel/quarkus/component/asn1/deployment/Asn1Processor.java
aldettinger: @essobedo Both side, we might be good to merge. Worst case, we could experiment and fix on main.

PR #5408, opened 2023-10-06
perf-tool: upgrade hyperfoil version

aldettinger:


oscerd (approved):

jamesnetherton (approved):
Should we also update the `maven.compiler` properties? It's pointing at JDK 11...

aldettinger:
> Should we also update the `maven.compiler` properties? It's pointing at JDK 11...

Sounds right, still let's put this in todo list for another pr. Need a bit more thought, I think we would just lose the capacity to run perf-regression for old versions, 2.x maybe. That might not be an issue, maybe we could target to compare only active branch, like today 2.16.x, 3.2x and 3.4.0?

aldettinger:
Failing tests are not related, merging then.

PR #5409, opened 2023-10-06
Remove redundant cxf.version property

jamesnetherton:
I believe this is a relic of when we used to have CXF maven plugins in our tests. But that's since been removed in favor of Quarkus codegen.


oscerd (approved):

aldettinger (approved):
ok, the cxf version is pulled from quarkus-cxf-bom :+1:

ppalaga (approved):

Issue #5432, opened 2023-10-16
Ftp tests fail in FIPS environment


JiriOndrusek:
### Bug description

If you run ftp tests on machine with FIPS, you encounter a failure:

```
Caused by: java.security.UnrecoverableKeyException: failed to decrypt safe contents entry: java.security.NoSuchAlgorithmException: Cannot find any provider supporting PBEWithSHA1AndRC2_40
```

or 

```
Caused by: com.jcraft.jsch.JSchException: SSH_MSG_DISCONNECT: 9 getKexProposal() no resolved signatures available
```

PR #5433, opened 2023-10-16
Ftp fails in the FIPS because of the not supported key.

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/5432


osmman:
Hello, I think that we should not use JKS in FIPS environment, because it is not standartized and is deprecated. Plus it is not certified by FIPS. We should use PKCS12 keystore.

JiriOndrusek:
Thanks for the tip! That makes sense and I'll change the PR accordingly.

oscerd (approved):

JiriOndrusek:
Failure is not related to the changes (which affects only ftp module) -> `[INFO] Camel Quarkus :: Integration Tests :: Debug ........ FAILURE [ 20.200 s]`

PR #5477, opened 2023-10-31
Register To*Stream classes for reflection

turing85:


aldettinger (approved):

extensions-support/xalan/deployment/src/main/java/org/apache/camel/quarkus/support/xalan/deployment/XalanNativeImageProcessor.java
zhfeng, line 1: @turing85 Is it possible to add some tests for `To*Stream` ? so we can see these reflection are work.

turing85: @zhfeng not easily, no. I know that I have an application that triggers this case. I do not know, however, what part of the application is responsible for the classes to be used. And I cannot share the application I have.

zhfeng: @turing85 I see. And is there any `<xsl:message>` to write some output XML contents? I try to debug the XSLT compiler and it seems that in this case, `ToXMLStream` is used in the generated translet class.

turing85: Hmm... would have to check. I am not sure whether there is an output. I was able to trigger the classes yesterday with transformations setting an explicit `<xsl:output... />`, but this alone did not trigger the error. I can say that the application having this error uses custom extension functions.

turing85: Adding a `<xsl:message>foo</xsl:message>` does not trigger it either.

zhfeng: It might write some XML content, like `<xsl:message><test>foo</test></xsl:message>`?

turing85: hmm.... let me test...

turing85: Nope. No change. Still works.

zhfeng: OK. I will merge this PR and maybe find a test later.

zhfeng (approved):

Issue #5482, opened 2023-11-03
Cxf soap tests fail in FIPS environment


JiriOndrusek:
### Bug description

If you run `integration-test-groups/cxf-soap/` in FIPS environment, following tests fail:

* `cxf-soap-ws-security-server` 

```
Errors: 
 CxfWssSecurityPolicyServerCxfWayTest.encrypetdSigned:53 » SOAPFault unsupported key transport encryption algorithm: No such algorithm: "RSA/ECB/OAEPWithSHA1AndMGF1Padding"
 CxfWssSecurityPolicyServerTest.encrypetdSigned:52 » SOAPFault unsupported key transport encryption algorithm: No such algorithm: "RSA/ECB/OAEPWithSHA1AndMGF1Padding"
```

* `cxf-soap-ws-trust`

```
Errors: 
 CxfWsTrustTest.programmaticSts:107 » SOAPFault unsupported key transport encryption algorithm: No such algorithm: "RSA/ECB/OAEPWithSHA1AndMGF1Padding"
 CxfWsTrustCxfWayTest.programmaticSts:112 » SOAPFault unsupported key transport encryption algorithm: No such algorithm: "RSA/ECB/OAEPWithSHA1AndMGF1Padding"
```

PR #5485, opened 2023-11-06
Cxf-soap tests fail in FIPS environment (fix for JVM)

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/5482


JiriOndrusek:
Native failure has to be investigated further, reported as https://github.com/apache/camel-quarkus/issues/5486

JiriOndrusek:
All certficates and keystores have to be generated via openssl to work correctly in the FIPS, therefore I removed the keytool-plugin and added a description, how to generate certificates. 

I found several bad passwords in the callbacks and fixed them (I'm not sure why it was working  in non-FIPS environment - probably the passwords were not required by the cxf in non-FIPS) .

jamesnetherton:
> For the FIPS-enabled environment, the profile `fips` has to be used. The reason is that the `BC-FIPS` is used instead of the `BC`

The `fips` Maven profile has:

```
<properties>
    <security.procide>quarkus.security.security-providers=BCFIPS</security.procide>
</properties>
```

What does that do?

JiriOndrusek:
Sorry I forgot to remove it on all places. It shouldn'tbe there

JiriOndrusek:
> > For the FIPS-enabled environment, the profile `fips` has to be used. The reason is that the `BC-FIPS` is used instead of the `BC`
> 
> The `fips` Maven profile has:
> 
> ```
> <properties>
>     <security.procide>quarkus.security.security-providers=BCFIPS</security.procide>
> </properties>
> ```
> 
> What does that do?

Removed. Thanks for noticing!

oscerd (approved):

ppalaga:
Great work, thanks for taking care, @JiriOndrusek!

> All certficates and keystores have to be generated via openssl

Do you happen to know what aspect of the generated keystore is not FIPS-compliant with keytool-maven-plugin? At the first sight, the plugin mojos seem to support all we need: keysize, keyalgorithm, storetype...

JiriOndrusek:
> Great work, thanks for taking care, @JiriOndrusek!
> 
> > All certficates and keystores have to be generated via openssl
> 
> Do you happen to know what aspect of the generated keystore is not FIPS-compliant with keytool-maven-plugin? At the first sight, the plugin mojos seem to support all we need: keysize, keyalgorithm, storetype...

I probably mislead a little bit with my comment. It is possible to run successfully the `keytool-maven-plugin`, but I haven't found a way of generating such strong certificates to satisfy runtime in FIPS

JiriOndrusek:
> It looks good.
> 
> So we configure openssl with the `.cnf` configuration file. Is it need for fips purpose ? Or something else please ?

TBH I'm using *.cnf mainly because the default values functionality (so user is not forced to fill all - e.g. CN which is "referenced" from the tests. I'll look into an option with no *.cnf files in the future to be sure whether it is mandatory.

JiriOndrusek:
There are 2 open issues:
- problem in the native (reported as https://github.com/apache/camel-quarkus/issues/5486) I'll investigate it .
- the option of not using BCFIPS in FIPS and use only SunPKCS11-NSS-FIPS provider. I think that it won't be possible, because quarkus-cxf-rt-ws-security brings BC as a transitive dependency, therefore requires BC. But I think that I'll discover the answer during investigation of the native failure. The native issue seems like a BC class is needed, when BC is replaced with BC-FIPS

aldettinger (approved):
Thanks for answer @JiriOndrusek . That's look good start and investigation can be done in further prs. Related to `.cnf` file, it seems good point to have default value provided, beyond no strong opinion on this.

ppalaga:
> > Do you happen to know what aspect of the generated keystore is not FIPS-compliant with keytool-maven-plugin? At the first sight, the plugin mojos seem to support all we need: keysize, keyalgorithm, storetype...
> 
> I probably mislead a little bit with my comment. It is possible to run successfully the `keytool-maven-plugin`, but I haven't found a way of generating such strong certificates to satisfy runtime in FIPS.
> 
> I remember seeing some not-supported algorithms for signing or similar, when I was trying to use keystore only. I don't remember the detail but will save them during investigation of other FIPS issues.

This sounds like an answer: https://stackoverflow.com/a/50398011 

> PKCS#12 stores the private key encrypted with a password derived key. It looks like keytool uses pbeWithSHAAnd128BitRC2-CBC (pkcs-12PbeIds 5), an PBES1 algorithm for doing so. Even the keytool.exe of Oracle Java 9 does use this algorithm as you can verify by uploading a .p12 file to the online ASN.1 decoder decoding a sample PKCS#12 file.
> 
> If I read the [PKCS#12 standard](https://www.rfc-editor.org/rfc/rfc7292) correctly PBES1 was long ago superseeded by the "newer" version of the key derivation system named "PBES2" (mostly PBKDF2 based) with should be used instead. But keytool does not make use of it. This is my interpretation of the error message.
> 
> Therefore the certificate and the key may be acceptable, but the PKCS#12 container is not acceptable. You may try to extract key and certificate and save them in a new PKCS#12 file using a current software like OpenSSL (or you simply generate the whole PKCS#12 file directly using OpenSSL).

JiriOndrusek:
> > > Do you happen to know what aspect of the generated keystore is not FIPS-compliant with keytool-maven-plugin? At the first sight, the plugin mojos seem to support all we need: keysize, keyalgorithm, storetype...
> > 
> > 
> > I probably mislead a little bit with my comment. It is possible to run successfully the `keytool-maven-plugin`, but I haven't found a way of generating such strong certificates to satisfy runtime in FIPS.
> > I remember seeing some not-supported algorithms for signing or similar, when I was trying to use keystore only. I don't remember the detail but will save them during investigation of other FIPS issues.
> 
> This sounds like an answer: https://stackoverflow.com/a/50398011
> 
> > PKCS#12 stores the private key encrypted with a password derived key. It looks like keytool uses pbeWithSHAAnd128BitRC2-CBC (pkcs-12PbeIds 5), an PBES1 algorithm for doing so. Even the keytool.exe of Oracle Java 9 does use this algorithm as you can verify by uploading a .p12 file to the online ASN.1 decoder decoding a sample PKCS#12 file.
> > If I read the [PKCS#12 standard](https://www.rfc-editor.org/rfc/rfc7292) correctly PBES1 was long ago superseeded by the "newer" version of the key derivation system named "PBES2" (mostly PBKDF2 based) with should be used instead. But keytool does not make use of it. This is my interpretation of the error message.
> > Therefore the certificate and the key may be acceptable, but the PKCS#12 container is not acceptable. You may try to extract key and certificate and save them in a new PKCS#12 file using a current software like OpenSSL (or you simply generate the whole PKCS#12 file directly using OpenSSL).

Thanks for the explanation. I can confirm the behavior, as I tried JKS keystores at first, but they are not accepted by FIPS-enabled system. I had to move to PKCS12

integration-test-groups/cxf-soap/cxf-soap-ws-security-server/README.adoc
```diff
@@ -16,4 +16,50 @@ We test in two ways how the SOAP service endpoints are deployed: the Camel way a
...
+
+When prompted for password, type `password`.
+When prompted whether to trust the certificate, type `yes`.
+
+```
+openssl pkcs12 -export -in alice.crt -inkey alice.key -certfile cxfca.crt -name "alice" -out alice.p12 -passout pass:password -keypbe aes-256-cbc -certpbe aes-256-cbc
+openssl pkcs12 -export -in bob.crt -inkey bob.key -certfile cxfca.crt -name "bob" -out bob.p12 -passout pass:password -keypbe aes-256-cbc -certpbe aes-256-cbc
+
+keytool -import -trustcacerts -alias bob -file bob.crt -keystore alice.p12
+keytool -import -trustcacerts -alias alice -file alice.crt -keystore bob.p12
```
ppalaga, line 64:
```suggestion
```

This won't work based on the info from https://stackoverflow.com/a/50398011 
It is also not necessary because we do `openssl pkcs12 -export ... -out alice.p12` (and the same for Bob) above.

ppalaga, line 64: I'd prefer having a single runnable shell script so that it is easier to check whether it works and so that it is easier to copy it to Quarkus CXF :-) The comments can be moved there too.

integration-test-groups/cxf-soap/cxf-soap-ws-security-server/README.adoc
```diff
@@ -16,4 +16,50 @@ We test in two ways how the SOAP service endpoints are deployed: the Camel way a
...
+=== Generate certificates
+```
+openssl req -new -subj '/O=apache.org/OU=eng (NOT FOR PRODUCTION)/CN=cxfca' -x509 -key cxfca.key -out cxfca.crt
+
+openssl req -new -subj '/O=apache.org/OU=eng (NOT FOR PRODUCTION)/CN=alice' -key alice.key -out alice.csr
+openssl x509 -req -in alice.csr -CA cxfca.pem -CAkey cxfca.key -CAcreateserial -out alice.crt
+
+openssl req -new -subj '/O=apache.org/OU=eng (NOT FOR PRODUCTION)/CN=bob' -key bob.key -out bob.csr
+openssl x509 -req -in bob.csr -CA cxfca.pem -CAkey cxfca.key -CAcreateserial -out bob.crt
+```
```
ppalaga, line 52: Could you please have a look what is the validity duration of the generated certs? If it is something relatively short like a couple of weeks or moths, I'd vote for increasing it to something reasonably long (20 years?) so that we do not have to regen them too often. What do others think?

JiriOndrusek: good point. I 'll change the validity to several years, the default validity seems to be 1 months, which would require the regeneration of the certificates too often.

integration-test-groups/cxf-soap/cxf-soap-ws-security-server/README.adoc
JiriOndrusek: I agree. I kept this approach as I saw it on different places in CQ, but putting all commands together makes sense..

ppalaga: I am blind, sorry, I thought `openssl pkcs12 -export` and `keytool -import` were doing the same thing which they apparently don't. Sorry for the confusion. 

It would be really interesting if `keytool -import` really works. It would mean that keytool is able to import to a proper PKCS12 container, although it is not able to create it.

JiriOndrusek: That is my impression, that keytool is able to import to existing PKCS12 container. At least the tests are working and if I reme,mber correctly, this import is necessary - but I'll verify it during today and let you know.

ppalaga: Well if it really works, I wonder if we could/should store an empty PKCS12 container created by openssl in the repo and do the rest with keytool plugin. Not that I really insist on doing that. Just an idea. 

If we stay with the shell script, can't the `keytool -import` be done with openssl? The shell script would thus have less dependencies.

JiriOndrusek: I understand, I originally tried to achieve that with openssl only,  but without a success . But I've just found a guide describing how to do that. so I  will try, whether the result is  as described,

JiriOndrusek: @ppalaga I put the commands together with `echo *` as a separator (and to show hint if needed), is it ok? (It can be copied and run as one block on linux)

JiriOndrusek: @ppalaga Eventhough it is possible to generate a PKCS12 with more certificates (with a workaround by concatenating the certificates), there is  only 1 alias  present in the keystore. Cxf is not able to load appropriate certificate from such keystore using alias and there is no API on `openssl` CLI for setting different aliases. Therefore the solution with keytool is in my POV the only usable one.

ppalaga: @llowinge and his friends might perhaps want to have the commands in a single script to be able to execute it from their CI (to make sure that also the cert generation works on FIPS enabled systems)?

ppalaga: OK, thanks for explaining!

JiriOndrusek: you mean too to extract the current commands into a single file. I understand now, no problem, I'll prepare it.

JiriOndrusek: Script  is in separated file.

zhfeng (approved):

JiriOndrusek:
The change in CXF is prepared as https://github.com/apache/cxf/pull/1660, as soon as the CXF PR is merge, I'll rework this PR to leverage `CustomizedAlgorithmSuite`

JiriOndrusek:
This PR is using BCFIPS, which is not the right call anymore. Once the following [change](https://github.com/apache/cxf/pull/1660) is merged into cxf and leveraged by the CQ, this PR should be reworked.

JiriOndrusek:
Superseeded by https://github.com/apache/camel-quarkus/pull/5980

PR #5509, opened 2023-11-14
Clean up usage of hard coded hosts that use containers

jamesnetherton:


oscerd (approved):

zhfeng (commented):
It also needs to do in `PahoMqtt5TestResource.java` https://github.com/apache/camel-quarkus/blob/5f4d5b383a9f06aca74dfb9027262b73ace3db93/integration-tests/paho-mqtt5/src/test/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5TestResource.java#L80-L81 and https://github.com/apache/camel-quarkus/blob/5f4d5b383a9f06aca74dfb9027262b73ace3db93/integration-tests/paho-mqtt5/src/test/java/org/apache/camel/quarkus/component/paho/mqtt5/it/PahoMqtt5TestResource.java#L102-L109

jamesnetherton:
> It also needs to do in PahoMqtt5TestResource.java

Thanks, I made work like the other paho test. The downside is that when using a custom `HostnameVerifier` `org.eclipse.paho.mqttv5.client.internal.NetworkModuleService` does reflection on an internal JDK class `java.net.URI`. So I had to `--add-opens` to deal with that.

zhfeng (approved):
Yeah, that makes sense.

integration-tests/paho-mqtt5/src/main/resources/application.properties
```diff
@@ -15,3 +15,6 @@
 ## limitations under the License.
 ## ---------------------------------------------------------------------------
 quarkus.native.resources.includes=*.jks
+
+# Required by NetworkModuleService.setURIField
+quarkus.native.additional-build-args=-J--add-opens=java.base/java.net=ALL-UNNAMED
```
zhfeng, line 20: Do we need to add some notes in docs of `paho-mqtt5`? is it only needed in ssl cases?

jamesnetherton: I went back to look at it again. And the test is actually passing without it.

Maybe I previously had something misconfigured that triggered the problem code in `NetworkModuleService`.

I can revert the `--add-opens` and create follow up issue to investigate further as there's potentially an issue there for some scenarios (to be verified). WDYT?

zhfeng: Sure, it looks good to me.

PR #5573, opened 2023-11-30
Use native builder image pull strategy 'missing' to reduce interactions with quay.io

jamesnetherton:


zhfeng (approved):

jamesnetherton:
> So, at each ci build, on first container build the image is pulled only once, and then reused ? 

Yes, correct. Pulled once on each CI node running the native build 

> And then we have a chance to get a potential container build image update on next build ?

Yeah - CI nodes are ephemeral. So we start with a clean slate on every workflow run.

aldettinger (commented):
Looks good.
LGTM

Issue #5617, opened 2023-12-20
Camel 4.4 - HDFS component removed


oscerd:
### Description

The extensions need to be removed as well

PR #5624, opened 2024-01-02
Reference Camel SNAPSHOT docs as 4.3.0 is not published on the website

jamesnetherton:
Works around some failures observed in the Camel website build.


jamesnetherton:
I think for this to work we'd also need to implement https://github.com/apache/camel-quarkus/issues/5617. Given HDFS already seems to be marked as [deprected](https://github.com/apache/camel-quarkus/blob/main/extensions-jvm/hdfs/runtime/src/main/resources/META-INF/quarkus-extension.yaml#L34), there's probably no harm in doing the change now.

oscerd (approved):

aldettinger (approved):

Issue #5582, opened 2023-12-01
Adding extension for SMB


spatnity:
### Describe the feature here

 Adding a new Camel component extension to the project i.e SMB. It is a new camel component that lets you interact with SMB file shares

PR #5644, opened 2024-01-11
Quarkus extension for smbComponent

spatnity:


spatnity:
#5582 Adding extension for SMB

extensions/smb/deployment/src/main/java/org/apache/camel/quarkus/component/smb/deployment/JakartaEnablement.java
```diff
@@ -0,0 +1,94 @@
...
+import org.eclipse.transformer.action.impl.SignatureRuleImpl;
+import org.eclipse.transformer.util.FileUtils;
+import org.objectweb.asm.ClassReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Required due to net.engio.mbassy.dispatch.el.ElFilter referencing javax.el packages.
+ *
+ * TODO: Remove this class if this issue is resolved: https://github.com/bennidi/mbassador/issues/169
```
aldettinger, line 43: Could it be interesting to create a Camel Quarkus issue also so that we don't forget ?

integration-tests/smb/src/test/java/org/apache/camel/quarkus/component/smb/it/SmbTestResource.java
```diff
@@ -0,0 +1,61 @@
...
+public class SmbTestResource implements QuarkusTestResourceLifecycleManager {
+
+    private static final int SMB_PORT = 445;
+
+    private GenericContainer<?> container;
+
+    @Override
+    public Map<String, String> start() {
+        try {
+            container = new GenericContainer<>(new ImageFromDockerfile("localhost/samba:camel", false)
```
aldettinger, line 35: So we use a locally built samba container ? Isn't there a kind of official image out there so that we don't need to maintain it ?

@jamesnetherton Probably we would like to avoid ImageFromDockerFile if possible ?

jamesnetherton: > So we use a locally built samba container ? Isn't there a kind of official image out there so that we don't need to maintain it ?
> 
> @jamesnetherton Probably we would like to avoid ImageFromDockerFile if possible ?

It's a direct copy from how the Camel SMB test does things. It at least enables the test data to be set up easily.

There probably is an alternative. I'd have to check for a good candidate. Not sure if we need to do that now or follow up aftwards?

aldettinger: Indeed, it maps the camel way. Guess it was done that way because there is no official image.
On the top of my head, we had also used ImageFromDockerFile in mail test where it worked for a while.
So probably, this is suitable for platform tests and also for people testing with different processors instruction sets.

aldettinger (approved):
It looks great job @spatnity :+1:

extensions/smb/deployment/src/main/java/org/apache/camel/quarkus/component/smb/deployment/JakartaEnablement.java
spatnity: Yeah we could create one.May I create @jamesnetherton ?

jamesnetherton: https://github.com/apache/camel-quarkus/issues/5646

jamesnetherton (approved):
Nice work @spatnity 😎

spatnity:
> Nice work @spatnity 😎
Thankyou! @jamesnetherton thanks for supporting:)

Issue #5620, opened 2023-12-31
Camel 4.4 - Allow to configure camel.variable.xxx


davsclaus:
### Describe the feature here

Camel 4.4 introduced variables which can be configured from application.properties.
https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-19749

However for Quarkus and Spring Boot then this requires to add code to support this

A SB example is at
https://github.com/apache/camel-spring-boot-examples/tree/main/variables

And the commit in SB to support this is:
https://github.com/apache/camel-spring-boot/commit/671e772eb2240ff0e4537f5e5d4330714cb73f24

JiriOndrusek:
There are 3 parts of the solution:

1. Code in the camel-quarkus main: which is already merged in the `camel-main` branch as a part of upgrade to Camel 4.4
2. Basic test coverage: https://github.com/apache/camel-quarkus/pull/5654
3. Example: https://github.com/apache/camel-quarkus-examples/pull/194

jamesnetherton:
Bit late - but done here https://github.com/apache/camel-quarkus-examples/pull/194.

PR #5654, opened 2024-01-17
Variables - test coverage

JiriOndrusek:
Partially fixes https://github.com/apache/camel-quarkus/issues/5620
(the other part of the solution is the example - https://github.com/apache/camel-quarkus-examples/pull/194)

This PR adds basic variables coverage into `integration-test-groups/foundation`.

One test (`customVariablerepository`) is not executed in the native move, because it is not possible to remove a CDI bean for one test. IMO the tested functionality uses functions which were already part of the Camel before, therefore there is no reason to test this in native.

### DIFF
--- a/integration-test-groups/foundation/variables/src/main/java/org/apache/camel/quarkus/variables/it/VariablesResource.java
+++ b/integration-test-groups/foundation/variables/src/main/java/org/apache/camel/quarkus/variables/it/VariablesResource.java
@@ -1,0 +1,186 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.variables.it;
+
+import java.util.List;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.core.Response;
+import org.apache.camel.CamelContext;
+import org.apache.camel.Exchange;
+import org.apache.camel.ProducerTemplate;
+import org.apache.camel.component.mock.MockEndpoint;
+
+@Path("/variables")
+@ApplicationScoped
+public class VariablesResource {
+
+    @Inject
+    ProducerTemplate producerTemplate;
+
+    @Inject
+    CamelContext context;
+
+    @Path("/setLocalVariable")
+    @POST
+    public String setLocalVariable(String body) throws Exception {
+        MockEndpoint end = context.getEndpoint("mock:setLocalVariableEnd", MockEndpoint.class);
+        end.expectedMessageCount(1);
+        end.expectedVariableReceived(VariablesRoutes.VARIABLE_NAME, VariablesRoutes.VARIABLE_VALUE);
+
+        producerTemplate.requestBody("direct:setLocalVariableStart", body, String.class);
+
+        // make sure we got the message
+        end.assertIsSatisfied();
+
+        // lets get the variable value
+        List<Exchange> exchanges = end.getExchanges();
+        Exchange exchange = exchanges.get(0);
+
+        return exchange.getVariable(VariablesRoutes.VARIABLE_NAME, String.class);
+    }
+
+    @Path("/setGlobalVariable")
+    @POST
+    public Response setGlobalVariable(String body) throws Exception {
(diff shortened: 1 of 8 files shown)


oscerd (approved):

integration-test-groups/foundation/variables/src/main/java/org/apache/camel/quarkus/variables/it/VariablesResource.java
```diff
@@ -0,0 +1,195 @@
...
+        Exchange midExchange = exchanges.get(0);
+
+        return midExchange.getVariable(VariablesRoutes.VARIABLE_NAME, String.class) + ","
+                + context.getVariable(VariablesRoutes.VARIABLE_NAME, String.class);
+    }
+
+    @Path("/customGlobalRepository")
+    @POST
+    public Response customGlobalRepository(String body) throws Exception {
+        //        context.getRegistry().bind(GLOBAL_VARIABLE_REPOSITORY_ID, new MyGlobalRepo());
```
aldettinger, line 138: Should we remove this one if useless ?

integration-test-groups/foundation/variables/src/main/java/org/apache/camel/quarkus/variables/it/VariablesResource.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    @Path("/convert")
+    @POST
+    public String convert(String body) throws Exception {
+        MockEndpoint end = context.getEndpoint("mock:convertEnd", MockEndpoint.class);
+        end.expectedMessageCount(1);
+
+        producerTemplate.requestBody("direct:convertStart", body, String.class);
+        //        fluentProducerTemplate.to("direct:convertStart").withVariable(VariablesRoutes.VARIABLE_NAME, body).send();
```
aldettinger, line 166: Same here, could we remove that ?

integration-test-groups/foundation/variables/src/main/java/org/apache/camel/quarkus/variables/it/VariablesResource.java
```diff
@@ -0,0 +1,195 @@
...
+    @POST
+    public Response filter(String body, @PathParam("city") String city) throws Exception {
+        MockEndpoint end = context.getEndpoint("mock:filterEnd", MockEndpoint.class);
+        end.expectedMessageCount(1);
+
+        producerTemplate.requestBodyAndHeader("direct:filterStart", body, "city", city, String.class);
+
+        try {
+            Exchange exhange = end.assertExchangeReceived(0);
+            end.reset();
```
aldettinger, line 187: Didn't we have some kind of race condition issue generating flaky tests in the past ? A bit like 2 threads invoking the filter method concurrently and then 1 thread call reset while the other thread had the exchange.

Not sure it applies here anyway, still it would it be good to find another way to test without reset ?

integration-test-groups/foundation/variables/src/main/java/org/apache/camel/quarkus/variables/it/VariablesResource.java
JiriOndrusek: I missed that, will remove, thanks!

integration-test-groups/foundation/variables/src/main/java/org/apache/camel/quarkus/variables/it/VariablesResource.java
JiriOndrusek: It seems, that I reverted more code then anticipated, because I was removing that a while ago

integration-test-groups/foundation/variables/src/main/java/org/apache/camel/quarkus/variables/it/VariablesResource.java
JiriOndrusek: If there is a possibility of flakiness, I think that I refactor the test class and ensure that calls to this resource  are synchronized (by joining 2 test methods into 1)

integration-test-groups/foundation/variables/src/test/java/org/apache/camel/quarkus/variables/it/CustomRepositoryVariablesTest.java
```diff
@@ -0,0 +1,59 @@
...
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.QuarkusTestProfile;
+import io.quarkus.test.junit.TestProfile;
+import io.restassured.RestAssured;
+import io.restassured.http.ContentType;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test is executed in JVM mode only because the profile enabling customVariableRepositoty can not be enabled for
```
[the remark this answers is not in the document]
JiriOndrusek: done

integration-test-groups/foundation/variables/src/main/java/org/apache/camel/quarkus/variables/it/VariablesResource.java
aldettinger: Also, I remember a technic where we had one distinct mock per expected message. So that you never call reset... That was far, hard to remember if this applies to the case at hand :)

JiriOndrusek:
> Looks pretty decent @JiriOndrusek
> 
> Concerning mock.reset(), I would vote to find another way if possible. If too difficult, maybe we just add a comment to have context in case the test is proven to be flaky on ci one day.

I removed mock.reset by joining 2 rest calls into 1 test method, which ensures order, therefore no `reset()`  is required.

integration-test-groups/foundation/variables/src/main/java/org/apache/camel/quarkus/variables/it/VariablesResource.java
JiriOndrusek: I skipped `reset()`  by rearranging test method differently.

PR #5680, opened 2024-01-25
Use /bin/bash for release-util sign.sh script

jamesnetherton:


oscerd (approved):
I think we should do this also in other repositories. Thanks for noticing and sorry.

jamesnetherton:
> I think we should do this also in other repositories. Thanks for noticing and sorry.

It's fine - it likely only matters for awkward folks like me who choose to use Debian *nix variants where /bin/sh != /bin/bash.

oscerd:
> > I think we should do this also in other repositories. Thanks for noticing and sorry.
> 
> It's fine - it likely only matters for awkward folks like me who choose to use Debian *nix variants where /bin/sh != /bin/bash.

Well, it's always a use case, so better to fix. I'll take care of projects using the script!

PR #5683, opened 2024-01-26
Update jasypt.adoc

lordrip:
Fix Title order affecting Camel website

image

PR #5684, opened 2024-01-26
Fix Jasypt docs indentation level

jamesnetherton:
Replaces #5683.

I tried a local build of the Camel website and this seems to be enough to fix the warnings about indentation.


lordrip (approved):
Awesome, thanks a lot @jamesnetherton

apupier (approved):

PR #5711, opened 2024-02-02
Jasypt dev UI

jamesnetherton:
Not sure if this is a good idea or not. But it was an excuse to play with the Dev UI 😁 .

I was thinking it could be an alternative to the [Camel Jasypt tooling](https://camel.apache.org/components/next/others/jasypt.html#_tooling). Avoids the need to manually run stuff from the command line etc.

There's an example of how it works [here](https://gist.github.com/jamesnetherton/1a72af01b56b8c90f9efa3781b6cd77d#file-cq-jasypt-devui-md).


oscerd (approved):

ppalaga (approved):
Nice!

PR #5726, opened 2024-02-06
Make alternative-jdk job a matrix build

jamesnetherton:
The `integration-tests-alternative-jdk` CI job has been failing for reasons that I cannot determine. I suspect it's due to running out of disk space. Hence this attempt at reclaiming some space.

### DIFF
--- a/.github/workflows/camel-master-cron.yaml
+++ b/.github/workflows/camel-master-cron.yaml
@@ -36,6 +36,7 @@ jobs:
     outputs:
       matrix: ${{ steps.set-native-matrix.outputs.matrix }}
       examples-matrix: ${{ steps.set-examples-matrix.outputs.examples-matrix }}
+      alternate-jvm-matrix: ${{ steps.set-alternate-jvm-matrix.outputs.alternate-jvm-matrix }}
     env:
       MAVEN_OPTS: -Xmx4500m
     steps:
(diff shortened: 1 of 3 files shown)


jamesnetherton:
Well... the job completed at least. But this change breaks the JMS & Infinispan tests because the 'common' bits they depend on get cleaned up.

Not sure what else we can do aside from splitting the job up.

oscerd (approved):

aldettinger:
> Well... the job completed at least. But this change breaks the JMS & Infinispan tests because the 'common' bits they depend on get cleaned up.
> 
> Not sure what else we can do aside from splitting the job up.

Ah shame, so it's probably not a good idea. No other ideas than splitting on my side.
There does not seem to have that much articles about disk saving technics out there.

jamesnetherton:
Lets see if cleaning up the Downloaded maven repo archive helps...

jamesnetherton:
Tweaked things to split the alt-jdk job into two parts.

jamesnetherton:
@aldettinger CI checks passed. WDYT about the change?

aldettinger:
> @aldettinger CI checks passed. WDYT about the change?

This is the best solution so far. A neat improvement could be one day to cut at other letter than "n", as the first job takes 52m and the second one 22m.

Out of curiosity, what about infra capacity ? We can add as many jobs from matrix as we want and they are spread across available nodes ?

aldettinger (approved):

jamesnetherton:
> This is the best solution so far. A neat improvement could be one day to cut at other letter than "n", as the first job takes 52m and the second one 22m.

Yeah it's not super optimized in the current state.

> Out of curiosity, what about infra capacity ? We can add as many jobs from matrix as we want and they are spread across available nodes ?

This change actually takes us over the limit of concurrent jobs. The native Group 13 tests are left in the pending state until one of the other jobs completes. Which usually happens in around 15 - 20 minutes. So it's not a big delay.

jamesnetherton:
Applied the same changes to the nightly workflows.

PR #5727, opened 2024-02-07
Unshade Google and Micrometer related packages from kudu-client

jamesnetherton:
@ppalaga WDYT about this? Do you think it's safe to unshade more libraries from the Kudu client?

The only thing left shaded with this change is `com.sangupta:murmur`.


ppalaga:
Could we perhaps unshade also com.sangupta:murmur?

jamesnetherton:
> com.sangupta:murmur

Latest commit unshades murmur.

ppalaga (approved):

PR #5763, opened 2024-02-16
Add quarkus-artemis dependencies to the BOM

jamesnetherton:
@zhfeng WDYT?

I was thinking that maybe we should  just explicitly add the bits of quarkus-artemis we actually need. Instead of importing the BOM, which causes all `io.quarkiverse` & `org.apache.activemq` dependencies to get added to the BOM.

### DIFF
--- a/poms/bom-test/pom.xml
+++ b/poms/bom-test/pom.xml
@@ -56,13 +56,6 @@
                 <scope>import</scope>
             </dependency>
             <dependency>
-                <groupId>io.quarkiverse.artemis</groupId>
-                <artifactId>quarkus-artemis-bom</artifactId>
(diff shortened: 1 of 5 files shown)


oscerd (approved):

jamesnetherton:
Hmmm - now I think a bit more, maybe we should just import `quarkus-artemis-bom` and have all of its content in our BOM. Otherwise we maybe mess up user apps if they import `camel-quarkus-bom` & `quarkus-artemis-bom`.

E.g you may end up with conflicting versions depending on the order things are imported.

jamesnetherton:
Need to think a bit more on this. Thus closing for now.

Issue #5742, opened 2024-02-09
Jt400: make native work


JiriOndrusek:
### Describe the feature here

The target is to have jt400 extension work in native. Althought there no tests present so far, the native build fails with

```
 Discovered unresolved type during parsing: com.ibm.as400.access.UnixSocketUser. This error is reported at image build time because class com.ibm.as400.access.CurrentUser is registered for linking at image build time by command line
```

Problem could be fixed without the existence of the tests. It should be possible to move jt400 tests & extension into native  ones.

PR #5782, opened 2024-02-21
Jt400: make native work

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/5742

PR makes extension jt400 compilable for native. (which was failing before, even if there are no tests)


oscerd (approved):

jamesnetherton:
We should probably tidy up a bit after the 3.8.0. But lets merge so that we have a baseline included in the release.

PR #5783, opened 2024-02-21
feat: introduce a CamelModelReifierFactoryBuildItem spi to inject a custom ModelReifierFactory

lburgazzoli:

### DIFF
--- a/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelContextRecorder.java
+++ b/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelContextRecorder.java
@@ -38,6 +38,7 @@ import org.apache.camel.spi.CamelContextCustomizer;
 import org.apache.camel.spi.ComponentNameResolver;
 import org.apache.camel.spi.FactoryFinderResolver;
 import org.apache.camel.spi.ModelJAXBContextFactory;
+import org.apache.camel.spi.ModelReifierFactory;
 import org.apache.camel.spi.ModelToXMLDumper;
 import org.apache.camel.spi.ModelToYAMLDumper;
 import org.apache.camel.spi.PackageScanClassResolver;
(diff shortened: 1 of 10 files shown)


extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelContextRecorder.java
```diff
@@ -75,8 +77,7 @@ public RuntimeValue<CamelContext> createContext(
         extendedCamelContext.addContextPlugin(VariableRepositoryFactory.class, new DefaultVariableRepositoryFactory(context));
         extendedCamelContext.setRegistry(registry.getValue());
 
-        //initialize reifier factory (later then implemented in the Camel)
-        context.setModelReifierFactory(context.createDefaultModelReifierFactory());
+        context.setModelReifierFactory(modelReifierFactory.getValue());
```
aldettinger, line 80: We are near the release, still it might be possible to merge as this pr might not have big consequence for users.

There is just one unclear area. Before this pr, it looks `context.createDefaultModelReifierFactory()` could resolve `ModelReifierFactory` [as a service](https://github.com/apache/camel/blob/main/core/camel-core-engine/src/main/java/org/apache/camel/impl/DefaultCamelContext.java#L849). Not sure it that would bring any scenario with undesired side effect.

@jamesnetherton Any idea ?

jamesnetherton: > Any idea ?

I'm assuming for our case we only ever care about the default reifier factory or the custom one provided by the camel-k extension. So it should be safe.

aldettinger: Yeah, and there are probably not that much other developers playing with such advanced scenarios.

PR #5807, opened 2024-02-27
Use correct quarkus-extension-maven-plugin in beanio extension

jamesnetherton:


aldettinger (approved):
Indeed, it's the only extension having that.

aldettinger:
A clue [here](https://github.com/apache/camel-quarkus/blame/main/extensions-jvm/beanio/runtime/pom.xml#L34) that the origin is probably back and forth between supported and not supported. Should we update the jvmSince version in this case ?

jamesnetherton:
> Should we update the jvmSince version in this case ?

Yes, I actually thought we had done that in the initial PR but I guess we missed it.

Latest commit updates it to 3.8.0.

PR #5823, opened 2024-02-29
[3.8.x] Manage software.amazon.awssdk:endpoints-spi

ppalaga:


jamesnetherton:
Not sure why none of the CI workflows have run...

PR #5897, opened 2024-03-19
Promote Qdrant extension to native

aldettinger:


oscerd (approved):

extensions/qdrant/runtime/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+		<camel.quarkus.jvmSince>3.9.0</camel.quarkus.jvmSince>
+		<camel.quarkus.nativeSince>3.9.0</camel.quarkus.nativeSince>
+	</properties>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.apache.camel.quarkus</groupId>
+			<artifactId>camel-quarkus-core</artifactId>
+		</dependency>
+		<dependency>
```
aldettinger, line 45: @jamesnetherton @ppalaga  A dependency on camel-quarkus-grpc is needed to have the native tests run locally.
Would you have any idea to integrate this cleanly in the source tree ? We need  kind of support project ?

jamesnetherton: Adding `quarkus-grpc-common` should be enough, no?

aldettinger: Indeed, it might be sufficient. There is an issue but it's not related to qdrant :+1:

aldettinger:
Closing in favor of #5948

PR #5904, opened 2024-03-20
Upgrade to Quarkus CXF 3.9.0

ppalaga:


oscerd (approved):

ppalaga:
@aldettinger if this PR passes as is, then I think there is no need to stop the CQ 3.9.0 release train. QCXF 3.8.1 (that we currently use in CQ) and 3.9.0 are quite similar.

aldettinger:
@ppalaga Let see how it goes, the release is not working anyway. So there might be an opportunity integrate this if I need to restart from scratch.

ppalaga:
@aldettinger passed

PR #5905, opened 2024-03-21
Manage com.sun.xml.fastinfoset:FastInfoset

zhfeng:
`com.sun.xml.fastinfoset:FastInfose`t is needed for the support of `@org.apache.cxf.annotations.FastInfoset` annotation which makes CXF endpoints and clients exchange FastInfoset encoded binary messages instead of text XML. This is documented in a minimalistic way around here: https://docs.quarkiverse.io/quarkus-cxf/dev/reference/index.html#_supported_cxf_annotations


oscerd (approved):

ppalaga (approved):
Thanks @zhfeng

Issue #5899, opened 2024-03-20
jt400 `com.ibm.as400.access.AS400` should be registered for runtime reinitialization


jamesnetherton:
### Bug description

There's a large static initializer block:

https://github.com/IBM/JTOpen/blob/98e74fae6d212563a1558abce60ea5c73fcfc0c0/src/main/java/com/ibm/as400/access/AS400.java#L171-L313

Without rerunning it at runtime, the original values discovered at build time are likely baked permanently into the native image. Which is potentially undesirable.

PR #5908, opened 2024-03-21
jt400AS400 should be registered for runtime reinitialization

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/5899

PR also adds missing *IT class and a few native buildItems.

**I set flat-classpath to true**, which is a workaround before finding a correct way of excxluding mock classes if `-Dskip-mock-tests=true` is set.


oscerd (approved):

JiriOndrusek:
@jamesnetherton I'd like to try to fix the flat path correctly (to turn it on only for mocked tests) so I haven't created PR for 3.8.x so far.

aldettinger (approved):
Could be a good experiment, please refrain from merging to main until we are in a better position with the 3.9.0 camel-quarkus release.

zhfeng (approved):

PR #5956, opened 2024-04-04
Add tests for FastCamelContext JSON schema lookups

jamesnetherton:


oscerd (approved):

extensions-core/core/deployment/pom.xml
```diff
@@ -71,6 +71,11 @@
             <artifactId>camel-timer</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-zipfile</artifactId>
+            <scope>test</scope>
+        </dependency>
```
aldettinger, line 78: The pr looks good, not sure about the global context though. Is it that the zipfile JSON schema could not be loaded ? Why is that then ?

jamesnetherton: There is some (perhaps temporary) inconsistency in terms of where the JSON schema files are located.

https://issues.apache.org/jira/browse/CAMEL-20546

So the current implementation would fail to resolve things properly. Hence these changes + tests to verify it.

aldettinger: Ok make sense :+1:

aldettinger (approved):

Issue #5966, opened 2024-04-05
Http: tests are not working in FIPS environment


JiriOndrusek:
### Bug description

Keystore and trusstore used in the tests does not work in FIPS environment.

Issue #5967, opened 2024-04-05
Improve keystore generation


JiriOndrusek:
### Describe the feature here

Http-grouped: It should be possible to generate keystore and truststore during build. (if not disabled)

JiriOndrusek:
The feature for generating the certificates could be created in a general way, so other modules can re-use it

JiriOndrusek:
One way of general solution might be adopting the Quarkus way by using https://github.com/cescoffier/certificate-generator

jamesnetherton:
> One way of general solution might be adopting the Quarkus way by using https://github.com/cescoffier/certificate-generator

Yeah, I saw that project mentioned on Quarkus Zulip some time ago and wondered whether we could start using it.

If possible, I think it'd be good to try and use a single tool for cert generation in the project. Currently we've got a mix of shell scripts, maven plugins etc which becomes a bit of a maintenance pain.

JiriOndrusek:
> Indeed https://github.com/cescoffier/certificate-generator would be great to replace them all. I did not know it exist, thanks for the pointer. IIRC, we were forced to use openssl for writing PKCS12 stores, because Java's `java.security.KeyStore` did not support some sort of store encoding required by FIPS. Looking at https://github.com/cescoffier/certificate-generator/blob/main/certificate-generator/src/main/java/me/escoffier/certs/CertificateUtils.java#L200-L270 it seems that it is using `java.security.KeyStore` for writing the stores. But you mentioned recently @JiriOndrusek that JKS works on FIPS too? I wonder whether we need to try to fix certificate-generator's PKCS12 impl to comply with FIPS?

With one of the recent java updates (I'm not sure which concrete build it was), certificates generated by the java keystore  work in FIPS environment. (I noticed it several weeks ago) Therefore the certificate-generator project should cover all our needs.

I'm currently trying to solve FIPS for kafka tests and I'm using the certificate-generator, which seems to be easily usable. Once I'm sure that all works, I'll ping you on the PR to discuss whether we want to use such approach on the whole project.

JiriOndrusek:
I tried generator-projext in kafka and all looks good.
My second attempt would be cxf-soap to see whether the complex generation can be achieved by the project.

PR #5968, opened 2024-04-05
Http: use FIPS complaiant keystore and truststore

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/5966


JiriOndrusek:
I add *.pkcs12 and *.ext into ignored files for license check.

integration-test-groups/http/common/generate-certs.sh
```diff
@@ -0,0 +1,67 @@
...
+mkdir -p "$workDir"
+mkdir -p "$destinationDir"
+
+# Certificate authority
+openssl genrsa -out "$workDir/ca.key" $keySize
+openssl req -x509 -new -subj '/O=apache.org/OU=eng (NOT FOR PRODUCTION)/CN=ca' -key "$workDir/ca.key" -nodes -out "$workDir/ca.pem" -days $days -extensions v3_req
+openssl req -new -subj '/O=apache.org/OU=eng (NOT FOR PRODUCTION)/CN=ca' -x509 -key "$workDir/ca.key" -days $days -out "$workDir/ca.crt"
+
+for actor in localhost; do
+  # Generate keys
```
aldettinger, line 53: So actor = localhost ? What's the purpose of the loop then ?

ppalaga: It is copied from a place where there were more actors. The loop could be removed, but OTOH, maybe we'll want to test mTLS one day?

ppalaga:
I'd personally vote for not having those certs checked in in git but rather executing the script as a part of Maven build as we do in QCXF

https://github.com/quarkiverse/quarkus-cxf/blob/e7cb0f50eceb03af8adea31c5403dd8b5554e2b7/integration-tests/mtls/pom.xml#L78-L96

In that way, we can stay sure that the script works and it will also keep working 10001 days from now.

ppalaga:
> > executing the script as a part of Maven build
> 
> How would that work on Windows?

It works flawlessly on GH Actions Windows hosts, when shell is set to bash and openssl is installed (no idea how to install it, it's there OOtB on GHA Windows runners). The same works locally for sure. Is that a satisfactory answer?

jamesnetherton:
> Is that a satisfactory answer?

Yes it probably uses WSL, which anyone can install.

zhfeng (approved):

integration-test-groups/http/common/generate-certs.sh
JiriOndrusek: Exactly, I copied the code from cxf and it looked better to keep the loop in case we need to ad generation of more certificates

JiriOndrusek:
> I'd personally vote for not having those certs checked in in git but rather executing the script as a part of Maven build as we do in QCXF

Yes., I agree, there is a small hickup, that several modules extend the `common` module, therefire the code for generation hasa to be present several times. I'd like to think about a nicer solution to do not duplicate code. <therefore I created an issue to improve this thing - https://github.com/apache/camel-quarkus/issues/5967  and I forgot to linkl those issues together. @ppalaga @jamesnetherton should I apply the generation (with duplicated code), or is it ok to wait the the improve?

JiriOndrusek:
TBH I even thought aboyt placing the generation script in one place in QC repository and use it via several modules. (In theory that should work and would be nicer in my POV) - to not have copies of  the generation scripts in several modules

jamesnetherton:
>  @ppalaga @jamesnetherton should I apply the generation (with duplicated code), or is it ok to wait the the improve?

I have no strong opinion about it. But at some point we should probably discuss our strategy for certificate generation. IMO we should try to make it uniform across the entire project and have it so we can override things like the CN, validity period etc.

integration-test-groups/http/common/generate-certs.sh
aldettinger: ok, plus one day, this kind of script might be used for other extensions and we would like to keep it generic.

zhfeng:
Can we open a new issue for the certification generation?

JiriOndrusek:
> Can we open a new issue for the certification generation?

I already created the issue (a few days ago) - https://github.com/apache/camel-quarkus/issues/5967
So I think that we can merge this PR and I'll continue using the new ticket)

PR #5978, opened 2024-04-08
Remove findbugs exclusion from quarkus-grpc-common in gRPC extension

jamesnetherton:


aldettinger (approved):
so, they don't break the enforcer check anymore ? should we then remove it from there https://github.com/apache/camel-quarkus/blob/main/extensions/qdrant/runtime/pom.xml#L41 too ?

jamesnetherton:
> should we then remove it

I think it depends on the findbugs usage. If a qdrant app can run ok in dev mode, then it's probably fine to leave things as-is. 

For gRPC it fails in dev mode with the exclusion because the grpc-java codebase is littered with `javax.annotation.CheckReturnValue` in various places.

zhfeng (approved):

aldettinger:
> > should we then remove it
> 
> I think it depends on the findbugs usage. If a qdrant app can run ok in dev mode, then it's probably fine to leave things as-is.
> 
> For gRPC it fails in dev mode with the exclusion because the grpc-java codebase is littered with `javax.annotation.CheckReturnValue` in various places.

Ok, looks to run good. Many thanks for the tip @jamesnetherton

Issue #5979, opened 2024-04-08
Cxf-soap: tests are not working in FIPS environment


JiriOndrusek:
### Bug description

When running tests from `cxf-soap-grouped` in FIPs environment, there are errors:

- [ERROR]   CxfWsTrustTest.programmaticSts:107 » SOAPFault unsupported key transport encryption algorithm: No such algorithm: "RSA/ECB/OAEPWithSHA1AndMGF1Padding"
- [ERROR]   CxfWsTrustCxfWayTest.programmaticSts:112 » SOAPFault unsupported key transport encryption algorithm: No such algorithm: "RSA/ECB/OAEPWithSHA1AndMGF1Padding"
- [ERROR]   CxfWssSecurityPolicyServerCxfWayTest.encrypetdSigned:53 » SOAPFault unsupported key transport encryption algorithm: No such algorithm: "RSA/ECB/OAEPWithSHA1AndMGF1Padding"
- [ERROR]   CxfWssSecurityPolicyServerTest.encrypetdSigned:52 » SOAPFault unsupported key transport encryption algorithm: No such algorithm: "RSA/ECB/OAEPWithSHA1AndMGF1Padding"

PR #5980, opened 2024-04-08
Cxf-soap: tests are not working in FIPS environment

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/5979

@ppalaga I thought, that the regenaration of keystores + some additional logic handling FIPS in tests would be needed (similar to https://github.com/quarkiverse/quarkus-cxf/pull/1297 and https://github.com/quarkiverse/quarkus-cxf/pull/1174). With the current simple fix the test suite works in FIPS environment. 

- There are no tests verifying that non-FIPS configuration fails -> which is the reason of not requiring special handling in the  test, so it is OK
- I'm not sure about the keystores though:

1. On one side - keystores can not be used "fully" to encrypt a message, because the proces should fail in FIPS
2. On the other side, it is not necessary to test such usecase, because it is covered by Quarkus-cxf, which is leveraged here.

WDYT? 

a. Should we keep the fix this small? (I think so)
b. Should we consider making the tests more covering (to use keystores properly)? (I'm not sure of it)

### DIFF
--- a/integration-test-groups/cxf-soap/cxf-soap-ws-security-server/src/main/resources/encrypt-sign-policy.xml
+++ b/integration-test-groups/cxf-soap/cxf-soap-ws-security-server/src/main/resources/encrypt-sign-policy.xml
@@ -45,7 +45,7 @@
                     </sp:RecipientToken>
                     <sp:AlgorithmSuite>
                         <wsp:Policy>
-                            <sp:Basic256/>
+                            <sp:CustomAlgorithmSuite/>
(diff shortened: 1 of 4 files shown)


JiriOndrusek:
Failure of CI (MongoDB Grouped and Debezium) is not caused by this change

zhfeng (approved):

jamesnetherton:
@JiriOndrusek I assume this PR is still needed? If so, please rebase.

PR #6012, opened 2024-04-19
Add Wasm extension

jamesnetherton:


oscerd (approved):

integration-tests/wasm/rust/Cargo.lock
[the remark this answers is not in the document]
jamesnetherton: Depends if we want reproducable builds I guess.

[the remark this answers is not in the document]
zhfeng: I don't understand. It is used by Cargo to do the rust build, is it right?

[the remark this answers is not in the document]
jamesnetherton: Yeah - it's like package.lock in the JavaScript world or go.sum with Go. It records the state of the downloaded packages, their checksums etc. So when someone else runs the build, they get an identical result (in theory).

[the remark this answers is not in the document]
zhfeng: OK, I get it.

integration-tests/wasm/rust/build.sh
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -euxo pipefail
+
+rustup target add wasm32-unknown-unknown
+
+cargo build --target wasm32-unknown-unknown --release
```
zhfeng, line 23: what is about `wasm32-unknown-unknown`? Should user change it on different arch platform?

jamesnetherton: It basically means use a 32Bit address space and compile & execute on any target (I think).

I tested on Mac OS (aarch64) and Linux (x86_64) and it's working fine. The CI build will tell us the state of things on Windows.

extensions/wasm/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,32 @@
...
+---
+name: "Camel Wasm"
+description: "Invoke Wasm functions"
+metadata:
+  icon-url: "https://raw.githubusercontent.com/apache/camel-website/main/antora-ui-camel/src/img/logo-d.svg"
+  guide: "https://camel.apache.org/camel-quarkus/latest/reference/extensions/wasm.html"
+  categories:
+  - "integration"
+  status:
+  - "stable"
```
lburgazzoli, line 32: As today the [component support level is experimental](https://github.com/apache/camel/blob/7676ee4dec59ac6ff6b7069c8ac42686212bd952/components/camel-wasm/pom.xml#L37), I don't recall how it maps to extensions status so wonder if this is correct

jamesnetherton: Thanks - I'll fix it up after the CI build completes.

zhfeng: Hmm, I think it should be `experimental` as well in extensions.

integration-tests/wasm/rust/build.sh
jamesnetherton: The Windows job passed.

extensions/wasm/runtime/src/main/resources/META-INF/quarkus-extension.yaml
jamesnetherton: It's updated now.

The maven plugins that determine the status probably need to be made smarter and interrogate the component metadata.

I can create a follow up issue for that.

zhfeng: Yeah, this file could be generated by `cq-maven-plugin`.

zhfeng (approved):

PR #6023, opened 2024-04-22
Fixup Milvus JVM only state

jamesnetherton:
Just for if / when we add native support. The bits will be in the correct state for the maven tooling to tweak.


oscerd (approved):
Thanks and sorry. I probably missed the steps in the guide.

Issue #6029, opened 2024-04-23
Jt400: possible missing resource in the native


JiriOndrusek:
### Bug description

This is a reminder about a behavior I noticed with JT400 native run. 

If you add following code into resource:
```
            Jt400Component comp = context.getComponent("jt400", Jt400Component.class);
            comp.stop();
            comp.getConnectionPool().close();
```
The execution of  ` comp.getConnectionPool().close();` should fail in the native because o f missing resource. I suspect that a bundle with error description fails to load.

PR #6030, opened 2024-04-23
Jt400: possible missing resource in the native

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/6029

### DIFF
--- a/extensions/jt400/deployment/src/main/java/org/apache/camel/quarkus/component/jt400/deployment/Jt400Processor.java
+++ b/extensions/jt400/deployment/src/main/java/org/apache/camel/quarkus/component/jt400/deployment/Jt400Processor.java
@@ -18,6 +18,7 @@ package org.apache.camel.quarkus.component.jt400.deployment;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.ListResourceBundle;
 import java.util.regex.Pattern;
 
 import com.ibm.as400.access.AS400;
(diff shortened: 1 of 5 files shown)


extensions/jt400/deployment/src/main/java/org/apache/camel/quarkus/component/jt400/deployment/Jt400Processor.java
```diff
@@ -86,6 +89,16 @@ void reflectiveClasses(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses
 
     }
 
+    @BuildStep
+    void resourceBundles(BuildProducer<NativeImageResourceBundleBuildItem> imageResourceBundles,
+            CombinedIndexBuildItem combinedIndex) {
+        IndexView index = combinedIndex.getIndex();
+
+        index.getAllKnownSubclasses(LIST_RESOURCE_BUNDLE_NAME).stream()
```
jamesnetherton, line 97: Probably best to filter on a package prefix of `com.ibm.as400`.

Otherwise we register everything on the classpath extending `ListResourceBundle`. Which is probably not what we want.

zhfeng:
@JiriOndrusek is it possible to add an unit test?

extensions/jt400/deployment/src/main/java/org/apache/camel/quarkus/component/jt400/deployment/Jt400Processor.java
JiriOndrusek: fixed

JiriOndrusek:
> @JiriOndrusek is it possible to add an unit test?

I'm thinking about that, I can try. Although the test would cover a scenario which does not make sense

JiriOndrusek:
@zhfeng test is added

JiriOndrusek:
@zhfeng I have to check again, whether the added test can not affect existing one when executed in different order

JiriOndrusek:
> @zhfeng I have to check again, whether the added test can not affect existing one when executed in different order

I run the tests ordered with new method at the first place and at last place and there was no issue.

JiriOndrusek:
@jamesnetherton As I think about the test coverage, it makes sense to me to move the test into  `jt400-mocked`, because it does not require a real service and it would be executed by CI. (of course U=I have to try , whether it is possible). WDYT?

jamesnetherton:
> move the test into `jt400-mocked`

Sounds ok to me. Or if we need to have it duplicated in both mocked and non-mocked, that's fine too.

JiriOndrusek:
> > move the test into `jt400-mocked`
> 
> Sounds ok to me. Or if we need to have it duplicated in both mocked and non-mocked, that's fine too.

Duplication (even if should be avoided) sounds like the right solution
.

JiriOndrusek:
I added the new test intojt400-mocked (as duplication)
Therefore it should be fulfilled that at leas one of `jt400` and `jt400-mocked` is executed.

PR #6042, opened 2024-04-25
[LRA] Fix test when running with DOCKER_HOST

avano:
please also backport to 3.8

### DIFF
--- a/integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
+++ b/integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
@@ -58,7 +58,18 @@ public class LraTestResource implements QuarkusTestResourceLifecycleManager {
                 lraPort = container.getMappedPort(LRA_PORT);
             }
 
+            // If the test runs with a remote docker, it needs to know the IP of the machine where the test app runs
+            if (System.getenv("DOCKER_HOST") != null) {
+                hostname = System.getProperty("lra.appNode.ip");
+                if (hostname == null) {
+                    throw new RuntimeException(
+                            "You need to supply application node IP when running LRA test with a remote docker instance using lra.appNode.ip property");
+                }
+            }
+
             return CollectionHelper.mapOf(
+                    // In case of remote docker, expose the endpoint at 0.0.0.0, as the default is "localhost" only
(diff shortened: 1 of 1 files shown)


integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java
```diff
@@ -58,7 +58,18 @@ public Map<String, String> start() {
                 lraPort = container.getMappedPort(LRA_PORT);
             }
 
+            // If the test runs with a remote docker, it needs to know the IP of the machine where the test app runs
+            if (System.getenv("DOCKER_HOST") != null) {
+                hostname = System.getProperty("lra.appNode.ip");
```
jamesnetherton, line 63: Just curious - what actually uses the `lra.appNode.ip` property? I don't seen any other references to it.

avano: the `hostname` variable [is used](https://github.com/avano/camel-quarkus/blob/lra/integration-tests/lra/src/test/java/org/apache/camel/quarkus/component/lra/it/LraTestResource.java#L76) in the `camel.lra.local-participant-url` property

jamesnetherton: Ah I totally missed that, thanks.

So, instead of having to manually pass along a system property, would it be enough to use `DockerClientFactory.instance().dockerHostIpAddress()`?

avano: this is not the IP where a docker container runs (rhel sidecar), but the IP where the test application runs (windows). the LRA coordinator runs in the docker container on the remote machine and it needs to be able to invoke the endpoint from the test application

while it is possible to get an IP automatically, it may not work in all envs correctly - for example in our openstack it needs to use an internal openstack ip, as the machines are not reachable using public ips, so I assumed passing the IP manually would be the safest option

jamesnetherton: Ok, lets leave it as-is then. Thanks for the explanation 👍

zhfeng (approved):

PR #6082, opened 2024-05-14
Update license-maven-plugin to 4.5

zhfeng:


jamesnetherton (approved):
Thanks!

License checks passed so we should be good to merge.

Issue #6090, opened 2024-05-15
Kafka tests fails in FIPS environment


JiriOndrusek:
### Bug description

kafka-ssl, kafka-sasl, kafla-sasl-ssl fail in FIPS environment.

PR #6091, opened 2024-05-15
kafka-ssl in fips

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/6090
contains part of https://github.com/apache/camel-quarkus/issues/5967

-  **kafka** test module works in FIPS
- **kafka-sasl** uses plaintext sasl, therefore can not work in FIP (sasl scenario for FIPS is covered by kafka-sasl-ssl), the test is **disabled** in FIPS environment
- **kafka-ssl** and **kafka-sasl-ssl** are both fixed by this PR

### Workaround because of strimzi kafka container. 

Password-based encryption support in FIPs mode was implemented in the Red Hat build of OpenJDK 17 update 4.
OpenJdk17 should be present in the image according to this [blog post](https://strimzi.io/blog/2023/01/25/running-apache-kafka-on-fips-enabled-kubernetes-cluster/). Unfortunately, even the newest image ([3.7.0](https://quay.io/repository/strimzi-test-container/test-container?tab=tags&tag=latest-kafka-3.2.1%20bash)) uses java 11.
I had to add a step in the `camel-quarkus-integration-tests-support-kafka` module when I was creating a custom image with OpenJDK 17. This workaround happens only in FIPS mode and only once. (The custom image with Java 17 is cached and the next runs should use it)

### Adaptation of certificate-generator project

[Certificate-generator](https://github.com/cescoffier/certificate-generator/tree/main) project brings support for the generation of certificates easily and clearly. The project can be used as a common approach to certificate generation for the CQ (see discussion in this [ticket](https://github.com/apache/camel-quarkus/issues/5967))

CQ requires the certificates to be customized in case the external docker host is used (see the [code](https://github.com/apache/camel-quarkus/blob/3.10.0/integration-tests-support/kafka/src/main/java/org/apache/camel/quarkus/test/support/kafka/KafkaTestSupport.java#L63-L93) for kafka).
To achieve the same functionality when using certificate-generator project I had to create a special annotation `TestCertificates` with the extension implementation (`TestCertificateGenerationExtension`). The extension is based on the original extension brought by certificate-generator project and customizes _cn_ and _SubjectAlternativeName_ in case the docker host is not localhost.

The new annotation (`TestCertificates`) is placed into a new module **camel-quarkus-integration-tests-support-certificate**


integration-tests/kafka-sasl-ssl/src/test/java/org/apache/camel/quarkus/kafka/sasl/KafkaSaslSslTest.java
```diff
@@ -22,10 +22,17 @@
...
 import io.restassured.RestAssured;
 import io.restassured.path.json.JsonPath;
+import me.escoffier.certs.Format;
+import me.escoffier.certs.junit5.Certificate;
+import org.apache.camel.quarkus.test.support.certificate.TestCertificates;
 import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+@TestCertificates(certificates = {
```
JiriOndrusek, line 32: @ppalaga This annotation instructs certificate-generator project to generate certificates.
(it should be possible to test more certificate types, by providing some parameters for the test methods. I haven't tried that yet)

oscerd (approved):

integration-tests-support/kafka/pom.xml
```diff
@@ -52,6 +52,14 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-junit4-mock</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel</groupId>
+            <artifactId>camel-util</artifactId>
```
jamesnetherton, line 57: Do we need `camel-util`?

integration-tests-support/test-support/src/main/java/org/apache/camel/quarkus/test/FipsModeUtil.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package org.apache.camel.quarkus.test;
+
+import java.security.Provider;
+import java.security.Security;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+public class FipsModeUtil {
```
jamesnetherton, line 25: Should have a private constructor if it's a utility class

integration-tests/kafka-ssl/pom.xml
```diff
@@ -78,6 +78,10 @@
             <artifactId>quarkus-junit4-mock</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-integration-tests-support-certificate</artifactId>
```
jamesnetherton, line 83: Should be `<scope>test</scope>`?

pom.xml
```diff
@@ -86,6 +86,7 @@
         <bcprov-ext-jdk18on.version>1.78</bcprov-ext-jdk18on.version><!-- TODO: Remove this. Required as there is no 1.78.1 version released -->
         <brotli.version>0.1.2</brotli.version><!-- @sync org.apache.httpcomponents.client5:httpclient5-parent:${httpclient5.version} prop:brotli.version -->
         <caffeine.version>3.1.5</caffeine.version><!-- @sync io.quarkus:quarkus-bom:${quarkus.version} dep:com.github.ben-manes.caffeine:caffeine -->
+        <certificate.generator.version>0.5.0</certificate.generator.version>
```
jamesnetherton, line 89: Nitpick - Can you move this with the other test dependency properties:

https://github.com/apache/camel-quarkus/blob/7eebe0453ff401cc083f838aa247ed285688d800/pom.xml#L171-L180

integration-tests-support/certificate/src/main/java/org/apache/camel/quarkus/test/support/certificate/TestCertificateGenerationExtension.java
```diff
@@ -0,0 +1,143 @@
...
+            }
+
+            certificateFiles.addAll(generator.generate(request));
+        }
+    }
+
+    private Optional<String> resolveDockerHost() {
+        String dockerHost = DockerClientFactory.instance().dockerHostIpAddress();
+        if (!dockerHost.equals("localhost") && !dockerHost.equals("127.0.0.1")) {
+            String imageName = ConfigProvider.getConfig().getValue("eclipse-temurin.container.image", String.class);
```
jamesnetherton, line 118: The stuff that was using `eclipse-temurin.container.image` I assume is not needed anymore given that the cert generator will take care of what it was being used for?

So maybe any references to this method can just be replaced with `DockerClientFactory.instance().dockerHostIpAddress()`?

integration-tests-support/kafka/src/main/java/org/apache/camel/quarkus/test/support/kafka/KafkaTestResource.java
```diff
@@ -18,38 +18,96 @@
...
+                    //to customize the image.
+                    // This workaround can be removed once the strimzi container with openjdk 17 is released.
+                    // According to https://strimzi.io/blog/2023/01/25/running-apache-kafka-on-fips-enabled-kubernetes-cluster/
+                    // image should exist
+                    j17container = new GenericContainer(
+                            new ImageFromDockerfile(customImageName, false)
+                                    .withDockerfileFromBuilder(builder -> builder
+                                            .from("quay.io/strimzi-test-container/test-container:latest-kafka-3.2.1")
+                                            .env("JAVA_HOME", "/usr/lib/jvm/jre-17")
+                                            .env("PATH", "/usr/lib/jvm/jre-17/bin:$PATH")
```
jamesnetherton, line 82: Any idea if there's any open issues with Strimzi to see if they can switch to JDK 17 in their container images?

If not, we should maybe create one.

jamesnetherton:
If possible, it might be a good idea to test that this works ok in the Quarkus Platform before we merge this work.

integration-tests-support/certificate/src/main/java/org/apache/camel/quarkus/test/support/certificate/TestCertificateGenerationExtension.java
JiriOndrusek: I started container containe to access method `container.getHost()`. I was not sure if `DockerClientFactory.instance().dockerHostIpAddress()` returns the same value. Probably yes, in that case I'll remove the container.

integration-tests-support/kafka/src/main/java/org/apache/camel/quarkus/test/support/kafka/KafkaTestResource.java
JiriOndrusek: I found this blog post, which indicates, that the issue was fixed - https://strimzi.io/blog/2023/01/25/running-apache-kafka-on-fips-enabled-kubernetes-cluster/
But I tried the newest image and the java is still 11. So I'm not sure what is wrong.

pom.xml
JiriOndrusek: sure, I missed the separation for test versions

integration-tests/kafka-ssl/pom.xml
JiriOndrusek: yes, my bad, test scope should be there

integration-tests-support/kafka/pom.xml
JiriOndrusek: I used `CollectionHelper` from `camel-util` and you are right, I can refactor the code to not use it, therefore I can remove the dependency

JiriOndrusek:
> If possible, it might be a good idea to test that this works ok in the Quarkus Platform before we merge this work.

I suppose that running platform with CQ locally should be enough, right?

JiriOndrusek:
I also noticed, that I have to do small fix in the  `TestCertificateGenerationExtension`, because the docker customization is not necessary for each certificate generation task (so I plan to ass a parameter to annotation `TestContainers`, so I would test docker only in necessary cases

jamesnetherton:
> I suppose that running platform with CQ locally should be enough, right?

Yeah, you can just tweak the CQ version in the platform to 3.11.0-SNAPSHOT, build the project and run a few of the CQ tests.

integration-tests-support/kafka/src/main/java/org/apache/camel/quarkus/test/support/kafka/KafkaTestResource.java
jamesnetherton: Here's the reason I guess:

https://github.com/strimzi/test-container-images/blob/main/images/base/Dockerfile#L6

JiriOndrusek: I'll report the issue to the strimzi and propose a fix. We can remove workaround once the issue is fixed

JiriOndrusek: Thanks for the link!

JiriOndrusek: raised a ticket https://github.com/strimzi/test-container-images/issues/25 and proposed a fix https://github.com/strimzi/test-container-images/pull/26.
Here is a ticket for removing the workaround https://github.com/apache/camel-quarkus/issues/6097

integration-tests-support/kafka/pom.xml
JiriOndrusek: Removed

integration-tests/kafka-ssl/pom.xml
JiriOndrusek: added

integration-tests-support/test-support/src/main/java/org/apache/camel/quarkus/test/FipsModeUtil.java
JiriOndrusek: added

pom.xml
JiriOndrusek: moved

JiriOndrusek:
> > I suppose that running platform with CQ locally should be enough, right?
> 
> Yeah, you can just tweak the CQ version in the platform to 3.11.0-SNAPSHOT, build the project and run a few of the CQ tests.

I built quarkus-platform with CQ 3.11.0-SNAPSHOT and successfully run all 4 kafka test modules.

jamesnetherton (approved):
I'm ok with merging this and using it as a base to build on & improve to use the cert generator in other places in the project.

JiriOndrusek:
@jamesnetherton  The support module was renamed to `camel-quarkus-integration-tests-support-certificate-generator`

PR #6098, opened 2024-05-16
Use more correct convention with configuring datasource

llowinge:


jamesnetherton:
Seems this failed.

I think we should just remove any `quarkus.datasource.postgres` config and just have `quarkus.datasource.jdbc.telemetry=true`. 

That should be guaranteed to work everywhere.

llowinge:
@jamesnetherton Done.

jamesnetherton (approved):

PR #6130, opened 2024-05-29
Fix #5824 to add camel trace config properties

### LINKED ISSUE #5824: Camel 4.5 - Trace config
### Describe the feature here

The backlog tracer has its configuration moved into its own group like the debug already have
https://issues.apache.org/jira/browse/CAMEL-20472

Its probably easier to see what we did for CSB to learn what may need to be done in CEQ
https://github.com/apache/camel-spring-boot/commit/3e8de47443debcf6c1372f53312961fbd6e0be37

There is a `DebugConfig.java` in CEQ today, so maybe a `TraceConfig.java` needs to be created. As trace is very similar to how debug config is.

zhfeng:
Fixes #5824


extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelContextRecorder.java
```diff
@@ -107,4 +108,34 @@ public void registerManagementEventBridge(RuntimeValue<CamelContext> camelContex
...
+            tracer.setRemoveOnDump(config.trace.removeOnDump);
+            tracer.setBodyMaxChars(config.trace.bodyMaxChars);
+            tracer.setBodyIncludeStreams(config.trace.bodyIncludeStreams);
+            tracer.setBodyIncludeFiles(config.trace.bodyIncludeFiles);
+            tracer.setIncludeExchangeProperties(config.trace.includeExchangeProperties);
+            tracer.setIncludeExchangeVariables(config.trace.includeExchangeVariables);
+            tracer.setIncludeException(config.trace.includeException);
+            tracer.setTraceRests(config.trace.traceRests);
+            tracer.setTraceTemplates(config.trace.traceTemplates);
+            tracer.setTracePattern(config.trace.tracePattern.orElse(""));
```
davsclaus, line 135: these should be null instead of empty string

zhfeng: The same with `setTraceFilter`. it should be `null` as well?
```java
tracer.setTraceFilter(config.trace.traceFilter.orElse(""));
```

zhfeng: @davsclaus set with `null` on both of them.

davsclaus: yes both

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
```diff
@@ -406,4 +412,96 @@ public static class EventBridgeConfig {
...
+        public boolean includeExchangeProperties;
+
+        /**
+         * Whether to include the exchange variables in the traced message
+         */
+        @ConfigItem(defaultValue = "true")
+        public boolean includeExchangeVariables;
+
+        /**
+         * Trace messages to include exception if the message failed
```
aldettinger, line 478:
```suggestion
         * Whether to include the exception in the traced message in case of failed exchange.
```

extensions-core/core/deployment/src/test/java/org/apache/camel/quarkus/core/runtime/CamelTraceTest.java
```diff
@@ -0,0 +1,94 @@
...
+        assertThat(tracer.getBacklogSize()).isEqualTo(100);
+        assertThat(tracer.getTracePattern()).isNull();
+        assertThat(tracer.getTraceFilter()).isNull();
+    }
+
+    private BacklogTracer getBacklogTracer(CamelContext camelContext) {
+        BacklogTracer tracer = null;
+        if (registry != null) {
+            // lookup in registry
+            Map<String, BacklogTracer> map = registry.findByTypeWithName(BacklogTracer.class);
```
aldettinger, line 84: What happen if we don't look in the registry ? Shouldn't the tracer be accessible from `getContextPlugin` always ? Maybe, this is related to stand by mode ?

zhfeng: Hmm, I borrow some codes from [DefualtChannel::getOrCreateBacklogTracer()](https://github.com/apache/camel/blob/main/core/camel-base-engine/src/main/java/org/apache/camel/impl/engine/DefaultChannel.java#L297-L320), @davsclaus any idea?

aldettinger: The history lead up to camel 3 refactoring, might be there since even longer. As we are in line with camel-core should be fine to validate the pr.

Still it would be interesting to get the explanation if anyone remember :)

PR #6191, opened 2024-06-14
Use LazySecretKeysHandler for Jasypt Config

### LINKED ISSUE #6190: Jasypt SecretKeyHandler config cannot use configuration coming from Configurable Config sources
### Bug description

From https://github.com/quarkusio/quarkus/issues/40767

radcortez:
- Fixes #6190

This uses `LazySecretKeysHandler` to allow the Jasypt `SecretKeyHandler` to only initialize when a value requires the handler. This also allows for Configurable Config sources to contribute configuration to the handler.

Because we don't know when the handler is going to be initialized, we create a separate instance of the Encryptor for the DevUI.

### DIFF
--- a/extensions/jasypt/deployment/src/test/java/org/apache/camel/quarkus/component/jasypt/JasyptPasswordNotProvidedTest.java
+++ b/extensions/jasypt/deployment/src/test/java/org/apache/camel/quarkus/component/jasypt/JasyptPasswordNotProvidedTest.java
@@ -17,6 +17,9 @@
 package org.apache.camel.quarkus.component.jasypt;
 
 import io.quarkus.test.QuarkusUnitTest;
+import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.component.jasypt.JasyptPropertiesParser;
+import org.apache.camel.component.properties.PropertiesComponent;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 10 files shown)


zhfeng (approved):

oscerd (approved):

extensions/jasypt/deployment/src/test/java/org/apache/camel/quarkus/component/jasypt/JasyptPasswordNotProvidedTest.java
```diff
@@ -25,11 +28,27 @@
...
             .setExpectedException(IllegalStateException.class)
-            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class).addClass(JasyptRoutes.class));
 
     @Test
     void passwordNotProvidedThrowsException() {
         // Nothing to test as we just verify the application fails to start
     }
+
+    public static final class JasyptRoutes extends RouteBuilder {
```
jamesnetherton, line 40: Just curious why we add `RouteBuilder` classes to tests that do nothing other that assert the application fails to start correctly?

Can we remove them from here and the other tests they got added to?

jamesnetherton (commented):
Thanks so much for this @radcortez.

I left a question about some of the test changes.

extensions/jasypt/deployment/src/test/java/org/apache/camel/quarkus/component/jasypt/JasyptPasswordNotProvidedTest.java
radcortez: Because the secret key is now lazy, it does not initialize when the Config starts (on application start), but when it is called. Without the route, the application starts fine (instead of the expected exception) because the secret handle is never invoked, so we need to force it.

jamesnetherton (approved):

PR #6216, opened 2024-06-20
Log the Camel Quarkus version on startup

jamesnetherton:
I opened this in draft to get some feedback.

For micro release of Quarkus, we may end with something like this in the console logs:

```
INFO  [org.apa.cam.qua.cor.CamelBootstrapRecorder] (main) Apache Camel Quarkus 3.12.0 is starting
...
INFO  [io.quarkus] (main) my-cool-app 1.0.0 on JVM (powered by Quarkus 3.12.3)
```

Do folks think this could be a source of confusion for users? That the CQ version is potentially out of alignment compared to the version of the platform BOM.

### DIFF
--- a/extensions-core/core/deployment/pom.xml
+++ b/extensions-core/core/deployment/pom.xml
@@ -94,6 +94,12 @@
     </dependencies>
 
     <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+            </resource>
+        </resources>
         <plugins>
             <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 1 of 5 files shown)


zbendhiba:
@jamesnetherton The PR alignes the Camel Quarkus version to minor quarkus platform version, is that correct?

jamesnetherton:
> The PR alignes the Camel Quarkus version to minor quarkus platform version, is that correct?

The CQ version displayed in the logs will just be whatever the platform is [aligned](https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L61) with. So it may not match the platform version perfectly.

zhfeng:
Well, I doubt that this could confuse tht users, there are camel version and quarkus version, now camel-quarkus version. How about show all of these version one line?

```
(main) Apache Camel Quarkus 3.12.0 (based on Camel 4.6.0 and Quarkus 3.12.0) is starting
```

aldettinger:
+1 for the proposal from @zhfeng.

jamesnetherton:
Thanks for the feedback all. Seems the general idea is acceptable.

I'll stick with my original proposal of `Apache Camel Quarkus 3.12.0 is starting` initially. We can always revisit and update it in future.

Issue #6088, opened 2024-05-15
Crypto does not work in FIPS


JiriOndrusek:
### Bug description

Crypto extension does not work in FIPS  enabled system.
I'm disabling the tests in fips, please enable them once fixed.

PR #6241, opened 2024-06-28
Extracion of crypto-pgp and making crypto work on FIPS

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/6088

Crypto component was split into `crypto` and `crypto-pgp` to support FIPS. See the [commit](https://github.com/apache/camel/pull/14651/commits/2161144a8c06b23732aa8947e02e738a251bdef8). (this is the reason of why this PR is opened against `camel-main`)

- `crypto-pgp` is not registering BC as a provider; component contains only PGPDataFormat
- `crypto` contains all other (CryptoDataFormat, components)

**How to run `crypto` on FIPS** the BC dependency has to be excluded and BCFIPS added (or another BC implementation for FIPS) .

**Limitation:** Because of the BCFIPS, it is not possible to use crypto and crypto-pgp together on FIPS system (if BCFIPS is utilized)

This PR contains:

- new extension based on `crypto-pgp` component (uses BC). The component and all tests are extracted from the `crypto` component. All tests work on FIPS system.
- fixes in bouncycastle-support to work correctly on FIPS system
-  added `fips` profile in crypto integration test module, to allow execution on FIPS systems
-  some *.adoc addition for FIPS in crypto and crypto-pgp etensions
- it is not possible to utilize `certificate-generator-support` because the tests require `DES` and the `certificate-generator-support` uses `RSA`.


JiriOndrusek:
@ppalaga FYI

aldettinger:
The failure seems merely linked to uncommitted changes.

docs/modules/ROOT/pages/reference/extensions/crypto-pgp.adoc
```diff
@@ -0,0 +1,55 @@
...
+endif::[]
+
+[id="extensions-crypto-pgp-camel-quarkus-limitations"]
+== Camel Quarkus limitations
+
+[id="extensions-crypto-pgp-limitations-fips"]
+=== FIPS
+
+It may not be possible to run `crypto` and `crypto-pgp` extensions together on FIPS enabled system.
+For example if `crypro` uses `BCFIPS` provider and `crypto-pgp` uses `BC` provider, it is no possible to have bot providers on one classpath.
```
aldettinger, line 54:
```suggestion
For example if `crypto` uses `BCFIPS` provider and `crypto-pgp` uses `BC` provider, it is not possible to have both providers on the same classpath.
```

docs/modules/ROOT/pages/reference/extensions/crypto.adoc
```diff
@@ -45,6 +45,51 @@ ifeval::[{doc-show-user-guide-link} == true]
...
+
+Extension requires BouncyCastle provider and also utilizes the quarkus security extension (see https://quarkus.io/guides/security-customization#registering-security-providers[security providers registration doc])
+If there is no `BC*` provider registered (by `quarkus.security.security-providers` property).
+The `BC` provider is registered.
+
+[id="extensions-crypto-limitations-fips"]
+=== FIPS
+
+It may not be possible to run `crypto` and `crypto-pgp` extensions together on FIPS enabled system.
+For example if `crypro` uses `BCFIPS` provider and `crypto-pgp` uses `BC` provider, it is no possible to have bot providers on one classpath.
```
aldettinger, line 90:
```suggestion
For example if `crypto` uses `BCFIPS` provider and `crypto-pgp` uses `BC` provider, it is not possible to have both providers on the same classpath.
```

extensions-support/bouncycastle/runtime/src/main/java/org/apache/camel/quarkus/support/bouncycastle/BouncyCastleRecorder.java
```diff
@@ -35,11 +34,19 @@ public class BouncyCastleRecorder {
...
             // TODO: Fix BuildStep execution order so that this is not required
             // https://github.com/apache/camel-quarkus/issues/3472
-            provider = new BouncyCastleProvider();
-            Security.addProvider(provider);
+            try {
+                provider = (Provider) Thread.currentThread().getContextClassLoader()
+                        .loadClass("org.bouncycastle.jce.provider.BouncyCastleProvider").getConstructor().newInstance();
+                Security.addProvider(provider);
+            } catch (Exception e) {
+                // Might happen if bouncy castle in not present (in FIPS system)
```
aldettinger, line 48: Then provider stays <code>null</code> on fips system ? In other words, it throws on new <code>Instance()</code> or <code>addProvider()</code> ?

extensions/crypto-pgp/runtime/src/main/doc/limitations.adoc
```diff
@@ -0,0 +1,4 @@
+=== FIPS
+
+It may not be possible to run `crypto` and `crypto-pgp` extensions together on FIPS enabled system.
+For example if `crypro` uses `BCFIPS` provider and `crypto-pgp` uses `BC` provider, it is no possible to have bot providers on one classpath.
```
aldettinger, line 4:
```suggestion
For example if `crypto` uses `BCFIPS` provider and `crypto-pgp` uses `BC` provider, it is no possible to have both providers on the same classpath.
```

extensions/crypto/runtime/src/main/doc/limitations.adoc
```diff
@@ -0,0 +1,10 @@
+=== Security Provider
+
+Extension requires BouncyCastle provider and also utilizes the quarkus security extension (see https://quarkus.io/guides/security-customization#registering-security-providers[security providers registration doc])
+If there is no `BC*` provider registered (by `quarkus.security.security-providers` property).
+The `BC` provider is registered.
+
+=== FIPS
+
+It may not be possible to run `crypto` and `crypto-pgp` extensions together on FIPS enabled system.
+For example if `crypro` uses `BCFIPS` provider and `crypto-pgp` uses `BC` provider, it is no possible to have bot providers on one classpath.
```
aldettinger, line 10:
```suggestion
For example if `crypto` uses `BCFIPS` provider and `crypto-pgp` uses `BC` provider, it is no possible to have both providers on the same classpath.
```

extensions/joor/deployment/src/main/java/org/apache/camel/quarkus/component/joor/deployment/JoorProcessor.java
```diff
@@ -119,6 +119,19 @@ void collectExpressions(JoorExpressionConfig config,
         }
     }
 
+    /**
+     * Indicates whether the jOOR expressions should be compiled at build time.
+     */
+    public static final class CompileAtBuildTime implements BooleanSupplier {
```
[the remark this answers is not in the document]
aldettinger: Actually, this is just code reordering, this change could be discarded safely.

extensions-support/bouncycastle/deployment/src/main/java/org/apache/camel/quarkus/support/bouncycastle/deployment/BouncyCastleSupportProcessor.java
```diff
@@ -54,23 +70,45 @@ ReflectiveClassBuildItem registerForReflection(CombinedIndexBuildItem combinedIn
...
+    /**
+     * Indicates whether FIPS provider is registered via quarkus.security.
+     */
+    static final class FipsProviderConfigured implements BooleanSupplier {
+        SecurityConfig securityConfig;
+
+        @Override
+        public boolean getAsBoolean() {
+            return securityConfig.securityProviders().orElse(Collections.emptySet()).stream()
+                    .filter(p -> p.toLowerCase().contains("fips")).findAny().isPresent();
```
jamesnetherton, line 95:
```suggestion
            return securityConfig.securityProviders().orElse(Collections.emptySet()).stream()
                    .anyMatch(p -> p.toLowerCase().contains("fips"));
```

integration-tests/crypto/src/test/java/org/apache/camel/quarkus/component/crypto/it/AbstractCryptoTest.java
```diff
@@ -0,0 +1,93 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import static org.apache.camel.quarkus.component.crypto.it.CryptoResource.MESSAGE;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@DisabledIfFipsMode //todo close https://github.com/apache/camel-quarkus/issues/6088
```
aldettinger, line 33: So, this comment is not needed anymore ? We just need to close #6088 ?

extensions-support/bouncycastle/runtime/pom.xml
```diff
@@ -55,7 +55,7 @@
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcpkix-jdk18on</artifactId>
         </dependency>
-    </dependencies>
+   </dependencies>
```
jamesnetherton, line 58: Nitpick - restore original indentation.

aldettinger:
That's quite a pr. Many thanks for transparently explaining the intent in the first note and putting so much useful comment in the code. It really helps one without much fips knowledge to review :)

docs/modules/ROOT/pages/reference/extensions/crypto.adoc
```diff
@@ -45,6 +45,51 @@ ifeval::[{doc-show-user-guide-link} == true]
...
+[id="extensions-crypto-usage"]
+== Usage
+[id="extensions-crypto-usage-fips"]
+=== FIPS
+
+When running `crypto` extension on FIPS enabled system the BC-FIPS has to be utilized.
+
+* Exclude all `bc*` dependencies. Example:
+```
+                    <exclusions>
```
jamesnetherton, line 57: Can we strip the large bits of leading whitespace from these code snippets? If we leave it, then it might look a bit weird if the Camel website renders it as-is.

docs/modules/ROOT/pages/reference/extensions/crypto-pgp.adoc
JiriOndrusek: > It is still not clear to me what happens when bcprov is excluded from `crypto-pgp` and replaced with BCFIPS. Have you tried that by any chance?

There might be more problems related to this question. The major one is that **bcpg** depends on **bcprov**. Class [BcKeyFingerprintCalculator](https://github.com/bcgit/bc-java/blob/main/pg/src/main/java/org/bouncycastle/openpgp/operator/bc/BcKeyFingerprintCalculator.java#L10) references [org.bouncycastle.crypto.Digest](https://github.com/bcgit/bc-java/blob/main/core/src/main/java/org/bouncycastle/crypto/Digest.java); The same class is not part of the **bcfips**. (I checked the jar downloaded by maven, and you can see it e.g in this [fork](https://github.com/tashiscool/bc-fips/tree/master/core/src/main/java/org/bouncycastle/crypto))

**Therefore it is not possible to replace bcprov with bcfips.**

- The only workaround I'm aware of is to have a different bouncycastle fips implementation, which uses different packages. In this case the **bcprov** and **custom_bcfips** can coexist in the project and should  work together)
- For illustration, the **bcprov** package `org.bouncycastle` contains 11 sub-packages; **bcfip's** `org.bouncycastle` contains 2 sub-packages. (So I expect more possible issues.)

docs/modules/ROOT/pages/reference/extensions/crypto.adoc
JiriOndrusek: fixed

extensions/crypto-pgp/runtime/src/main/doc/limitations.adoc
JiriOndrusek: fixed

extensions/joor/deployment/src/main/java/org/apache/camel/quarkus/component/joor/deployment/JoorProcessor.java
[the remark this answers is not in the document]
JiriOndrusek: discarded

extensions-support/bouncycastle/deployment/src/main/java/org/apache/camel/quarkus/support/bouncycastle/deployment/BouncyCastleSupportProcessor.java
JiriOndrusek: fixed

docs/modules/ROOT/pages/reference/extensions/crypto.adoc
JiriOndrusek: fixed

JiriOndrusek:
@ppalaga 

> It is still not clear to me what happens when bcprov is excluded from `crypto-pgp` and replaced with BCFIPS. Have you tried that by any chance?

There might be more problems related to this question. The major one is that **bcpg** depends on **bcprov**. Class [BcKeyFingerprintCalculator](https://github.com/bcgit/bc-java/blob/main/pg/src/main/java/org/bouncycastle/openpgp/operator/bc/BcKeyFingerprintCalculator.java#L10) references [org.bouncycastle.crypto.Digest](https://github.com/bcgit/bc-java/blob/main/core/src/main/java/org/bouncycastle/crypto/Digest.java); The same class is not part of the **bcfips**. (I checked the jar downloaded by maven, and you can see it e.g in this [fork](https://github.com/tashiscool/bc-fips/tree/master/core/src/main/java/org/bouncycastle/crypto))

**Therefore it is not possible to replace bcprov with bcfips.**

- The only workaround I'm aware of is to have a different bouncycastle fips implementation, which uses different packages. In this case the **bcprov** and **custom_bcfips** can coexist in the project and should  work together)
- For illustration, the **bcprov** package `org.bouncycastle` contains 11 sub-packages; **bcfip's** `org.bouncycastle` contains 2 sub-packages. (So I expect more possible issues.)

extensions-support/bouncycastle/runtime/src/main/java/org/apache/camel/quarkus/support/bouncycastle/BouncyCastleRecorder.java
JiriOndrusek: added condition for loading fips provider

docs/modules/ROOT/pages/reference/extensions/crypto-pgp.adoc
JiriOndrusek: fixed

extensions/crypto/runtime/src/main/doc/limitations.adoc
JiriOndrusek: fixed

extensions/crypto-pgp/runtime/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-crypto-pgp-parent</artifactId>
+        <version>3.13.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-crypto-pgp</artifactId>
+    <name>Camel Quarkus :: PGP :: Runtime</name>
+    <description>Encrypt and decrypt messages using Java Cryptographic Extension (JCE) and PGP.</description>
```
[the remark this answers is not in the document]
JiriOndrusek: > there is any strong relationship between Bouncy Castle PGP support and Java Cryptographic Extension

I changed the description as suggested., as it makes message clearer from mu POV/

integration-tests/crypto/src/test/java/org/apache/camel/quarkus/component/crypto/it/AbstractCryptoTest.java
JiriOndrusek: removed

aldettinger (approved):
@JiriOndrusek The ci failure seems linked to needed regeneration, beyond looks good

JiriOndrusek:
> @JiriOndrusek The ci failure seems linked to needed regeneration, beyond looks good

Yes, I forgot to regenerate POMS whe I was maintaining camel-main, hopefully now it is ok

ppalaga:
> @ppalaga
> 
> > It is still not clear to me what happens when bcprov is excluded from `crypto-pgp` and replaced with BCFIPS. Have you tried that by any chance?
> 
> There might be more problems related to this question. The major one is that **bcpg** depends on **bcprov**. Class [BcKeyFingerprintCalculator](https://github.com/bcgit/bc-java/blob/main/pg/src/main/java/org/bouncycastle/openpgp/operator/bc/BcKeyFingerprintCalculator.java#L10) references [org.bouncycastle.crypto.Digest](https://github.com/bcgit/bc-java/blob/main/core/src/main/java/org/bouncycastle/crypto/Digest.java); The same class is not part of the **bcfips**. (I checked the jar downloaded by maven, and you can see it e.g in this [fork](https://github.com/tashiscool/bc-fips/tree/master/core/src/main/java/org/bouncycastle/crypto))
> 
> **Therefore it is not possible to replace bcprov with bcfips.**

Good to know, thanks for the information!

extensions-support/bouncycastle/runtime/src/main/java/org/apache/camel/quarkus/support/bouncycastle/BouncyCastleRecorder.java
```diff
@@ -35,11 +34,28 @@ public class BouncyCastleRecorder {
...
         if (provider == null) {
             // TODO: Fix BuildStep execution order so that this is not required
             // https://github.com/apache/camel-quarkus/issues/3472
-            provider = new BouncyCastleProvider();
-            Security.addProvider(provider);
+            try {
+                provider = (Provider) Thread.currentThread().getContextClassLoader()
+                        .loadClass(SecurityProviderUtils.BOUNCYCASTLE_PROVIDER_CLASS_NAME).getConstructor().newInstance();
+                Security.addProvider(provider);
+            } catch (Exception e) {
```
ppalaga, line 47: I wonder whether it would not be better to catch something more specific here ClassNotFoundException or similar. Because Exception catches many other exceptions, such as NPE that we perhaps want to bubble up?

JiriOndrusek: My idea is that any failure cause RuntimeException with the message `Neither BC nor BCFIPS provider can be registered` + exception message. Which is correct also with i.e, NPE

JiriOndrusek: (if provider is not loaded, the execution would fail in the next block of code)

ppalaga: > My idea is that any failure would cause RuntimeException with the message `Neither BC nor BCFIPS provider can be registered` + exception message. Which is correct also with i.e, NPE

Yeah, that's what the code currently does. My question is whether "any failure" including NPEs inside the try block should be caught. If we want to detect the availability of bcprov/bcfips by loading some of its classes, then we should catch only  exceptions related to classloading. Other exceptions which may signal other unrelated error conditions in the code should IMO be left uncaught so that users see the real issue including the stack trace. Note that the stack traces of both exceptions are lost with the current implementation. That's IMO fine for the classloading exceptions (because those are kinda expected) but not fine for everything else.

JiriOndrusek: I refactored the catch block and I',m catching only the exceptions related to initialization of provider. No unexpected exception is caught now.

JiriOndrusek:
I switched this PR to draft and I'll rebase  to `main`, as soon as `Upgrade to Camel 4.7.0 is merged` (https://github.com/apache/camel-quarkus/pull/6264)

extensions/crypto/runtime/src/main/doc/usage.adoc
```diff
@@ -0,0 +1,23 @@
+=== FIPS
+
+When running `crypto` extension on FIPS enabled system the BC-FIPS has to be utilized.
```
ppalaga, line 3: I think that now after the module split, the statement "BC-FIPS has to be utilized" is too strong. I would say "any FIPS-compliant Java Security Provider (such as BC-FIPS) hast to be used instead of `bcprov`" is right as of Camel 4.7.0.

Actually I wonder, whether there is any reason (beyond backwards compatibility) to remove BC from the Crypto component in Camel? The Camel crypto module not only compiles but also its tests are passing after the removal of all BC dependencies. `SunJCE` provider might be just fine for the basic usage. Note that I have not looked whether the tests perform some real crypto-operations, but I strongly hope so. 

If we remove BC from the crypto component, then the component should keep working out of the box also on FIPS enabled systems, no?

JiriOndrusek: Interesting idea, I'll check if we can remove BC from camel-crypto. If it would be possible/approved, then definitely the statement  `any FIPS-compliant Java Security Provider (such as BC-FIPS) hast to be used instead of bcprov` would be much better.

JiriOndrusek: @ppalaga I checked the options and you are right. Camel-crypto does not require BC at all. I will propose such change in near future to the Camel. In the meantime I'll exclude the BC from camel-crypto in the CQ BOM (and I will revert it back as soon as the relevant Camel bits reaches the CQ)
I have a question what to suggest to users:
1 - In non-FIPS environment all worjs out of the box.
2 - In the FIPS environment, If user wants to use BCFIPS, following dependencies have to be added

```
                <dependency>
                    <groupId>org.bouncycastle</groupId>
                    <artifactId>bc-fips</artifactId>
                </dependency>
                <dependency>
                    <artifactId>camel-quarkus-support-bouncycastle</artifactId>
                </dependency>
```

If another fips-compliant security provider is used, it has to be registered properly.

(I tested the scenario with BCFIPS on FIPS machine)

Does it make sense to you, to suggest  both approaches? (BCFIPS and different fips compliant provider)
(the other option is to suggest use of BCFIPS only)

JiriOndrusek: Fixed with explanation in the main thread

ppalaga: > 1 - In non-FIPS environment all worjs out of the box.

I'd vote for being more explaining even for non-FIPS. We should note the that

* BC dependency was removed from crypto in Camel 3.8.0. 
* If users need algorithms or hash functions provided by BC, they need to add what? - camel-quarkus-support-bouncycastle AND bcprov? camel-quarkus-support-bouncycastle will do what? - register BC as a Security Provider and add native support (correct me if I am wrong)?
* When adding other non-default security providers to class path, users should also register them.

> 2 - In the FIPS environment

Sounds right.

JiriOndrusek:
The crypto/crypto-pgp changes are present in Camel 3.7.0.
I rebased this PR against `main` and I think that it could be approved/merged. All the questions and suggestions are answered or fixed.

JiriOndrusek:
@ppalaga WDYT? Can we merge this PR?

ppalaga (approved):
Looks great, thanks @JiriOndrusek !

PR #6253, opened 2024-07-04
Remove OpenAPI V2 integration test from camel-k-maven-plugin

jamesnetherton:


jamesnetherton:
`functional-extension-tests` tests job passed which is good enough to merge this.

PR #6262, opened 2024-07-09
Remove prefix from shell code snippets

jamesnetherton:
Not sure what others think about this.

From my POV, having the `$` is a pain if you want to copy / paste commands from the shell code snippets.


oscerd (approved):

aldettinger (approved):
This is indeed something of value :+1:

ppalaga:
`$` is nicer visually, because one sees, where a command starts, but you are right that it is not so nice when copying. I would not mind removing them, but I just stumbled across this piece of Antora docs: 

> Console blocks
> The default UI also adds a clipboard button to all console blocks. A console block is either a literal paragraph that begins with a $ or a source block with the language console.
>
>The script provided by the default UI will automatically strip the $ prompt at the beginning of each line and join the lines with &&.

https://docs.antora.org/antora-ui-default/code-blocks/#console-blocks

It is a feature of default Antora UI, so it may or may not work in the same way in Camel custom UI. 
We should perhaps try to switch from `[source,shell]` to `[source,console]`

jamesnetherton:
> it may or may not work in the same way in Camel custom UI

Yes, there is no copy button support unfortunately https://github.com/apache/camel-website/issues/1137.

ppalaga:
Good to know. Let's get rid of the dollars then.

Issue #6288, opened 2024-07-23
Improve quarkus-langchain4j version alignment


aldettinger:
### Describe the feature here

Camel Quarkus bom currently only manages a small subset of quarkus-langchain4j dependencies.
For instance, developers using the llm specific dependency need to maintain some quarkus-langchain4j versions manually, which is error prone.
Let's try to have Camel Quarkus come with a default proposal for all those versions.

aldettinger:
Given feedback from prs and discussions, we opted for improving the documentation.
In LangChain4j related extensions docs, users will be recommended to import the `quarkus-langchain4j-bom`.

PR #6289, opened 2024-07-23
Improve Quarkus LangChain4j version alignment #6288

aldettinger:

### DIFF
--- a/docs/modules/ROOT/pages/reference/extensions/langchain4j-chat.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/langchain4j-chat.adoc
@@ -43,3 +43,26 @@ Or add the coordinates to your existing project:
 ifeval::[{doc-show-user-guide-link} == true]
 Check the xref:user-guide/index.adoc[User guide] for more information about writing Camel Quarkus applications.
 endif::[]
+
+[id="extensions-langchain4j-chat-quarkus-langchain4j-bom"]
+== Ensure Quarkus LangChain4j dependencies version alignment with the Bom
+
+In order to ensure alignment across all Quarkus and LangChain4j related dependencies, it is recommended to import the Quarkus LangChain4j bom as below:
+[source,xml]
+----
+<dependencyManagement>
+  <dependencies>
+    <dependency>
+      <groupId>io.quarkiverse.langchain4j</groupId>
+      <artifactId>quarkus-langchain4j-bom</artifactId>
+      <version>${insert your version here}</version>
+      <type>pom</type>
+      <scope>import</scope>
+    </dependency>
+  </dependencies>
+  ...
+</dependencyManagement>
+----
+
+Note that the import order is paramount when using maven `dependencyManagement`.
+As such, one might need to import the `quarkus-langchain4j-bom` before other related Camel and Quarkus boms.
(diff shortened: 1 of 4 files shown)


poms/bom/src/main/generated/flattened-reduced-pom.xml
```diff
@@ -8428,6 +8412,78 @@
...
+        <version>0.32.0</version>
+      </dependency>
+      <dependency>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-core</artifactId>
+        <version>0.32.0</version>
+        <type>test-jar</type>
+        <classifier>tests</classifier>
+      </dependency>
+      <dependency>
```
aldettinger, line 8442: @jamesnetherton On the top of your mind, would any of the following dependencies below generate issues on the platform ? Like aws ? infinispan ?

jamesnetherton: I think the platform generator plugin would force versions that are 'owned' by their respective platform members. E.g Infinispan versions would come from `quarkus-core` & AWS versions from`quarkus-amazon-services`.

Importing `langchain4j-bom` in its entirety is perhaps a bit heavyweight. It's adding some stuff that I'd prefer to not have in the BOM, like lombok, jsoup and a test-jar. Maybe we can tweak the `cq-maven-plugin` config to restrict what's included to `dev.langchain4j:*` (maybe even `*` is too much).

Or, if there's just a known set of dependencies that we think are used commonly, maybe we just add those individually to the BOM.

aldettinger: Need more thinking indeed, jsoup & lombok would be too high a price.

Actually, I'm not even sure it would have the desired effect.
Like, let's say this pr is integrated and a next quarkus-platform release embed it.
Would the dependency management include the `quarkus-langchain4j-bom` ? Like  an application depending on quarkus-langchain4j-ollama would have the right version pulled from dependency management ?

jamesnetherton: > Would the dependency management include the `quarkus-langchain4j-bom`

Everything in `camel-quarkus-bom` should in theory make its way into the platform generated `quarkus-camel-bom`. So the langchain4j bits should get propagated.

You can verify it locally by building the platform with the CQ SNAPSHOT.

aldettinger: After investigations, `jsoup` and `lombok` are used at deep langchain4j level. Managing them seems unmaintanable. Unless there are other ideas, the best way would be too let users force the versions by importing the `quarkus-langchain4j-bom`.

jamesnetherton: We could maybe add a section to the docs about importing the BOM in user apps. You could add an asciidoc attribute to `antora.yml` for the langchain4j version so that it could be used in a code snippet.

aldettinger: That looks like an interesting idea. Let investigate to update this pr.

jamesnetherton (approved):

aldettinger:
Just a small fix coming and it should be ready to go.

PR #6292, opened 2024-07-24
Disable commons-pool2 JMX MBean registration in native mode

jamesnetherton:
Long term we maybe want to have this logic available elsewhere so it can be shared. @zhfeng WDYT?


aldettinger (approved):

zhfeng (approved):

zhfeng:
Thanks @jamesnetherton - Yeah, we need a common extension to share it exactly. But it seem too heavy to create a quarkivers extension of `commos-pool2` just including this substitution?

jamesnetherton:
> But it seem too heavy to create a quarkiverse extension of `commos-pool2` just including this substitution?

Yeah that was my thought too.

Maybe we can try to explore some other solution eventually.

PR #6296, opened 2024-07-25
[Cassandra] Increase stack size for test container

avano:
Trying to run the tests on aarch64 RHEL and the container does not start with:

```
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

The Java thread stack size specified is too small. Specify at least 448k
```

this is an [issue](https://issues.apache.org/jira/browse/CASSANDRA-15446) in cassandra that is not fixed yet


aldettinger (approved):
:+1:

oscerd (approved):

jamesnetherton (approved):

Issue #4127, opened 2022-09-22
Splunk: extend test coverage by TLS


JiriOndrusek:
This is follow-up issue for https://github.com/apache/camel-quarkus/issues/4085

See [comment](https://github.com/apache/camel-quarkus/pull/4124#issuecomment-1254886684) for more information.

JiriOndrusek:
I tried to run the splunk https tests by following https://docs.splunk.com/Documentation/Splunk/9.2.0/Security/ConfigTLSCertsS2S


The first task is to configure splunk server to SSL. It is possible by provide necessary config parameters and by restarting afterwards. The config parameter might by added to the configuration file by copying (overwriting following file - [/opt/splunk/etc/system/local/server.conf)](https://github.com/opt/splunk/etc/system/local/server.conf)) (see [my POC](https://github.com/JiriOndrusek/camel-quarkus/blob/splunk-ssh-POC/integration-tests/splunk/src/test/java/org/apache/camel/quarkus/component/splunk/it/SplunkTestResource.java#L104))

There are several options. (You can validate the the certificates by running `openssl s_client -connect localhost:32880` or `openssl s_client -connect localhost:32880 -CAfile cacert.pem`)
• If certificates are kept as default -> splunk server generates the certificates during the start. 
• You can use keytool approach (similar to cxf tests) to generate certificates. Server pem has to be concatenated from private key and certificate. The proper method is to run following commands (but the same result should be also achieved when key and crt files are concatenated (I didn't verify that)
```
openssl pkcs12 -export -out combined.p12 -inkey localhost-key.pem -in localhost.pem -certfile splunkca.pem
openssl pkcs12 -in combined.p12 -out combined.pem -nodes
```

• You can generate certificates via annotations -> but in this case, I encountered a problem with self-signed certificate, therefore I left that way.


According to openSsl verification b**oth optoins (1 and 2) shows that server responds with a proper certificate.**

----

The second part of the job is to configure splunk client properly for ssh. The splunk component lacks ssl configuration options (I reported the issue as https://issues.apache.org/jira/browse/CAMEL-20393).
Unfortunately there is no clear way of fixing the issue. The underlying splunk client (..servicemix.bundles.splunk.1.9.5_1 creates sslContext statically from javax.next.sslSSLContext.getInstance("TLS"); with no option to create a custom ssl context.

[...]

PR #6302, opened 2024-07-30
Splunk: extend test coverage by TLS

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/4127

requires https://github.com/apache/camel/pull/14970
(I'm keeping this PR as draft, because the camel PR is not merged at this time and the CI should fail without it)

- Strict validation of certificates for client is turned off, because it is not possible to customize the sslContext for the client
- This PR introduces parent of splunkTest, whih allows to run test with scheme https or http
- Bunch of splunk configuration was refacored
- Certificate generation was added (keytool approch has to be used, because the *.pem files are concatenated in the *TestResource class)
- Ppositive and negative tests for SSL are added


jamesnetherton:
Instead of having separate tests for SSL Vs Non-SSL, I wonder if we just simplify things and test SSL only?

That's the approach we have used in other places. I assume in the majority of user scenarios that their Splunk instances will be SSL / TLS secured.

JiriOndrusek:
> Instead of having separate tests for SSL Vs Non-SSL, I wonder if we just simplify things and test SSL only?
> 
> That's the approach we have used in other places. I assume in the majority of user scenarios that their Splunk instances will be SSL / TLS secured.

I have to keep the non-ssl tests as I use them as negative ones for ssl.
But I can simplify the tests, so I would keep only one test class (for ssl), and therefore the splunk server would start only once

JiriOndrusek:
Superseeded by https://github.com/apache/camel-quarkus/pull/6314

PR #6304, opened 2024-07-31
Fix #6301 to re-sort the beans

### LINKED ISSUE #6301: Bean autowiring is non-deterministic when multiple beans exist with no @Priority
### Bug description

Relates to #5464. The code changes for that feature assume that if there are multiple beans for a given type, that they will have an `@Priority` assigned, so that the desired bean is returned from registry lookups.

IMO this is only reliable for beans declared from user code. The majority of beans created by Quarkus extensions do not have any `@Priority` (AFAIK). E.g like `DataSource` beans.

Thus, when Camel attempts to do autowiring, it will potentially inject a 'random' bean into components, since there is no `@Priority` enforced.

zhfeng:
Fix 
- #6301

### DIFF
--- a/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
+++ b/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
@@ -17,6 +17,7 @@
 package org.apache.camel.quarkus.core;
 
 import java.lang.annotation.Annotation;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -27,9 +28,11 @@ import java.util.Set;
 
 import io.quarkus.arc.Arc;
 import io.quarkus.arc.ArcContainer;
+import io.quarkus.arc.InjectableBean;
 import io.quarkus.arc.InstanceHandle;
 import io.smallrye.common.annotation.Identifier;
 import jakarta.enterprise.inject.AmbiguousResolutionException;
+import jakarta.enterprise.inject.Default;
 import jakarta.enterprise.inject.literal.NamedLiteral;
 import jakarta.enterprise.inject.spi.Bean;
 import jakarta.enterprise.inject.spi.BeanManager;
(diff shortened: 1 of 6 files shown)


oscerd (approved):

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
```diff
@@ -163,8 +165,13 @@ public <T> T findSingleByType(Class<T> type) {
             } else {
                 handles = container.listAll(type);
             }
+
             if (handles.size() > 0) {
-                return handles.get(0).get();
+                if (handles.stream().filter(h -> !h.getBean().getKind().equals(InjectableBean.Kind.SYNTHETIC)).count() > 0) {
```
jamesnetherton, line 170: What's the significance of the `SYNTHETIC` beans kind where we can just take the first one? Is that safe if multiple results are returned?

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
```diff
@@ -163,8 +165,13 @@ public <T> T findSingleByType(Class<T> type) {
...
+                    return handles.get(0).get();
+                } else {
+                    return container.select(type, Default.Literal.INSTANCE).get();
```
jamesnetherton, line 173: Is it safe to call `get()` if no default bean exists? IIRC it'll throw an exception if not.

jamesnetherton (commented):
Thanks @zhfeng. I left a couple of question comments.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
zhfeng: Synthetic beans are usually provided by extensions. I'm thinking another way with resorting the handels?

The orignal `handles` is sorted by `priority` and when `priority` is equal, the position is random. How about to introduce a new `Comparator` if `priority` is the same, the default bean is win?

jamesnetherton: > introduce a new Comparator if priority is the same, the default bean is win?

Yeah, sounds ok.

Probably faster than the alternative, which would be to collect all priorities in a set and then check the size.

zhfeng: OK, I updated with the new `Comparator`.

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
zhfeng: Remove it.

jamesnetherton (approved):

PR #6320, opened 2024-08-06
Avoid deploying redundant modules during the release

jamesnetherton:
This shrinks the number of modules deployed at release time from 1435 to 1274.

I ran the full release against a local Nexus instance and the output from both the `release:prepare` and `release:perform` stages was as expected. The list of deployed artifacts is here:

https://gist.github.com/jamesnetherton/f19fcec764d08f59c39933835f3d1dfb

I also tested the release with the Quarkus Platform. It built fine and I ran a small selection of tests, which passed.

WDYT @ppalaga?


oscerd (approved):

ppalaga (approved):
@jamesnetherton it looks great, thanks a lot for the effort!

PR #6331, opened 2024-08-08
Mark modules using legacy @ConfigRoot as such

gsmet:
The new extension annotation processor coming with 3.14 requires that you mark the modules using the legacy @ConfigRoot approach (i.e. without @ConfigMapping) with a specific annotation processor parameter.

This is important for the generation of the config documentation (which is enabled by default).

This can be merged pre-3.14 update but will result in a warning when building. With the advantage of having the build with main passing.

I think I have marked all of them but in any case, if you have an error, you now know what to do.

Fixes https://github.com/quarkusio/quarkus/issues/42417

### DIFF
--- a/extensions-core/core/runtime/pom.xml
+++ b/extensions-core/core/runtime/pom.xml
@@ -117,6 +117,9 @@
                             <version>${quarkus.version}</version>
                         </path>
                     </annotationProcessorPaths>
+                    <compilerArgs>
+                        <arg>-AlegacyConfigRoot=true</arg>
+                    </compilerArgs>
                 </configuration>
             </plugin>
         </plugins>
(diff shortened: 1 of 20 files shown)


gsmet:
At your disposal if you have any questions about this change!

gsmet:
I added more details here: https://github.com/quarkusio/quarkus/wiki/Migration-Guide-3.14#extension-developers

Note that by applying just this patch, you will get a warning when compiling the test classes (as given you don't have a config annotation, the annotation processor will not be triggered and you will get a warning telling you the parameter is unused).
Ideally, you would adjust to only run the annotation processor for the main classes as described in the doc but I didn't want to be too aggressive on your codebase :).
Feel free to do it if you think it makes sense.

jamesnetherton (approved):

jamesnetherton:
Thanks @gsmet!

I can follow up with any additional tweaks.

PR #6381, opened 2024-08-23
[3.8.x] Ban com.sun.xml.bind:jaxb-core and jaxb-impl to avoid clash with org.glassfish.jaxb:jaxb-core and jaxb-runtime

ppalaga:
Those two groups of artifacts have lots of overlapping classes. The Glassfish ones should be enough. Let's see if all tests are going to pass and then I'd do the same in Camel.

### DIFF
--- a/poms/bom/pom.xml
+++ b/poms/bom/pom.xml
@@ -1501,6 +1501,12 @@
                 <groupId>org.apache.camel</groupId>
                 <artifactId>camel-jaxb</artifactId>
                 <version>${camel.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>com.sun.xml.bind</groupId>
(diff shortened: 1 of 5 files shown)


ppalaga:
This change could potentially break user apps if they rely on classes available in `com.sun.xml.bind:jaxb-*` that are not available in the Glassfish artifacts. Therefore switching to draft.

ppalaga:
The failure in Camel K plugin is fixable, but anyway, I think we should better not merge this to 3.8.x, unless others think differently?

aldettinger:
> merge this to

Having the release branches stable should be a priority. Unless this pr fixes a big problem, I would vote not to merge in 3.8.x.

ppalaga:
We agreed not to merge this in 3.8.x

PR #6382, opened 2024-08-23
Ban com.sun.xml.bind:jaxb-core and jaxb-impl to avoid clash with org.glassfish.jaxb:jaxb-core and jaxb-runtime

ppalaga:
Those two groups of artifacts have lots of overlapping classes. The Glassfish ones should be enough. Let's see if all tests are going to pass and then I'd do the same in Camel.

### DIFF
--- a/poms/bom/pom.xml
+++ b/poms/bom/pom.xml
@@ -1565,6 +1565,12 @@
                 <groupId>org.apache.camel</groupId>
                 <artifactId>camel-jaxb</artifactId>
                 <version>${camel.version}</version>
+                <exclusions>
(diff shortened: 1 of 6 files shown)


ppalaga:
All test have passe 🎉, so I'd be in favor of merging this PR for main, and rather not merging the [one for 3.8.x](https://github.com/apache/camel-quarkus/pull/6381)

Here is the proper fix for Camel: https://github.com/apache/camel/pull/15298

jamesnetherton (approved):

zhfeng (approved):

PR #6394, opened 2024-08-28
Revert back to QuarkusTestResource

jamesnetherton:
Given this comment just landed on the Quarkus 3.14.x branch:

https://github.com/quarkusio/quarkus/blob/3.14/test-framework/common/src/main/java/io/quarkus/test/common/WithTestResource.java#L16-L18

And ongoing discussion:

https://groups.google.com/g/quarkus-dev/c/rS8-WN6b7XQ

I think we should revert back to `QuarkusTestResource` for 3.15.x. We can move forwards on `WithTestResource` for 3.16 and beyond.


aldettinger (approved):
Well caught :+1:

PR #6406, opened 2024-09-03
 Github Actions Security Best practices: Pin Actions to Full lenght C…

oscerd:
…ommit SHA - Assign milestone workflow

### DIFF
--- a/.github/workflows/assign-issue-milestone.yaml
+++ b/.github/workflows/assign-issue-milestone.yaml
@@ -35,7 +35,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Assign Closed Issues To Latest Milestone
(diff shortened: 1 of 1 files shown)


ppalaga:
@oscerd could you please explain, why is this a good practice?

oscerd:
@ppalaga https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

It's suggested to do this for third party, it's not crucial, but even for dependabot it doesn't change that much, it will point the new commit sha on new releases.

ppalaga:
Makes sense, thanks a lot!

oscerd:
I forgot to add the link to the hardening guide. Sorry.

PR #6415, opened 2024-09-03
[Tests] FTP: Use restrictions to bypass cert setup issues

avano:
Currently both `FtpTestResource` and `FtpsTestResource` are instantiated before the first test runs.

The current method did not work when executing only the native tests (`-Pnative -Dtest="\!*" -Dsurefire.failIfNoSpecifiedTests=false`) without jvm tests. When running with `-Pnative` only (so jvm tests first, then native), the problem did not occur, as the `FtpsIT` did not actually use the certificates defined in `FtpsTest`, but reused the certificates generated by the invocation of `SftpTest` in jvm mode.

When running with `-Pnative -Dtest="\!*" -Dsurefire.failIfNoSpecifiedTests=false`, the "workflow" without this change was:
1. start running FtpIT
2. instantiate FtpTestResource
3. instantiate SftpTestResource
4. run FtpIT
5. start running FtpsIT
6. generate certificates needed for SftpTestResource, but that is already initialized

With this change, the order is:
1. start running FtpIT
2. instantiate FtpTestResource
3. run FtpIT
4. start running FtpsIT
5. generate certificates needed for SftpTestResource
6. instantiate SftpTestResource

please also backport to 3.8 when merged

### DIFF
--- a/integration-tests/ftp/src/test/java/org/apache/camel/quarkus/component/ftp/it/FtpTest.java
+++ b/integration-tests/ftp/src/test/java/org/apache/camel/quarkus/component/ftp/it/FtpTest.java
@@ -25,7 +25,7 @@ import org.junit.jupiter.api.Test;
 import static org.hamcrest.CoreMatchers.is;
 
 @QuarkusTest
-@QuarkusTestResource(FtpTestResource.class)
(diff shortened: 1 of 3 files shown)


avano:
@JiriOndrusek fyi

aldettinger (approved):
Well done.

PR #6419, opened 2024-09-04
Use quarkus-micrometer-registry-prometheus instead of the plain io.micrometer dependency

jamesnetherton:
Anticipates a change coming in Quarkus 3.14.2.


aldettinger (approved):

ppalaga:
Do we need to release 3.14.1?

PR #6506, opened 2024-09-23
Force Dependabot to get dependency metadata only from Maven Central

jamesnetherton:
Try to avoid Dependabot resolving metadata from the Atlassian repos which often have stale data, not in sync with Maven Central.


aldettinger (approved):
Oh, well spotted @jamesnetherton :+1:

Issue #6534, opened 2024-09-24
Create camel-quarkus-langchain4j extension


aldettinger:
### Describe the feature here

Provide access to quarkus-langchain4j high level api.

PR #6544, opened 2024-09-24
Add Camel Quarkus LangChain4j extension #6534

aldettinger:


tooling/scripts/test-categories.yaml
```diff
@@ -135,6 +135,7 @@ group-08:
   - crypto
   - crypto-pgp
   - jq
+  - langchain4j
```
jamesnetherton, line 138: Probably better to move it into `group-02` since it's completing way before the others at the moment.

jamesnetherton (approved):

tooling/scripts/test-categories.yaml
aldettinger: Should be ok in fixup commit.

PR #6555, opened 2024-09-26
Upgrade Quarkus Langchain4j to 0.19.0.CR1

jamesnetherton:
Main motiviation for this is to fix some dependency / compatibility issues on `quarkus-main`.


aldettinger (approved):
Fair enough, plus should unblock other fixes.

PR #6571, opened 2024-10-01
Make BindToRegistry work outside of RouteBuilder classes

jamesnetherton:


integration-test-groups/foundation/bean/src/main/java/org/apache/camel/quarkus/component/bean/bind/BindToRegistryOnMethodBean.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.bean.bind;
+
+import org.apache.camel.BindToRegistry;
+
+public class BindToRegistryOnMethodBean {
+    @BindToRegistry
```
aldettinger, line 22: So, `BindToRegistryBean` class is included via annotated field and method ? Would it be better for test coverage to use distinct classes ala `BindToRegistryBean1` `BindToRegistryBean2` ? With better names of course :)

aldettinger (approved):

integration-test-groups/foundation/bean/src/main/java/org/apache/camel/quarkus/component/bean/bind/BindToRegistryOnMethodBean.java
jamesnetherton: > use distinct classes

I could, but the bean type doesn't matter so much (IMO).

I think what we want to verify is whether a bean was bound to registry with the expected name and is resolvable in a Camel route or via a manual lookup.

PR #6639, opened 2024-10-15
Fix Azure tests

llowinge:

### DIFF
--- a/integration-test-groups/azure/azure-eventhubs/src/test/java/org/apache/camel/quarkus/component/azure/eventhubs/it/AzureEventhubsTest.java
+++ b/integration-test-groups/azure/azure-eventhubs/src/test/java/org/apache/camel/quarkus/component/azure/eventhubs/it/AzureEventhubsTest.java
@@ -26,11 +26,13 @@ import io.quarkus.test.junit.QuarkusTest;
 import io.restassured.RestAssured;
 import io.restassured.http.ContentType;
 import io.restassured.response.Response;
+import io.restassured.response.ValidatableResponse;
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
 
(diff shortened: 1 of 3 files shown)


oscerd (approved):

integration-test-groups/azure/azure-eventhubs/src/test/java/org/apache/camel/quarkus/component/azure/eventhubs/it/AzureEventhubsTest.java
```diff
@@ -72,9 +74,10 @@ void produceConsumeEvents() {
                         .body(
                                 "body", is(messageBody),
                                 "headers.CamelAzureEventHubsEnqueuedTime", notNullValue(),
-                                "headers.CamelAzureEventHubsOffset", greaterThanOrEqualTo(0),
```
aldettinger, line 75: @llowinge looks there is a strange issue here :)
What is the error message please ? Is it a known bug in hamcrest matchers ?

By any chance, would using something like `greaterThanOrEqualTo(0L)` or `greaterThanOrEqualTo(Long.valueOf(0L))` work ?

llowinge: The problem is, that camel can return through Azure Long or Integer class. With real instance, it returns Long and the test fails. Unfortunately your suggestions don't work, because it fails with typing Integer to Long class (which is not possible). I had to do it this way, to support both Integer and Long classes.

aldettinger: It is a bad situation indeed. It would need kind of real instance only assertion or something like asserting that it matches any of a list of matchers.

Fair enough with the solution and many thanks for sharing :+1:

aldettinger (approved):

jamesnetherton (approved):

PR #6640, opened 2024-10-15
[ARM] Skip tests for unsupported services

avano:
please backport to 3.15 afterwards


oscerd (approved):

zhfeng (approved):

jamesnetherton:
We should maybe check if the platform is MacOS in all cases. I can run amd64 containers via Rosetta. So in theory all of these tests are runnable.

In that case, it be better to turn the check into a JUnit `DisabledIf` annotation / condition. See other examples here:

https://github.com/apache/camel-quarkus/tree/main/integration-tests-support/test-support/src/main/java/org/apache/camel/quarkus/test

avano:
thanks James, I don't have a mac to test, but since it works for you, I'll rework

aldettinger (approved):

avano:
@jamesnetherton fixed. feel free to suggest a better name for the annotation

jamesnetherton (approved):

PR #6653, opened 2024-10-17
Add tests for JQ built in functions

jamesnetherton:
This will fail as it needs a Camel [change](https://issues.apache.org/jira/browse/CAMEL-21357). Thus, opening in draft for feedback.


aldettinger:
Looks good.

Issue #6657, opened 2024-10-18
Jdbc/mysql Not it is inot necessary to use external DB for the FIPS testing


JiriOndrusek:
### Bug description

It is possible to run the test with a MySql DB spawned by the docker in the same machine, even on the FIPS system.
Therefore the FIPS testing should be made easier.

PR #6658, opened 2024-10-18
Mysql testResource instead of devservices in FIPS

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/6657

The FIPS testing is much more easier if simple test container is used.


integration-test-groups/jdbc/mysql/src/test/java/org/apache/camel/quarkus/component/jdbc/mysql/MysqlTestResource.java
```diff
@@ -0,0 +1,80 @@
...
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+
+/**
+ * This is a workaround bevause of the FIPS environment.
```
aldettinger, line 29:
```suggestion
 * This is a workaround because of the FIPS environment.
```

JiriOndrusek:
I  switched to draft as a different approach should be used.

JiriOndrusek:
The behavior had to be changed, it is not possible to disable devservice from the TestResource. I kept the FIPS profile, which stops dev service. TestResource starts the container only in case of stopped devservice

integration-test-groups/jdbc/mysql/src/test/java/org/apache/camel/quarkus/component/jdbc/mysql/MysqlTestResource.java
JiriOndrusek: fixed

aldettinger (approved):

jamesnetherton (approved):

PR #6707, opened 2024-10-28
Fix #6701 to support OpenAPI spec with yaml format

### LINKED ISSUE #6701: camel-quarkus-rest-openapi: Add support for OpenAPI specifications written in YAML format
### Describe the feature here

Currently, `camel-quarkus-rest-openapi` only detects and handle OpenAPI specifications with a `.json` suffix. However, OpenAPI specifications may be represented either in JSON or YAML format. Support for detecting and handling files with a `.yaml` suffix should be added.

zhfeng:


jamesnetherton (approved):
Not sure if we also want to consider the `.yml` extension too.

zhfeng:
Thanks @jamesnetherton - Hmm, good catch, @calleandersson, is it needed to add `.yml` as well?

calleandersson:
The following is written at https://yaml.org/faq.html:
_"Is there an official extension for YAML files?
Please use ".yaml" when possible."_

Personally, `.yaml` is enough for my needs but if `.yml` is commonly used it might be good to support it too but I have no strong oppinions on that.

Issue #6705, opened 2024-10-27
Better integration with Quarkus Amazon extension


scrocquesel:
### Describe the feature here

After reading the [documentation](https://camel.apache.org/camel-quarkus/3.15.x/reference/extensions/aws2-s3.html#extensions-aws2-s3-additional-camel-quarkus-configuration), I see an opportunity to enhance the user experience.

First, an `S3Presigner` bean should have been explicitly declared. However, due to a recently discovered bug, the Quarkus Amazon extension has historically created an `S3Presigner` bean even without an injection point. This means users are not required to add a dummy `S3Presigner` bean. This behavior will be removed in the next major version. I introduced a configuration flag that defaults to the legacy behavior to maintain backward compatibility in the next minor version.

Second, the Quarkus Amazon extension is now structured to allow other extensions to "request" a client bean without an injection point. Consequently, this approach should eliminate the need for users to add dummy beans and manually configure the client type.
By using [conditional dependencies](https://quarkus.io/guides/conditional-extension-dependencies), we can detect the presence of `quarkus-amazon-s3` and activate an extension that requires `S3Client` and `S3Presigner` beans. It would also be possible to embed an `application.properties` file for configuring the client type property. I don’t think Camel Quarkus users are likely to use an HTTP client implementation other than apache.

Third, starting version 2.19, Quarkus amazon will support [named clients](https://docs.quarkiverse.io/quarkus-amazon-services/dev/common-features.html). This was asked by our users to be able to connect to multiple queues with different credentials.

Last, I noticed that we’re not fully aligned on the AWS services we support, and even where we do overlap, this isn’t reflected in your documentation (cloudwatch, eventbridge, iam, kms, kinesis, lambda, secretmanager). I can work to close the gap of the missing services.

scrocquesel:
@jamesnetherton Have you make any decision about removing quarkus amazon references ?

jamesnetherton:
> @jamesnetherton Have you make any decision about removing quarkus amazon references ?

Yes, I will do it this week.

jamesnetherton:
Removed docs in https://github.com/apache/camel-quarkus/pull/6804.

PR #6734, opened 2024-10-31
Quarkus Amazon S3 integration proof of concept

scrocquesel:
Relates to #6705

This a proof of concept of how it could be possible to seamlessly integrate with the quarkus amazon extension.

### DIFF
--- a/extensions/aws2-s3/deployment-quarkus-amazon-s3-integration/src/main/java/org/apache/camel/quarkus/component/aws2/s3/deployment/QuarkusAmazonS3IntegrationProcessor.java
+++ b/extensions/aws2-s3/deployment-quarkus-amazon-s3-integration/src/main/java/org/apache/camel/quarkus/component/aws2/s3/deployment/QuarkusAmazonS3IntegrationProcessor.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.aws2.s3.deployment;
+
+import io.quarkus.amazon.common.deployment.RequireAmazonClientInjectionBuildItem;
+import io.quarkus.amazon.common.runtime.ClientUtil;
+import io.quarkus.deployment.annotations.BuildProducer;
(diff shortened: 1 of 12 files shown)


extensions/aws2-s3/runtime/pom.xml
```diff
@@ -59,6 +70,13 @@
             <groupId>software.amazon.awssdk.crt</groupId>
             <artifactId>aws-crt</artifactId>
```
scrocquesel, line 71: this should not be necessary, we provide graalvm substitutions https://github.com/quarkiverse/quarkus-amazon-services/blob/main/s3/runtime/src/main/java/io/quarkiverse/amazon/s3/runtime/S3CrtSubstitutions.java

Could be made available in a separate artifact for shared usage

extensions/aws2-s3/deployment-quarkus-amazon-s3-integration/src/main/java/org/apache/camel/quarkus/component/aws2/s3/deployment/QuarkusAmazonS3IntegrationProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+public class QuarkusAmazonS3IntegrationProcessor {
+
+    private static final DotName S3_CLIENT = DotName.createSimple(S3Client.class);
+    private static final DotName S3_PRESIGNER = DotName.createSimple(S3Presigner.class);
+
+    @BuildStep
+    void integrate(BuildProducer<RequireAmazonClientInjectionBuildItem> requireClientInjectionProducer) {
+
+        // request a default bean for S3Client and S3Presigner.
+        requireClientInjectionProducer.produce(new RequireAmazonClientInjectionBuildItem(
```
scrocquesel, line 36: I can move the build item to a dedicated spi artifact to reduce coupling

ppalaga:
@scrocquesel thanks for the PR. 

Is my understanding correct that he goal is to improve the user experience for applications depending on both Camel Quarkus 
 and Quarkus Amazon? With this change in place, Camel Quarkus would further work without depending on Quarkus Amazon, right?

scrocquesel:
> @scrocquesel thanks for the PR. 
> 
> Is my understanding correct that he goal is to improve the user experience for applications depending on both Camel Quarkus 
>  and Quarkus Amazon? With this change in place, Camel Quarkus would further work without depending on Quarkus Amazon, right?

Right. That's the idea.

scrocquesel:
> The question is whether this would be worth the effort. I was actually considering removing the Quarkus Amazon references & tests from the project. It's mostly a legacy from the early days of the project.
> 
> The vast majority of our users will be using the vanilla Camel component extensions as-is.

Oh great, this would simplify a lot.

We are left with the issue of referencing different versions of the AWS SDK. This could be problematic for the Quarkus Platform, as Quarkus Amazon tends to stay more up to date, and certain versions of the AWS SDK (like the recent 2.29.0) could break native compilation. I know you are copy/pasting some graalvm substitution and maybe this could be improved in some way ?

jamesnetherton:
> We are left with the issue of referencing different versions of the AWS SDK

We could leave the `quarkus-amazon-services-bom` import in our test BOM and that'd ensure we are testing with whatever version of the AWS SDK is used in the platform.

> Quarkus Amazon tends to stay more up to date

True. However, Camel is upgrading to the latest AWS SDK via dependabot (pretty much daily upgrades). So things hopefully wont get too far behind to the extent we have major compatibilty issues.

Issue #5244, opened 2023-09-01
Fop native failures due to pdfbox 3 upgrade


JiriOndrusek:
### Bug description

Since pdfbox was upgraded from 2.x to 3.x in camel (https://issues.apache.org/jira/browse/CAMEL-19796), some native failures happen. The native profile is disabled until the issue is fixed.

I encountered following error:

```
2023-08-31T11:20:48.9327060Z Caused by: com.oracle.graal.pointsto.constraints.UnresolvedElementException: Discovered unresolved method during parsing: org.apache.fontbox.cff.CFFParser.parse(byte[]). This error is reported at image build time because class org.apache.fop.fonts.truetype.OTFFile is registered for linking at image build time by command line
```

and (if above one is fixed)

```
Error: Classes that should be initialized at run time got initialized during image building:
 java.awt.Toolkit the class was requested to be initialized at run time (from feature io.quarkus.runner.Feature.beforeAnalysis with 'java.awt'). To see why java.awt.Toolkit got initialized use --trace-class-initialization=java.awt.Toolkit
java.awt.Rectangle the class was requested to be initialized at run time (from feature io.quarkus.runner.Feature.beforeAnalysis with 'java.awt'). To see why java.awt.Rectangle got initialized use --trace-class-initialization=java.awt.Rectangle

```

Once issues are fixed, native profile should be restored,

fugerit79:
Good morning.

Any news on this issue? @JiriOndrusek @aldettinger @ppalaga?
(Sorry for desturbing, I mentioned the issue assignee and the latest committers : https://quarkus.io/extensions/org.apache.camel.quarkus/camel-quarkus-fop/?tab=community)

Would be nice to have it working again.

Thanks in advance.

Issue #5234, opened 2023-08-30
Tika pdf failure after upgrade to pdfBox in Camel, requires new quarkiverse-tika


JiriOndrusek:
### Bug description

Camel upgraded pdfbox to 3.x (https://issues.apache.org/jira/browse/CAMEL-19796).

Pdfbox 3.x is not back compatible with 2.x, therefore quarkiverse-tika used by tika extension fails with the new pdfbox.

Apache tika is aware of the new version of pdfbox and the upgrade ticket is already in progress - see https://issues.apache.org/jira/browse/TIKA-3347

As soon as an new version of apache tika is released it has to be adopted by quarkiverse-tika and this new version has to be adopted by camel-quarkus.

I'm disabling tika tests using pdfbox until provblem is solved (on `camel-main`)

JiriOndrusek:
@ppalaga @jamesnetherton We have to wait for the new release of quarkiverse-tika supporting the new pdfbox. There is probably not much other options (the pdfbox involves i.e. fop, tika, pdf extensions, therefore keeping pdfbox on 2.x may be complicated. WDYT?

jamesnetherton:
> We have to wait for the new release `quarkiverse-tika`

Or we propose to revert the upgrade in Camel. I assume the same issue exists there if you try to bring the `tika` & `pdf` components together in the same app?

JiriOndrusek:
> > We have to wait for the new release `quarkiverse-tika`
> 
> Or we propose to revert the upgrade in Camel. I assume the same issue exists there if you try to bring the `tika` & `pdf` components together in the same app?

Error starts to happen  if tika parses pdf -> other functionality should work even with pdf extension together.

Reverting of the Camel change and pstponing it until quarkiverse-tika would supports 3.x would be an easy solution

oscerd:
There is no point in postponing the change in Camel. Not for special reasons, but just because we cannot base the core development on what Quarkus/Quarkiverse does. It's dangerous and not healthy.

JiriOndrusek:
@oscerd I'll add some tests covering pdf into tika component, because I can not see pf file there - https://github.com/apache/camel/tree/main/components/camel-tika/src/test/resources

If some troubles emerges, we can discuss what to do next.
Does it sounds ok?

oscerd:
Yes, it is.

oscerd:
Also if you check in the SBOM for camel, pdfbox is used explicitly only in camel-pdf and camel-fop.

[...]

PR #6744, opened 2024-11-04
Documentation for 'Fop native failures due to pdfbox 3 upgrade' #1

### LINKED ISSUE #6743: Documentation for issue 5244 : Fop native failures due to pdfbox 3 upgrade
### Description

After upgrading to pdfbox 3, it seems the native build is not working anymore, as for the issues:  

[camel-quarkus : Fop native failures due to pdfbox 3 upgrade #5244](https://github.com/apache/camel-quarkus/issues/5244)

*Maybe it should be reported in documentation?*

I suggest that because at first I read the [camel-quarkus-fop documentation](https://camel.apache.org/camel-quarkus/next/reference/extensions/fop.html) and it seemed native profile was available.

Only after i tried and it did not work a searched for the open issue where it is stated otherwise : 
<https://github.com/apache/camel-quarkus/issues/5244>

As it is open since more than one year, maybe the fact should be noted in the limitations.

The pull request is on the [came-quarkus-fop] documentation as I found the issue when trying to use it.
Here is a [reproducer](https://github.com/fugerit79/camel-quarkus-examples/blob/1-quarkus-camel-fop-example/rest-json/NATIVE.md)

Do you agree?

Thanks in advance.

fugerit79:
fugerit79/camel-quarkus/issues/1
#5244
#5234

This PR's goal is to 
Fix #6743 

And document on [camel-quarkus-fop limitations page](https://camel.apache.org/camel-quarkus/next/reference/extensions/fop.html) the fact that native profile is currenlty disabled.

[X] An issue should be filed for the change unless this is a trivial change (fixing a typo or similar). One issue should ideally be fixed by not more than one commit and the other way round, each commit should fix just one issue, without pulling in other changes.
[X] Each commit in the pull request should have a meaningful and properly spelled subject line and body. Copying the title of the associated issue is typically enough. Please include the issue number in the commit message prefixed by #.
[X] The pull request description should explain what the pull request does, how, and why. If the info is available in the associated issue or some other external document, a link is enough.
[X] Phrases like Fix #<issueNumber> or Fixes #<issueNumber> will auto-close the named issue upon merging the pull request. Using them is typically a good idea.
[X] Contributor guide is your good friend: https://camel.apache.org/camel-quarkus/latest/contributor-guide.html

### DIFF
--- a/docs/modules/ROOT/pages/reference/extensions/fop.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/fop.adoc
@@ -57,6 +57,8 @@ ignored.
 * If custom fonts are used, font cache has to be disabled because of https://github.com/oracle/graal/issues/460[Graal VM issue #460].
 Please set the https://xmlgraphics.apache.org/fop/2.1/configuration.html[FOP configuration property] `use-cache` to `false`.
 
+NOTE: Native build for module *camel-quarkus-fop* is affected by the link:https://github.com/apache/camel-quarkus/issues/5244[Fop native failures due to pdfbox 3 upgrade #5244], native profile is not working and disabled until the issue is fixed.
+
 Please file an https://github.com/apache/camel-quarkus/issues/new[issue] if you are missing some specific output format
 in native mode.
 
--- a/extensions/fop/runtime/src/main/doc/limitations.adoc
+++ b/extensions/fop/runtime/src/main/doc/limitations.adoc
@@ -8,5 +8,7 @@ ignored.
 * If custom fonts are used, font cache has to be disabled because of https://github.com/oracle/graal/issues/460[Graal VM issue #460].
 Please set the https://xmlgraphics.apache.org/fop/2.1/configuration.html[FOP configuration property] `use-cache` to `false`.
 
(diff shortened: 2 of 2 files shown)


zhfeng:
Thanks @fugerit79 - can you run the following command locally to regen the docs module as well.
```
./mvnw -f extensions/fop/runtime/pom.xml clean install
```

fugerit79:
Thanks @zhfeng , the docs module has been generated.

Is there anything else I can do?

jamesnetherton:
This may actually not be needed in the next release. FOP in Camel 4.9.0 will use pdfbox 3.x.

fugerit79:
@jamesnetherton as you think, but I lost some time before understanding why the native profile was not working.
Is there any due date for the release? And are we sure the native profile will work again for camel-quarkus-fop?

If there are no proof that the native build will work again for sure, I would rather update the documentation and then update it again.

So I guess would better to document the issue right now.

jamesnetherton:
> And are we sure the native profile will work again for camel-quarkus-fop

I need to investigate it. But I don't see why it can't be made to work when we have the correct pdfbox version.

Anyway, I'll this for now until it's fixed.

jamesnetherton (approved):

PR #6751, opened 2024-11-06
Remove redundant HostUtils substitutions

jamesnetherton:
This was probably required in the distant past but is likely no longer required. Lets see...


aldettinger:
uncommitted changes in group_01, probably need a regen

aldettinger (approved):

PR #6781, opened 2024-11-13
[3.15.x] Upgrade QCXF to 3.15.3

zhfeng:


JiriOndrusek (approved):

aldettinger (approved):

ppalaga:
@zhfeng this is the idea: https://github.com/apache/camel-quarkus/pull/6782 

Instead of removing all bannedDependencyResources, we just remove findbugs from camel-quarkus-banned-dependencies.xml using your elegant technique with passing the `allow-findbugs` property via Maven filtering.

The question at the bottom line is whether we should not give up banning findbugs altogether, given we allow it somewhere. The original motivation was to prevent multiple versions of findbugs in the transitive closure, thus making the downstream packaging easier. I think the machinery to allow it somewhere and ban elsewhere starts to be too complicated. In other words I would not mind stopping to ban findbugs altogether.

jamesnetherton:
> I would not mind stopping to ban findbugs altogether

Agreed. The ban / unban stuff will just get increasingly harder to maintain.

zhfeng:
Thanks @ppalaga -  Yeah, I agree as well.

PR #6803, opened 2024-11-19
Upgrade Quarkus Amazon Services to 2.20.0

jamesnetherton:


jamesnetherton:
I'll fix up the failure - annoying as it means there is a native mode breakage if we don't release CQ 3.17 to the platform.

aldettinger (approved):

Issue #6866, opened 2024-12-12
[Quarkus LangChain4j 0.23.x] AI services should be resolvable by name


aldettinger:
### Describe the feature here

Ai service defined like:

```
@ApplicationScoped
@Named("aiServiceResolvedByName")
public interface AiServiceResolvedByName {
...
}
```

Should be invokable with bean syntax ala `.bean("aiServiceResolvedByName");`.

aldettinger:
Today, error below is produced:

```
Expected: is "AiServiceResolvedByName has been resolved"
  Actual: dev.langchain4j.memory.chat.MessageWindowChatMemory@328a59bf

	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
	at org.codehaus.groovy.reflection.CachedConstructor.invoke(CachedConstructor.java:73)
	at org.codehaus.groovy.runtime.callsite.ConstructorSite$ConstructorSiteNoUnwrapNoCoerce.callConstructor(ConstructorSite.java:108)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallConstructor(CallSiteArray.java:57)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callConstructor(AbstractCallSite.java:263)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callConstructor(AbstractCallSite.java:277)
	at io.restassured.internal.ResponseSpecificationImpl$HamcrestAssertionClosure.validate(ResponseSpecificationImpl.groovy:512)
	at io.restassured.internal.ResponseSpecificationImpl$HamcrestAssertionClosure$validate$1.call(Unknown Source)
	at io.restassured.internal.ResponseSpecificationImpl.validateResponseIfRequired(ResponseSpecificationImpl.groovy:696)
	at io.restassured.internal.ResponseSpecificationImpl.this$2$validateResponseIfRequired(ResponseSpecificationImpl.groovy)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
[...]

PR #6867, opened 2024-12-12
Add test for AI service resolution by name #6866

aldettinger:


zbendhiba (approved):
I like the invoking an AIService by name feature !

PR #6957, opened 2025-01-30
Update kubernetes.adoc

rhaetor:
Removed downstream extra content, which is no longer required.


oscerd (approved):

jamesnetherton:
Everything under `reference/extensions` is generated, so you'll also need to make the same change to the source:

https://github.com/apache/camel-quarkus/blob/main/extensions/kubernetes/runtime/src/main/doc/configuration.adoc?plain=1#L1-L4

jamesnetherton (approved):

PR #6995, opened 2025-02-11
Remove GitHub pull request template

jamesnetherton:
Not sure what others think. From my POV the PR template generates pointless noise.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,21 +1,3 @@
-<!--
-
-    Licensed to the Apache Software Foundation (ASF) under one or more
-    contributor license agreements.  See the NOTICE file distributed with
-    this work for additional information regarding copyright ownership.
-    The ASF licenses this file to You under the Apache License, Version 2.0
-    (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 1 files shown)


aldettinger:
No strong advice. It may be handy for new contributors so that old contributors won't need to explain ;)
Maybe a few infos would at least deserve to be stored in contributor guide ?

```
An issue should be filed for the change unless this is a trivial change (fixing a typo or similar). One issue should ideally be fixed by not more than one commit and the other way round, each commit should fix just one issue, without pulling in other changes.
Each commit in the pull request should have a meaningful and properly spelled subject line and body. Copying the title of the associated issue is typically enough. Please include the issue number in the commit message prefixed by #.
The pull request description should explain what the pull request does, how, and why. If the info is available in the associated issue or some other external document, a link is enough.
 Please run mvn process-resources -Pformat (and amend the changes if necessary) before sending the pull request.
```

jamesnetherton:
Thanks for the feedback. The main part of the noise is the comment header that got added recently. So I think I'll just remove it and leave everything else as-is.

aldettinger (approved):

PR #7028, opened 2025-02-19
Align dependencies shared with Quarkus Google Cloud Services for libraries-bom 26.50.0

jamesnetherton:
Motivation for this is:

If we run ahead of libs used by Quarkus Google Cloud Services, we hit issues in the Quarkus Platform.

I will try to see if they can upgrade, but there's an issue in the latest firestore libs that probably prevents it. So we probably have to wait until the next `libraries-bom` release.


oscerd (approved):

aldettinger (approved):
Make sense, thanks for useful comment @jamesnetherton :+1:

jamesnetherton:
Not sure why CXF tests decided to fail. But seems unrelated to this change.

Issue #7112, opened 2025-03-10
Intermittent failure of AzureServiceBusTest


jamesnetherton:
I've seen frequent failures on CI where one of the Service Bus containers does not seem to start successfully. Here's the error from the logs.

Full logs can be found [here](https://github.com/apache/camel-quarkus/actions/runs/13749076245/job/38447909093).

```
fail: Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService[103]
      Health check sqlserver with status Unhealthy completed after 14705.7236ms with message '(null)'
      Microsoft.Data.SqlClient.SqlException (0x80131904): A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: TCP Provider, error: 35 - An internal exception was caught)
       ---> System.Net.Sockets.SocketException (00000001, 11): Resource temporarily unavailable
         at System.Net.Dns.GetHostEntryOrAddressesCore(String hostName, Boolean justAddresses, AddressFamily addressFamily, Nullable`1 startingTimestamp)
         at System.Net.Dns.GetHostAddresses(String hostNameOrAddress, AddressFamily family)
         at Microsoft.Data.SqlClient.SNI.SNITCPHandle.GetHostAddressesSortedByPreference(String serverName, SqlConnectionIPAddressPreference ipPreference)+MoveNext()
         at Microsoft.Data.SqlClient.SNI.SNITCPHandle.Connect(String serverName, Int32 port, TimeoutTimer timeout, SqlConnectionIPAddressPreference ipPreference, String cachedFQDN, SQLDNSInfo& pendingDNSInfo)
         at Microsoft.Data.SqlClient.SNI.SNITCPHandle..ctor(String serverName, Int32 port, TimeoutTimer timeout, Boolean parallel, SqlConnectionIPAddressPreference ipPreference, String cachedFQDN, SQLDNSInfo& pendingDNSInfo, Boolean tlsFirst, String hostNameInCertificate, String serverCertificateFilename)
         at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
         at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, SqlCommand command, Boolean callerHasConnectionLock, Boolean asyncClose)
[...]

PR #7117, opened 2025-03-10
Refactored azure-servicebus emulator to not use docker compose

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/7112

I think that the failure might be caused by the fixed time-out for waiting until sql edge container sends responses. 
I refactored the docker-compose solution to 2 docker containers, when the wait period is more configurable.

Lets see if this fix helps to avoid intermittent failures.


jamesnetherton (approved):

jamesnetherton:
Lets merge and see how things go.

PR #7128, opened 2025-03-12
Fix dependency convergence failures with camel-quarkus-google-secret-…

zhfeng:
…manager


jamesnetherton (approved):

jamesnetherton:
Looking at the failure, seems this dependency might actually be required.

zhfeng:
Yes, it happens in the native build. Is it an optional dependency?

```
com.oracle.svm.core.util.UserError$UserException: Discovered unresolved type during parsing: com.google.api.gax.httpjson.HttpJsonTransportChannel
```

jamesnetherton:
> Yes, it happens in the native build. Is it an optional dependency?

Seem's it is a mandatory dependency. We could manage it ourselves in the BOM and have the version sync'd like:

```
<google-gax-httpjson.version>???</google-gax-httpjson.version><!-- @sync com.google.cloud:google-cloud-secretmanager:${google-cloud-secretmanager-version} dep:com.google.api:gax-httpjson -->
```

zhfeng:
OK, I will take a look.

PR #7134, opened 2025-03-13
Support Camel default route resource locations

jamesnetherton:
Two things fixed here:

* Loading of default route resource locations and setup of dev mode hot reload
* Native mode capability to resolve classpath resources from glob patterns

Opening as draft initially to see if anything breaks.


aldettinger (approved):

jamesnetherton:
Setting back to draft as there's a few more things I want to check & test before merging.

Issue #7188, opened 2025-03-26
Camel 4.11 - Add camel-dfdl extension


davsclaus:
### Describe the feature here

This is a new component in Camel 4.11. If native is too complex at start then a jvm only is appreciated.

igarashitm:
Working branch is here, which is currently based on camel-main branch
https://github.com/igarashitm/camel-quarkus/tree/7188

I'll reopen a PR once main branch is based on camel 4.11

jamesnetherton:
@igarashitm Camel 4.11.0 upgrade has been merged to `main`.

PR #7197, opened 2025-03-27
feat(extension): Add camel-dfdl JVM only extension

igarashitm:
Fixes: #7188

Opening as a draft - there is one issue on transitive dependencies (Scala)
For some reason, Scala 2.13 is brought in while Daffodil requires 2.12 (CNFE scala.Serializable). It doesn't happen in Camel upstream, `mvn dependency:tree` on `camel-dfdl` shows only Scala 2.12.

I added `dependencyManagement` here and found it makes tests pass, however it would be better to enforce it on extension side. No luck so far, even `mvn dependency:tree -Dverbose` doesn't reveal what is bringing Scala 2.13 in.
[`camel-main...igarashitm:camel-quarkus:7188?expand=1`#diff-7b389ad4ee](https://github.com/apache/camel-quarkus/compare/camel-main...igarashitm:camel-quarkus:7188?expand=1#diff-7b389ad4eedc6d5b2de065759c8d0648b616a3d2e76d7d00ae08cca734f0b83cR41-R51)

[x] An issue should be filed for the change unless this is a trivial change (fixing a typo or similar). One issue should ideally be fixed by not more than one commit and the other way round, each commit should fix just one issue, without pulling in other changes.
[x] Each commit in the pull request should have a meaningful and properly spelled subject line and body. Copying the title of the associated issue is typically enough. Please include the issue number in the commit message prefixed by #.
[x] The pull request description should explain what the pull request does, how, and why. If the info is available in the associated issue or some other external document, a link is enough.
[x] Phrases like Fix #<issueNumber> or Fixes #<issueNumber> will auto-close the named issue upon merging the pull request. Using them is typically a good idea.
[x] Please run mvn process-resources -Pformat (and amend the changes if necessary) before sending the pull request.
[x] Contributor guide is your good friend: https://camel.apache.org/camel-quarkus/latest/contributor-guide.html


jamesnetherton:
> For some reason, Scala 2.13 is brought in

Because `quarkus-bom` enforces that version and we cannot override it. Users would have to override it manually in `dependencyManagement` within their apps. I can't think of a better way of handling it at the moment.

igarashitm:
@jamesnetherton oh OK that makes sense, and agree, in that case there's no other option. We might want to document about it somewhere.

jamesnetherton:
> And the `influxdb-java` updating to 2.25 in GitHub workflow, it doesn't happen in my local...

TBH - I would wait to integrate this until after Camel 4.11 is upgraded on the `main` branch. `camel-main` is an endlessly moving target that makes adding changes like this a pain.

PR #7246, opened 2025-04-10
Add GitHub workflow to perform dependency convergence checks

jamesnetherton:
Opening as draft to get feedback on whether we think this is a good idea or not.

This adds a workflow to do dependency convergence checks against a 'super app' that declares all `org.apache.camel.quarkus` extension dependencies. Failures are reported in a GitHub issue.

Some preliminary results can be seen [here](https://gist.github.com/jamesnetherton/7d57e17613633b65095f0febb6ab8699).

It may help to spot potential issues. For example, at the above link you can see `opentelemetry-semconv` is flagged. This is because Quarkus switched to using a different `groupId` for that dependency, so the dep Camel Otel uses is no longer managed by `quarkus-bom`. Unfortunately we have not noticed this change.

### DIFF
--- a/.github/workflows/check-dependency-convergence.yml
+++ b/.github/workflows/check-dependency-convergence.yml
@@ -1,0 +1,177 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: Check Dependency Convergence
+
+on:
+  schedule:
+    # Run every sunday at 12PM
+    - cron: '0 12 * * 0'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+env:
+  LANG: en_US.UTF-8
+  MAVEN_OPTS: -Xmx3000m
+  CQ_MAVEN_ARGS: -V -ntp -e -Daether.connector.http.connectionMaxTtl=120
+  ISSUE_ID: 7348
+
+jobs:
+  check-dependency-convergence:
+    if: github.repository == 'apache/camel-quarkus'
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Set up JDK 17
+        uses: actions/setup-java@c5195efecf7bdfc987ee8bae7a71cb8b11521c00 # v4.7.1
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Set Build Info
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/check-dependency-convergence.yml
```diff
@@ -0,0 +1,76 @@
...
+concurrency:
+  group: ${{ github.ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+env:
+  LANG: en_US.UTF-8
+  MAVEN_OPTS: -Xmx3000m
+  CQ_MAVEN_ARGS: -V -ntp -e -Daether.connector.http.connectionMaxTtl=120
+  # TODO: Create a reporting issue
+  ISSUE_ID: ???
```
zhfeng, line 35: So this is the similar with the `camel-main` or `quarkus-main` building?

jamesnetherton: Yes, it will build the `main` branch, perform convergence checks against a generated project and report failures to a dedicated issue.

tooling/scripts/create-superapp.groovy
```diff
@@ -0,0 +1,97 @@
...
+def rootPom = new XmlParser().parse(new File('pom.xml'))
+def camelQuarkusVersion = rootPom.version.text()
+def quarkusVersion = rootPom.properties.'quarkus.version'?.text()
+def mavenEnforcerPluginVersion = rootPom.properties.'maven-enforcer-plugin.version'?.text()
+
+// Get org.apache.camel.quarkus dependencies from camel-quarkus-bom
+def cqBomPath = "${System.properties['user.home']}/.m2/repository/org/apache/camel/quarkus/camel-quarkus-bom/${camelQuarkusVersion}/camel-quarkus-bom-${camelQuarkusVersion}.pom"
+def bom = new XmlParser().parse(new File(cqBomPath))
+def cqBomDependencies = bom.dependencyManagement.dependencies?.dependency?.findAll {
+    it.groupId.text().startsWith('org.apache.camel.quarkus') && !it.artifactId.text().contains('-support')
```
ppalaga, line 29: In case this still selects too many artifacts, it might be worthwhile to try selecting all `org.apache.camel.quarkus:camel-quarkus-*-deployment` artifacts and then remove `-deployment` from artifactId. That should find only extensions.

ppalaga:
Very useful, thanks @jamesnetherton! I was always scared to do this for all extensions, because I feared too many issues, some of those perhaps not solvable at all... The list in your gist is quite long

jamesnetherton:
> I feared too many issues, some of those perhaps not solvable at all... The list in your gist is quite long

Yes, exactly. I wonder if we should just evaluate the results and choose what we think is worth fixing? I don't really want to end up having to manage a million extra third party deps in our BOM to resolve all of these issues.

ppalaga:
I agree, we should take care for some low hanging fruits, but not spend too much time on fixing all. 
I think what matters most is that a super-app consisting of extensions known for not having any issues should not shrink over time. 
So we might introduce a manually curated super-app that must work.
If we have time, we could add some additional functionality, that would randomly add some of the extensions that did not work in the past to the superapp and fail the workflow if it works unexpectedly. In that case that random extension can be added to the manually curated superapp. WDYT?

jamesnetherton:
So start superapp from a known good baseline and introduce some randomness to check for failures?

Yes, that sounds sensible.

aldettinger:
The general idea is very good :+1:  Not sure about the randomness, still worth experimenting and evaluate over tiùme.

jamesnetherton:
I finally got back to working on this.

> I feared too many issues, some of those perhaps not solvable at all... The list in your gist is quite long

I rethought this and decided it'd be simpler to just have some GAV excludes for stuff that we want to ignore. So I hacked up a script that runs in the workflow for that.

This leaves us with only a single [failure for otel](https://github.com/apache/camel-quarkus/issues/7298). Which I intend to fix shortly. Afterwards, we'll be starting at a 'clean' baseline.

We should probably also fix issues for the `com.google.*` deps TBH. I can create an issue(s) for that.

jamesnetherton:
Let's merge and see how this goes. If it gets annoying or becomes not useful then we can rework / remove it.

ppalaga:
Great work, thanks a lot, @jamesnetherton!

PR #7261, opened 2025-04-14
Fix #7254 to support named entityManagerFactory in camel-quarkus-jpa

### LINKED ISSUE #7254: Camel Quarkus JPA can not work with the named persistence unit
### Bug description

When `quarkus-hibernate-orm` is config with a named persistence unit like
```
quarkus.hibernate-orm."test".packages=org.acme.entity
quarkus.hibernate-orm."test".datasource=test_db
quarkus.hibernate-orm."test".log.sql=true
quarkus.hibernate-orm."test".database.generation=none

```

The `entityManagerFactory=#test` does not work. It can not find the righ `EntityManagerFactory`.

zhfeng:

### DIFF
--- a/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
+++ b/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import io.quarkus.arc.Arc;
 import io.quarkus.arc.ArcContainer;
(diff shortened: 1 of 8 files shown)


oscerd (approved):

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/NamedAnnotationLiteral.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package org.apache.camel.quarkus.core;
+
+import jakarta.enterprise.util.AnnotationLiteral;
+
+/**
+ * An interface for annotation literals that are named
+ */
+public interface NamedAnnotationLiteral {
+    AnnotationLiteral getLiteral(String name);
```
ppalaga, line 25: Some kinds of beans might require that we select them through multiple qualifiers, no? To solve that possible future use case, this method should rather return an array of `AnnotationLiteral`. But because, I guess, there is no such case in sight, it is fine to keep the current solution as is and change it only once we hit such a case. I hope everybody agrees?

ppalaga (approved):

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/RuntimeBeanRepository.java
```diff
@@ -80,6 +81,17 @@ private static <T> Set<T> getReferencesByType(BeanManager manager, Class<T> type
...
+        beans = beans.stream()
+                .filter(bean -> {
+                    if (bean instanceof InjectableBean injectableBean) {
+                        return !injectableBean.getKind().equals(InjectableBean.Kind.SYNTHETIC)
+                                || bean.getTypes().contains(type);
+                    } else {
+                        return true;
+                    }
+                }).collect(Collectors.toSet());
+
```
zhfeng, line 94: @jamesnetherton  @ppalaga I replace to check the beans which are generated by the extensions are `SYNTHETIC`. We don't need to introduce a `NamedAnnotationLiteral` build item. The `EntityManagerFactory` class from `quarkus-hibernete-orm` is also with `@Named` annotation.

ppalaga (approved):
Thanks @zhfeng!

PR #7305, opened 2025-04-25
feat(observability-services): use management port

squakez:
This PR enables the management endpoints on the management port 9876. This is aligning the default used for all runtimes. Beside that, there is a commit required to fix the platform-http component after a change on Camel core.


jamesnetherton:
You'll probably need to adjust the tests so that they use the correct port when accessing the health and metrics endpoints.

squakez:
@jamesnetherton I've fixed the tests. I'll have a look to see if there is any more testing error.

jamesnetherton:
Looks good to me. I merged it manually.

PR #7378, opened 2025-05-19
Remove rest-assured groovy exclusions

jamesnetherton:
Probably a relic of the past. It breaks on `quarkus-main` after Quarkus upgraded `rest-assured`. Hence I'm removing it.


aldettinger (approved):
oh, good catch.

Issue #7340, opened 2025-05-07
Remove the substitution in salesforce extension once org.cometd.java:cometd-* is based on jetty 12.0.20+


JiriOndrusek:
The library `org.cometd.java:cometd-*` (currently 8.0.7) is compiled agains jetty version 12.0.17 (see [code](https://github.com/cometd/cometd/blob/8.0.7/pom.xml#L96)).
Camel upgraded jetty to 12.0.20 in [commit](https://github.com/apache/camel/commit/3a85ce1c5fbd2332ac2d5fde14b7b7fc86ba322e).

Since that the salesforce test throws a native error 
```
Error: Discovered unresolved method during parsing: org.eclipse.jetty.util.ajax.JSON$StringSource.<init>(java.lang.String). This error is reported at image build time because class org.cometd.common.JettyJSONContext is registered for linking at image build time by command line and command line.
Error encountered while parsing org.cometd.common.JettyJSONContext.parse(JettyJSONContext.java:49) 
Parsing context:
   at BufferingJSONAsyncParserSubstitute.complete(BufferingJSONAsyncParser.java:41)
   at org.cometd.client.http.jetty.JettyHttpClientTransport$ResponseListener.onComplete(JettyHttpClientTransport.java:285)
```

Once the newer comethd based on jetty 12.0.20+ is used, the [substitution](https://github.com/apache/camel-quarkus/commit/905858c302d7c5872e34e592358c4955f04d75e1#diff-fb79eb651cec2eba10821a51e1c223da3570ab16bb9627520aa38bd9f0c5b54d) can be removed.

jamesnetherton:
@JiriOndrusek we can probably remove the workaround. Camel has upgraded to cometd 8.0.8.

https://github.com/cometd/cometd/blob/8.0.8/pom.xml#L90

JiriOndrusek:
Thanks for the information! I'll take care of that

JiriOndrusek:
Fixed in `camel-maim` in commit https://github.com/apache/camel-quarkus/commit/701f4b467ca139e318054f6d81477a58fdfdcaa5.

PR #7407, opened 2025-06-02
[3.20.x] Upgrade Camel to 4.10.5

jamesnetherton:


jamesnetherton:
Seems Camel 4.10.5 has some misalignment with Jetty and Cometd. So we see this issue https://github.com/apache/camel-quarkus/issues/7340. We'll need to work around it.

Issue #7259, opened 2025-04-12
Camel 4.12 - Add camel-weaviate extension


davsclaus:
### Describe the feature here

https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-21943

jamesnetherton:
Done in https://github.com/apache/camel-quarkus/pull/7497.

PR #7420, opened 2025-06-04
[camel-main] Weaviate extension

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/7259

Extension requires camel 4.13.0. (https://github.com/apache/camel/pull/18272 and https://github.com/apache/camel/pull/18251)

Tests can run  via docker image or against real API.


JiriOndrusek:
@jamesnetherton I'd keep this PR(draft) opened until we merge the camel-main into main, to avoid separating this changes in the camel-main maintenance. WDYT?

integration-tests/weaviate/src/main/java/org/apache/camel/quarkus/component/weaviate/it/WeaviateResource.java
```diff
@@ -0,0 +1,113 @@
...
+        if (body instanceof List) {
+            body = ((List<?>) body).stream().map(o -> o instanceof Double ? ((Double) o).floatValue() : o)
+                    .collect(Collectors.toList());
+        }
+
+        Exchange response = context.createFluentProducerTemplate()
+                .to(getUrl())
+                .withBody(body)
+                .withHeaders(headers)
+                .request(Exchange.class);
```
jamesnetherton: General comment. No need to change things unless you want to. But...

I think it's better to get into the habit of writing actual Camel routes within a `RouteBuilder` class, instead of using the `ProducerTemplate` to send messages directly to the component under test. That way it aligns better with what users are more likely to do in their own apps.

jamesnetherton (approved):
Nice work - LGTM

jamesnetherton:
> @jamesnetherton I'd keep this PR(draft) opened until we merge the camel-main into main, to avoid separating this changes in the camel-main maintenance. WDYT?

Yeah, I'm fine with that if it makes life easier to maintain `camel-main`. The alternative might be to move this commit before the Camel upgrade commit. Choose whatever is easier for you.

integration-tests/weaviate/src/main/java/org/apache/camel/quarkus/component/weaviate/it/WeaviateResource.java
JiriOndrusek: I understand. TBH I switched from routes to fluent producer and back several time, as I was not able to decide what is better. I'll rework the code to use routes, as your point makes sense to me.

JiriOndrusek: fixed

JiriOndrusek:
Lets merge it, with some re-thinking I don't see this change making maintenance of `camel-main` harder.

PR #7431, opened 2025-06-10
Support resolving beans by name and qualifiers

jamesnetherton:
I ended up refactoring `RuntimeBeanRepository` more than I would usually like.

* Reorganised things so that the core public methods are at the beginning
* Simplify bean lookups and rely more on the Arc container APIs instead of `BeanManager`
* Added support for resolving beans by name using qualifiers

Opening in draft initially to see if anything breaks.

### DIFF
--- a/integration-test-groups/foundation/bean/src/test/java/org/apache/camel/quarkus/component/bean/BeanTest.java
+++ b/integration-test-groups/foundation/bean/src/test/java/org/apache/camel/quarkus/component/bean/BeanTest.java
@@ -246,9 +246,8 @@ public class BeanTest {
         RestAssured.given()
                 .get("/bean/allBeanInstances")
                 .then()
-                .body("size()", is(5))
+                .body("size()", is(4))
                 .body(allOf(
-                        containsString("defaultBean"),
                         containsString("overridingBean"),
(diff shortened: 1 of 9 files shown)


integration-test-groups/foundation/bean/src/test/java/org/apache/camel/quarkus/component/bean/BeanTest.java
```diff
@@ -246,9 +246,8 @@ public void notReducedTest() {
         RestAssured.given()
                 .get("/bean/allBeanInstances")
                 .then()
-                .body("size()", is(5))
+                .body("size()", is(4))
                 .body(allOf(
-                        containsString("defaultBean"),
```
jamesnetherton, line 251: Not sure why previously we had this behaviour. When fetching all beans for a type it would return a default bean, even if it had been overridden. Which seems wrong to me.

With this change, that wont happen. Hopefully it does not break anything.

aldettinger: Probably a left over, let's see the test suite.

zhfeng (approved):

PR #7458, opened 2025-06-20
Rework upload-source.sh to fetch sources and signatures from repository.apache.org

jamesnetherton:
Avoids local (re)signing of sources, which with the current scripts, can potentially end up using the wrong key.


aldettinger:
Good catch, not sure why releasers would be resigning again before this pr anyway.

PR #7553, opened 2025-07-28
Avoid using LaunchMode static method

gsmet:
It will be made non-static in 3.26 for consistency with the other methods.
While this will raise a warning with earlier version, it will work with both new and old versions.
Note that it is not binary compatible though, so we will need a 3.26 release - which is probably planned anyway.


jamesnetherton (approved):
Thanks!

PR #7563, opened 2025-07-30
Stabilize throttle test

llowinge:


jamesnetherton:
Native mode failure:

```
java.lang.ClassNotFoundException: org.apache.camel.processor.ThrottlerRejectedExecutionException
```

`ThrottlerRejectedExecutionException` needs to be registered for reflection. I forget if we typically just do this in the test or configure it in our extensions so that it works automatically OOTB. I'll check...

jamesnetherton:
I think we can add `ThrottlerRejectedExecutionException` here:

https://github.com/apache/camel-quarkus/blob/baceec93b1478ba91889c466f9115e843443d2b5/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelNativeImageProcessor.java#L76

llowinge:
After discussion with @jamesnetherton we will not add that exception as default. Only if user want to use that exception, it should be manually registered for native, given class registration is not for free and without usage it doesn't make sense.

PR #7566, opened 2025-07-31
Deprecate / remove some redundant config and test modules

jamesnetherton:
Some tidying & depreciation of stuff that (IMO) is now redundant.

**Deprecate RoutesDiscoveryConfig, CamelRoutesLoaderBuildItems and remove redundant tests**

AFAIK, this was only ever required for camel-k. Since the camel-k-runtime is deprecated in favour of the plain CQ runtime, I think we can deprecate this config and eventually remove it in CQ 4.x.

**Deprecate CamelRoutesCollectorBuildItem and simplify testing**

CamelRoutesCollectorBuildItem is probably a relic of the past. AFAIK nobody is producing their own RoutesCollector at build time to override the default & I'm not sure we should even be supporting that anyway.

**Remove redundant custom-log-component tests**

It was just checking a component instance produced at build time & bound to the registry was usable. We already do this in various places (vertx, quartz and many others). 

**Simplify testing of MainListener**

I didn't see the need for a dedicated itest module for this. We already configure a custom MainListener in the core which gets tested everywhere.


aldettinger (approved):
Yeah, it's probably time to deprecate. And it could be removed later on so that users have time to give feedback if needed.

oscerd (approved):

PR #7569, opened 2025-07-31
Rebalance native CI tests group 8

jamesnetherton:


jamesnetherton:
Hmmm - this forces group-3 to be slowest 🙃.

I'll set to draft and see if we can rebalance most of the groups.

PR #7619, opened 2025-08-11
Remove not needed Infinispan exclusions which was needed for differen…

llowinge:
…t versions of Infinispan between Camel and Quarkus


jamesnetherton:
You'll need to update the generated BOMs by running `mvn clean install` in the `poms/bom`module.

llowinge:
Thanks @jamesnetherton !

Issue #7396, opened 2025-05-26
Debezium provides modules for Quarkus/native support


JiriOndrusek:
### Describe the feature here

The debezium [enhancement](https://github.com/debezium/debezium/pull/6392)  introduces a module(s) with the native support for quarkus. As soon as those modules are released, it might be possible to remove native code from debezium extensions and use those modules instead.  Also a lot of tickets about debezium failures in nativecould be closed (some from the [query](https://github.com/apache/camel-quarkus/issues?q=is%3Aissue%20state%3Aopen%20debezium))

jamesnetherton:
Since we have #7634 I think we can close this one.

PR #7634, opened 2025-08-14
[camel-main] Fixed debezium native (except Oracle)

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/7396 

Follow-up issue (about debezium oracle) is https://github.com/apache/camel-quarkus/issues/7638

### DIFF
--- a/extensions-support/debezium/deployment/src/main/java/org/apache/camel/quarkus/support/debezium/deployment/DebeziumSupportProcessor.java
+++ b/extensions-support/debezium/deployment/src/main/java/org/apache/camel/quarkus/support/debezium/deployment/DebeziumSupportProcessor.java
@@ -16,6 +16,33 @@
  */
 package org.apache.camel.quarkus.support.debezium.deployment;
 
+import io.debezium.connector.common.BaseSourceTask;
+import io.debezium.embedded.async.ConvertingAsyncEngineBuilderFactory;
+import io.debezium.engine.DebeziumEngine;
+import io.debezium.pipeline.notification.channels.LogNotificationChannel;
+import io.debezium.pipeline.notification.channels.SinkNotificationChannel;
+import io.debezium.pipeline.notification.channels.jmx.JmxNotificationChannel;
+import io.debezium.pipeline.signal.actions.StandardActionProvider;
+import io.debezium.pipeline.signal.channels.FileSignalChannel;
+import io.debezium.pipeline.signal.channels.KafkaSignalChannel;
+import io.debezium.pipeline.signal.channels.SourceSignalChannel;
+import io.debezium.pipeline.signal.channels.jmx.JmxSignalChannel;
+import io.debezium.pipeline.signal.channels.process.InProcessSignalChannel;
+import io.debezium.pipeline.txmetadata.DefaultTransactionMetadataFactory;
+import io.debezium.schema.SchemaTopicNamingStrategy;
+import io.debezium.snapshot.lock.NoLockingSupport;
+import io.debezium.snapshot.mode.AlwaysSnapshotter;
+import io.debezium.snapshot.mode.ConfigurationBasedSnapshotter;
+import io.debezium.snapshot.mode.InitialOnlySnapshotter;
+import io.debezium.snapshot.mode.InitialSnapshotter;
+import io.debezium.snapshot.mode.NeverSnapshotter;
+import io.debezium.snapshot.mode.NoDataSnapshotter;
+import io.debezium.snapshot.mode.RecoverySnapshotter;
+import io.debezium.snapshot.mode.SchemaOnlyRecoverySnapshotter;
+import io.debezium.snapshot.mode.SchemaOnlySnapshotter;
+import io.debezium.snapshot.mode.WhenNeededSnapshotter;
+import io.debezium.snapshot.spi.SnapshotLock;
+import io.debezium.storage.file.history.FileSchemaHistory;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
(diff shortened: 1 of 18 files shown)


extensions-support/debezium/deployment/src/main/java/org/apache/camel/quarkus/support/debezium/deployment/DebeziumSupportProcessor.java
```diff
@@ -24,45 +50,104 @@
...
-        nativeImage.produce(
+    @BuildStep()
+    void registerNativeImageResources(BuildProducer<NativeImageResourceBuildItem> resources) {
+        resources.produce(new NativeImageResourceBuildItem(
+                "META-INF/services/io.debezium.embedded.async.ConvertingAsyncEngineBuilderFactory"));
+        resources.produce(
                 new NativeImageResourceBuildItem("META-INF/services/io.debezium.engine.DebeziumEngine$BuilderFactory"));
-    }
+        resources.produce(new NativeImageResourceBuildItem("META-INF/services/io.debezium.spi.snapshot.Snapshotter"));
+        resources.produce(new NativeImageResourceBuildItem("META-INF/services/io.debezium.spi.snapshot.Snapshotter"));
```
jamesnetherton: Seems we have `META-INF/services/io.debezium.spi.snapshot.Snapshotter` duplicated.

In general, we should perhaps use `ServiceProviderBuildItem` instead of `NativeImageResourceBuildItem`, passing in the Service class name from Class.getName(). That way you detect breakages if APIs change / get removed etc.

extensions/debezium-mongodb/deployment/src/main/java/org/apache/camel/quarkus/component/debezium/mongodb/deployment/DebeziumMongodbProcessor.java
```diff
@@ -32,8 +37,18 @@ FeatureBuildItem feature() {
...
+                MongoDbConnector.class,
+                MongoDbConnectorTask.class,
+                MongoDbSourceInfoStructMaker.class,
+                DefaultMongoDbAuthProvider.class,
+                SelectAllSnapshotQuery.class).build());
+    }
+
+    @BuildStep
+    void addDependencies(BuildProducer<IndexDependencyBuildItem> indexDependency) {
+        indexDependency.produce(new IndexDependencyBuildItem("io.debezium", "debezium-connector-mongodb"));
```
jamesnetherton, line 52: Question: There is no Jandex usage in `DebeziumMongodbProcessor`, so where do we use these indexed classes. Is it in `DebeziumSupportProcessor`?

debezium-grouped-native.log
```diff
@@ -0,0 +1 @@
+[INFO] Scanning for projects...
```
jamesnetherton: I guess we can remove this.

jamesnetherton (commented):
Thanks so much for looking into this and getting native working again 💪.

I left a few comments / questions.

debezium-grouped-native.log
JiriOndrusek: sure, I missed that:)

extensions-support/debezium/deployment/src/main/java/org/apache/camel/quarkus/support/debezium/deployment/DebeziumSupportProcessor.java
JiriOndrusek: I'll refactor it as you suggest

extensions/debezium-mongodb/deployment/src/main/java/org/apache/camel/quarkus/component/debezium/mongodb/deployment/DebeziumMongodbProcessor.java
JiriOndrusek: yes, correct

extensions-support/debezium/deployment/src/main/java/org/apache/camel/quarkus/support/debezium/deployment/DebeziumSupportProcessor.java
JiriOndrusek: @jamesnetherton I've just relized that the `META-INF/services/io.debezium.embedded.async.ConvertingAsyncEngineBuilderFactory` is a better sokution.
In case the app contains only `debezium-postgresql' the file contains i.e.
```
io.debezium.connector.postgresql.snapshot.lock.NoSnapshotLock
io.debezium.connector.postgresql.snapshot.lock.SharedSnapshotLock
```
the same file for debezium oracle contains
```
io.debezium.connector.oracle.snapshot.lock.NoSnapshotLock
io.debezium.connector.oracle.snapshot.lock.SharedSnapshotLock
```

therefore we an keel those `META-INF` in debezium-support or we can list all services for each flavor of debezium.
What do you prefer?
I'd say that keeping meta-inf files (as is) would cause small number of issues, I'd say....

jamesnetherton: We could maybe have common code in `debezium-support` and use `ServiceProviderBuildItem.allProvidersFromClassPath` for `ConvertingAsyncEngineBuilderFactory ` etc. That way it'd work for both Postgres and Oracle.

https://github.com/quarkusio/quarkus/blob/main/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ServiceProviderBuildItem.java#L78C44-L78C69

I don't want to over complicate things. So if you think it's better to keep things as-is, then lets do that and we can optimize later.

JiriOndrusek: I've created an issue (with the follow up - because of the oracle - https://github.com/apache/camel-quarkus/issues/7638
and I added the instruction to refactor the service items  during the follow-up works. Is it ok for you?

jamesnetherton: > Is it ok for you

Ok 👍.

debezium-grouped-native.log
JiriOndrusek: removed

JiriOndrusek:
@jamesnetherton  
I updated the PR with the native support of mssql, mysql and oracle.
Though there is a small issue with oracle when the `debezium-grouped` is executed, reported as https://github.com/apache/camel-quarkus/issues/7638
I think that this PR can be merged (once CI confirms the status)

JiriOndrusek:
@jamesnetherton There is one thing I'm not sure about:
- the mssql test requires MySQL connector-j(see https://github.com/apache/camel-quarkus/blob/main/integration-test-groups/debezium/mssql/pom.xml#L51)
- the library does not work in the native as is. I had to do som substitution and runtime registration.

I don't think that the code belongs to debezium-mssql, so I created a new module `extesnion-support/mysql-connector`.
I'm not sure about the location, If you think that there might be a better place, I'll refactor it.

jamesnetherton:
> * the mssql test requires MySQL connector-j(see https://github.com/apache/camel-quarkus/blob/main/integration-test-groups/debezium/mssql/pom.xml#L51)

Oh, I messed up there 😅. You can remove it from mssql (please do it on the `main` branch) where it should not be required. It's only for the MySQL bits. I only want to add more -support modules unless its really necessary.

All of the native support for it should come from `quarkus-jdbc-mysql`.

jamesnetherton:
Just noticed there's changes to `ZipfileProcessor.java` in this PR. Are they still needed? If so, they should be in their own commit.

Once that's solved, then I think this is ok to merge. Might be best to do that before the nightly job tries to auto update `camel-main` & create conflicts etc.

JiriOndrusek:
> Just noticed there's changes to `ZipfileProcessor.java` in this PR. Are they still needed? If so, they should be in their own commit.
> 
> Once that's solved, then I think this is ok to merge. Might be best to do that before the nightly job tries to auto update `camel-main` & create conflicts etc.

My bad, the zipfile change should be part of the https://github.com/apache/camel-quarkus/commit/41195dcb2bade7d1ac7724beabd8a7f1aebdb7de.

I'll fix that.

JiriOndrusek:
It's probably even easier, the change is not required at all. All required is already in https://github.com/apache/camel-quarkus/commit/41195dcb2bade7d1ac7724beabd8a7f1aebdb7de
(this was just a dead end, which somehow ended in this PR)

JiriOndrusek:
Zipfile bits are removed

JiriOndrusek:
CI is green🎉

PR #7660, opened 2025-08-20
Fix order of applying system properties to camel main

llowinge:


llowinge:
The test passed 30x in my personal GH action https://github.com/llowinge/camel-quarkus/actions/runs/17091967449/job/48468390017

Issue #7638, opened 2025-08-14
Debezium: oracle native test fails if is executed as a part of debezium-grouped


JiriOndrusek:
With the PR https://github.com/apache/camel-quarkus/pull/7634, all debezium-services work in the native mode if executed as standalone.
If debezium-grouped is started in the native, the poracle one fails with:
```
Caused by: java.sql.SQLException: No suitable driver found for jdbc:oracle:thin:@localhost:32928/FREE
	at java.sql@21.0.8/java.sql.DriverManager.getConnection(DriverManager.java:708)
	at java.sql@21.0.8/java.sql.DriverManager.getConnection(DriverManager.java:191)
	at io.debezium.jdbc.JdbcConnection.lambda$patternBasedFactory$0(JdbcConnection.java:198)
	at io.debezium.jdbc.JdbcConnection$ConnectionFactoryDecorator.connect(JdbcConnection.java:136)
	at io.debezium.jdbc.JdbcConnection.establishConnection(JdbcConnection.java:925)
	at io.debezium.jdbc.JdbcConnection.connection(JdbcConnection.java:909)
	at io.debezium.jdbc.JdbcConnection.connection(JdbcConnection.java:903)
	at io.debezium.jdbc.JdbcConnection.queryAndMap(JdbcConnection.java:647)
	at io.debezium.jdbc.JdbcConnection.queryAndMap(JdbcConnection.java:520)
	at io.debezium.connector.oracle.OracleConnection.resolveOracleDatabaseVersion(OracleConnection.java:160)
```

This should be solved.

+ refactor the service providers according the https://github.com/apache/camel-quarkus/pull/7634#discussion_r2276159844

JiriOndrusek:
I started investigating the issue. There is an interesting fact:

- Oracle fails when is grouped with any other debezium flavor
- Addition of another debezium extension is enough for the failure. You can for example add 

```
        <dependency>
            <groupId>org.apache.camel.quarkus</groupId>
            <artifactId>camel-quarkus-debezium-postgres</artifactId>
        </dependency>
````
into `integration-test-groups/debezium/oracle` and the native execution fails with the error from summary

The problem seems to be similar to https://github.com/quarkusio/quarkus/issues/19055

**Comment by JiriOndrusek** [comment]
The same behavior happens also if 
```
     <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-jdbc-postgresql</artifactId>
        </dependency>
```
[...]

PR #7665, opened 2025-08-20
Fixed Oracle native failure in grouped tests

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/7638

This is probably not the final solution. The simple workaround is provided (see limitations.adoc and comment in `DebeziumOracleProcessor`.

I think, that at least the same workaround should be implementable as a part of the extension and should not require any user manual change. (Unfortunately all my  different attempts failed)


JiriOndrusek:
Thanks @jamesnetherton! I fixed the PR in a way that there is no action required for the user.

PR #7685, opened 2025-09-01
Adds Opensearch tests on JVM mode.

svkcemk:


oscerd (approved):

integration-tests-jvm/opensearch/src/main/java/org/apache/camel/quarkus/component/opensearch/it/OpensearchRoutes.java
```diff
@@ -0,0 +1,50 @@
...
+package org.apache.camel.quarkus.component.opensearch.it;
+
+import org.apache.camel.builder.RouteBuilder;
+
+public class OpensearchRoutes extends RouteBuilder {
+
+    @Override
+    public void configure() throws Exception {
+        from("direct:indexDoc")
+                .toD("opensearch://opensearch?operation=Index");
```
jamesnetherton: Nitpick - I guess we don't need `toD` for anything here. I don't see any dynamic URIs being used to we can just use the standard `.to(..)` for everything.

svkcemk: Thanks @jamesnetherton I agree I just missed!

PR #7715, opened 2025-09-08
Do not rely on Quarkus dev service when testing as we start our own container

ppalaga:


integration-test-groups/jdbc/mysql/src/main/resources/application.properties
```diff
@@ -21,5 +21,6 @@
 
 quarkus.datasource.mysql.db-kind=mysql
 quarkus.datasource.mysql.jdbc.max-size=8
+quarkus.datasource.devservices.enabled=false
```
jamesnetherton, line 24: Does this work when the `jdbc-grouped` itest module is run? All of the 'grouped' module `application.properties` will get combined, so this would effectively disable dev services for all of those JDBC tests.

ppalaga: Good point. Given that all tests passed, I'd assume, we do not rely on any of those DB dev services. 

I am seeing `<quarkus.datasource.mysql.devservices.enabled>false</quarkus.datasource.mysql.devservices.enabled>` in poms. I saw the dev container starting though. `quarkus.datasource.mysql.devservices.enabled` does not perhaps work?
I do not have much time to dig deeper and I'd be thankful if you @jamesnetherton or somebody else could come up with some solution.

Actually, I hit this issue when running the tests on a host exceeding dockerhub limits. Those default Quarkus dev containers seem to be pulled from there. So we should perhaps suggest Quarkus not to be pulling from there by default.

jamesnetherton: If the tests pass, maybe we just merge it? WDYT?

I don't mind following up later to see if we can improve it a bit.

> host exceeding dockerhub limits

Probably a discussion to be had elsewhere, but it should really be proxing to a mirror where the images are already cached & not pulling directly from dockerhub on each build.

PR #7723, opened 2025-09-10
Do not rely on Postgres dev service because we have our own TestResource

ppalaga:


oscerd (approved):

jamesnetherton:
> we have our own TestResource

For the otel tests, we don't. Not sure if it makes sense to add one. We only set up a DB to test JDBC driver telemetry works with Camel JDBC.

ppalaga:
Yeah, thanks, I noticed all of that too, when I saw the error messages. Next time, I'll have to think more before sending a PR. Sorry for the spam.

PR #7769, opened 2025-09-24
Add langchain4j-embeddings native support

jamesnetherton:


jamesnetherton:
Looks like I need to make the native binary lookup smarter for Linux...

oscerd (approved):

PR #7772, opened 2025-09-24
Add retry logic to Debezium Oracle container startup

jamesnetherton:
Speculative fix for frequent debezium-oracle test failures. For some reason the default 5 minute startup timeout isn't always enough.


jamesnetherton:
Raising the timeout didn't work, so I added some basic retry logic for container startup.

PR #7838, opened 2025-10-15
Sanitize storing of user-supplied github context data in pre-build-checks step

jamesnetherton:


.github/workflows/ci-build.yaml
```diff
@@ -128,8 +128,8 @@ jobs:
               echo "continue-build=true" >> $GITHUB_OUTPUT
             else
               mkdir ./dependabot-pr
-              echo ${{ github.head_ref }} > ./dependabot-pr/BRANCH_REF
-              echo ${{ github.event.pull_request.head.sha }} > ./dependabot-pr/PR_HEAD_SHA
+              printf '%s\n' "${{ github.head_ref }}" > ./dependabot-pr/BRANCH_REF
+              printf '%s\n' "${{ github.event.pull_request.head.sha }}" > ./dependabot-pr/PR_HEAD_SHA
```
ppalaga, line 132: Is this supposed to be a protection against some sort of an attack? Why is printf more safe than echo?

jamesnetherton: I thought maybe to guard against being able to sneak in `\` sequences that could potentially alter the output. But now that I think about it more. I don't think `head_ref` or `sha` are vulnerable to such things.

PR #7865, opened 2025-10-22
Add tests for transformer features and support Transformer in native mode

### LINKED ISSUE #5478: Add support & test coverage for `transformer` features
### Description

E.g the functionality documented in:

https://camel.apache.org/manual/transformer.html

It's likely some things will not work OOTB in native mode. Like `@DataTypeTransformer` classpath scanning. We probably need to do such processing at build time.

I added some initial basic tests here:

https://github.com/apache/camel-quarkus/tree/main/integration-test-groups/foundation/transformer

gansheer:
Closes #5478

Improve test coverage for `transformer` features:
* Custom transformer (.withJava())
* DataFormat transformer (.withDataFormat())
* Endpoint transformer (.withUri())
* @DataTypeTransformer annotation (unnamed and named)

Add support in native mode with discovery of subclasses of org.apache.camel.spi.Transformer .

Note: I am not sure if I should move the registration code for reflection in `CamelNativeImageProcessor` or use something like `@BuildStep(onlyIf = NativeOrNativeSourcesBuild.class)`. Not sure what the best practice is in this case so I kept it in `Camel Processor`, don't hesitate to comment review on this.

### DIFF
--- a/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelProcessor.java
+++ b/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelProcessor.java
@@ -43,6 +43,8 @@ import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.ApplicationArchivesBuildItem;
 import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.NativeImageResourceBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+import io.quarkus.deployment.pkg.steps.NativeOrNativeSourcesBuild;
 import io.quarkus.maven.dependency.ArtifactKey;
 import io.quarkus.runtime.RuntimeValue;
 import io.smallrye.common.annotation.Identifier;
@@ -108,6 +110,8 @@ class CamelProcessor {
             "org.apache.camel.Predicate");
     private static final DotName CONVERTER_TYPE = DotName.createSimple(
             "org.apache.camel.Converter");
+    private static final DotName TRANSFORMER_TYPE = DotName.createSimple(
+            "org.apache.camel.spi.Transformer");
 
     private static final Set<DotName> UNREMOVABLE_BEANS_TYPES = CamelSupport.setOf(
             ROUTES_BUILDER_TYPE,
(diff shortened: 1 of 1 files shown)


extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelProcessor.java
```diff
@@ -431,6 +432,37 @@ CamelComponentNameResolverBuildItem componentNameResolver(
...
+                .collect(Collectors.toSet());
+
+        if (!transformerClasses.isEmpty()) {
+            LOGGER.debug("Found @DataTypeTransformer annotated classes: {}", transformerClasses);
+            packageScanClass.produce(new CamelPackageScanClassBuildItem(transformerClasses));
+
+            // Register transformer classes for reflection so they can be instantiated at runtime
+            transformerClasses.forEach(className -> reflectiveClass.produce(
+                    ReflectiveClassBuildItem.builder(className)
+                            .methods()
```
jamesnetherton: Nitpick - if Camel does not need reflective access to any methods on the transformer classes, then you can remove `.methods()`. It may be enough to only enable constructor reflection (enabled by default with `ReflectiveClassBuildItem`).

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelProcessor.java
```diff
@@ -431,6 +432,37 @@ CamelComponentNameResolverBuildItem componentNameResolver(
...
+     * This enables transformer.scan() to work in native mode.
+     */
+    @BuildStep
+    void discoverDataTypeTransformers(
+            CombinedIndexBuildItem combinedIndex,
+            BuildProducer<CamelPackageScanClassBuildItem> packageScanClass,
+            BuildProducer<ReflectiveClassBuildItem> reflectiveClass) {
+
+        IndexView view = combinedIndex.getIndex();
+        DotName dataTypeTransformer = DotName.createSimple("org.apache.camel.spi.DataTypeTransformer");
```
jamesnetherton: Wonder if instead of searching for `@DataTypeTransformer`, maybe we just find all subclasses of `org.apache.camel.spi.Transformer`? That way users would not need to tag any custom classes with `@RegisterForReflection`.

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelProcessor.java
```diff
@@ -431,6 +432,37 @@ CamelComponentNameResolverBuildItem componentNameResolver(
...
+
+        Set<String> transformerClasses = view.getAnnotations(dataTypeTransformer)
+                .stream()
+                .filter(ai -> ai.target().kind() == AnnotationTarget.Kind.CLASS)
```
jamesnetherton: I think we should also filter transformer classes that are part of the Camel core + components. They get loaded automatically from the `META-INF/services` files.

jamesnetherton:
Overall it looks good to me.

> use something like @BuildStep(onlyIf = NativeOrNativeSourcesBuild.class)

I think we should in this case, since this change is specific to native support. There's no point in having the build step run for non-native.

jamesnetherton (approved):

Issue #7862, opened 2025-10-22
Support Java 25


apupier:
### Describe the feature here

Java 25 is an LTS version, it will be nice to support it.

Current state:
* Apache Camel still have some blocker to support Java 25 https://issues.apache.org/jira/browse/CAMEL-22114 
* Quarkus is still work in progress https://github.com/orgs/quarkusio/projects/59
* I was able to do the local compilation with Java 25 of Camel Quarkus

It would be nice to start playing tests on the CI with Java 25

PR #7889, opened 2025-10-24
Provide specific GitHub workflow to build and test JDK 25 on CI

apupier:
part of #7862

I mainly copied the ci-build.yaml and modified:
- use the JDK 25 instead of JDK 17
- remove integration-tests-alternative-jdk section
- change triggers: launch one time per week for now, allow to launch manually and launch for PR on a jdk25 branch

I think it will be useful to start having tests running on CI to have an overview of what is the statu and better to keep it separated from normal flow until it is ready to not pollute "normal" flow

### DIFF
--- a/.github/workflows/jdk25-build.yaml
+++ b/.github/workflows/jdk25-build.yaml
@@ -15,67 +15,16 @@
 # limitations under the License.
 #
 
-name: Camel Quarkus CI
+name: JDK 25 Build
 
 on:
-  push:
-    branches:
-      - main
-      - camel-main
-      - quarkus-main
-      - "[0-9]+.[0-9]+.x"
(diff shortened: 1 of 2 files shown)


jamesnetherton:
Please add the workflow file into `paths-ignore` for `ci-build.yaml` & `pr-validate.yml`.

> I think it will be useful to start having tests running on CI to have an overview of what is the statu and better to keep it separated from normal flow until it is ready to not pollute "normal" flow

Yeah, I think this is an ok compromise, given that we are already 'maxed out' on the number of concurrent jobs in the main CI workflow. It'll be good to get an initial set of JDK 25 issues and starting working through them.

jamesnetherton (approved):

Issue #6041, opened 2024-04-25
Workflow to validate documentation changes against the Camel website


jamesnetherton:
### Description

Not sure how feasible this is, but it might be worth a POC.

If there are adoc changes on a PR, then we could trigger a workflow to validate them against the Camel website. E.g by running the scripts mentioned in the docs module README.

https://github.com/apache/camel-quarkus/blob/main/docs/README.adoc

PR #7891, opened 2025-10-24
ci: Add github workflow to validation documentation on PR

gansheer:
Ref #6041 

This is the same workflow as the one added to the apache/camel repository: https://github.com/apache/camel/pull/19650

It validates that any change in the doc is valid with in syntax with using the doc local/partial build https://github.com/apache/camel-quarkus/blob/main/docs/README.adoc in 1 to 3 mn to execute.

It does not execute any advanced check like the dead links for now.

### DIFF
--- a/.github/workflows/pr-doc-validation.yml
+++ b/.github/workflows/pr-doc-validation.yml
@@ -1,0 +1,60 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: PR doc validation
+on:
+  pull_request:
+    branches:
+      - main
+      - "[0-9]+.[0-9]+.x"
+    paths:
+      - '**.adoc'
+      - '!**/README.adoc'
+
+concurrency:
+  group: ${{ github.ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  check-doc:
+    if: github.repository == 'apache/camel-quarkus'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout camel-quarkus repo
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+          submodules: recursive
+          path: camel-quarkus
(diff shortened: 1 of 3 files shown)


gansheer:
The general  goal is to have at least this workflow in the repository involved with the documentation site generated by camel-website repository.

Some questions:
* It there a workflow that checks that any modification in a extension documentation has been built into docs/** ?
* Would there be an point to have more complete documentation check that could be executed before or at the beginning of a release process ? (or on demand)

.github/workflows/pr-doc-validation.yml
```diff
@@ -0,0 +1,58 @@
...
+# limitations under the License.
+#
+
+name: PR doc validation
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - '**.adoc'
```
jamesnetherton, line 25: Optional nitpick - Maybe exclude all `README.adoc` as they're not really related to docs hosted on the website

.github/workflows/pr-doc-validation.yml
```diff
@@ -0,0 +1,58 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: PR doc validation
+on:
+  pull_request:
+    branches:
+      - main
```
jamesnetherton, line 22: Assuming this workflow can work on maintenance branches, then I guess we could have:

```suggestion
      - main
      - "[0-9]+.[0-9]+.x"
```

jamesnetherton:
> It there a workflow that checks that any modification in a extension documentation has been built into docs/**

Yes, the initial build step checks that there are no uncommitted files left dangling from generated content. And fails the build if there are.

> Would there be an point to have more complete documentation check that could be executed before or at the beginning of a release process ? (or on demand)

It'd maybe be too late at that stage. Usually if we break the website we get notified pretty quickly and fix it before releases etc.

Do we know how long a full docs build (with link checks etc) would take?

jamesnetherton:
Please add the workflow file into `paths-ignore` for `ci-build.yaml` & `pr-validate.yml`.

gansheer:
> > Would there be an point to have more complete documentation check that could be executed before or at the beginning of a release process ? (or on demand)
> 
> It'd maybe be too late at that stage. Usually if we break the website we get notified pretty quickly and fix it before releases etc.
> 
> Do we know how long a full docs build (with link checks etc) would take?

The full docs build with a complete check would take approximatively 40min to run and would look like this:
```
      - name: Validate documentation
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        shell: bash
        run: |
          cd camel-website
          yarn workspaces foreach --all install
          cd ../camel-quarkus/docs
          yarn install
          ./local-build.sh full
          cd ../../camel-website
          yarn build:hugo
          yarn checks
```
It does not work correctly yet because I am missing some configuration and as a result the checks get too strict. Also the full build involves building the documentation from all repositories and would end up in failure with any error if any of the other repositories has its own error.
There are improvments to be made in the full build and check process I think before adding it to a sub-repo.

.github/workflows/pr-doc-validation.yml
gansheer: Done

.github/workflows/pr-doc-validation.yml
gansheer: Done

jamesnetherton (approved):

aldettinger (approved):

PR #7937, opened 2025-11-03
Add ActiveMQ 6 extension

### LINKED ISSUE #6104: Camel 4.7 - camel-activemq6
### Describe the feature here

In Camel 4.7 we now have a camel-activemq6 component which is a copy of camel-activemq but for 6.x brokers

gansheer:
Closes #6104

Camel components activemq6 is a copy of activemq with the difference of dependency client:
* camel-activemq uses activemq-client-jakarta (5.x)
* camel-activemq6 uses activemq-client (6.x)

This is the same for this extension: it is a copy of activemq extension with a different camel component dependency.


jamesnetherton:
@gansheer please rebase from `main` so we test it with Camel 4.16.0.

PR #7939, opened 2025-11-05
Upgrade camel to 4.16.0

JiriOndrusek:
Upgrade to Camel 4.16.0

### DIFF
--- a/docs/modules/ROOT/pages/reference/extensions/pqc.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/pqc.adoc
@@ -23,9 +23,9 @@ Post Quantum Computing Signature and Verification component.
 == What's inside
 
 * PQC (Post-Quantum Cryptography)
-* PQC Algorithms
+* xref:{cq-camel-components}::pqc-component.adoc[PQC Algorithms component], URI syntax: `pqc:label`
 
-
+Please refer to the above links for usage and configuration details.
 
 [id="extensions-pqc-maven-coordinates"]
 == Maven coordinates
--- a/tooling/maven-plugin/src/main/java/org/apache/camel/quarkus/maven/UpdateExtensionDocPageMojo.java
+++ b/tooling/maven-plugin/src/main/java/org/apache/camel/quarkus/maven/UpdateExtensionDocPageMojo.java
@@ -252,7 +252,10 @@ public class UpdateExtensionDocPageMojo extends AbstractDocGeneratorMojo {
                         getLog().warn("Failed to determine component link overrides for " + name);
                     }
                 }
-
+                //see https://github.com/apache/camel-quarkus/issues/7942
+                if (name.equals("pqc") && "dataformat".equals(kind)) {
+                    return "https://issues.apache.org/jira/browse/CAMEL-22652"; // pqc-dataformat.adoc is missing in camel
+                }
                 return xrefPrefix + name + (!"other".equals(kind) ? "-" + kind : "") + ".adoc";
             }
         });


JiriOndrusek:
There is a missing pqc-dataformat.adoc. Our `update-extension-doc-page` is expecting the `pqc-dataformat.adoc` to exist and generates a xref to this file.  -> camel-website validation fails
I put a workaround, so that link is not generated.  The workaround should be removed as soon as the doc is added. See //see https://github.com/apache/camel-quarkus/issues/7942

aldettinger:
@JiriOndrusek https://issues.apache.org/jira/browse/CAMEL-22586 seems fixed in 4.16.0, so the workaround is still needed ? Or there is a follow up issue maybe ?

JiriOndrusek:
@aldettinger  I linked bad ticket in the CQ.
Dataformat exists, but the doc page is missing https://issues.apache.org/jira/browse/CAMEL-22652

JiriOndrusek:
I see that Andrea already fixed that , so perhaps I can remove the workaround. Let me try it.

aldettinger (approved):
It seems fixed in 4.17, so the workaround would probably need to stay.

JiriOndrusek:
Workaround is removed, lets wait for CI to confirm it.

JiriOndrusek:
I'll remove the workaround in `camel-main`, it is still required in `main`

oscerd (approved):

extensions/xchange/deployment/src/main/java/org/apache/camel/quarkus/component/xchange/deployment/XchangeProcessor.java
```diff
@@ -138,7 +138,7 @@ void registerProxyClasses(BuildProducer<NativeImageProxyDefinitionBuildItem> nat
 
     @BuildStep
     void runtimeInitializedClasses(BuildProducer<RuntimeInitializedClassBuildItem> runtimeInitializedClasses) {
-        Stream.of(Order.class.getName())
+        Stream.of(Order.class.getName(), "org.apache.http.impl.auth.NTLMEngineImpl")
```
jamesnetherton, line 141: Is this definitely required? I don't see the Apache HTTP client on the runtime classpath for camel-xchange.

integration-tests/jira/pom.xml
```diff
@@ -112,7 +112,7 @@
                 </plugins>
             </build>
         </profile>
-        <profile>
+                <profile>
```
jamesnetherton: Nitpick - restore original formatting.

jamesnetherton:
> I'll remove the workaround in `camel-main`, it is still required in `main`

Our `antora.yml` points at the `next` docs version. So if the problem is fixed on the Camel core `main` branch, it should work ok.

In any case, there's a alternative method of controlling the doc component links:

https://camel.apache.org/camel-quarkus/next/contributor-guide/extension-documentation.html#_overriding_generated_component_links

extensions/xchange/deployment/src/main/java/org/apache/camel/quarkus/component/xchange/deployment/XchangeProcessor.java
JiriOndrusek: without the registration, there is an error:
```
Error: Detected an instance of Random/SplittableRandom class in the image heap. Instances created during image generation have cached seed values and don't behave as expected. If these objects should not be stored in the image heap, you can use 

    '--trace-object-instantiation=java.security.SecureRandom'

to find classes that instantiate these objects. Once you found such a class, you can mark it explicitly for run time initialization with 

    '--initialize-at-run-time=<culprit>'

to prevent the instantiation of the object.
The object was probably created by a class initializer and is reachable from a static field. You can request class initialization at image runtime by using the option --initialize-at-run-time=<class-name>. Or you can write your own initialization methods and call them explicitly from your main entry point.
Detailed message:
Trace: Object was reached by
  trying to constant fold static field org.apache.http.impl.auth.NTLMEngineImpl.RND_GEN
    at org.apache.http.impl.auth.NTLMEn
    ```

Comment (jamesnetherton) [reply extensions/xchange/deployment/src/main/java/org/apache/camel/quarkus/component/xchange/deployment/XchangeProcessor.java:141]: Ok - thanks for confirming. Then I guess we have to keep it.

**Comment by JiriOndrusek** [comment]
> Member

~~The override mechanism is handy. I set both links (for dataformat and component) to xref camel component.~~

~~The other option is to remove xrefs at all~~

File: integration-tests/jira/pom.xml (hunk 2 above)
Comment (JiriOndrusek) [reply integration-tests/jira/pom.xml]: fixed

**Comment by JiriOndrusek** [comment]
@jamesnetherton  the overriding mechanism you mentioned works correctly, unfortunately there is no option to remove only xref to one kind. Let me explain.

* (1) Originally the xrefs are
```
* xref:{cq-camel-components}:dataformats:pqc-dataformat.adoc[PQC (Post-Quantum Cryptography) data format]
* xref:{cq-camel-components}::pqc-component.adoc[PQC Algorithms component], URI syntax: `pqc:label`
```
and the first one is not valid.

* (2) The overriding mechanism allows to remove both of them to
```
* PQC (Post-Quantum Cryptography)
* PQC Algorithms
```

* (3) The original workaround (which was inside the plugin  itself) removed only the first xref, so the result was
```
* PQC (Post-Quantum Cryptography)
* xref:{cq-camel-components}::pqc-component.adoc[PQC Algorithms component], URI syntax: `pqc:label`
```

I think that both options are acceptable for now (both options - 2 & 3 -  are workarounds)
I kept option (3) as it does not require altering the plugin.
WDYT?

JiriOndrusek:
@jamesnetherton I kept the original workaround with the plugin.
And I think that we can merge this PR as soon as CI is green.

PR #7940, opened 2025-11-05
perf-regression: Upgrade hyperfoil-maven-plugin to 0.28.0

aldettinger:


jamesnetherton:
Not sure this can work given this release has a minimum requirement of JDK 21.

aldettinger:
> Not sure this can work given this release has a minimum requirement of JDK 21.

Yes, this is at least a first issue. Maybe it could be interesting to catch more logs to check for another possible issue.
I've once caught part of an error message mentioning something related to issue while running in a container.
It could be good to clarify this also.

aldettinger:
Conclusion, there is only the unsupported class version issue. Nothing special in ci nor container.
Let's try to gain insights from the hyperfoil community about the minimum JDK required version in [HYPERFOIL-629](https://github.com/Hyperfoil/Hyperfoil/issues/629) and adapt.

Issue #6433, opened 2024-09-06
Investigate Windows build slowness


jamesnetherton:
### Description

Not sure why, but some tests take almost twice as long to complete compared to Linux.

Picking Jasypt as a random example.

Linux:

```
Camel Quarkus :: Integration Tests :: Jasypt ....... SUCCESS [ 31.549 s]
```

Windows:

```
Camel Quarkus :: Integration Tests :: Jasypt ....... SUCCESS [ 52.278 s]
```

If we can figure out why and maybe fix it, it'd probably cut the Windows CI job build time by 30% - 40%.

llowinge:
@jamesnetherton I would like to investigate two things. Firstly i'll add timestamps to maven build log (so i can compare which Maven phase takes the largest time). Especially i'm concerned about the `clean` phase (given this https://github.com/apache/maven-mvnd/issues/536).

llowinge:
Maven phases comparsion:
Linux (1.298s):
```
09:31:38,449 [INFO] ---< org.apache.camel.quarkus:camel-quarkus-integration-test-jasypt >---
09:31:38,449 [INFO] Building Camel Quarkus :: Integration Tests :: Jasypt 3.27.0-SNAPSHOT [2/13]
09:31:38,449 [INFO]   from /home/runner/work/camel-quarkus/camel-quarkus/integration-tests/jasypt/pom.xml
09:31:38,449 [INFO] --------------------------------[ jar ]---------------------------------
09:31:38,727 [INFO] 
09:31:38,727 [INFO] --- clean:3.5.0:clean (default-clean) @ camel-quarkus-integration-test-jasypt ---
09:31:38,728 [INFO] 
09:31:38,728 [INFO] --- resources:3.3.1:copy-resources (copy-camel-quarkus-enforcer-rules-xslt) @ camel-quarkus-integration-test-jasypt ---
09:31:38,731 [INFO] Copying 1 resource from ../../tooling/enforcer-rules to target/enforcer-rules
09:31:38,734 [INFO] 
09:31:38,734 [INFO] --- enforcer:3.5.0:enforce (enforce-java-version) @ camel-quarkus-integration-test-jasypt ---
09:31:38,735 [INFO] Skipping Rule Enforcement.
09:31:38,735 [INFO] 
09:31:38,735 [INFO] --- enforcer:3.5.0:enforce (camel-quarkus-enforcer-rules) @ camel-quarkus-integration-test-jasypt ---
[...]

PR #7963, opened 2025-11-13
[relates #6433] Improve Windows build speed

llowinge:


jamesnetherton:
Seems an average of 2 - 3 mins saving. Hard to find real gains on infra that looks to be inherently slower compared to the *nix runners. Since it's a pretty harmless change I see no reason not to merge it.

PR #7998, opened 2025-11-21
Switch LangChain4j model testing from llama3.1 to granite4:3b

jamesnetherton:


jamesnetherton:
> Shouldn't we update the Readmes too?

Yes, good catch. I added a mention of `granite4:3b`. I left the reference to `orca-mini` because some tests still use it and it's generally good to have multiple `ChatModel` beans in the mix for testing.

PR #8039, opened 2025-12-04
Restore Milo extension

jamesnetherton:
Restores Milo and adds native support.

I will create an issue to follow up with more tests later. There is some oddness in the component that needs digging into first around server credentials and host binding.


aldettinger:
Small glitch:
```
Imports are not sorted in /home/runner/work/camel-quarkus/camel-quarkus/extensions/milo/deployment/src/main/java/org/apache/camel/quarkus/component/milo/deployment/MiloProcessor.java
```

aldettinger (approved):

PR #8074, opened 2025-12-16
Override lz4-java for flink & aws2-kinesis with at.yawk.lz4 variant

jamesnetherton:
Other lingering references to the old `org.lz4` dependency will get fixed up when Quarkus eventually upgrades kafka-clients or does overrides in their kafka extension.

### DIFF
--- a/extensions-jvm/flink/runtime/pom.xml
+++ b/extensions-jvm/flink/runtime/pom.xml
@@ -42,6 +42,16 @@
         <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-flink</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.lz4</groupId>
+                    <artifactId>lz4-java</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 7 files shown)


oscerd:
Is this something for Camel core too?

jamesnetherton:
> Is this something for Camel core too?

Yes, probably. Although the scope is a lot wider there because it's a transitive of kafka-clients which appears on the classpath in many places. Like all the Debezium components etc.

oscerd:
I think it would be good to have an issue for this, so we could look around.

jamesnetherton:
> I think it would be good to have an issue for this, so we could look around.

https://issues.apache.org/jira/browse/CAMEL-22788

gansheer (approved):

davsclaus:
camel-rocketmq also has lz4-java

ppalaga:
> camel-rocketmq also has lz4-java

Yeah, we have agreed with @jamesnetherton that we need to add org.lz4:lz4-java to enforcer bans to be sure it is not going to leak in any way.

ppalaga:
https://github.com/apache/camel-quarkus/pull/8077 amends this PR by adding the ban

PR #8077, opened 2025-12-17
Override lz4-java for flink & aws2-kinesis with at.yawk.lz4 variant

ppalaga:


poms/bom/pom.xml
```diff
@@ -1643,6 +1661,12 @@
                 <groupId>org.apache.camel</groupId>
                 <artifactId>camel-ibm-watson-language</artifactId>
                 <version>${camel.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.lz4</groupId>
+                        <artifactId>lz4-java</artifactId>
+                    </exclusion>
+                </exclusions>
```
apupier: I do not see it as a dependency of camel-ibm-watson-language:
image

ppalaga: Thanks, good catch! I have no idea how it got there...

ppalaga:
b1fc8c8: addressed the feedback of @apupier

PR #8089, opened 2025-12-23
Upgrade Debezium to 3.4.0.Final

jamesnetherton:
We may need to wait for Camel 4.17 for this to be mergeable. Let's see...


jamesnetherton:
CI says it's good, thus merging.

PR #8100, opened 2026-01-05
Provide basic test for pdf:merge

apupier:


gansheer (approved):

apupier:
> If the intent is to test `merge` only. Maybe a later improvement could be to create first and second pdfs with pdfbox api ? 

In fact, it would cause the code to be longer as there si a more direct APi already available in the test for the simple case of text.

> Or even generating them once and store as test resources ?

I preferred to not follow this path to have better diff and more control but for some complex pdf might be easier

PR #8106, opened 2026-01-07
Upgrade test container images

apupier:
when same image is used in Camel core, aligned with the version used in Camel 4.16


oscerd (approved):

apupier:
removed the upgrades which doesn't work directly and I will create specific issues for each of them so that we can already merge all the straightforward ones

PR #8131, opened 2026-01-12
Add .sdkmanrc

ppalaga:


ppalaga:
bae99a6: changed Maven version to 3.9.11 to avoid multiple staging repos when releasing

apupier (approved):

PR #8142, opened 2026-01-14
Fix potential master-file race with cluster data deletion and lock acquisition on Windows

jamesnetherton:


integration-tests/master-file/src/test/java/org/apache/camel/quarkus/component/master/it/MasterFileTestResource.java
```diff
@@ -0,0 +1,48 @@
...
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+import org.apache.commons.io.FileUtils;
+import org.jboss.logging.Logger;
+
+public class MasterFileTestResource implements QuarkusTestResourceLifecycleManager {
+    private static final Logger LOG = Logger.getLogger(MasterFileTestResource.class);
+
+    @Override
+    public Map<String, String> start() {
+        // We need to clean up cluster data early as possible, before the app under test is started to avoid potential races on lock files with Windows
```
apupier: What is creating the cluster data folder? Something else than the app under test?

jamesnetherton: No it's just the app(s) under test.

apupier: So it is a race condition between MasterFileTest and MasterFileIT?
What about cleaning the folder with a tearDown/afterAll? (to avoid the need for a specific lifecycle class) Would it work?

jamesnetherton: >  cleaning the folder with a tearDown/afterAll

I decided against that because you loose the ability to inspect the files after the test(s) have run.

But if you think it's a better way. I can update the PR for that.

apupier: no strong opinion.
I usually prefer to clean the file created by the test to avoid interaction with others and comment/uncomment code if needed for investigation. In my experience, I spent more time investigating why a test was flaky due to interaction between different one than looking to output of a test which was already written.

jamesnetherton: I switched to AfterEach. Probably cleaner compared to what I originally proposed.

PR #8149, opened 2026-01-14
Upgrade Quarkus to 3.31.0.CR1

jamesnetherton:
Example project builds are expected to fail because they are not yet adjusted for Testcontainers 2.x. I have a branch for this, which I'll aim to get merged immediately after this PR.

https://github.com/apache/camel-quarkus-examples/tree/testcontainers-2.x


jamesnetherton:
Seems `Langchain4jAgentTest` can't match the correct WireMock stub in one of its tests. Not seen that before, so I need to investigate. If it's not a quick fix, I'll likely disable the failing test and follow up on it later.

PR #8156, opened 2026-01-15
Update MacOS build tips in contributor user guide

JinyuChen97:
update some mac build tips for camel quarkus doc

### DIFF
--- a/docs/modules/ROOT/pages/contributor-guide/index.adoc
+++ b/docs/modules/ROOT/pages/contributor-guide/index.adoc
@@ -46,7 +46,11 @@ A build with integration tests in both the JVM mode and the native mode:
 mvn clean install -Pnative
 ----
 
-TIP: You may want to install and use https://github.com/mvndaemon/mvnd[`mvnd` - the Maven Daemon] for faster builds.
+TIP: You may want to install and use https://github.com/mvndaemon/mvnd[`mvnd` - the Maven Daemon] for faster builds. When using `mvnd` on MacOS, make sure the mvnd version matches your installed Maven version (`mvn`) to avoid compilation failures due to missing dependencies.
+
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/contributor-guide/index.adoc
```diff
@@ -46,7 +46,11 @@ A build with integration tests in both the JVM mode and the native mode:
 mvn clean install -Pnative
 ----
 
-TIP: You may want to install and use https://github.com/mvndaemon/mvnd[`mvnd` - the Maven Daemon] for faster builds.
+TIP: You may want to install and use https://github.com/mvndaemon/mvnd[`mvnd` - the Maven Daemon] for faster builds. When using `mvnd` on MacOS, make sure the mvnd version matches your installed Maven version (`mvn`) to avoid compilation failures due to missing dependencies.
+
+TIP: For building native images on MacOS, you usually **do not need** to use `-Dquarkus.native.container-build`. This option is primarily intended for Linux systems, where the native image build runs inside a Docker container. On MacOS, GraalVM can typically generate the native image directly, so omitting this option simplifies the build and avoids unnecessary errors.
+
+TIP: Some extensions may not yet support native image generation on MacOS. If an integration test fails when running `mvnd clean install -Dnative` due to failure to build the native image, this is expected and not necessarily a problem with your environment. Running the same test in JVM mode (`mvn clean install`) should work correctly.
```
jamesnetherton: I think we should be explicit about what we expect to fail with MacOS native builds. IIRC that's currently just limited to extensions that require AWT. 

```suggestion
TIP: Extensions that depend on Java AWT are not yet supported for native image generation on MacOS. If an integration test fails when running `mvnd clean install -Dnative` due to missing AWT support, this is expected and not necessarily a problem with your environment. Running the same test in JVM mode (`mvn clean install`) should work correctly.
```

jamesnetherton (commented):
I added a minor suggestion / improvement comment.

Otherwise it LGTM.

docs/modules/ROOT/pages/contributor-guide/index.adoc
JinyuChen97: done

PR #8162, opened 2026-01-16
Add cli-debug extension

### LINKED ISSUE #8040: Camel 4.17 - camel-cli-debug extension
### Describe the feature here

There is a new camel-cli-debug component in camel-core
https://issues.apache.org/jira/browse/CAMEL-22748

And we need a CEQ extension. Its just a placeholder as there is no java code needed.

CSB has a starter
https://github.com/apache/camel-spring-boot/tree/main/dsl-starter/camel-cli-debug-starter

jamesnetherton:
Fixes #8040


extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/main/CamelMain.java
```diff
@@ -99,6 +103,21 @@ public MainConfigurationProperties getMainConfigurationProperties() {
...
+        if (camelContext.isSourceLocationEnabled()) {
+            for (RoutesBuilder route : getMainConfigurationProperties().getRoutesBuilders()) {
+                if (route instanceof ResourceAware ra && ra.getResource() == null) {
+                    Resource resource = ResourceHelper.resolveResource(camelContext, "source:" + route.getClass().getName());
+                    if (resource != null && resource.exists()) {
+                        ra.setResource(resource);
+                    }
+                }
+            }
+        }
```
jamesnetherton, line 117: @davsclaus I had to add this to ensure non-registry discovered `RouteBuilder` impls (i.e ones that are not CDI beans) have their `source:` resources configured.

Not sure if we should push a similar change into `camel-main`?

davsclaus: Ah yeah so if you add those route builder directly in java code - yeah that would be good to have in core as well

PR #8164, opened 2026-01-16
Rename camel-quarkus-junit5 to camel-quarkus-junit

### LINKED ISSUE #8130: Rename camel-quarkus-junit5 as camel-quarkus-junit
Quarkus 3.31.0 will upgrade to JUnit 6.x and will deprecate / rename `quarkus-junit-5` to `quarkus-junit`. I think we should do something in Camel Quarkus so that the GAV is JUnit version agnostic.

After #8125 is merged, we can use `camel-test-junit6` in a new `camel-quarkus-junit` module. Then deprecate `camel-quarkus-junit5` and have it include `camel-quarkus-junit`, until we've given reasonable time for users to migrate to the new GAV.

jamesnetherton:
Fixes #8130


jamesnetherton:
Seems the `relocation` rule I added to `camel-quarkus-junit5` confuses the cq-maven-plugin. If I can't make it work I could  fallback on adding a direct dependency to `camel-quarkus-junit` instead of doing relocations.

jamesnetherton:
I took the fallback option. Probably not worth adjusting pom-tuner and cq-maven-plugin to support relocations just for this change.

PR #8167, opened 2026-01-16
Activate traceobjectinstantiation for hazelcast native compilation to

apupier:
gathe rmore information when it fails on CI


apupier:
unable to reproduce the error on the CI after a bunch of run

PR #8197, opened 2026-01-22
Support extract metadata for camel docling in native

apupier:
it has a very limited scope, it requires
https://issues.apache.org/jira/browse/CAMEL-22888 to be able to test more


extensions/docling/deployment/src/main/java/org/apache/camel/quarkus/component/docling/deployment/DoclingProcessor.java
```diff
@@ -73,5 +74,7 @@ void registerForReflection(
         reflectiveClass.produce(ReflectiveClassBuildItem.builder(doclingCoreBuilderClasses.toArray(new String[0]))
                 .methods(true)
                 .build());
+
+        reflectiveClass.produce(ReflectiveClassBuildItem.builder(DocumentMetadata.class).constructors().methods(true).build());
```
jamesnetherton: Just to clarify - are we registering `DocumentMetadata` because something in Camel or elsewhere does reflective operations on it?

Or are we only adding it because it's used with `ObjectMapper` in the tests?

integration-tests/docling/src/test/java/org/apache/camel/quarkus/component/docling/it/DoclingTest.java
```diff
@@ -127,12 +133,62 @@ public void extractText() {
...
+
+                    @Override
+                    public void describeTo(Description description) {
+                        if (documentMetadata != null) {
+                            description.appendText(documentMetadata.toString());
+                        } else {
+                            description.appendText("Wasn't able to parse the answer");
+                        }
+                    }
+                });
```
jamesnetherton: If you just want to do assertions on a JSON response. There is a much easier way:

https://github.com/rest-assured/rest-assured/wiki/usage#example-1---json

integration-tests/docling/src/test/java/org/apache/camel/quarkus/component/docling/it/DoclingTest.java
```diff
@@ -141,7 +197,33 @@ public void extractMetadata() {
...
+
+                    private DocumentMetadata documentMetadata;
+
+                    @Override
+                    public boolean matches(Object actual) {
+                        ObjectMapper mapper = new ObjectMapper();
+                        try {
+                            documentMetadata = mapper.readValue((String) actual, DocumentMetadata.class);
+                            return documentMetadata.getFileName().startsWith("docling-test")
+                                    && documentMetadata.getFilePath().contains("docling-test");
```
jamesnetherton: See previous comment about JSON response matching.

integration-tests/docling/src/main/java/org/apache/camel/quarkus/component/docling/it/DoclingResource.java
```diff
@@ -206,13 +210,38 @@ public Response extractMetadata(String documentContent) throws IOException {
         java.nio.file.Path tempFile = Files.createTempFile("docling-test", ".md");
         Files.writeString(tempFile, documentContent);
         try {
-            String result = producerTemplate.requestBody("direct:extractMetadata", tempFile.toString(), String.class);
-            return Response.ok(result).build();
+            DocumentMetadata result = producerTemplate.requestBody("direct:extractMetadata", tempFile.toString(),
+                    DocumentMetadata.class);
+            ObjectWriter ow = new ObjectMapper().writer().withDefaultPrettyPrinter();
+            String json = ow.writeValueAsString(result);
+            return Response.ok(json).build();
```
jamesnetherton: I you use `quarkus-resteasy-jackson` or `quarkus-resteasy-jsonb`, you can set the `Response` entity to `result` and it'll get serialized to JSON automatically. Thus no need for `ObjectMapper`.

Same for `extractMetadataFromPdf `.

extensions/docling/deployment/src/main/java/org/apache/camel/quarkus/component/docling/deployment/DoclingProcessor.java
apupier: I think it is because I request a DocumentMetadata.class as output type of the producerTemplate.requestBody https://github.com/apache/camel-quarkus/pull/8197/files#diff-d6794a6a6cc2c6724d73ed97f0f830c8f7527d08968a193abee6ed89042f7442R213

so technically in the test but I think that if someone is trying to use DocumentMetdata directly they will have to register it.

What do you think?

jamesnetherton: I've not tried it, but I don't think that leads to reflection being used. I think it just casts the result to the type you provide.

From what I can see in `DoclingProducer`, it's using `new DocumentMetadata()`, so there's no reflection involved.

integration-tests/docling/src/main/java/org/apache/camel/quarkus/component/docling/it/DoclingResource.java
apupier: when adding quarkus-resteast-jackson or quarkus-resteasy-jsonb, I got this error at compilation:
```
java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
	[error]: Build step io.quarkus.arc.deployment.ArcProcessor#validate threw an exception: jakarta.enterprise.inject.spi.DeploymentException: Found 2 deployment problems: 
[1] Unsatisfied dependency for type org.apache.camel.CamelContext and qualifiers [@Default]
	- injection target: org.apache.camel.quarkus.component.docling.it.DoclingResource#context
	- declared on CLASS bean [types=[java.lang.Object, org.apache.camel.quarkus.component.docling.it.DoclingResource], qualifiers=[@Any, @Default], target=org.apache.camel.quarkus.component.docling.it.DoclingResource]
[2] Unsatisfied dependency for type org.apache.camel.ProducerTemplate and qualifiers [@Default]
	- injection target: org.apache.camel.quarkus.component.docling.it.DoclingResource#producerTemplate
	- declared on CLASS bean [types=[java.lang.Object, org.apache.camel.quarkus.component.docling.it.DoclingResource], qualifiers=[@Any, @Default], target=org.apache.camel.quarkus.component.docling.it.DoclingResource]
```

apupier: rerverted these changes and still got this compilation error. rebuilding everything, i guess there is something not clean in my environment oO

integration-tests/docling/src/test/java/org/apache/camel/quarkus/component/docling/it/DoclingTest.java
apupier: a lot easier effectively, we lost the type-safe side but a lot more concise. thanks

integration-tests/docling/src/main/java/org/apache/camel/quarkus/component/docling/it/DoclingResource.java
apupier: working fine after full rebuild

extensions/docling/deployment/src/main/java/org/apache/camel/quarkus/component/docling/deployment/DoclingProcessor.java
apupier: I registered for reflection only the test side then

PR #8220, opened 2026-01-28
Docling test using local docling cli

apupier:
I recommend to review one commit at a time, as one is only a refatcor. I can also split in 2 PRs if it is more convenient.

Add docling test with cli mode

the test is not activated by default to avoid the need to configure the
CI with local docling on command-line.


integration-tests/docling/src/test/java/org/apache/camel/quarkus/component/docling/it/DoclingTest.java
```diff
@@ -220,4 +221,19 @@ public void convertToJsonAsync() {
...
+        String testContent = "# Test Document\nThis is a test document for JSON conversion with CLI.";
+
+        RestAssured.given()
+                .contentType(ContentType.TEXT)
+                .body(testContent)
+                .when()
+                .post("/docling/convert/json/cli")
+                .then()
+                .statusCode(200)
+                .body(not(emptyString()));
```
apupier, line 237: it will be benefical to check fo rmore than just not an empty string. i will implement that when improving for all the other tests too

PR #8222, opened 2026-01-29
Improve assertions for Docling tests

apupier:


integration-tests/docling/src/test/java/org/apache/camel/quarkus/component/docling/it/DoclingTest.java
```diff
@@ -88,7 +88,7 @@ public void convertToMarkdown() {
                 .post("/docling/convert/markdown")
                 .then()
                 .statusCode(200)
-                .body(not(emptyString()));
+                .body(containsString("This is a test document for conversion."));
```
aldettinger, line 91: @apupier So with Docling, the extraction results are deterministic ?

apupier: in theory no, but with simple enough document, it seems yes

PR #8227, opened 2026-01-30
Add OpenAI extension

jamesnetherton:
The `openai` component is not in the 4.17.0 catalog. This scenario happens occasionally, so I added a small 'enhancement' to the `check-extension-pages` mojo. So that it can continue to generate the expected doc resources. I intend to immediately remove the additional plugin config once this change lands on the `camel-main` branch.


docs/modules/ROOT/pages/reference/extensions/openai.adoc
```diff
@@ -0,0 +1,94 @@
...
+endif::[]
+
+[id="extensions-openai-usage"]
+== Usage
+[id="extensions-openai-usage-structured-output-with-output-class-in-native-mode"]
+=== Structured output with output class in native mode
+
+When using structured output with the `outputClass` option in native mode, you must ensure that the target class is registered for reflection.
+
+This can be done with the `@RegisterForReflection` annotation or configuration property `quarkus.camel.native.reflection.include-patterns`. For example:
```
aldettinger, line 47: Don't we generally reference the centralized native doc part in such case ?

jamesnetherton: Most extension docs give some native mode examples and then link to the native mode guide at the end. Which is the pattern I followed here.

integration-tests/openai/src/test/java/org/apache/camel/quarkus/component/openai/it/OpenaiTest.java
```diff
@@ -0,0 +1,179 @@
...
+        // Send the streaming request
+        RestAssured.given()
+                .contentType(ContentType.TEXT)
+                .body("Stream the numbers 1 to 10 on a new line each time and nothing else.")
+                .post("/openai/chat/streaming")
+                .then()
+                .statusCode(204);
+
+        // Assert the streamed results
+        IntStream.rangeClosed(1, 10)
```
aldettinger: Do we rely on message ordering for expectation ? If yes, not sure the route would preserve order, not sure of the opposite though.

aldettinger (approved):
Just a few neat questions. The comments and doc are really helpful for review 👏

integration-tests/openai/src/test/java/org/apache/camel/quarkus/component/openai/it/OpenaiTest.java
jamesnetherton: We do rely on ordering for this test...

Unless the LLM gets the order of numbers wrong (unlikely), won't the streamed chunks come in the expected 1-10 order?

Or do you think the messages arriving at the `seda` endpoint might be out of order (I've not seen it happen so far in local testing)?

aldettinger: I was more fearing a kind of race condition with wireTap thread pool but nothing demonstrated. If we could not simply switch to non order expectations now, we are not taking a big risk anyway. Even if flaky, it would not happen a lot :crossed_fingers: .

docs/modules/ROOT/pages/reference/extensions/openai.adoc
```diff
@@ -0,0 +1,94 @@
...
+----
+@RegisterForReflection
+public class MyStructuredOutputClass {
+    ...
+}
+----
+
+[source,java]
+----
+public class MyJQRoutes extends RouteBuilder {
```
apupier: `MyJQRoutes`? A copy-paste from JQ doc?

integration-tests/openai/src/test/java/org/apache/camel/quarkus/component/openai/it/OpenaiTest.java
jamesnetherton: Ok - thanks for the feedback. Lets see how it goes. If it does prove to be flaky on CI, I will rework it.

docs/modules/ROOT/pages/reference/extensions/openai.adoc
jamesnetherton: Whoops - I thought I already fixed that. I'll change it.

jamesnetherton:
I need to fix things up for Windows where I added test file deletion:

```
The process cannot access the file because it is being used by another process 
```

docs/modules/ROOT/pages/reference/extensions/openai.adoc
jamesnetherton: Should be fixed in the latest commit.

integration-tests/openai/src/test/java/org/apache/camel/quarkus/component/openai/it/OpenaiTest.java
jamesnetherton: @aldettinger I think you are right about this!

Seems the latest Windows failure is due to unexpected ordering.

I'll try to rework the test to make it more reliable.

aldettinger: Ok, that said the probability should have been quite low. Anyway, no order in the route is not a big deal, just need to relax assertion.

jamesnetherton: I reworked the test assertion(s) so that it can deal with results coming out of sequence.

aldettinger: Looks good, well done :clap:

PR #8237, opened 2026-02-03
Improve file-watch extension docs and test coverage

jamesnetherton:


apupier (approved):

jamesnetherton:
Looks like I need to restore the original state of the `Builder` class substitution and make it active on all platforms (I was testing only on macOS).

PR #8243, opened 2026-02-04
chore: Minor code quality cleanups and modernizations in core runtime

sanjana2505006:
This PR removes an unnecessary doInit() override in FastTypeConverter that only calls super. 

Based on maintainer feedback, other proposed modernizations (lambdas) have been removed from this PR to avoid potential memory overhead in the Quarkus runtime.

### DIFF
--- a/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/FastTypeConverter.java
+++ b/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/FastTypeConverter.java
@@ -29,11 +29,6 @@ public class FastTypeConverter extends DefaultTypeConverter {
     }
 
     @Override
-    protected void doInit() throws Exception {
-        super.doInit();
-    }
-
-    @Override
     public void loadCoreAndFastTypeConverters() throws Exception {
         for (TypeConverterLoader loader : getCamelContext().getRegistry().findByType(TypeConverterLoader.class)) {
             LOG.debug("TypeConverterLoader: {} loading converters", loader);


sanjana2505006:

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelContextRecorder.java
```diff
@@ -83,9 +83,12 @@ public RuntimeValue<CamelContext> createContext(
         ExtendedCamelContext extendedCamelContext = context.getCamelContextExtension();
         context.setApplicationContextClassLoader(tccl);
         extendedCamelContext.addContextPlugin(FactoryFinderResolver.class, factoryFinderResolver.getValue());
-        extendedCamelContext.addContextPlugin(RuntimeCamelCatalog.class, new CamelRuntimeCatalog(config.runtimeCatalog()));
```
apupier, line 86: I guess you have something not aligned in automatic code style compared to the configuration of the project as several lines are splitted despite there was no change.

jamesnetherton:
> please review and let me know if any changes are needed

Hi! Thanks for contributing. I'd prefer that the PR is reduced to only removing `doInit` from `FastTypeConverter`.

The other parts I think we can live without. If we were to follow the conventional Quarkus wisdom with regards to Lambdas, we'd probably use them a lot less, given they come with some slight memory overhead.

sanjana2505006:
@jamesnetherton @apupier 
Whenever you have time, please take a look at this PR and let me know if any changes are needed from my side.

PR #8255, opened 2026-02-05
chore: core runtime modernization and cleanup

sanjana2505006:
I was looking through the core runtime module and spotted a couple of small opportunities to modernize the code and make it a bit cleaner.

Here’s what I changed:
- FastCamelContext.java: Switched loadJsonSchema to use `try-with-resources`. It’s just safer and standard practice for handling streams.
- CamelMain.java: Replaced a raw `System.out.println` with a proper `LOG.warn()` when unknown options are encountered. It felt out of place to have console print statements there.

### DIFF
--- a/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/main/CamelMain.java
+++ b/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/main/CamelMain.java
@@ -189,7 +189,8 @@ public final class CamelMain extends MainCommandLineSupport implements HasCamelC
             }
         }
         if (!valid) {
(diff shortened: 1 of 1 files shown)


sanjana2505006:

jamesnetherton:
> @jamesnetherton Whenever you get a chance, please review and let me know if any changes are needed

The change to `FastCamelContext` would be ok. The other for `CamelMain` I don't think we need. It purposely uses `System.out.println` because the output is combined with [other information](https://github.com/apache/camel/blob/585f77d400b994a4483b200af33c8bd2d7b81cf0/core/camel-main/src/main/java/org/apache/camel/main/MainCommandLineSupport.java#L186-L193) that we don't want to be wrapped in log formatting.

PR #8258, opened 2026-02-05
chore: formatting and list modernization

sanjana2505006:
I was browsing through the core extension code and noticed a few spots where we could tidy things up a bit.
Here is what I changed this time:

- PathFilterTest.java: Swapped a bunch of `Arrays.asList` calls to `List.of.` It makes the test setup a little less verbose and uses the modern immutable lists.

- CamelConfig.java: Dropped a redundant `public` modifier from an interface method just to align with the rest of the file.

### DIFF
--- a/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
+++ b/extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
@@ -278,7 +278,7 @@ public interface CamelConfig {
          *
          * @asciidoclet
          */
-        Optional<List<String>> includePatterns();
+        public Optional<List<String>> includePatterns();
 
         /**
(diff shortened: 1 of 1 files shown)


sanjana2505006:

extensions-core/core/runtime/src/main/java/org/apache/camel/quarkus/core/CamelConfig.java
```diff
@@ -278,7 +278,7 @@ interface ReflectionConfig {
          *
          * @asciidoclet
          */
-        public Optional<List<String>> includePatterns();
+        Optional<List<String>> includePatterns();
```
apupier: I think that it is an API break and we should go through a deprecation phase first if we really want to hide it

jamesnetherton: It's defined in an interface so implicitly public.

PR #8259, opened 2026-02-05
chore: minor logging cleanup in CamelMain

sanjana2505006:
I spotted a tiny logging nit in CamelMain.java where we were using string concatenation instead of placeholders.
I switched it to use `{}` style parameterization. It's just a small cleanup to keep things consistent with the rest of the project (and avoids unnecessary string creation).


sanjana2505006:
@apupier  Whenever you get a chance, please review and let me know if any changes are needed

sanjana2505006:
Closing this in favour of #8261, as I have consolidated the logging fix there along with the other modernisation improvements in a much cleaner branch. This will avoid confusion with the unrelated CI failures here. Thank you

PR #8264, opened 2026-02-06
chore: replace string concatenation with parameterized logging

sanjana2505006:
This is a minor cleanup PR to improve logging efficiency and readability. I've replaced string concatenation with SLF4J-style parameterized logging (using `{}`).

This approach is preferred as it avoids the overhead of constructing the long log string if the debug/info level isn't actually enabled at runtime. It also helps keep the logging style consistent across the codebase.
Verified that the logging output remains exactly the same!


sanjana2505006:
@apupier  Whenever you get a chance, please review and let me know if any changes are needed

PR #8268, opened 2026-02-07
chore: modernize List usage and minor style fixes

sanjana2505006:
I was looking through the codebase and noticed a few places where we could use `List.of` instead of `Arrays.asList` for better consistency and conciseness.

I also spotted a double semicolon in a test file and some unused imports, so I cleaned those up while I was at it.


sanjana2505006:
@apupier Whenever you get a chance, please review and let me know if any changes are needed

PR #8278, opened 2026-02-10
Align dev.langchain4j dependencies with Quarkus LangChain4j 1.7.0

jamesnetherton:


jamesnetherton:
Setting as draft as there's a potential native mode issue with `langchain4j-agent`.

PR #8289, opened 2026-02-11
Disable Groovy invokedynamic compilation

### LINKED ISSUE #8122: [JDK25] Groovy native application fails to start due to Unsupported method java.lang.invoke.MethodHandleNatives.setCallSiteTargetNormal
### Bug description

https://github.com/quarkiverse/quarkus-groovy/issues/353

jamesnetherton:
Fixes #8122

### DIFF
--- a/docs/modules/ROOT/pages/reference/extensions/groovy.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/groovy.adoc
@@ -76,3 +76,12 @@ from("direct:start")
 `greeting.message` will be evaluated once at build time and its value will be permanently stored in the native image.
 It is not possible to override the value of the property at runtime. Attempting to do so will result in an exception being thrown.
 
+[id="extensions-groovy-limitations-invokedynamic-support-in-native-mode"]
+==== InvokeDynamic support in native mode
+
+This extension disables Groovy https://docs.groovy-lang.org/next/html/documentation/invokedynamic-support.html[InvokeDynamic] support during script compilation.
+
+Please note that if scripts or their dependencies make any explicit reference to `InvokeDynamic`, they will not work since the underlying logic required to support the feature is not available in GraalVM / Mandrel >= 25
+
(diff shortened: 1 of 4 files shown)


apupier:
Can you precise what is the impact for end-users? They cannot use some functionalities of Groovy in native mode?

jamesnetherton:
The reason for the change is that some Camel Groovy script operations lead to usage of invokedymanic and a code path to `java.lang.invoke.MethodHandleNatives.setCallSiteTargetNormal`.  That method is [deleted](https://github.com/oracle/graal/blob/70b9a98abc80bbdc54145dd84edfbfe94b6e836b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/methodhandles/Target_java_lang_invoke_MethodHandleNatives.java#L208-L209) in GraalVM.

It's a problem for things like this:

```groovy
from("direct:scriptGroovy")
        .script()
        .groovy("exchange.getMessage().setBody('Hello ' + exchange.getMessage().getBody(String.class) + ' from Groovy!')");
```

Because the `String` cast from `.getBody(String.class)` results in `invokedymanic` usage.

> Can you precise what is the impact for end-users? They cannot use some functionalities of Groovy in native mode?

TBH it's a little bit unknown because we can't realistically test everything you can do with Groovy APIs.

There may be a performance impact. However, in the case of most Camel integrations, I'd hope that it's relatively low. I assume most users are unlikely to be writing large complicated scripts. Typical usage is limited to simple expressions. 

As far as our test suite goes, I don't observe any big difference compared to the previous state.

This is only for native mode. JVM works as per vanilla Camel.

apupier:
thanks for the explanation.
I think it is worthy to add a note in the documentation that all Groovy scripts which are invoking invokeDynamic (even through dependencies) won't work anymore due to JDK 25 restriction.

I agree that it should not be a very common use case.

jamesnetherton:
> I think it is worthy to add a note in the documentation

I added some notes about it to the Groovy extension docs.

PR #8296, opened 2026-02-13
Attempt to reclaim more disk space from GitHub actions runners

jamesnetherton:


apupier:
Can you provide a bit more context? We hit a limit? How much space it will save?

On main branch, I noticed that we have:
https://github.com/apache/camel-quarkus/actions/runs/21978061889/job/63495270195#step:3:11
```
Free disk space before space reclaimed
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       145G   54G   92G  37% /
Reclaiming disk space...
Free disk space after space reclaimed
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       145G   28G  118G  19% /
```

Also have you already evaluated to use an existing github action runner like https://github.com/marketplace/actions/disk-space-reclaimer? (but maybe not that worthy?)

apupier (approved):

.github/reclaim-disk-space.sh
```diff
@@ -22,12 +22,16 @@ df -h /
...
+sudo rm -rf /etc/skel/.rustup \
+     rm -rf /home/packer/.rustup \
+     rm -rf /home/runner/.rustup \
+     rm -rf /opt/ghc \
      rm -rf /opt/hostedtoolcache/CodeQL \
      rm -rf /opt/pipx \
      rm -rf /usr/lib/google-cloud-sdk \
      rm -rf /usr/local/.ghcup \
      rm -rf /usr/local/go \
+     rm -rf /usr/local/julia1.12.4 \
```
apupier: maybe we should use rm -rf /usr/local/julia* to be sure that it continue to work with newer version of Julia

jamesnetherton: Yes, good idea. I've made that change in the latest commit.

jamesnetherton:
> Can you provide a bit more context?

As the build gets larger, we consume more disk space. I've started to see sporadic failures in some of the nightly builds and in the Quarkus Platform.

The overall disk footprint of our build is huge. The Maven repo is between 2GB & 4GB. The generated JARs & native apps in the target directories consume a lot of space. Then there's over 30GB of test containers (Docling alone is over 6GB).

> How much space it will save?

Not much TBH. Around 1GB - 1.5GB. But it's enough for now.

> Have you already evaluated to use an existing github action runner like https://github.com/marketplace/actions/disk-space-reclaimer? 

Yes, a long time back. I'd prefer we do something custom. We can't use non-approved actions in ASF projects without copying the code to our repo. IMO, it's not worth it.

gansheer:
Just so you know the jlumbroso/free-disk-space action is in the approved list (even if I get having something custom with how old the last version is) :
https://github.com/apache/infrastructure-actions/blob/main/approved_patterns.yml

jamesnetherton:
> Just so you know the jlumbroso/free-disk-space action is in the approved list

Yeah, seems somewhat undermaintained 😞.

I'll merge this to restore short-term stability and we can discuss future alternatives if needed.

PR #8300, opened 2026-02-16
Refactor: Modernize Stream API usage in AWS2 extensions and Opensearch tests

sanjana2505006:
This PR modernizes the codebase by adopting cleaner Java Stream API patterns.

**Changes:**
- Replaced `collect(Collectors.toList())` with `.toList()` in `Aws2EksProcessor`, `Aws2AthenaProcessor`, and `Aws2DdbProcessor`.
- Optimized array stream handling in `OpensearchResource` by using `Arrays.stream()` directly.

**Verification:**
- Validated codestyle with `mvn formatter:format impsort:sort`
- Ran `mvn validate`
- Verified local compilation and processed integration tests.

PR #8302, opened 2026-02-16
Modernize AWS2 processors and fix Kinesis BOM version

sanjana2505006:
I started off by switching the AWS2 deployment processors to use `.toList()` instead of the old `Collectors.toList()` pattern. I ended up touching all 19 of them just to keep things consistent across the components.

While I was testing that, I hit a snag with the `aws2-kinesis` build. It looks like the `amazon-kinesis-client` was actually missing its version in the BOM, which only became an issue once I started cleaning up properties. I've added the version back to the root POM and the BOM so the build is stable again.

I also took a look at those CDI errors in the `opensearch` JVM tests. It seems like a systemic issue with how those tests handle bean discovery on this branch, so I've reverted the IT code to its original state but kept the Stream API updates.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -267,7 +267,7 @@
         <nats.container.image>mirror.gcr.io/nats:2.11.6</nats.container.image>
         <opensearch.container.image>mirror.gcr.io/opensearchproject/opensearch:3.1.0</opensearch.container.image>
         <openssh-server.container.image>mirror.gcr.io/linuxserver/openssh-server:version-9.7_p1-r4</openssh-server.container.image>
-        <oracle-debezium.container.image>mirror.gcr.io/gvenzl/oracle-free:23.9-slim-faststart</oracle-debezium.container.image>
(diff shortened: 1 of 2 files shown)


sanjana2505006:

jamesnetherton:
Isn't this PR the same as https://github.com/apache/camel-quarkus/pull/8300 but with a bunch of broken / unwanted changes to `pom.xml`?

sanjana2505006:
> Isn't this PR the same as #8300 but with a bunch of broken / unwanted changes to `pom.xml`?


Thank you for catching that, @jamesnetherton.
My branch was a bit out of sync, so it picked up some weird regressions in the `pom.xml`
I've just pushed a fix to revert those.

Even though #8300 does some general modernization, I kept this one separate because it hits 19 different AWS2 processors specifically and fixes a Kinesis BOM version issue that was breaking my local build.

I'm happy to merge these changes into #8300 if that's easier for you or keep this focused on AWS. Let me know what you prefer!

jamesnetherton:
Closing because it's mostly identical to https://github.com/apache/camel-quarkus/pull/8300.

Issue #7543, opened 2025-07-24
Hashicorp Vault tests fail to start due to LinkageError


jamesnetherton:
### Bug description

Not sure what's going on here. I see the following if I run the hashicorp-vault tests locally. But they seem to be passing when run on GitHub CI.

```
Caused by: java.lang.LinkageError: loader constraint violation: loader 'app' wants to load interface org.apache.commons.logging.Log. A different interface with the same name was previously loaded by io.quarkus.bootstrap.classloading.QuarkusClassLoader @36510e73. (org.apache.commons.logging.Log is in unnamed module of loader io.quarkus.bootstrap.classloading.QuarkusClassLoader @36510e73, parent loader 'app')
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:862)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:760)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:681)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:639)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:525)
	at org.springframework.http.HttpLogging.forLog(HttpLogging.java:68)
	at org.springframework.http.HttpLogging.forLogName(HttpLogging.java:57)
	at org.springframework.http.converter.AbstractHttpMessageConverter.<init>(AbstractHttpMessageConverter.java:53)
	at org.springframework.http.converter.ByteArrayHttpMessageConverter.<init>(ByteArrayHttpMessageConverter.java:44)
	at org.springframework.vault.client.VaultClients.createRestTemplate(VaultClients.java:111)
	at org.springframework.vault.client.VaultClients.createRestTemplate(VaultClients.java:88)
	at org.springframework.vault.client.RestTemplateBuilder.createTemplate(RestTemplateBuilder.java:227)
[...]

PR #8325, opened 2026-02-20
Fixes (with workaround) hashicorp-vault test running localy. Relates to #7543

JiriOndrusek:
Workaround for https://github.com/apache/camel-quarkus/issues/7543

 When hashicorp test runs on the system, which contains this module's /target/classes folder, test  fails.
 Reason is explained in the ticket https://github.com/apache/camel-quarkus/issues/7543

 Artifact created by this module does not contain anything (except license) in the target/classes folder.
 Workaround loads and installs  the licences from the different folder (target/license) and therefore does not trigger the Quarkus classloader.

### DIFF
--- a/extensions-support/spring/shade/core/pom.xml
+++ b/extensions-support/spring/shade/core/pom.xml
@@ -82,6 +82,33 @@
                     </execution>
                 </executions>
             </plugin>
+            <!--
+                 Workaround for https://github.com/apache/camel-quarkus/issues/7543.
+                 This module produces an artifact that does not include any compiled classes only license files, which causes tests to fail when the empty target/classes directory is detected.
+                 The workaround explicitly loads and installs the license files from the target/license directory.
+            -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-resources-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>legal-resources</id>
+                        <phase>process-resources</phase>
+                        <goals>
+                            <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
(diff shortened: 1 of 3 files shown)


JiriOndrusek:
@jamesnetherton I think that https://github.com/apache/camel-quarkus/issues/7543 can be closed, but follow-up ticket would be nice (to solve the problem "cleanly" - this is more like a workaround)

gansheer (approved):

extensions-support/spring/shade/core/pom.xml
```diff
@@ -82,6 +82,35 @@
                     </execution>
                 </executions>
             </plugin>
+            <!-- Workaround for https://github.com/apache/camel-quarkus/issues/7543
+                 See more information in th ticket.
+                 Shortly, when hashicorp test runs onb the system, which contains this module/target/classes, it fails.
+                 Reason is explained in the ticker https://github.com/apache/camel-quarkus/issues/7543
+                 Artifact created by this module does mot contain anything (except license) in the target/classes folder.
+                 Workaround loads and installs  the licences from the folder target/license
+                 -->
```
jamesnetherton:
```suggestion
            <!--
                 Workaround for https://github.com/apache/camel-quarkus/issues/7543.

                 This module produces an artifact that does not include any compiled classes only license files, which causes tests to fail when the empty target/classes directory is detected.

                 The workaround explicitly loads and installs the license files from the target/license directory.
            -->
```

jamesnetherton (approved):
Added some suggested wording tidy-ups.

extensions-support/spring/shade/core/pom.xml
JiriOndrusek: Thanks for the rewording my comment! It's fixed in the PR now.

Issue #8277, opened 2026-02-10
camel-quarkus-opentelemetry2 disconnected vertx-http spans


jamesnetherton:
### Bug description

I discovered this by switching the [Observability example project](https://github.com/apache/camel-quarkus-examples/tree/main/observability) from `camel-quarkus-http` to `camel-quarkus-vertx-http`.

Here's a typical view of how tracing looks. All 10 spans should be connected together, but they are split:

<img width="500" height="500" alt="vertx-http-traces" src="https://github.com/user-attachments/assets/4c023679-8e36-4377-92b8-3ba72d54551d" />

I think it's caused when the Vert.x HTTP Client creates its own internal spans. When tracing is suppressed on the client, then all of the Camel route spans are connected properly.

I tested with a standalone `camel-main` application and this issue does not occur when the Otel agent is used.

The workaround will be to disable internal tracing for the HTTP client. You loose a span for the request but we still have the `vertx-http` endpoint span, which contains mostly identical info.

jamesnetherton:
The other thing to mention is that even with a plain Quarkus app with using `quarkus-reactive-routes` and the Vert.x HTTP Client, the spans all end up forming the expected hierarchy.

So there is perhaps some combination of platform-http / vertx-http in Camel & Camel Quarkus causing issues that needs investigating.

PR #8331, opened 2026-02-23
Disable internal HTTP client tracing for vertx-http

jamesnetherton:
Relates to #8277.


jamesnetherton:
Hmmm - native test failed. I don't see that locally, so will need to investigate.

jamesnetherton:
Rebased and set the PR back to 'ready for review' state.

I am not sure what happened with the initial failure. The CI logs suggest multiple instances of the native test app got launched. I have run the tests 50+ times on my CQ fork and have not been able to reproduce the failure.

I added some additional debug logic in case it happens again.

PR #8363, opened 2026-03-02
feat(micrometer): enable logging metrics on shutdown

squakez:
This PR enables the possibility to log metrics when an application shuts down. It requires changes available in Camel core provided by https://github.com/apache/camel/pull/21673

Ref [CAMEL-23089](https://issues.apache.org/jira/browse/CAMEL-23089)

### DIFF
--- a/extensions/micrometer/runtime/src/main/java/org/apache/camel/quarkus/component/micrometer/CamelMicrometerConfig.java
+++ b/extensions/micrometer/runtime/src/main/java/org/apache/camel/quarkus/component/micrometer/CamelMicrometerConfig.java
@@ -105,6 +105,22 @@ public interface CamelMicrometerConfig {
      */
     Optional<String> routePolicyExcludePattern();
 
+    /**
+     * Log metrics when application is shutting down. (default, `false`).
+     *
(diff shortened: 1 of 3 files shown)


extensions/micrometer/runtime/src/main/java/org/apache/camel/quarkus/component/micrometer/CamelMicrometerConfig.java
```diff
@@ -105,6 +105,22 @@ public interface CamelMicrometerConfig {
...
+    @WithDefault("false")
+    boolean logMetricsOnShutdown();
+
+    /**
+     * List of metrics (comma separated) to log when application is shutting down. You can use `*` character to log any
+     * metrics containing the wildcard, for example `camel.exchanges.*` (default to all metrics available).
+     *
+     * @asciidoclet
+     */
+    Optional<String> logMetricsOnShutdownFilters();
```
jamesnetherton: You can avoid the manual String split and rely on the builtin SmallRye Config converters.

```suggestion
    Optional<String[]> logMetricsOnShutdownFilters();
```

PR #8364, opened 2026-03-02
Add Qdrant exception and auth tests

llowinge:

### DIFF
--- a/integration-tests/qdrant/src/test/java/org/apache/camel/quarkus/component/qdrant/it/QdrantAuthTest.java
+++ b/integration-tests/qdrant/src/test/java/org/apache/camel/quarkus/component/qdrant/it/QdrantAuthTest.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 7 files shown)


llowinge:
@apupier Please review if you will have a time. Thanks

integration-tests/qdrant/src/test/java/org/apache/camel/quarkus/component/qdrant/it/QdrantAuthTest.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Test
+    void invalidApiKeyShouldBeRejected() {
+        // Verify that authentication is actually enforced with wrong API key
+        RestAssured.put("/qdrant/apiKey/invalid")
+                .then()
+                .statusCode(200)
+                .body(containsString("Authentication correctly failed"));
+    }
```
apupier, line 46: at first read, I was wondering why there is a status code 200 when the api key is invalid, but given it s relatively specific use case, I'm not sure we can write it in a better way, I mean having the 403 or something like that as i'm not sure it is surfacing up from the underlying calls. Just a note that it made me tick when reading it at first, in case you have a second thought on it.

aldettinger (approved):

integration-tests/qdrant/src/test/java/org/apache/camel/quarkus/component/qdrant/it/QdrantAuthTest.java
llowinge: thanks, i've changed it to 403 to have it more clear and readable

PR #8366, opened 2026-03-03
feat(observability-svc): default logging metrics on shutdown

squakez:
Follows #8363 (to be rebased after that one is merged).

### DIFF
--- a/extensions/observability-services/runtime/src/main/resources/application.properties
+++ b/extensions/observability-services/runtime/src/main/resources/application.properties
@@ -16,6 +16,8 @@
 ## ---------------------------------------------------------------------------
 quarkus.management.enabled=true
 quarkus.management.port=9876
+quarkus.camel.metrics.log-metrics-on-shutdown=true
+quarkus.camel.metrics.log-metrics-on-shutdown-filters=app.info,camel.exchanges.*
 quarkus.micrometer.export.prometheus.path=/observe/metrics
 quarkus.smallrye-health.root-path=/observe/health
 # Default is disabled: the user must enable it to push traces to the given server


oscerd (approved):

extensions/observability-services/runtime/src/main/resources/application.properties
```diff
@@ -16,6 +16,8 @@
 ## ---------------------------------------------------------------------------
 quarkus.management.enabled=true
 quarkus.management.port=9876
+quarkus.camel.metrics.log-metrics-on-shutdown=true
+quarkus.camel.metrics.log-metrics-on-shutdown-filters=app.info,camel.exchanges.*
```
jamesnetherton, line 20: Nitpick - would probably be better adding it to the micrometer itest module, since that's where most of the metrics testing lives.

squakez: Testing this one could be a little cumbersome as it requires to check the application log after the Camel route is stopped. I tried to do the same on the core and I ended up testing the logic, but the check on the log after the route is complete. It could be nice, but I am not sure I'd be able to put something in place that easily. Do you have any example by any chance?

jamesnetherton: > Do you have any example by any chance?

Example config here:

https://github.com/apache/camel-quarkus/blob/bbe10ba1e78203bd6c7e0cac7bcb06efcfabf34c/integration-tests-jvm/cli-debug/src/main/resources/application.properties#L17-L25

If you know the log category, then you can have a dedicated log file for the messages you're interested in.

Then in the tests, it's just a case of reading the log file content and doing some assertions on it (perhaps wrapped with `Awaitility.await`) .

squakez: I've just tried that, but the problem is that the feature is adding content when it shuts down the application. So, when we're in the test, that content is not yet available. It's the same test closing the application that trigger that log. I've tried to run this "AfterAll" but whenever I try, the app is still alive, hence, no content is there. Not sure how I can proceed easily.

squakez: I've re-read the debugger test, maybe I can come up with the very same approach, let me retry

squakez: No, same stuff. Using the same logic in the debugger example suffer the same problem. As the application is stopped by the test, it cannot reach those log values. Please, advice how you want to proceed.

jamesnetherton: Ok, lets merge and I will try to follow up later if I can find a way of testing it.

Issue #8370, opened 2026-03-03
langchain4j-agent: issue when using OnnxBertBiEncoder models in rag scenarios


aldettinger:
### Bug description

langchain4j-agent supports rag scenarios, so computing embeddings should be supported out of the box. However, there is an issue when using OnnxBertBiEncoder models. It could be worked around by adding a dependency to the camel-quarkus-langchain4j-embeddings extension, though it should not be needed to do so.

aldettinger:
The issue occurs only in native mode.

Stack traces as below could be produced:

```
2026-03-04 09:59:11,903 ERROR [io.quarkus.runtime.Application] (main) Failed to start application: java.lang.RuntimeException: Failed to start quarkus
	at io.quarkus.runner.ApplicationImpl.doStart(Unknown Source)
	at io.quarkus.runtime.Application.start(Application.java:112)
	at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:127)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:79)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:50)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:143)
	at io.quarkus.runner.GeneratedMain.main(Unknown Source)
Caused by: java.lang.UnsatisfiedLinkError: ai.djl.huggingface.tokenizers.jni.TokenizersLibrary.encode(JLjava/lang/String;Z)J [symbol: Java_ai_djl_huggingface_tokenizers_jni_TokenizersLibrary_encode or Java_ai_djl_huggingface_tokenizers_jni_TokenizersLibrary_encode__JLjava_lang_String_2Z]
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jni.access.JNINativeLinkage.getOrFindEntryPoint(JNINativeLinkage.java:152)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jni.JNIGeneratedMethodSupport.nativeCallAddress(JNIGeneratedMethodSupport.java:54)
	at ai.djl.huggingface.tokenizers.jni.TokenizersLibrary.encode(Native Method)
	at ai.djl.huggingface.tokenizers.HuggingFaceTokenizer.encode(HuggingFaceTokenizer.java:242)
	at ai.djl.huggingface.tokenizers.HuggingFaceTokenizer.encode(HuggingFaceTokenizer.java:253)
[...]

PR #8371, opened 2026-03-03
langchain-agent: fix issue when using OnnxBertBiEncoder models in rag…

aldettinger:
… scenarios #8370


aldettinger:
Yeah, that was a good first choice. Hope it helps to have a second look.

PR #8380, opened 2026-03-05
Fix propagation of config values from QuarkusTestProfile in AWS Secre…

llowinge:
…t Manager native tests and enable CamelContextSqsReloadTest as Native test


jamesnetherton:
Relates to https://github.com/quarkusio/quarkus/issues/52916.

PR #8382, opened 2026-03-05
[camel-main] Require JDK 21 minimum

jamesnetherton:
I will create a separate PR for `main` that updates the `camel-master-cron.yaml` scheduled workflow.

I didn't bother touching the `quarkus-lts-ci-build.yaml` workflow. At some point in the future we'll likely have a mix of supported LTS releases where some have a JDK 17 baseline and others have JDK 21. I'll leave it as a future task to handle that.

### DIFF
--- a/.github/workflows/check-dependency-convergence.yml
+++ b/.github/workflows/check-dependency-convergence.yml
@@ -40,11 +40,11 @@ jobs:
     permissions:
       issues: write
     steps:
-      - name: Set up JDK 17
+      - name: Set up JDK 21
         uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
         with:
           distribution: 'temurin'
-          java-version: '17'
+          java-version: '21'
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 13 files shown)


tooling/perf-regression/cq-perf-regression-sample-base/pom.xml
```diff
@@ -30,11 +30,11 @@
     </parent>
     <artifactId>cq-perf-regression-sample</artifactId>
     <properties>
-        <!-- Let's not upgrade hyperfoil-maven-plugin as long as camel-quarkus supports JDK17: https://github.com/apache/camel-quarkus/issues/8027 -->
+        <!-- Let's not upgrade hyperfoil-maven-plugin as long as camel-quarkus supports JDK21: https://github.com/apache/camel-quarkus/issues/8027 -->
```
apupier: I think this one should nto be updated. And we can create another task to upgrade hyperfoil

jamesnetherton: Ok - I put the file back to how it is now on `main`.

jamesnetherton:
I nobody objects, I am going to merge this. I don't want to get caught up with unrelated build failures due to changes on the camel main branch. 

I think we've proved the build works in the initial CI run.

PR #8395, opened 2026-03-09
Upgrade Quarkus Amazon Services to 3.15.0

jamesnetherton:


jamesnetherton:
I will investigate the native mode build failure:

```
Caused by: java.lang.ClassNotFoundException: software.amazon.awssdk.crt.checksums.CRC64NVME
```

jamesnetherton:
Latest commit adds some native mode fixes for AWS checksums that are only present if `aws-crt` is on the classpath.

PR #8403, opened 2026-03-10
Use dynamic db-kind for all configuration profiles in JPA tests

jamesnetherton:


jamesnetherton:
Realised I also need to do the same for SQL tests. So latest commit fixes that.

zhfeng (approved):
Good catch!

PR #8404, opened 2026-03-10
Fix propagation of config values from QuarkusTestProfile in Azure Key…

llowinge:
… Vault native tests


jamesnetherton:
Relates to https://github.com/quarkusio/quarkus/issues/52916.

PR #8411, opened 2026-03-11
Avoid using reserved word ApiResponse in rest-openapi specs

jamesnetherton:


integration-tests/rest-openapi/src/main/openapi/petstore.json
apupier, line 1: is it a file coming from an official example?
If yes, it would mean that each person that we are no more testing with the version that a lot of people will use to do their first tests, isn't it?

jamesnetherton: > is it a file coming from an official example

It was probably copied from the Camel core component tests. But, yes, likely originates from the swagger-petstore project.

>  it would mean that each person that we are no more testing with the version that a lot of people will use to do their first tests, isn't it?

Yes, true. Not sure if tweaking the name of an API response is a big deal though. I assume if folks see a bunch of compilation warnings, they'd modify the specs to get rid of them.

PR #8419, opened 2026-03-12
 [Merge when supporting Camel 4.19.0] chore(extensions): deprecate camel-tracing based extensions

squakez:
Ref https://github.com/apache/camel/pull/21898


jamesnetherton:
There's usually no need to do this manually, so long as the component deprecation status is set correctly in the Camel Catalog.

We have tooling that will automatically adjust the extension metadata based on that.

squakez:
> There's usually no need to do this manually, so long as the component deprecation status is set correctly in the Camel Catalog.
> 
> We have tooling that will automatically adjust the extension metadata based on that.

Ah, nice, I was not aware of the mechanism. Closing the PR then.

PR #8422, opened 2026-03-12
Use Awaitility for elasticsearch-rest assertions with mutating and destrucive operations

jamesnetherton:
Speculative fix for intermittent failures of elasticsearch-rest tests.


jamesnetherton:
Test failed so I guess there's something else wrong.

jamesnetherton:
I borrowed the health check logic from `integration-tests-jvm/elasticsearch` to make sure the cluster state is as expected before running tests.

PR #8430, opened 2026-03-16
Downgrade camel-main scheduled workflow to JDK 17 minimum

jamesnetherton:
This reverts commit 50953c1153fbcf772bc898d466c132eaaa624df3.


apupier:
I' not sure that we can do a full revert as JDk 25 will surely be supported, isn't it?

jamesnetherton:
> I' not sure that we can do a full revert as JDk 25 will surely be supported, isn't it?

We have the weekly workflow for JDK 25. So we could perhaps use that for verification. WDYT?

PR #8436, opened 2026-03-18
Add http non proxy host tests usecase

svkcemk:

### DIFF
--- a/integration-test-groups/http/http/src/test/java/org/apache/camel/quarkus/component/http/http/it/HttpTest.java
+++ b/integration-test-groups/http/http/src/test/java/org/apache/camel/quarkus/component/http/http/it/HttpTest.java
@@ -16,6 +16,8 @@
  */
 package org.apache.camel.quarkus.component.http.http.it;
 
+import java.util.stream.Stream;
+
 import io.quarkus.test.common.QuarkusTestResource;
 import io.quarkus.test.junit.QuarkusTest;
 import io.restassured.RestAssured;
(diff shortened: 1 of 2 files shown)


integration-test-groups/http/http/src/test/java/org/apache/camel/quarkus/component/http/http/it/HttpTest.java
```diff
@@ -91,4 +93,57 @@ public void compression() {
...
+        RestAssured.given()
+                .queryParam("non-proxy-hosts", "*localhost*")
+                .queryParam("proxy-port", fakeProxyPort)
+                .queryParam("proxy-host", proxyHost)
+                .when()
+                .get("/test/client/{component}/nonProxy", component())
+                .then()
+                .statusCode(500);
+
+    }
```
apupier: What about splitting in different tests?

It will help having a clearer view on which parts are failign and which are passing in case of one failure.
It woudl also help to see in junit eport what is tested and not.
The comments might even become redundant with test name method and could be removed.

svkcemk: Yeah I can do that I thought of keeping all non proxy host tests together!

apupier: Good idea to group them together.
it is possible to use Nested tests https://docs.junit.org/6.0.3/writing-tests/nested-tests.html to group them

svkcemk: @apupier I tried to clean up the tests using `ParameterizedTest`could you mind to take a look please

PR #8438, opened 2026-03-18
Upgrade Quarkus to 3.33.0

jamesnetherton:


jamesnetherton:
Pushed a fix for failing servlet tests:

```
Caused by: io.smallrye.config.ConfigValidationException: Configuration validation failed:
	java.util.NoSuchElementException: SRCFG00011: Could not expand value java.io.tmpdir in property quarkus.camel.servlet.my-servlet.multipart.location
	at io.smallrye.config.SmallRyeConfig.buildMappings(SmallRyeConfig.java:172)
	at io.smallrye.config.SmallRyeConfig.<init>(SmallRyeConfig.java:126)
	at io.smallrye.config.SmallRyeConfigBuilder.build(SmallRyeConfigBuilder.java:785)
	at io.quarkus.runtime.generated.SharedConfig.<clinit>(Unknown Source)
```

I think what we have been doing in the servlet config for years is wrong. Doesn't really make sense to have config that's `BUILD_AND_RUN_TIME_FIXED` defaulted to `java.io.tmpdir`.

The solution is to just pass a `null` default for the location on the multipart config, and the Undertow servlet internals from `quarkus-vertx-http` will use `java.io.tmpdir` at runtime.

gansheer (approved):

PR #8444, opened 2026-03-19
fix: extension metadata status is a string value

gansheer:
Ref https://github.com/quarkusio/quarkus/issues/52765

It appears that the right type for status value is a string.
* metadata extension template has been fixed
* metadatas has been updated


gansheer:
@jamesnetherton Should the change be also done on 3.27/ 3.20 ?

jamesnetherton:
> @jamesnetherton Should the change be also done on 3.27/ 3.20 ?

We could do it for 3.27. I am not sure whether we will have further releases of 3.20.x.

gansheer:
> 
> We could do it for 3.27. I am not sure whether we will have further releases of 3.20.x.
>
Noted I will take care of the backports.

Issue #7733, opened 2025-09-15
Intermittent failure of GroupedAws2KinesisTest#failingDefaultCredentialsProviderTest


llowinge:
### Bug description

In this run https://github.com/apache/camel-quarkus/actions/runs/17657311248 the test `GroupedAws2KinesisTest#failingDefaultCredentialsProviderTest` failed in Alternative JDK 21 with:

```
org.opentest4j.AssertionFailedError: Expected java.lang.AssertionError to be thrown, but nothing was thrown.
	at org.apache.camel.quarkus.test.support.aws2.BaseAWs2TestSupport.failingDefaultCredentialsProviderTest(BaseAWs2TestSupport.java:99)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at io.quarkus.test.junit.QuarkusTestExtension.runExtensionMethod(QuarkusTestExtension.java:1000)
	at io.quarkus.test.junit.QuarkusTestExtension.interceptTestMethod(QuarkusTestExtension.java:848)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

and full log:

<details>
  <summary>
full log
  </summary>

```
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.205 s -- in org.apache.camel.quarkus.component.aws2.kinesis.it.GroupedAws2KinesisFirehoseTest
[INFO] Running org.apache.camel.quarkus.component.aws2.kinesis.it.GroupedAws2KinesisTest
2025-09-11 21:38:54,617 INFO  [org.apa.cam.qua.tes.sup.aws.Aws2TestResource] (docker-java-stream--1160732461) STDOUT: 2025-09-11T21:38:54.614  INFO --- [et.reactor-0] localstack.request.aws     : AWS kinesis.GetRecords => 200
2025-09-11 21:38:54,618 INFO  [org.apa.cam.qua.tes.sup.aws.Aws2TestResource] (docker-java-stream--1160732461) STDOUT: 2025-09-11T21:38:54.615  INFO --- [et.reactor-3] localstack.request.aws     : AWS kinesis.GetRecords => 200
[...]

PR #8470, opened 2026-03-25
[3.33.x] Upgrade Quarkus to 3.33.1

jamesnetherton:
Also added backport for https://github.com/apache/camel-quarkus/pull/8469.

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -28,7 +28,7 @@ asciidoc:
     camel-version: 4.18.0 # replace ${camel.version}
     camel-docs-version: 4.18.x
     camel-quarkus-version: 3.33.1 # replace ${camel-quarkus.version}
-    quarkus-version: 3.33.0 # replace ${quarkus.version}
+    quarkus-version: 3.33.1 # replace ${quarkus.version}
     graalvm-version: 23.1.2 # replace ${graalvm.version}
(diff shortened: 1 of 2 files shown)


apupier:
Does it mean that we shoudl skip Camel Quarkus 3.33.0 and move directly to 3.33.1 to align with Quarkus version? Or it is fine as long as we are aligned with the first 2 digits?

jamesnetherton:
> Does it mean that we shoudl skip Camel Quarkus 3.33.0 and move directly to 3.33.1 to align with Quarkus version? Or it is fine as long as we are aligned with the first 2 digits?

Not sure I understood the question, but...

Quarkus Core 3.33.1 is released and will get updated in the Quarkus Platform. So we should do likewise and upgrade here. We will continue to update to the latest Quarkus 3.33.x releases when they become available. And will eventually do new Camel Quarkus 3.33.x releases (if needed) in line with the Quarkus Platform LTS schedule.

Assuming there are no breaking changes in the Quarkus patch releases, then we don't need to release unless we have something of value to deliver to users.

apupier:
ok, I thought the Camel Quarkus 3.33.0 was not yet released officially. My bad.

jamesnetherton:
Failing `GroupedAws2KinesisTest.failingDefaultCredentialsProviderTest` is a known [flake](https://github.com/apache/camel-quarkus/issues/7733).

PR #8476, opened 2026-03-26
langchain4j-agent: Reinstate some of the camel-quarkus-langchain4j tests

aldettinger:


aldettinger:
@jamesnetherton It would be great if you could check whether this is good candidate to reinstate with nowadays context. This is part of what was removed in August 2025.

PR #8523, opened 2026-04-13
Add ibm-watson-language native support and integration tests

### LINKED ISSUE #7991: Add ibm-watson-language native support and integration tests

JinyuChen97:
fixes #7991

Promote the ibm-watson-language extension from JVM-only to native mode support.

All Camel core functionalities of the ibm-watson-language extension were tested in a real IBM Cloud environment and passed. (Tests only run when a real API key and service URL are provided; otherwise, they are skipped.)

(Limitations) Two warnings were identified: the Watson SDK property files (java-sdk-version.properties and sdk-core-version.properties) cannot be loaded in native mode. This does not impact functionality, but results in an "unknown version" value in the User-Agent header.


extensions/ibm-watson-language/runtime/pom.xml
```diff
@@ -43,6 +44,15 @@
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-ibm-watson-language</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-gson</artifactId>
+        </dependency>
```
jamesnetherton, line 60: Can we have this instead?

```suggestion
        <dependency>
            <groupId>com.google.code.gson</groupId>
            <artifactId>gson</artifactId>
        </dependency>
```

extensions/ibm-watson-language/deployment/src/main/java/org/apache/camel/quarkus/component/ibm/watson/language/deployment/IbmWatsonLanguageProcessor.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Quarkus extension processor for IBM Watson Natural Language Understanding support.
+ * <p>
+ * <b>Known Limitation in Native Mode:</b>
+ * In GraalVM native mode, Watson SDK cannot load {@code java-sdk-version.properties} and
+ * {@code sdk-core-version.properties}, resulting in warning logs:
+ *
+ * <pre>
+ * WARNING [com.ibm.watson.common.SdkCommon] Could not load java-sdk-version.properties
+ * WARNING [com.ibm.cloud.sdk.core.util.RequestUtils] Could not load file: sdk-core-version.properties
```
jamesnetherton: The usual fix for this is to first find where those files live on the classpath. Then register `NativeImageResourceBuildItem` for them.

extensions/ibm-watson-language/deployment/src/main/java/org/apache/camel/quarkus/component/ibm/watson/language/deployment/IbmWatsonLanguageProcessor.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final String FEATURE = "camel-ibm-watson-language";
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    void registerForReflection(
+            io.quarkus.deployment.builditem.CombinedIndexBuildItem combinedIndex,
```
jamesnetherton: Not sure why, but Claude etc seem to like writing type references using the FQCN. Please take a scan through this class and replace it with imports so that the code is more readable.

apupier:
> Not sure why, but Claude etc seem to like writing type references using the FQCN

If an AI model was used, it should be mentioned in the commit message too. Specifying the model/software used.

extensions/ibm-watson-language/runtime/src/main/java/org/apache/camel/quarkus/component/ibm/watson/language/graal/GenericModelSubstitutions.java
```diff
@@ -0,0 +1,59 @@
...
+import com.oracle.svm.core.annotate.TargetClass;
+
+final class GenericModelSubstitutions {
+}
+
+/**
+ * Substitute DynamicModelTypeAdapterFactory to avoid Gson internal API dependency.
+ * <p>
+ * The IBM Watson SDK's DynamicModelTypeAdapterFactory uses Gson internal classes
+ * (com.google.gson.internal.$Gson$Types) which cannot be resolved at native image build time.
```
apupier: I'm wondering if it can be similar issue than https://github.com/IBM/java-sdk-core/issues/233 which would require only to upgrade the gson version

extensions/ibm-watson-language/runtime/pom.xml
JinyuChen97: Done

extensions/ibm-watson-language/deployment/src/main/java/org/apache/camel/quarkus/component/ibm/watson/language/deployment/IbmWatsonLanguageProcessor.java
JinyuChen97: I was try to use 
@BuildStep                                                                                                                                                                                                                                                                                    
      void registerResources(BuildProducer<NativeImageResourceBuildItem> resource) {                                                                                                                                                                                                                
        resource.produce(new NativeImageResourceBuildItem("java-sdk-version.properties"));                                                                                                                                                                                                        
        resource.produce(new NativeImageResourceBuildItem("sdk-core-version.properties"));                                                                                                                                                                                                        
      }. or NativeImageResourcePatternsBuildItem to match all *version.properties files but the warnings still persist in native mode. The issue looks like the watson sdk uses its own ClassLoader (SdkCommon.class.getClassLoader()) to load these properties, and in native mode it cannot access resources registered in our extension's context. Please correct me if I'm wrong.

extensions/ibm-watson-language/deployment/src/main/java/org/apache/camel/quarkus/component/ibm/watson/language/deployment/IbmWatsonLanguageProcessor.java
JinyuChen97: Done

extensions/ibm-watson-language/runtime/src/main/java/org/apache/camel/quarkus/component/ibm/watson/language/graal/GenericModelSubstitutions.java
JinyuChen97: yes, the issue solved by import ibm sdk core and upgrade it to 9.25.0. Thanks

extensions/ibm-watson-language/runtime/pom.xml
```diff
@@ -43,6 +44,20 @@
...
             <artifactId>camel-ibm-watson-language</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.ibm.cloud</groupId>
+            <artifactId>sdk-core</artifactId>
+            <version>9.25.0</version>
```
apupier: I think that this kind of version alignment must be done in the bom. Also it mean sthat we should spot where it was defined first and see if we can upgrade it there too

JinyuChen97: okay, the ibm sdk core version 9.23.1 defined in bom is enough to fix the gson issue. I exclude the 9.20.0 brought from ibm-watson-language and use the 9.23.1 one until ibm-watson-language upgrade their sdk core to 9.23.1, is this fine?

extensions/ibm-watson-language/runtime/pom.xml
jamesnetherton: You'll need to remove `camel-quarkus-gson-deployment` from the deployment module also.

extensions/ibm-watson-language/deployment/src/main/java/org/apache/camel/quarkus/component/ibm/watson/language/deployment/IbmWatsonLanguageProcessor.java
jamesnetherton: This worked fine for me:

```
@BuildStep
void addResources(BuildProducer<NativeImageResourceBuildItem> nativeImageResource) {
    nativeImageResource.produce(new NativeImageResourceBuildItem("java-sdk-version.properties"));
    nativeImageResource.produce(new NativeImageResourceBuildItem("sdk-core-version.properties"));
}
```

Your code snippet has the raw `BuildProducer resource`. Not sure if that perhaps makes it not work.

apupier (changes requested):
I guess current build failure is due to https://github.com/apache/camel-quarkus/pull/8523#discussion_r3079572708

extensions/ibm-watson-language/deployment/src/main/java/org/apache/camel/quarkus/component/ibm/watson/language/deployment/IbmWatsonLanguageProcessor.java
JinyuChen97: yes, it fixed the warning. Thanks

apupier (changes requested):
you need to rebuild the pom folder as reported by the build failure:

```
[7](https://github.com/apache/camel-quarkus/actions/runs/24405469029/job/71288929897?pr=8523#step:10:18)
	modified:   poms/bom/src/main/generated/flattened-full-pom.xml
	modified:   poms/bom/src/main/generated/flattened-reduced-pom.xml
	modified:   poms/bom/src/main/generated/flattened-reduced-verbose-pom.xml
```

jamesnetherton:
@JinyuChen97 Please rebase.

apupier (changes requested):
The cmmit title is just "fix pipeline issue" since last force push, I guess it eased the main goal of the change to add natives upport and integration tests

apupier:
the couchdb tests seems to fail consistently, even though I do not see the link with current PR

Maybe try to rebase on latest main branch?

extensions/ibm-watson-language/runtime/pom.xml
```diff
@@ -42,6 +43,25 @@
...
+                <exclusion>
+                    <groupId>com.ibm.cloud</groupId>
+                    <artifactId>sdk-core</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>com.ibm.cloud</groupId>
+            <artifactId>sdk-core</artifactId>
+        </dependency>
```
apupier: What is the reason to exclude the dependency in camel-ibm-watson-language and readd it on the quarkus component?

extensions/ibm-watson-language/runtime/pom.xml
```diff
@@ -42,6 +43,25 @@
...
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+        </dependency>
```
apupier: Why do we need to add gson dependency?

extensions/ibm-watson-language/runtime/pom.xml
JinyuChen97: That's because the ibm watson language is still using 9.20.0 java-sdk-core and I use the similar way like https://github.com/apache/camel-quarkus/pull/7284 to remove the version and add the 9.23.2 java-sdk-core from bom. But yes, I can just directly import it without remove it and the old sdk core should be overrided

JinyuChen97:
> the couchdb tests seems to fail consistently, even though I do not see the link with current PR
> 
> Maybe try to rebase on latest main branch?

hmmm, I'm not really sure if it related to the sdk-core version change? as it was changed by jiri before https://github.com/apache/camel-quarkus/pull/7284, but I found the bom sdk core version removed, I guess the com.ibm.cloud:cloudant containing sdk-core 9.23.1+ is used? 

updated: rebased and see if pipeline fail again

extensions/ibm-watson-language/runtime/pom.xml
JinyuChen97: after tested, it passed from ibm extension, so it can be removed.

apupier (changes requested):
couchdbtest are still failing consistently, I do not know how but it is affecting another component, surely some dependent library mismatch

JinyuChen97:
> couchdbtest are still failing consistently, I do not know how but it is affecting another component, surely some dependent library mismatch

the original couchdb has 9.25.0 version of sdk core import from camel, but the sdk core version 9.23.1 import in poms/bom effected the sdk core version been used in couchdb caused to some core dependencies of okhttp been exclude by bom configuration. (For 9.23.1 sdk core, okhttp 4.x.x is used). Upgrade to 9.25.0 sdk core fix this issue as okhttp 5.x.x is used.

apupier (approved):

jamesnetherton (approved):

PR #8535, opened 2026-04-15
Temporarily work around inaccessible Shibboleth Maven repositories

jamesnetherton:
I can't think of any better options that'd allow us to keep CI working...


jamesnetherton:
I still need to fix up the BOM generation.

jamesnetherton:
Seems this could work - would need to disable the `cxf-soap` example native build to have a fully green CI.

apupier:
Shibboleth should be back at least temporarily: https://shibboleth.net/pipermail/announce/2026-April/000378.html

jamesnetherton:
Seems to not be needed (for now...).

PR #8574, opened 2026-04-24
Pin RabbitMQ container image to rabbitmq:4.2.4-management-alpine

jamesnetherton:
Aligns with Camel and avoids breaking 4.3.x changes that are now in the floating image tag.


jamesnetherton:
`GooglePubsubTest` failure is not related - thus merging this.

PR #8580, opened 2026-04-25
fix(#8525): add virtual thread support integration tests

### LINKED ISSUE #8525: Add test coverage for virtual thread support
We have somewhat neglected virtual thread support in this project.

Camel 4.19.0 will bring a new configuration option `camel.main.virtualThreadsEnabled`, so it'd be good to be able to test it. We'll have to use `EnabledForJreRange` given the default CI baseline is JDK 17.

redinside-dev:
## Why

Camel Quarkus has neglected virtual thread support testing. Camel 4.19.0 introduces the \`camel.main.virtualThreadsEnabled\` configuration option, but there was no test coverage for it. Without virtual thread tests, regressions in virtual thread compatibility won't be caught.

## What

Add \`CoreMainVirtualThreadsTest\` integration test class gated with \`@EnabledForJreRange(min = 21)\` since default CI baseline is JDK 17. Tests use a dedicated \`VirtualThreadsTestProfile\` that sets \`camel.main.virtualThreadsEnabled=true\` only for this test class, ensuring isolation from other tests in the module.

The test class verifies:
- **Startup**: App starts successfully with virtual threads enabled
- **Route health**: Routes are present and CamelContext is healthy with VT
- **Property resolution**: Property placeholder resolution works with VT
- **Type conversion**: Custom type converters work with VT
- **Component resolution**: Custom components resolve correctly with VT
- **Reactive executor**: Reactive executor availability with VT
- **Thread pool factory**: Thread pool factory availability with VT
- **Graceful shutdown**: JMX connector service starts cleanly with VT

## Testing

Tests are gated with \`@EnabledForJreRange(min = 21)\` so they are automatically skipped on JDK < 21. They use a dedicated \`QuarkusTestProfile\` that overrides \`camel.main.virtualThreadsEnabled=true\` only for the VT test class, ensuring complete isolation from the default \`CoreMainTest\` suite.

## Related issues

Fixes apache/camel-quarkus#8525

### DIFF
--- a/integration-tests/main/src/test/java/org/apache/camel/quarkus/main/CoreMainVirtualThreadsTest.java
+++ b/integration-tests/main/src/test/java/org/apache/camel/quarkus/main/CoreMainVirtualThreadsTest.java
@@ -1,0 +1,196 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


jamesnetherton:
@anuragg-saxenaa If you point an AI agent at issues, please check to see whether they are already assigned, or ask if they can be worked on.

PR #8583, opened 2026-04-27
Add security-policy integration test for CAMEL-23250

gnodet:
## Summary

- Adds `integration-test-groups/foundation/security-policy` module to test Camel's security policy enforcement framework (CAMEL-23250) in Quarkus
- Uses `QuarkusTestProfile` with `SecurityPolicyResult` plugin inspection instead of process executor
- 4 test scenarios: warn detection, category override, allowed-properties exclusion, allow policy

## Test cases

1. **SecurityPolicyWarnTest** — default warn policy detects violations via `SecurityPolicyResult` plugin
2. **SecurityPolicyCategoryOverrideTest** — `insecure-dev-policy=allow` suppresses insecure:dev violations
3. **SecurityPolicyAllowedPropertiesTest** — `allowed-properties` excludes specific properties from enforcement
4. **SecurityPolicyAllowTest** — `policy=allow` suppresses all violations

## Prerequisites

Requires Camel 4.21.0+ which includes the security policy framework and the fix from https://github.com/apache/camel/pull/22779 (enforcement running on empty properties map).

## Test plan

- [x] All 4 tests pass in JVM mode
- [ ] CI verification

_Claude Code on behalf of Guillaume Nodet_

### DIFF
--- a/integration-test-groups/foundation/security-policy/src/main/java/org/apache/camel/quarkus/security/policy/SecurityPolicyResource.java
+++ b/integration-test-groups/foundation/security-policy/src/main/java/org/apache/camel/quarkus/security/policy/SecurityPolicyResource.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.security.policy;
+
+import java.util.stream.Collectors;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+import org.apache.camel.CamelContext;
+import org.apache.camel.main.SecurityPolicyResult;
+
+@Path("/security-policy")
+@ApplicationScoped
+public class SecurityPolicyResource {
+
+    @Inject
+    CamelContext camelContext;
+
+    @GET
+    @Path("/has-violations")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String hasViolations() {
+        SecurityPolicyResult result = camelContext.getCamelContextExtension()
+                .getContextPlugin(SecurityPolicyResult.class);
+        if (result == null) {
+            return "null";
+        }
+        return String.valueOf(result.hasViolations());
+    }
+
+    @GET
+    @Path("/violation-categories")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String violationCategories() {
+        SecurityPolicyResult result = camelContext.getCamelContextExtension()
+                .getContextPlugin(SecurityPolicyResult.class);
+        if (result == null) {
+            return "";
+        }
+        return result.getViolations().stream()
+                .map(v -> v.category())
+                .collect(Collectors.joining(","));
+    }
+
+    @GET
+    @Path("/violation-property-keys")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String violationPropertyKeys() {
+        SecurityPolicyResult result = camelContext.getCamelContextExtension()
+                .getContextPlugin(SecurityPolicyResult.class);
+        if (result == null) {
+            return "";
+        }
+        return result.getViolations().stream()
+                .map(v -> v.propertyKey())
+                .collect(Collectors.joining(","));
+    }
+
+    @GET
+    @Path("/violation-policies")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String violationPolicies() {
+        SecurityPolicyResult result = camelContext.getCamelContextExtension()
+                .getContextPlugin(SecurityPolicyResult.class);
+        if (result == null) {
+            return "";
+        }
+        return result.getViolations().stream()
+                .map(v -> v.policy())
+                .collect(Collectors.joining(","));
+    }
+}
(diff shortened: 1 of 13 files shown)


jamesnetherton:
>Uses process executor pattern (like main-unknown-args-fail) to verify startup behavior under different security policies

We don't need to start a secondary process with `QuarkusProcessExecutor` just to pass custom configuration options. We can use `QuarkusTestProfile` to achieve the same thing.

I think it might be better (assuming it works ok) to create these tests in `integration-test-groups/foundation/security-policy`. That way we don't need to add yet another module under `integration-tests`.

apupier (changes requested):
you need to target the camel-main branch for features which are not yet part of a release of Camel

.github/workflows/quarkus-lts-ci-build.yaml
apupier, line 27: this change is unrelated to the PR

apupier:
I updated (force-push) the `camel-main` branch to have it based on the 4.21.0-SNAPSHOT.
Please rebase on top of it. it will simplify the review and also avoid mixing unrelated commits in this PR

jamesnetherton:
Every test class will need a corresponding IT class annotated with `QuarkusIntegrationTest`. Constructs like `@Inject` wont work in ITs, so the tests will need to be adjusted to get any required data for assertions, from the application under test via JAX-RS endpoints etc.

gnodet:
@jamesnetherton @apupier feedback addressed

jamesnetherton (approved):

jamesnetherton:
Changes look ok to me. Might be best to get this merged ASAP before `camel-main` branch starts to drift into failure due to changes in the Camel core.

apupier:
current failure are also happening on camel-main branch https://github.com/apache/camel-quarkus/actions/runs/25027412512/job/73304129024

AGENTS.md
```diff
@@ -196,6 +196,12 @@ class KafkaIT extends KafkaTest {
 }
 ```
 
+### Important Constraints
+
+- **No `@Inject` in `@QuarkusIntegrationTest`**: Integration tests run the application as a separate process, so CDI injection does not work. Expose application state via JAX-RS endpoints (using `quarkus-resteasy`) and assert via RestAssured in tests. See `integration-test-groups/foundation/core/` for the pattern.
+- **Every `@QuarkusTest` needs an IT class**: Each `@QuarkusTest` class must have a corresponding `@QuarkusIntegrationTest` class that extends it, to ensure tests also run in native mode.
+- **Grouped test modules share one native application**: Modules under `integration-test-groups/` are compiled into a single grouped native binary per group. Properties in one module's `application.properties` affect ALL modules in the group. Avoid setting properties that alter global Camel behavior (e.g. `camel.main.duration-max-messages`).
+
```
apupier: this is unrelated to the PR and commit message.

this should be provided in a different PR and can be provided directly on main branch. it will ease the maintenance of the camel-main branch

integration-test-groups/foundation/security-policy/src/test/java/org/apache/camel/quarkus/security/policy/SecurityPolicyWarnTest.java
```diff
@@ -0,0 +1,36 @@
...
+@QuarkusTest
+class SecurityPolicyWarnTest {
+
+    @Test
+    void warnPolicyDetectsViolations() {
+        RestAssured.when().get("/security-policy/has-violations").then().body(is("true"));
+        RestAssured.when().get("/security-policy/violation-categories").then().body(containsString("insecure:dev"));
+        RestAssured.when().get("/security-policy/violation-property-keys").then().body(containsString("devConsoleEnabled"));
+        RestAssured.when().get("/security-policy/violation-policies").then().body(containsString("warn"));
+    }
```
apupier, line 47: I think it will be worthy to split with several different test methods, so that each of them will report a result in case of failure.

AGENTS.md
gnodet: _Claude Code on behalf of Guillaume Nodet_

Good point. Reverted from this PR and created a separate PR targeting main: #8588

integration-test-groups/foundation/security-policy/src/test/java/org/apache/camel/quarkus/security/policy/SecurityPolicyWarnTest.java
gnodet: _Claude Code on behalf of Guillaume Nodet_

Done — split into 4 separate test methods: `warnPolicyDetectsViolations`, `warnPolicyReportsInsecureDevCategory`, `warnPolicyReportsDevConsoleEnabledProperty`, `warnPolicyReportsWarnLevel`.

gnodet:
_Claude Code on behalf of Guillaume Nodet_

The 3 CI failures all share the same root cause:

```
ClassNotFoundException: org.jboss.marshalling.ByteInput
```

All 3 fail in `camel-quarkus-integration-test-milvus` during Quarkus class transformation (`ClassTransformingBuildStep`). The security-policy module (in group-03/foundation) is not affected.

Same [`issues`](https://github.com/apache/camel-quarkus/actions/runs/24995019250) on `camel-main` branch

@apupier @jamesnetherton that looks unrelated to this PR, do you confirm?

apupier:
> The 3 CI failures all share the same root cause:
> 
> ```
> ClassNotFoundException: org.jboss.marshalling.ByteInput
> ```
> 
> All 3 fail in `camel-quarkus-integration-test-milvus` during Quarkus class transformation (`ClassTransformingBuildStep`). The security-policy module (in group-03/foundation) is not affected.
> 
> Same [`issues`](https://github.com/apache/camel-quarkus/actions/runs/24995019250) on `camel-main` branch
> 
> @apupier @jamesnetherton that looks unrelated to this PR, do you confirm?

yes I confirmed https://github.com/apache/camel-quarkus/pull/8583#issuecomment-4333147965

integration-test-groups/foundation/security-policy/src/main/java/org/apache/camel/quarkus/security/policy/SecurityPolicyResource.java
```diff
@@ -0,0 +1,90 @@
...
+    CamelContext camelContext;
+
+    @GET
+    @Path("/has-violations")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String hasViolations() {
+        SecurityPolicyResult result = camelContext.getCamelContextExtension()
+                .getContextPlugin(SecurityPolicyResult.class);
+        if (result == null) {
+            return "null";
```
apupier, line 44: a bit weird to return null instead of empty string like the other methods but not a real issue

apupier (approved):

jamesnetherton:
I'll merge this. The Milvus fix from #8593 will get automatically rebased to `camel-main` by the nightly workflow if the build is green.

PR #8588, opened 2026-04-28
Document integration testing constraints in AGENTS.md

gnodet:
## Summary
- Add "Important Constraints" section under "Integration Tests" in AGENTS.md documenting:
  - `@Inject`/CDI does not work in `@QuarkusIntegrationTest` — use JAX-RS endpoints + RestAssured
  - Every `@QuarkusTest` class needs a corresponding `@QuarkusIntegrationTest` IT class
  - Grouped test modules share a single native application — avoid properties that alter global Camel behavior

## Test plan
- [ ] Verify AGENTS.md renders correctly on GitHub

_Claude Code on behalf of Guillaume Nodet_

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -196,6 +196,12 @@ class KafkaIT extends KafkaTest {
 }
 ```
 
+### Important Constraints
+
+- **No `@Inject` in `@QuarkusIntegrationTest`**: Integration tests run the application as a separate process, so CDI injection does not work. Expose application state via JAX-RS endpoints (using `quarkus-resteasy`) and assert via RestAssured in tests. See `integration-test-groups/foundation/core/` for the pattern.
+- **Every `@QuarkusTest` needs an IT class**: Each `@QuarkusTest` class must have a corresponding `@QuarkusIntegrationTest` class that extends it, to ensure tests also run in native mode.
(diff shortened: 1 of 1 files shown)


oscerd (approved):

apupier (approved):

gansheer (approved):

AGENTS.md
```diff
@@ -196,6 +196,12 @@ class KafkaIT extends KafkaTest {
 }
 ```
 
+### Important Constraints
+
+- **No `@Inject` in `@QuarkusIntegrationTest`**: Integration tests run the application as a separate process, so CDI injection does not work. Expose application state via JAX-RS endpoints (using `quarkus-resteasy`) and assert via RestAssured in tests. See `integration-test-groups/foundation/core/` for the pattern.
+- **Every `@QuarkusTest` needs an IT class**: Each `@QuarkusTest` class must have a corresponding `@QuarkusIntegrationTest` class that extends it, to ensure tests also run in native mode.
```
aldettinger, line 202: This might not always be true for JVM only tests ? Maybe we could quote 'when native mode testing is needed' ? Or 'do the same as other tests in the same folder' ?

Issue #8590, opened 2026-04-28
Intermittent failure of MicroProfileHealthSupervisedRouteTest.supervisedRouteTest on Windows


jamesnetherton:
### Bug description

```
org.awaitility.core.ConditionTimeoutException: 
Assertion condition defined as a Lambda expression in org.apache.camel.quarkus.component.microprofile.it.health.MicroProfileHealthSupervisedRouteTest 4 expectations failed.
JSON path status doesn't match.
Expected: is "DOWN"
  Actual: UP

JSON path checks.status.findAll().unique() doesn't match.
Expected: iterable containing ["UP", "DOWN"]
  Actual: <[UP]>

JSON path checks.find { it.name == 'camel-routes' }.status doesn't match.
Expected: is "DOWN"
  Actual: UP

JSON path checks.find { it.name == 'camel-consumers' }.status doesn't match.
Expected: is "DOWN"
  Actual: UP
 within 5 seconds.
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:1160)
	at org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:790)
	at org.apache.camel.quarkus.component.microprofile.it.health.MicroProfileHealthSupervisedRouteTest.supervisedRouteTest(MicroProfileHealthSupervisedRouteTest.java:38)
Caused by: java.util.concurrent.TimeoutException
	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:204)
	at org.awaitility.core.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:101)
	at org.awaitility.core.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:81)
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:103)
	... 5 more
```

PR #8591, opened 2026-04-28
Fix intermittent MicroProfileHealthSupervisedRouteTest failure on Windows

jamesnetherton:
The test was failing intermittently on Windows due to slow application startup causing a timing issue. The test expects to catch the health status in a "DOWN" state while routes are restarting, but on Windows the application can take 10+ seconds to start. By the time the test begins checking, the route has already exhausted all retry attempts (2.5 seconds with 10 × 250ms delay) and transitioned to "UP".

Changes:
- Increased backoff-delay from 250ms to 1000ms (total retry period now 10 seconds instead of 2.5 seconds)
- Increased await timeouts from 5 seconds to 15 seconds to provide more time to catch both DOWN and UP states on slower systems


aldettinger (approved):

apupier:
fixes #8590

PR #8595, opened 2026-04-28
Promote elasticsearch to native support

### LINKED ISSUE #4155: Elasticsearch: new extension
New extension based on new component introduced in Camel 3.19: https://issues.apache.org/jira/browse/CAMEL-17780

JinyuChen97:
- Move elasticsearch extension from extensions-jvm to extensions
- Move integration tests from integration-tests-jvm to integration-tests and add more tests
- imported quarkus-elasticsearch-java-client to support camel-quarkus-elasticsearch native mode
- All tests pass in both JVM and native modes (9/9)

Fixes #4155


jamesnetherton:
Please rebase. The project version is bumped to 3.36.0-SNAPSHOT so you'll need to clean up any references to 3.35.

integration-tests/elasticsearch/src/test/java/org/apache/camel/quarkus/component/elasticsearch/it/ElasticsearchTestResource.java
```diff
@@ -57,12 +57,15 @@ public Map<String, String> start() {
...
-                    // camel
+                    // Camel Elasticsearch component configuration
                     "camel.component.elasticsearch.host-addresses", hostAddresses,
                     "camel.component.elasticsearch.user", ELASTICSEARCH_USERNAME,
-                    "camel.component.elasticsearch.password", ELASTICSEARCH_PASSWORD);
+                    "camel.component.elasticsearch.password", ELASTICSEARCH_PASSWORD,
+                    // Quarkus Elasticsearch client configuration
+                    "quarkus.elasticsearch.hosts", hostAddresses,
+                    "quarkus.elasticsearch.username", ELASTICSEARCH_USERNAME,
+                    "quarkus.elasticsearch.password", ELASTICSEARCH_PASSWORD);
```
jamesnetherton: What's the reason to add these?

The Camel component configuration should be enough (ideally).

JinyuChen97: > What's the reason to add these?
> 
> The Camel component configuration should be enough (ideally).

The quarkus-elasticsearch-java-client dependency starts devServices if quarkus.elasticsearch.hosts is empty, causing tests to fail because the devServices container lacks the                        action.destructive_requires_name=false configuration needed for wildcard index deletion.                                                                                                               
The camel configuration was only used in the waitClusterReady() test method. I've updated it to read from quarkus configuration instead, quarkus creates the authenticated restclient bean, and camel autowires it automatically.

jamesnetherton: Can we not disable the dev service in `application.properties`:

```
quarkus.elasticsearch.devservices.enabled=false
```

JinyuChen97: > Can we not disable the dev service in `application.properties`:
> 
> ```
> quarkus.elasticsearch.devservices.enabled=false
> ```

Yes, quarkus.elasticsearch.devservices.enabled=false works in application.properties to disable DevServices, but quarkus.elasticsearch.hosts is still required from TestResource because TestContainers use random ports that are only known at runtime.                                                                                                                                                         
The restclient from quarkus needs quarkus.elasticsearch.hosts to connect, otherwise, the restclient defaults to localhost:9200 will caused connection refused. For both settings will be the same (return null) disable devservice or only hosts provides the address based on the quarkus source code? But provide host seems solved both(disable devservice and provide host), setting enabled=false in application.properties seems a bit redundant? Please correct me if I'm wrong.

`if (!config.enabled().orElse(true)) {
            // explicitly disabled
            log.debug("Not starting Dev Services for Elasticsearch, as it has been disabled in the config.");
            return null;
        }`
        
`for (String hostsConfigProperty : buildItemConfig.hostsConfigProperties) {
            // Check if elasticsearch hosts property is set
            if (ConfigUtils.isPropertyNonEmpty(hostsConfigProperty)) {
                log.debugf("Not starting Dev Services for Elasticsearch, the %s property is configured.", hostsConfigProperty);
                return null;
            }
        }`

jamesnetherton: Ok - so if I understand the problem correctly, the Quarkus Elasticsearch client bean is materialised even if we set no `quarkus.elasticsearch` config options (because of the host default). Which means it gets autowired into the Camel component.

If that's true, then it might be best to set:

```
camel.component.elasticsearch.autowired-enabled=false
```

We need to prove that the vanilla Camel component functionality & configuration can work properly.

JinyuChen97: the camel.component.elasticsearch.autowired-enabled=false is not working when put in application.properties and camel still autowiring, but it works and native tests passed when return by QuarkusTestResourceLifecycleManager.start().

jamesnetherton (approved):

PR #8617, opened 2026-05-05
extend elasticsearch test with SSL

### LINKED ISSUE #5174: Extend elasticsearch tests with SSL
### Describe the feature here

Currently we don't test SSL with elasticsearch component. After we will add support for Native, it could be good to test it.

JinyuChen97:
adding a new test module for elasticsearch ssl testing

Fixes #5174

### DIFF
--- a/integration-tests/elasticsearch/pom.xml
+++ b/integration-tests/elasticsearch/pom.xml
@@ -65,6 +65,11 @@
             <artifactId>awaitility</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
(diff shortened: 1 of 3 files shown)


jamesnetherton:
Might be simpler to take the same approach as the `elasticsearch-rest-client` tests and just make SSL default enabled for the existing tests. That way you don't need to start multiple containers and have dedicated tests etc.

Realistically, the majority of users aren't going to hooking their apps up to insecure ES instances (hopefully).

jamesnetherton (approved):

PR #8625, opened 2026-05-06
Add test coverage for Gmail DataType Transformers : update-message-labels and draft

### LINKED ISSUE #8531: Camel 4.19 add test coverage for Gmail  DataType Transformers : update-message-labels and draft
### Describe the feature here

Features
- https://issues.apache.org/jira/browse/CAMEL-23258
- https://issues.apache.org/jira/browse/CAMEL-23251

JinyuChen97:
Add test coverage for Gmail DataType Transformers : update-message-labels and draft

   - google-mail:update-message-labels transformer: Tests automatic label name-to-ID resolution when updating message labels via exchange variables

   - google-mail:draft transformer: Tests draft creation and update using GoogleMailStreamConstants headers (MAIL_FROM, MAIL_TO, MAIL_SUBJECT, etc.)

   - Threading metadata support: Tests draft replies with proper threadId and Message-ID headers for email threading

   All added tests pass with real google cloud environment in both JVM and native modes.

   Generated with Claude Code on behalf of Jinyu Chen

   fixes #8531

### DIFF
--- a/integration-tests/google/pom.xml
+++ b/integration-tests/google/pom.xml
@@ -48,6 +48,10 @@
             <artifactId>camel-quarkus-google-sheets</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-direct</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-resteasy</artifactId>
         </dependency>
@@ -113,6 +117,19 @@
             </activation>
             <dependencies>
                 <!-- The following dependencies guarantee that this module is built after them. You can update them by running `mvn process-resources -Pformat -N` from the source tree root directory -->
+                <dependency>
+                    <groupId>org.apache.camel.quarkus</groupId>
+                    <artifactId>camel-quarkus-direct-deployment</artifactId>
+                    <version>${project.version}</version>
+                    <type>pom</type>
+                    <scope>test</scope>
+                    <exclusions>
+                        <exclusion>
+                            <groupId>*</groupId>
+                            <artifactId>*</artifactId>
+                        </exclusion>
+                    </exclusions>
+                </dependency>
                 <dependency>
                     <groupId>org.apache.camel.quarkus</groupId>
                     <artifactId>camel-quarkus-google-calendar-deployment</artifactId>
--- a/integration-tests/google/src/main/java/org/apache/camel/quarkus/component/google/it/GoogleMailResource.java
+++ b/integration-tests/google/src/main/java/org/apache/camel/quarkus/component/google/it/GoogleMailResource.java
@@ -20,13 +20,17 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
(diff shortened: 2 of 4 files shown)


integration-tests/google/pom.xml
```diff
@@ -47,6 +47,10 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-google-sheets</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-direct</artifactId>
+        </dependency>
```
jamesnetherton, line 53: The -deployment dependency needs adding into the virtualDependencies profile. You can run this from the project root to add it:

```
mvn process-resources -Pformat -N
```

aldettinger: Looks done.

aldettinger:
The pr implementation looks good, maybe a few questions about the intent.

Do we need that level of test verbosity ? For instance, what does those camel-quarkus tests bring compared to camel tests ?
Labels and drafts are documented functionalities, so we generally double camel tests in this situation.
However, is there an interest in doubling tests for each and every data type transformer ?
Maybe we need a kind of foundation test covering a single data type transformer and that's all ?
Any thoughts ?

jamesnetherton:
> Do we need that level of test verbosity ? For instance, what does those camel-quarkus tests bring compared to camel tests ?

It's a good point. I'd say not, unless there's something about a feature that might cause it to behave differently in CQ. E.g different dependency versions, use of reflection or other stuff problematic for native mode.

> Maybe we need a kind of foundation test covering a single data type transformer

We already have that.

Given the work is already done, I am ok merging this. It was actually a useful exercise because it's likely nobody has run any of the Google tests for 6+ years 🙈 (they are disabled without the relevant credentials being set).

integration-tests/google/src/main/java/org/apache/camel/quarkus/component/google/it/GoogleMailResource.java
```diff
@@ -99,6 +104,63 @@ public Response readMail(@QueryParam("messageId") String messageId) {
...
+    public Response getMessageIdHeader(@QueryParam("messageId") String messageId) {
+        Map<String, Object> headers = new HashMap<>();
+        headers.put("CamelGoogleMail.userId", USER_ID);
+        headers.put("CamelGoogleMail.id", messageId);
+
+        try {
+            Message response = producerTemplate.requestBodyAndHeaders("google-mail://messages/get", null, headers,
+                    Message.class);
+            if (response != null && response.getPayload() != null && response.getPayload().getHeaders() != null) {
+                for (com.google.api.services.gmail.model.MessagePartHeader header : response.getPayload().getHeaders()) {
```
jamesnetherton: Please use imports in all cases and avoid FQCN references where possible.

aldettinger:
> Given the work is already done, I am ok merging this. It was actually a useful exercise because it's likely nobody has run any of the Google tests for 6+ years 🙈 (they are disabled without the relevant credentials being set).

Ok, that's fine to merge then. And it's always possible to clean up in another ticket anyway. Like we could keep the gmail tests and just remove the data type transformer part if make sense.

aldettinger (approved):

jamesnetherton (approved):

PR #8627, opened 2026-05-06
Add camel-quarkus-cli-connector as a conditional dev mode dependency

### LINKED ISSUE #8623: Add camel-quarkus-cli-connector as a conditional dev mode dependency
### Describe the feature here

Primarily for Camel JBang so that the various commands that depend on the cli-connector can work without the user doing anything or having to add additional logic into Camel JBang itself.

jamesnetherton:
Fixes #8623

Note - The `camel-quarkus-console` `<artifact>` removal was due to failures in `CamelQuarkusCoreDevUITest`. It's probably due a bug in Quarkus, I'd need to dig into it more another time. But `console` is a transitive of `cli-connector` so it's a safe change. I verified the dev-ui pages still work fine.

### DIFF
--- a/extensions-core/core/runtime/pom.xml
+++ b/extensions-core/core/runtime/pom.xml
@@ -102,7 +102,7 @@
                         <provides>org.apache.camel</provides>
                     </capabilities>
                     <conditionalDevDependencies>
-                        <artifact>org.apache.camel.quarkus:camel-quarkus-console:${project.version}</artifact>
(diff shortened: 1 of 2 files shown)


jamesnetherton:
Seems the `cli-connector` does not play well with dev mode continuous testing. I will investigate.

jamesnetherton:
Dev mode continuous testing issues should be fixed in the latest commit.

aldettinger (approved):

PR #8647, opened 2026-05-12
Add test coverage for virtual thread support

jamesnetherton:


integration-test-groups/foundation/seda/src/test/java/org/apache/camel/quarkus/component/seda/it/SedaVirtualThreadsIT.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package org.apache.camel.quarkus.component.seda.it;
+
+import io.quarkus.test.junit.QuarkusIntegrationTest;
+import org.junit.jupiter.api.Test;
+
+@QuarkusIntegrationTest
+class SedaVirtualThreadsIT extends SedaVirtualThreadsTest {
+    /**
+     * Overridden to remove JDK version restrictions as the JDK used in native mode should be >= 25
```
aldettinger, line 25: So, super is restricted to range JDK21+ ? The override remove the restriction ? Then how do we ensure that the native test is restricted to JDK25+ ?

jamesnetherton: > How do we ensure that the native test is restricted to JDK25+ 

It's implicit because the native Mandrel builder image defaults to JDK 25. If someone is building & running native tests with Mandrel / GraalVM < JDK 21, then it's invalid because we do not support that.

aldettinger: OK, thanks for clarification. It make sense.

aldettinger (approved):

PR #8650, opened 2026-05-13
Upgrade Quarkus to 3.36.0.CR1

jamesnetherton:


jamesnetherton:
I'll set this to draft, while I figure out how to work around issues caused by `narayana-jta` upgrade.

jamesnetherton:
There's a fix for the narayana-jta issue in Quarkus. I'm assuming it'll be backported to 3.36.0 so it's probably simpler to wait for that instead of doing workarounds.

PR #8684, opened 2026-05-21
Fixes #8621 Add KafkaOffsetBackingStore native support and integration tests 

### LINKED ISSUE #8621: Test Debezium with KafkaOffsetBackingStore offset storage
Currently all Debezium tests use the default `FileOffsetBackingStore`. We should test in at least one place with `offsetStorage=KafkaOffsetBackingStore`.

Relates to this [example route](https://github.com/apache/camel-quarkus/issues/8239#issuecomment-4386021000).

jomin7:
Fixes #8621 Add KafkaOffsetBackingStore native support and integration tests 

  Body:

  Summary

  Add native compilation support for KafkaOffsetBackingStore and integration tests for MongoDB and PostgreSQL Debezium connectors.

  In production/containerized deployments, storing Debezium offsets in a local file is impractical. KafkaOffsetBackingStore provides a distributed,
  fault-tolerant alternative by persisting offsets to a Kafka topic. This was not working in native mode due to missing GraalVM reflection registrations.

  Changes

  Native support (DebeziumSupportProcessor):
  - Register KafkaOffsetBackingStore for reflection
  - Register Kafka Connect converter classes (ByteArrayConverter, BooleanConverter, etc.)
  - Register Kafka serializer/deserializer classes needed at runtime
  - Register Kafka client infrastructure classes (JmxReporter, RangeAssignor, SASL classes) instantiated reflectively by KafkaBasedLog
  - Register JSON converter classes with constructors/methods access (used by KafkaOffsetBackingStore for offset serialization)

  Shared test support (AbstractDebeziumResource):
  - Add kafkaBootstrapServers(), receiveViaKafkaOffset(), and getKafkaOffsetEndpointUrl() endpoints to the base class for reuse across connectors

  Integration tests (MongoDB & PostgreSQL):
  - Add testKafkaOffsetBackingStore() test for each connector
  - Start a Strimzi Kafka cluster alongside the database container in test resources
  - Override getKafkaOffsetEndpointUrl() with connector-specific Debezium endpoint configuration

  Grouped module (debezium-grouped):
  - Add strimzi-test-container test dependency (sources are copied from individual modules at build time)

### DIFF
--- a/extensions-support/debezium/deployment/src/main/java/org/apache/camel/quarkus/support/debezium/deployment/DebeziumSupportProcessor.java
+++ b/extensions-support/debezium/deployment/src/main/java/org/apache/camel/quarkus/support/debezium/deployment/DebeziumSupportProcessor.java
@@ -51,10 +51,8 @@ import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
 import io.quarkus.deployment.builditem.IndexDependencyBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.NativeImageResourceBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
-import io.quarkus.deployment.builditem.nativeimage.RuntimeInitializedClassBuildItem;
 import io.quarkus.gizmo.Gizmo;
 import org.apache.camel.quarkus.support.debezium.DebeziumComponentObserver;
-import org.apache.kafka.common.security.authenticator.SaslClientAuthenticator;
 import org.apache.kafka.connect.json.JsonConverter;
 import org.apache.kafka.connect.source.SourceTask;
 import org.jboss.jandex.DotName;
@@ -75,11 +73,6 @@ public class DebeziumSupportProcessor {
     }
 
     @BuildStep
-    RuntimeInitializedClassBuildItem runtimeInitializedClasses() {
-        return new RuntimeInitializedClassBuildItem(SaslClientAuthenticator.class.getName());
-    }
-
-    @BuildStep
     public void configureKafkaComponentForDevServices(BuildProducer<AdditionalBeanBuildItem> additionalBean) {
         additionalBean.produce(AdditionalBeanBuildItem.unremovableOf(DebeziumComponentObserver.class));
     }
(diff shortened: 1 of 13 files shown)


apupier (commented):
Have you considered depending on https://quarkus.io/extensions/io.quarkus/quarkus-kafka-client/ ? it sounds like a lot of native configuration is similar to what they are doing.

jomin7:
@apupier Thanks for the suggestion. I looked into it.quarkus-kafka-client covers the Kafka client layer (producers, consumers), but the classes needed here are from
   the Kafka Connect layer (offset backing stores, Connect converters, etc.), which aren't covered by it. Manual reflection registration would still be required.

  There also seems to be a test execution error related to containerization. Will revisit the implementation and update.

jamesnetherton:
> the Kafka Connect layer (offset backing stores, Connect converters, etc.), which aren't covered by it

Right, but the common Kafka code for registering serializers etc is already handled by the `quarkus-kafka-client` extension. I think it'd cover everything from lines 167 - 185.

jomin7:
Hi @apupier, @jamesnetherton — thanks for the review feedback! I've updated the PR to add quarkus-kafka-client as a dependency on the debezium support extension. This lets us drop the SaslClientAuthenticator runtime initialization and reflection registration since quarkus-kafka-client already handles that. The Connect-layer classes (KafkaOffsetBackingStore, JSON/type converters) are still registered explicitly as they fall outside what the Kafka 
client extension covers. Please let me know if there's anything else that needs to be addressed.

extensions-support/debezium/deployment/src/main/java/org/apache/camel/quarkus/support/debezium/deployment/DebeziumSupportProcessor.java
```diff
@@ -144,6 +142,18 @@ void reflectiveClasses(CombinedIndexBuildItem combinedIndex, BuildProducer<Refle
...
+                "org.apache.kafka.connect.converters.ByteArrayConverter",
+                "org.apache.kafka.connect.converters.BooleanConverter",
+                "org.apache.kafka.connect.converters.DoubleConverter",
+                "org.apache.kafka.connect.converters.FloatConverter",
+                "org.apache.kafka.connect.converters.IntegerConverter",
+                "org.apache.kafka.connect.converters.LongConverter",
+                "org.apache.kafka.connect.converters.ShortConverter")
+                .constructors()
+                .methods()
+                .build());
```
jamesnetherton: Now that I think about it - probably a better way for this is to search the Jandex index for all classes that implement `org.apache.kafka.connect.storage.Converter`. That way we don't need to maintain a manual list of types.

jomin7: Thanks @jamesnetherton, I looked into using the Jandex index for this. The Converter implementations (ByteArrayConverter, BooleanConverter, etc.) live in connect-runtime. Indexing that jar via IndexDependencyBuildItem causes both CDI and RESTEasy to discover its internal JAX-RS resource classes (like InternalConnectResource), which blows up with unsatisfied constructor dependencies. ExcludedTypeBuildItem only covers CDI, and since RESTEasy independently scans the Jandex index for @Path classes, there's no way to selectively hide them from it.

As an alternative, I'm replacing the manual list with a build-time scan of the META-INF/services/org.apache.kafka.connect.storage.Converter service loader files from the classpath. This gets us the exact same result—no hardcoded list, auto-discovery of all implementations, and forward compatibility with future Kafka versions—but without needing to index connect-runtime at all.

Does that approach sound good to you?

jamesnetherton: Ok - sounds fine. There is also [ServiceProviderBuildItem](https://github.com/quarkusio/quarkus/blob/main/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ServiceProviderBuildItem.java) for this purpose.

jomin7: Sure.Thanks for the heads-up on ServiceProviderBuildItem

jomin7: Thanks @jamesnetherton, updated to use ServiceProviderBuildItem.allProvidersFromClassPath("org.apache.kafka.connect.storage.Converter") to auto-discover the Converter implementations from META-INF/services on the classpath. This replaces the hardcoded list and keeps it forward-compatible with future Kafka versions.

jamesnetherton (approved):

PR #8692, opened 2026-05-22
Replace external Maven Central URL with local test endpoint in HTTP tests

### LINKED ISSUE #8686: Do not hit repo.maven.apache.org from tests
### Description

We hit repo.maven.apache.org from https://github.com/apache/camel-quarkus/blob/main/integration-tests-support/http/src/main/java/org/apache/camel/quarkus/component/http/common/AbstractHttpResource.java#L34
We should choose some other, preferrably local URL, because Maven Central tends to send 429 on excessive usage these days.

JinyuChen97:
added maven-metadata/component local sever endpoint to mock the maven metadata response
added sslContextParameters to handle https connection to localhost for different http cllient resources tests

all jvm tests passed. (Some auth related native tests failed but seems not related to this change)

fixes #8686

### DIFF
--- a/integration-test-groups/http/http/src/main/java/org/apache/camel/quarkus/component/http/http/HttpResource.java
+++ b/integration-test-groups/http/http/src/main/java/org/apache/camel/quarkus/component/http/http/HttpResource.java
@@ -115,7 +115,9 @@ public class HttpResource extends AbstractHttpResource {
                         + "&proxyAuthHost=%s"
                         + "&proxyAuthPort=%d"
                         + "&proxyAuthUsername=%s"
-                        + "&proxyAuthPassword=%s", String.format(PROXIED_URL, "http"), proxyHost, proxyPort, USER_ADMIN,
+                        + "&proxyAuthPassword=%s"
+                        + "&sslContextParameters=#sslContextParameters", String.format(PROXIED_URL, "http"), proxyHost,
+                        proxyPort, USER_ADMIN,
                         USER_ADMIN_PASSWORD)
                 .request(String.class);
     }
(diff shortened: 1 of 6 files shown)


jamesnetherton:
Doesn't this change the original intent of the tests? The reason why the HTTP proxy tests were reaching out to an external resource, is because `localhost` was configured as 'noproxy' for test purposes.

If we're worried about rate limits on Maven Central, maybe we just switch the URL to use the Google Maven mirror?

https://maven-central.storage-download.googleapis.com/maven2

JinyuChen97:
> Doesn't this change the original intent of the tests? The reason why the HTTP proxy tests were reaching out to an external resource, is because `localhost` was configured as 'noproxy' for test purposes.
> 
> If we're worried about rate limits on Maven Central, maybe we just switch the URL to use the Google Maven mirror?
> 
> https://maven-central.storage-download.googleapis.com/maven2

Thanks for the quick feedback! From my understand and the code base in camel CamelProxyRoutePlanner class, the test flow for nonProxy test seems determined by the PROXIED_URL = "https://localhost:xxxx......" and the nonProxyHosts parameter in test request, if the target host is matched with nonProxyHosts parameter by exact match or wildcard matched, the request will not go through proxy, if they are not matched like the nonProxyHosts is 'example.com' but targethost is 'localhost' then it will need to go through the proxy server. The test here looks like is for testing different matching logics, whether external endpoint or internal endpoint seems is not the test purpose here? Please correct me if I'm wrong.

jamesnetherton:
On reflection and some local testing I think this is ok.

jamesnetherton (approved):

PR #8731, opened 2026-06-04
Improve AGENTS.md for catching corner cases of formatting

llowinge:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,6 +26,7 @@ Apache Camel Quarkus provides Quarkus extensions for Apache Camel components, en
 - Do NOT directly modify generated files under `docs/modules` or `src/main/generated`
 - Do NOT use dynamic class loading or reflection (impacts native compilation)
 - Do NOT introduce Spring Boot or Spring unless explicitly required
+- Do NOT commit changes without running `./mvnw process-resources -Pformat` first
 
 ## Technology Stack
 - Java 17+


jamesnetherton (commented):
Thanks for working on this! The intent is great — agents definitely need clearer guidance on formatting and license headers. However, after reviewing the actual Maven configuration, I think this PR adds unnecessary complexity that could cause more problems than it solves.

### Key Issues

**1. The existing command already does everything**

`./mvnw process-resources -Pformat` already runs `license-maven-plugin` format goal (via the `format` profile), along with POM sorting and metadata updates. Suggesting `license:format` as a separate step is redundant and creates confusion about which command to use.

**2. Over-zealous instructions will cause performance problems**

"Do NOT create ANY new file without running `license:format` afterwards" would cause an agent to invoke Maven after every single file creation. That's extremely slow and unnecessary. Formatting should happen once when the task is complete, not after each file.

**3. Incorrect file type claims**

The PR lists `*.adoc` and `*.yaml`/`*.yml` as needing license headers, but the `license-maven-plugin` configuration in `pom.xml` explicitly **excludes** both:
- `**/*.adoc` (line 567)
- `modules/**/*.yml` (line 570)

**4. Redundant validation step**

Running `./mvnw validate -Pcheck-format` immediately after `./mvnw process-resources -Pformat` adds no value — if format succeeded, check will pass; if format failed, Maven already errored. CI runs this check as a safety net anyway.

### Suggested Alternative

If agents are skipping formatting, just emphasize the existing workflow more clearly.

In the "What NOT to do" section, add:
```
- Do NOT commit changes without running `./mvnw process-resources -Pformat` first
```

And keep Section 6 simple:
```
### 6. Format & Style
Before committing, format code:

    ./mvnw process-resources -Pformat
```

This achieves the goal (making formatting unmissable) without adding redundant commands, incorrect file-type guidance, or instructions that would trigger excessive Maven invocations.

oscerd (approved):

llowinge:
@jamesnetherton Thanks, don't we miss the adding of licenses (if i would keep only the suggested changes)?

jamesnetherton:
> @jamesnetherton Thanks, don't we miss the adding of licenses (if i would keep only the suggested changes)?

The format profile takes care of it, no?

https://github.com/apache/camel-quarkus/blob/main/pom.xml#L944-L960

llowinge:
@jamesnetherton You are right. Then i'm not sure, why Claude didn't follow it. But i'll remove it, thanks.

jamesnetherton (approved):

PR #8732, opened 2026-06-04
Cache Atlassian Maven artifacts

jamesnetherton:
Try to work around sporadic Atlassian maven repo 403s.

Obvs it needs to populate the cache initially, which might fail with the 403. But once it's done, workflows should run ok.


jamesnetherton:
Pushed updates for other workflows.

apupier (approved):
let's try. it doesn't solve the root issue of relying on a non-reliable repository, and it will put us blind on potential complete removal of the repository but it will help to gain on stability on our CI

PR #8736, opened 2026-06-05
Add .claude to .gitignore

llowinge:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ perf-regression-report@*.txt
 
 # Claude
 .claude
+
+# Bob
+.bob
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ quarkus.log*
 
 # Camel Quarkus Performance Regression Reports
 perf-regression-report@*.txt
+
+# Claude
+.claude


aldettinger (approved):
Looks valid. At least, it could help with security for sure.

Having a second advice from someone knowing claude would be welcome.

jamesnetherton:
> Having a second advice from someone knowing claude would be welcome.

Unless we're planning to add project specific commands etc, then I think it's fine. Most project specific config lives [elsewhere](https://github.com/Open-Harness-Engineering/ai-agents-oss-known-projects) anyway.

Seems we now have ignore rules for Bob also.

jamesnetherton (approved):

aldettinger:
> > Having a second advice from someone knowing claude would be welcome.
> 
> Unless we're planning to add project specific commands etc, then I think it's fine. Most project specific config lives [elsewhere](https://github.com/Open-Harness-Engineering/ai-agents-oss-known-projects) anyway.
> 
> Seems we now have ignore rules for Bob also.

Yes, favoring other locations looks fine. Anyway, we could always refine later if needed.

PR #8742, opened 2026-06-08
ci: Cache Atlassian Maven artifacts in additional workflows & fix camel-main handling

jamesnetherton:
* Fix workflows where caching of Atlassian artifacts was missing
* Handle camel-main branch special case where the Atlassian artifact version needs to be gathered from the camel parent/pom.xml
* Remove old inline cache and maven repository restoration steps and use setup-maven-cache consistently across all workflows
* Fixes up version comments for `actions/cache`. Dependabot will increment them automatically on the next update.


jamesnetherton:
I'll merge so we can test the nightly camel-main job.

PR #8746, opened 2026-06-09
ci: Use zstd compression when taring Maven repo

jamesnetherton:
I noticed that the GitHub cache action uses zstd when it tars up the cache content. So I figured we could do something similar for the m2 build artifact.

Speeds up .m2 archive compression / decompression and reduces its size.

### DIFF
--- a/.github/actions/setup-maven-cache/action.yaml
+++ b/.github/actions/setup-maven-cache/action.yaml
@@ -41,6 +41,6 @@ runs:
       shell: bash
       run: |
         df -h /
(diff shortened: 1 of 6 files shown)


aldettinger (approved):

apupier (approved):

jamesnetherton:
I need to tune which artifacts are excluded from the repo zip.

jamesnetherton:
I removed the tar file exclusion logic. On thinking about it more, we potentially expose ourselves to future maintenance headaches for not much gain. It could be revisited in future if needed.

PR #8750, opened 2026-06-10
[jt400-stabilize] Change order of sending message

tveskrna:
Test testInquiryMessageQueue is unstable in CEQ QE pipeline. This change stabilizes it. Tested in our pipeline.
But I would like someone to check integrity of the test.

I moved sending message before route start and add Sleep to give camel some time.


tveskrna:
Stabilize it for 3.33 so opened for this branch as well

jamesnetherton (approved):

PR #8765, opened 2026-06-17
Genericise setup-maven-cache action and cache shibboleth repository artifacts

jamesnetherton:


aldettinger:
Looks good as far as I get.

To be sure, the dependency cache is saved at the end of each initial build even when third party dependencies versions have not changed ? Maybe there is an eviction that could happen at frequent interval or something alike ?

jamesnetherton:
> To be sure, the dependency cache is saved at the end of each initial build even when third party dependencies versions have not changed ?

Yes, correct. The save step runs every build, but actions/cache/save is a no-op if a cache entry with that exact key already exists. So I think a warning is emitted and the build moves on without any issues.

We can probably make save conditional on cache miss, but I'm not sure it's worth doing.

jamesnetherton:
With regards to eviction, GitHub evicts cache entries not accessed within 7 days. Which I think is fine for us given we have a consistent number of builds on most of our active branches.

aldettinger (approved):
Thanks for double checking the understanding correctness 👍 
It's already fine like that. Changing would probably bring unneeded complexity for no gain.

PR #8776, opened 2026-06-19
chore: use WireMock for docling integration tests to avoid 20GB container pull in CI

jamesnetherton:
Replace the real docling-serve container in DoclingTestResource with WireMock record/playback. Default mode (CI) replays pre-recorded stubs with no container. Setting camel.quarkus.start-mock-backend=false uses the real container directly. Running with WIREMOCK_RECORD=true starts the container and captures stubs for future replays.

Also use deterministic temp file names in DoclingResource so that recorded stubs match on replay.


aldettinger (approved):
Looks fair though the mapping files are quite huge. Maybe shrinking the pdf file would come in handy on the long term, let's see.

PR #8779, opened 2026-06-22
chore: document that Camel components ship with Jandex index

jamesnetherton:
Add a note to the native build troubleshooting section clarifying that Camel component JARs already include a Jandex index, so IndexDependencyBuildItem should not be added for them.


aldettinger (approved):
Looks good. Generally speaking, some people are saying that positive assertions are better than restrictions to guide agents... Not sure how to use that straight but maybe something to keep in mind :)

PR #8790, opened 2026-06-24
Add incremental build support to CI workflow

jamesnetherton:
Adds incremental build support to the CI workflow using Scalpel. Currently only for PRs. 

I have tested it across multiple scenarios on my fork and I think it works ok.

How it works:
  
- Scalpel runs during initial-mvn-install and produces a report of affected modules
- `IncrementalBuildMojo` generates the native test matrix, JVM test list, and functional test scope flags from the report
- Container image property changes in pom.xml are detected separately (Scalpel doesn't track these) by scanning *TestResource.java files for references, including transitive resolution through integration-tests-support/ modules
- Full builds automatically trigger for changes to: poms/build-parent/**, extensions-core/**, tooling/maven-plugin/**, .mvn/**
- Push events always run full builds; incremental is PR-only
- The full build can be forced by adding the `ci/disable-incremental` label to a PR.

Future optimisations:

- Move more workflow shell logic into the Maven Mojo
- Explore incremental builds for push events
- Relax full build triggers for extensions-core modules. Likely only needs to be applied to extensions-core/core
- Start thinking about making parts of our workflows reusable and have the core parts extracted into composite local actions

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -106,6 +106,12 @@ env:
   CQ_MAVEN_ARGS: -V -ntp -e -Daether.connector.http.connectionMaxTtl=120
   TESTCONTAINERS_RYUK_DISABLED: true
   CHECKOUT_REF: ${{ github.event_name == 'pull_request' && github.event.pull_request.user.login == 'dependabot[bot]' && github.head_ref || '' }}
+  # Scalpel configuration (defaults for all jobs; initial-mvn-install overrides SCALPEL_FULL_BUILD_TRIGGERS).
+  # Broad extensions-core/** trigger here is intentional: functional-extension-tests uses trim mode,
+  # which actively excludes modules from the build, so any extensions-core change should run full tests.
+  SCALPEL_FULL_BUILD_TRIGGERS: "-Dscalpel.fullBuildTriggers=poms/build-parent/**,poms/build-parent-it/**,extensions-core/**,tooling/maven-plugin/**,.mvn/**"
+  SCALPEL_EXCLUDE_PATHS: "-Dscalpel.excludePaths=**.adoc,**.md,docs/**,Jenkinsfile*,LICENSE.txt,NOTICE.txt,KEYS,camel-quarkus-sbom/**,.github/*.sh"
+  CQ_MAVEN_SKIP_CHECKS: "-Dformatter.skip -Dimpsort.skip -Denforcer.skip -Dcamel-quarkus.update-extension-doc-page.skip"
 
 jobs:
   pre-build-checks:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-build.yaml
```diff
@@ -299,36 +460,54 @@ jobs:
...
 
-            if [[ "x$(./mvnw org.apache.maven.plugins:maven-help-plugin:3.2.0:evaluate -Dexpression=ci.native.tests.skip -DforceStdout -q -f ${MODULE})" == "xtrue" ]]; then
-              JVM_MODULES+=("${MODULE}")
-            else
-              NATIVE_MODULES+=("${MODULE}")
-            fi
-          done
+              MODULE="integration-tests/$(echo ${MODULE} | sed 's/^[ \t]*//;s/[ \t]*$//')"
+
+              if [[ "x$(./mvnw org.apache.maven.plugins:maven-help-plugin:3.2.0:evaluate -Dexpression=ci.native.tests.skip -DforceStdout -q -f ${MODULE})" == "xtrue" ]]; then
```
apupier, line 487: note for another and later PR to not mix with this one and avoid conflict, there is version 3.5.1for maven-help-plugin

apupier (commented):
adding quite some complexity but the potential gain is worthy.

Could you point to the PR checks done with this version of the PR on your fork to review if we do not miss some use cases?
(reading the comment seems not but reviewing the code is not easy to check for it, at least for me)

jamesnetherton:
> Could you point to the PR checks done with this version of the PR on your fork to review if we do not miss some use cases?

Some open PRs here:

https://github.com/jamesnetherton/camel-quarkus/pulls

And a bunch of older closed ones here:

https://github.com/jamesnetherton/camel-quarkus/pulls?q=is%3Apr+is%3Aclosed

The workflow code was broadly the same across those tests.

aldettinger:
Hard to review indeed, still really worth it. It's better optimization than removing duplicate tests from camel and speeding up the full build by any optimization. The incremental build could actually compound with those if ever needed.

Keeping incremental build for prs only for quite a moment seems a good idea.

Here are the current understanding of a few scenarios for double checking:
 + Update quarkus version (probably near full build as scalpel would include a lot)
 + An `extensions-support` modified (scalpel should include only what's needed base on dependencies)
 + Changing .github files (An empty build will be triggered)
 + Test groups are rebalanced (An empty build will be triggered)
 + Changing an application.properties file (scalpel should include needed dependencies)

What do you think ?

jamesnetherton:
> Hard to review indeed

Yes, sorry for that. Our build is pretty complex already and these kind of changes inevitably end up increasing that complexity somewhat.

> Here are the current understanding of a few scenarios for double checking:
> 
> * Update quarkus version (probably near full build as scalpel would include a lot)
> * An `extensions-support` modified (scalpel should include only what's needed base on dependencies)
> * Changing .github files (An empty build will be triggered)
> * Test groups are rebalanced (An empty build will be triggered)
> * Changing an application.properties file (scalpel should include needed dependencies)

1. Update Quarkus version — Correct. Scalpel tracks BOM property changes, so a quarkus.version bump cascades through managed dependencies and flags most modules as affected. Effectively a near-full build. That's actually an enhancement I [added](https://github.com/maveniverse/scalpel/pull/36) in the latest release. Previously it had to worked around by adding the root pom.xml as a 'full build trigger', which is a bit heavyweight & bad for Dependabot PRs etc.

2. extensions-support modified — Correct. Scalpel tracks dependencies, so only test modules that depend on the changed support module would be affected.

3. Changing `.github` files — Most are listed in the workflow `paths-ignore` section and won't trigger the workflow at all. `ci-build.yaml` itself is not in paths-ignore (it can't be — it's the workflow file). Changes to it would trigger the
workflow to run. Scalpel would see it as a changed file with no downstream dependents — 0 affected modules, which falls back to a full build.

4. Test groups rebalanced — changes to `test-categories.yaml` would trigger the workflow and Scalpel would see it with no dependents — full build fallback.

5. Changing `application.properties` — Correct. Scalpel maps it to the owning module and includes that module plus dependents.

I probably should given a more concise summary of what's actually happening when in incremental mode...

The workflow reads a report generated by Scalpel that contains the modules affected by the changes on the PR. It then uses that info to figure out which modules need to be run in each of the separate jobs.

Some noteworthy details:

* For the native tests, it creates its own set of categories instead of referencing `test-categories.yml`
* The camel-quarkus-examples incremental build is a bit tricky because its a separate project. So we have some logic to see if any of the changed modules are part of the example project dependency tree, and if they are, the project is built and tested.

apupier (approved):
reviewed the opened PRs on the fork, seems fine.

Thanks a lot for this tremendous improvement

aldettinger (approved):
Yes, it's a bit of complexity that would highly likely produce awesome results, so worth investing 👍 

Many thanks for explanations, especially learned a lot with point 3 and fall back mechanisms. We could probably save a lot on tons of prs while still keeping full builds for push events.

jamesnetherton:
Lets merge and see how things work out. If it ends up being a disaster we can always revert back to the previous workflow state.

PR #8792, opened 2026-06-25
Upgrade hyperfoil-maven-plugin from 0.27.1 to 0.27.2

aldettinger:
Update the hyperfoil-maven-plugin version in the perf-regression tool from 0.27.1 to 0.27.2 to get latest bug fixes and improvements while maintaining Java 17 compatibility.

Version 0.28.0+ requires Java 21, so we stay on the 0.27.x series until Camel Quarkus upgrades its minimum Java requirement.

### DIFF
--- a/tooling/perf-regression/cq-perf-regression-sample-base/pom.xml
+++ b/tooling/perf-regression/cq-perf-regression-sample-base/pom.xml
@@ -30,8 +30,8 @@
     </parent>
     <artifactId>cq-perf-regression-sample</artifactId>
     <properties>
-        <!-- Let's not upgrade hyperfoil-maven-plugin as long as camel-quarkus supports JDK17: https://github.com/apache/camel-quarkus/issues/8027 -->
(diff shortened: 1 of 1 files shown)


jamesnetherton (approved):

aldettinger:
@jamesnetherton Incremental build wise, we probably fall into the zero modules selected, so falling back to full build ?

jamesnetherton:
> @jamesnetherton Incremental build wise, we probably fall into the zero modules selected, so falling back to full build ?

Yes, exactly. I think it's because the hyperfoil plugin declaration is in a sample app that is not connected to the main project build. So Scalpel knows nothing about it and all we can do is fall back to the full build.

aldettinger:
> > @jamesnetherton Incremental build wise, we probably fall into the zero modules selected, so falling back to full build ?
> 
> Yes, exactly. I think it's because the hyperfoil plugin declaration is in a sample app that is not connected to the main project build. So Scalpel knows nothing about it and all we can do is fall back to the full build.

Sure, still the fall back shows its utility with those corner cases :+1:

PR #8814, opened 2026-07-01
Remove unnecessary extension description overrides

### LINKED ISSUE #7343: Review extension description overrides
### Description

This is likely to mostly be a relic of the past, but we override extension descriptions in a lot of places.

Where possible, we should always try to use the description from the Camel component metadata in the catalog. Or send PRs to the Camel core to update descriptions if needed.

JiriOndrusek:
Fixes https://github.com/apache/camel-quarkus/issues/7343

 - Remove redundant `<description>` overrides from 129 extension `pom.xml` files so descriptions are inherited directly from the Camel catalog
  - Add "using AWS SDK version 2.x" suffix to 17 AWS extension descriptions (12 new + 5 updated to match catalog wording)
  - For 9 multi-component extensions (crypto, flatpack, mvel, rss, smooks, stax, thrift, wasm, pqc), set the description to the catalog component entry only, preventing redundant  concatenation of component + dataformat/language descriptions
  - Fix several stale or custom descriptions to match the current Camel 4.20.0 catalog (e.g. azure-storage-datalake, once, openai, vertx-http, vertx-websocket)

### DIFF
--- a/docs/modules/ROOT/pages/contributor-guide/extension-documentation.adoc
+++ b/docs/modules/ROOT/pages/contributor-guide/extension-documentation.adoc
@@ -30,6 +30,20 @@ The `update-extension-doc-page` mojo merges the https://raw.githubusercontent.co
 * `configuration`: Camel Quarkus specific configuration options (optional); the content of `src/main/doc/configuration.adoc` in the runtime module of the given extension.
 * `limitations`: Camel Quarkus specific limitations (optional); the content of `src/main/doc/limitations.adoc` in the runtime module of the given extension.
 
+== Extension description sync
+
+The `cq:update-quarkus-metadata` goal automatically syncs each extension's runtime `pom.xml` `<description>` with the Camel catalog.
+Extensions that need a custom description (e.g. multi-model extensions or extensions where the catalog description is not suitable) can opt out by setting:
+
+[source,xml]
+----
+<properties>
+    <cq.descriptionSync>false</cq.descriptionSync>
+</properties>
+----
+
+The property defaults to `true`.
+
 == Overriding generated component links
 
 Sometimes the generated component link xref within each extension doc page needs to be overridden.
--- a/extensions-jvm/ignite/runtime/pom.xml
+++ b/extensions-jvm/ignite/runtime/pom.xml
@@ -32,6 +32,7 @@
 
     <properties>
         <camel.quarkus.jvmSince>1.1.0</camel.quarkus.jvmSince>
+        <cq.descriptionSync>false</cq.descriptionSync>
     </properties>
 
     <dependencies>
--- a/extensions-jvm/mvel/runtime/pom.xml
+++ b/extensions-jvm/mvel/runtime/pom.xml
@@ -32,6 +32,7 @@
 
     <properties>
         <camel.quarkus.jvmSince>1.1.0</camel.quarkus.jvmSince>
+        <cq.descriptionSync>false</cq.descriptionSync>
     </properties>
 
     <dependencies>
--- a/extensions-jvm/smooks/runtime/pom.xml
+++ b/extensions-jvm/smooks/runtime/pom.xml
@@ -32,6 +32,7 @@
 
     <properties>
         <camel.quarkus.jvmSince>3.18.0</camel.quarkus.jvmSince>
+        <cq.descriptionSync>false</cq.descriptionSync>
     </properties>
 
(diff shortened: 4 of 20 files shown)


gansheer (approved):

gansheer:
Could these description come from creating a new extension with the org.l2x6.cq:cq-maven-plugin ?

JiriOndrusek:
> e org.l2x6.cq:cq-maven-plugin ?

You are right, when an extension is created via   `./mvnw cq:create -N -Dcq.artifactIdBase=???`, the description is copied from the catalog.
1 - which brings me to an idea, do we want to keep descriptions duplicated in non-generatad files? Like in case we try to avoid mistake/typos/omited words from Camel catalog. Wouldn't be the best approach to delete almost all descriptions in the runtime/pom.xml? 
 Thus using tooling to generate  current descriptions for the *.adoc  based on the catalog? -> which is what this PR does
If this approach is used, the **plugin can be enhanced** to not create a description field. Perhaps even put a comment like  
2 - I Can change the PR to keep descriptions in all extensions and make sure that they are the same as Catalog ones (except a few exception possibly) -> this approach is much harder to maintain

@gansheer , @jamesnetherton WDYT?

JiriOndrusek:
I suggest to do following:

Remove '<'description'>' from the cq:create runtime-pom.xml template. Do a one-time cleanup of existing
  unnecessary overrides. The catalog fallback in CqUtils.getDescription() already handles the rest - no
  logic is changed. Multi-model extensions have to add '<'description'>' manually - the plugin prints a waning (which should be rephrased).

jamesnetherton:
I agree with option 1. The catalog should be the single source of truth for descriptions.

The fallback logic in `CqUtils.getDescription()` already handles this cleanly — single-model extensions get the catalog description automatically, and multi-model extensions still require an explicit `<description>` (with the plugin warning when one is missing).

jamesnetherton:
Actually, thinking more - perhaps option 2 is better. We maybe should not remove descriptions entirely? 

They show up in tooling and various places like central.sonatype.com & mvnrepository.com. I'm also not sure what happens at release time. I think there might be nexus rules around descriptions being mandatory (although if its missing it probably inherits from the parent or something...).

JiriOndrusek:
> Actually, thinking more - perhaps option 2 is better. We maybe should not remove descriptions entirely?
> 
> They show up in tooling and various places like central.sonatype.com & mvnrepository.com. I'm also not sure what happens at release time. I think there might be nexus rules around descriptions being mandatory (although if its missing it probably inherits from the parent or something...).

In case the descriptions are necessary for other tasks/tools, we need  to keep them. In that case, I will only do a clean-up of the descriptions. 

Possibly we can add some kind of validation `cq:check-descriptions`, which could compare the description with the catalog and be executed during mvn ` process-resources -Pformat` to fix a difference automatically. (which almost simulates the "automagic" with an empty description. 

(and add <cq.descriptionSync>false</cq.descriptionSync> into pom.xml, to avoid syncing in case of custom description) - default value = **true**

jamesnetherton:
> process-resources -Pformat

Yeah, I think it'd be good to eventually have something added into that step. It is run by the CI build, so we would catch cases where things get out of sync.

JiriOndrusek:
FYI @jamesnetherton , @gansheer 

When the new cq-maven-plugin (with the fix https://github.com/l2x6/cq-maven-plugin/pull/902) is used in camel-quarkus.
- this PR specifies modules, which keeps their custom description (so for those modules, the sync is off)
- ./mvnw process-resources -Pformat -N is run, the descriptions would be synced and the result should look like https://github.com/JiriOndrusek/camel-quarkus/commit/6fc9055bff1784c16ff0194a8f4d5c8edb45e183

jamesnetherton:
We should add some docs for the `cq.descriptionSync` property to `extension-documentation.adoc`.

One question - could we merge this PR now without bumping the CQ plugin version? I ask because there's a known issue with the plugin that currently prevents us upgrading.

JiriOndrusek:
> We should add some docs for the `cq.descriptionSync` property to `extension-documentation.adoc`.
> 
> One question - could we merge this PR now without bumping the CQ plugin version? I ask because there's a known issue with the plugin that currently prevents us upgrading.

the PR  can be merged without bumping the plugin (I've just verified locally) and the doc page is updated as well

jamesnetherton (approved):

PR #8820, opened 2026-07-02
Add a2a extension

jamesnetherton:


integration-tests/a2a/src/main/java/org/apache/camel/quarkus/component/a2a/it/A2aResource.java
```diff
@@ -0,0 +1,320 @@
...
+        TaskPushNotificationConfig config = result.getMessage().getBody(TaskPushNotificationConfig.class);
+        return String.format("{\"id\":\"%s\",\"taskId\":\"%s\",\"url\":\"%s\"}",
+                config.getId(), config.getTaskId(), config.getUrl());
+    }
+
+    @Path("/push-config/list")
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @SuppressWarnings("unchecked")
+    public String listPushConfigs(@QueryParam("taskId") String taskId) {
```
aldettinger: Maybe rename to getPushConfigCount ?

integration-tests/a2a/src/test/java/org/apache/camel/quarkus/component/a2a/it/A2aTest.java
```diff
@@ -0,0 +1,575 @@
...
+                .then()
+                .statusCode(200)
+                .body(
+                        "state", is("COMPLETED"),
+                        "response", containsString("role=ROLE_USER"));
+    }
+
+    @Test
+    void jsonRpcParseError() {
+        RestAssured.given()
```
aldettinger: Is it Camel Quarkus responsibility to test this ?

integration-tests/a2a/src/test/java/org/apache/camel/quarkus/component/a2a/it/A2aTest.java
```diff
@@ -0,0 +1,575 @@
...
+                .extract()
+                .jsonPath()
+                .getString("taskId");
+
+        RestAssured.given()
+                .queryParam("taskId", taskId)
+                .post("/a2a/cancel-task")
+                .then()
+                .statusCode(200)
+                .body("error", notNullValue());
```
aldettinger, line 150: Is it the expected case that cancel returns an error ?

integration-tests/a2a/src/main/java/org/apache/camel/quarkus/component/a2a/it/A2aResource.java
```diff
@@ -0,0 +1,320 @@
...
+        }
+        Task task = result.getMessage().getBody(Task.class);
+        return String.format("{\"taskId\":\"%s\",\"contextId\":\"%s\",\"state\":\"%s\"}",
+                task.id(), task.contextId(), task.status().state().name());
+    }
+
+    @Path("/agent-card-roundtrip")
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public String agentCardRoundTrip() {
```
aldettinger: The purpose here is to ensure that full-agent-card is a valid JSON ? And full-agent-card is hardcoded resource ?

aldettinger: Maybe there is a special trick with A2AJsonMapper and Quarkus combination ?

extensions/a2a/deployment/src/main/java/org/apache/camel/quarkus/component/a2a/deployment/A2aProcessor.java
```diff
@@ -0,0 +1,65 @@
...
+        Set<String> modelClasses = combinedIndex.getIndex()
+                .getClassesInPackage("org.apache.camel.component.a2a.model")
+                .stream()
+                .map(ClassInfo::name)
+                .map(DotName::toString)
+                .collect(Collectors.toSet());
+
+        return ReflectiveClassBuildItem
+                .builder(modelClasses.toArray(new String[0]))
+                .methods(true)
```
aldettinger, line 62: If I understand Claude well, we would miss `.method(true)` here:

```
Missing .fields() in reflection registration for Jackson serialization
  The code registers model classes with .methods(true) only, but A2A model classes (Task, AgentCard, Message, etc.) are JSON DTOs serialized by Jackson (as seen in A2AJsonMapper usage). Jackson requires
  field-level reflection in native mode. Other extensions handling JSON models (azure-storage-blob, azure-eventhubs, azure-storage-datalake) use .fields() or .fields().methods(). Without this, JSON
  deserialization will fail in native mode with MissingReflectionRegistrationError.
```

Currently, not sure it's relevant, not sure it's not :)

aldettinger:
It looks producer streaming is [documented feature](https://github.com/apache/camel/blob/main/components/camel-ai/camel-a2a/src/main/docs/a2a-producer.adoc?plain=1#L131). Do we have a test for this ? If not, is it worth adding ? Or maybe too similar to consumer streaming functionality ?

On the other side, we could probably remove jsonRpcParseError, jsonRpcMethodNotFound and jsonRpcInvalidParams ?

integration-tests/a2a/src/main/java/org/apache/camel/quarkus/component/a2a/it/A2aResource.java
jamesnetherton: Method renamed as suggested.

integration-tests/a2a/src/test/java/org/apache/camel/quarkus/component/a2a/it/A2aTest.java
jamesnetherton: Fair point. Removed the test and also similar cases (jsonRpcMethodNotFound and jsonRpcInvalidParams).

integration-tests/a2a/src/test/java/org/apache/camel/quarkus/component/a2a/it/A2aTest.java
jamesnetherton: Renamed producerCancelTask to cancelCompletedTaskReturnsError and added a comment explaining the echo agent completes immediately, so canceling returns an error.

integration-tests/a2a/src/main/java/org/apache/camel/quarkus/component/a2a/it/A2aResource.java
jamesnetherton: Good catch. It is a redundant test. The JSON serialization / deseralization is already exercised elsewhere via code paths in the Camel component.

extensions/a2a/deployment/src/main/java/org/apache/camel/quarkus/component/a2a/deployment/A2aProcessor.java
jamesnetherton: The A2A model classes use a mix of patterns — AgentCard uses @JsonDeserialize(builder=...) with setter methods on the Builder, Task is a Java Record (Jackson uses the canonical constructor + accessor methods), and TaskPushNotificationConfig is a standard bean with getters/setters. In all cases Jackson operates through methods, not direct field access, so .methods(true) should be sufficient. The native integration tests pass with this configuration.

jamesnetherton:
> It looks producer streaming is [documented feature](https://github.com/apache/camel/blob/main/components/camel-ai/camel-a2a/src/main/docs/a2a-producer.adoc?plain=1#L131). Do we have a test for this ? If not, is it worth adding ? Or maybe too similar to consumer streaming functionality ?

Good point - test added for producer streaming.

aldettinger (approved):
Well done @jamesnetherton.

apupier (approved):

PR #8821, opened 2026-07-03
docs: add test coverage philosophy to CLAUDE.md

aldettinger:
Clarify when to add tests in Camel Quarkus vs relying on Camel core tests.

Key principles:
- Most component functionality is tested in Camel itself
- Camel Quarkus tests focus on Quarkus runtime compatibility
- Test documented features, Quarkus constraints, and native mode

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -177,6 +177,27 @@ After making changes, run the following from the project root to sync and regene
 ## Integration Tests
 Tests use the Quarkus test framework based on JUnit 6.
 
+### Test Coverage Philosophy
+
+**Camel Quarkus vs Camel Testing:**
+
+Most component functionality is tested in Camel itself. Camel Quarkus tests focus on validating that documented features work correctly in the Quarkus runtime environment (JVM and native mode).
+
+**Recommended test coverage:**
+- **Documented user-facing features** - Verify key features from the upstream Camel component documentation work in Quarkus (operations, data formats, protocols, configuration options)
+- **Quarkus-specific constraints** - Test limitations or behaviors unique to Quarkus (e.g., platform-http vs Undertow differences, native compilation requirements)
+- **Quarkus integration points** - Validate integration with Quarkus extensions (e.g., OAuth with Keycloak testcontainers, camel-oauth integration)
+- **Native mode compatibility** - Ensure resource loading from classpath, reflection registration, and JSON serialization work in native mode
+
+**Quick validation when adding an extension:**
+```bash
+# Review upstream Camel component docs
+open https://github.com/apache/camel/tree/main/components/camel-<category>/camel-<name>/src/main/docs
+
+# Compare documented features against your test methods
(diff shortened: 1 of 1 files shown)


jamesnetherton (commented):
Good addition — the Camel-vs-Camel-Quarkus testing boundary is non-obvious domain knowledge that an AI agent can't derive from code alone. The core philosophy is correct and fills a real gap. A few issues to address:

**1. The `open` command is human-oriented.** AI agents can't open a browser. Consider stating the path as a reference or using a command agents can execute (e.g., `find` or `gh api`).

**2. The URL template `camel-<category>/camel-<name>` is sometimes wrong.** Many components live directly under `components/camel-<name>/` without a category subdirectory (e.g., `camel-kafka`, `camel-sql`). An agent following this literally would get 404s.

**3. The `grep` command only searches `integration-tests/`.** Tests also live under `integration-test-groups/` and `integration-tests-jvm/`. An agent following this would conclude tests are missing when they exist elsewhere — the opposite of the section's intent. Consider:
```bash
find integration-tests/<name> integration-test-groups/*/<name> integration-tests-jvm/<name> -name "*Test.java" 2>/dev/null
```

**4. The glob `**/*Test.java` requires `shopt -s globstar`.** A `find` command is more portable and agent-friendly.

**5. Minor:** PR title references CLAUDE.md but the file is AGENTS.md.

jamesnetherton:
A few more suggestions to make this more effective for AI agents:

**6. Add explicit negative guidance.** Agents respond well to "do NOT" rules. Something like: "Do NOT duplicate tests that already exist in Camel core" would reinforce the philosophy directly.

**7. Tell the agent how to find upstream Camel tests, not just docs.** Knowing what Camel core already tests is how the agent decides what to skip. For example:
```bash
# Check what Camel core already tests for a component
find <camel-repo>/components/camel-<name>/src/test -name "*Test.java" | head -20
```

**8. Consider adding a concrete example.** A single grounded example helps an agent apply abstract guidance. E.g., "For `camel-kafka`, Camel core tests serialization and consumer groups; Camel Quarkus tests that producers work with CDI and that Kafka clients initialize correctly in native mode."

aldettinger:
**1. The tool selection should handle that, at least it does locally. Let's try to save some context tokens and refine in case of issue.

**2/3/4 Having an agent able to resolve where to find camel components should be enough. Need to think about that.

**5 DONE

**6 Explicit negative guidance is subject to debate, for instance in [Markus presentation](https://www.youtube.com/watch?v=Di4ii6Xsb9A). I would vote to keep the context minimal so far with positive guidance only and fall back to negative in case we fail.

**7/8 Having an agent able to resolve where to find the camel component should be enough. Need to think about it. Referring to camel own AGENTS.md is probably not the right solution, let's think.

aldettinger:
Two weeks later, there is still no idea how to do this. Let's keep the test coverage philosophy as human judgment for now and refine when we have a clearer approach. Closing this pr.

PR #8830, opened 2026-07-06
Add integration tests covering camel-mdc supported features

JinyuChen97:
Verifies that camel-quarkus correctly supports the camel-mdc extension across the full feature set defined in apache/camel MDCService:

- Default MDC fields: camel.exchangeId, camel.messageId, camel.routeId, camel.contextId, camel.threadId are populated on every exchange
- Backward-compat fields: camel.correlationId and camel.breadcrumbId are populated when the corresponding property / header is present
- Custom exchange header by exact name (custom-exchange-headers=myHeader)
- All exchange headers (custom-exchange-headers=*)
- Wildcard exchange headers (custom-exchange-headers=prefix*)
- Selected exchange properties by explicit names (custom-exchange-properties=p1,p2)
- All exchange properties (custom-exchange-properties=*)
- Wildcard exchange properties (custom-exchange-properties=prefix*)
- Multi-pattern comma+glob properties (custom-exchange-properties=test_*,more*)
- Async thread propagation via .threads(): MDC values survive the thread boundary and the async thread-id differs from the calling thread-id
- Async wiretap propagation: MDC values are carried into the wiretap thread and the wiretap runs on a distinct thread
- Processor MDC propagation: custom header and all default MDC fields are visible inside a processor invoked on the route

All tests pass in both JVM and native mode.

Co-authored-by: Bob <bob@ibm.com>

### DIFF
--- a/integration-tests/mdc/src/test/java/org/apache/camel/quarkus/component/mdc/it/MdcAsyncTest.java
+++ b/integration-tests/mdc/src/test/java/org/apache/camel/quarkus/component/mdc/it/MdcAsyncTest.java
@@ -17,33 +17,38 @@
 package org.apache.camel.quarkus.component.mdc.it;
 
 import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
 import io.restassured.RestAssured;
 import org.junit.jupiter.api.Test;
 
 import static org.hamcrest.CoreMatchers.containsString;
-import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.not;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 
 @QuarkusTest
-class MdcTest {
+@TestProfile(MdcAsyncTestProfile.class)
+class MdcAsyncTest {
 
     @Test
-    void testCustomHeader() {
-        RestAssured.get("/mdc/customHeader")
+    void testAsyncProcessingWithMdcPropagation() {
+        String response = RestAssured.get("/mdc/async")
                 .then()
                 .statusCode(200)
-                .body(equalTo("HELO"));
+                .body(containsString("asyncHeader:asyncValue"))
+                .body(containsString("asyncProp:asyncPropValue"))
+                .body(containsString("threadId:"), not(containsString("threadId:null")))
+                .extract().asString();
+
+        // MDC threadId on the async thread must differ from the calling thread
+        assertNotEquals(extractValue(response, "threadId"), extractValue(response, "callingThread"));
     }
 
-    @Test
-    void testDefaultMdcFields() {
-        RestAssured.get("/mdc/defaultFields")
-                .then()
-                .statusCode(200)
-                .body(containsString("exchangeId:"), not(containsString("exchangeId:null")))
-                .body(containsString("messageId:"), not(containsString("messageId:null")))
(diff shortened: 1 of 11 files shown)


jamesnetherton (commented):
Given the relative simplicity of the MDC component, we probably don't need to  duplicate a lot of what's already tested by the Camel component (IMO).

I'd suggest trimming the test scope to focus on Quarkus / Camel Quarkus specific functionality (config options etc) & anything that may impact native mode.

Keep:
 
- `testCustomHeader` + `testDefaultMdcFields` in MdcTest — proves the extension activates via quarkus.camel.mdc.enabled and that custom-exchange-headers config wires correctly
- `MdcPropertiesTest` — proves custom-exchange-properties config wiring (the other Quarkus config option)
- `MdcAsyncTest` — MDC propagation across .threads() is worth verifying since Quarkus manages thread pools differently and native mode could break this
- `MdcInterceptBeanTest` — likely worth testing with CDI beans

Remove:
 
- `MdcAllHeadersTest`, `MdcWildcardHeadersTest`, `MdcAllPropertiesTest`, `MdcWildcardPropertiesTest`, `MdcMultiPatternPropertiesTest` — these all test Camel's glob/wildcard pattern matching, not Quarkus config integration
- `MdcWiretapTest` — redundant with `MdcAsyncTest` (both verify cross-thread MDC propagation)
- `testBackwardCompatibilityFields` — correlationId / breadcrumbId population is a pure upstream Camel contract

Each TestProfile also triggers a full Quarkus restart, so this would cut ~8 unnecessary application restarts from CI.

integration-tests/mdc/src/test/java/org/apache/camel/quarkus/component/mdc/it/MdcAsyncTest.java
```diff
@@ -0,0 +1,54 @@
...
+                .statusCode(200)
+                .extract().asString();
+
+        assertThat(response).contains("asyncHeader:asyncValue");
+        assertThat(response).contains("asyncProp:asyncPropValue");
+        assertThat(response).contains("threadId:").doesNotContain("threadId:null");
+        // MDC threadId on the async thread must differ from the calling thread
+        String threadId = extractValue(response, "threadId");
+        String callingThread = extractValue(response, "callingThread");
+        assertThat(threadId).isNotEqualTo(callingThread);
```
jamesnetherton: You typically don't need to extract the response and feed it into assertions.

RestAssured already gives you the capability to do assertions on the response:

https://github.com/rest-assured/rest-assured/wiki/usage

So if you agree that it makes sense, we could remove assertj and rely on RestAssured.

integration-tests/mdc/src/test/java/org/apache/camel/quarkus/component/mdc/it/MdcAsyncTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@QuarkusTest
+@TestProfile(MdcAsyncTestProfile.class)
+class MdcAsyncTest {
+
+    @Test
+    public void testAsyncProcessingWithMdcPropagation() {
```
jamesnetherton: Nitpick (same for other tests)

```suggestion
   void testAsyncProcessingWithMdcPropagation() {
```

jamesnetherton (approved):

PR #8832, opened 2026-07-07
Next is 3.38.0-SNAPSHOT

jamesnetherton:


jamesnetherton:
I need to investigate further issues with syncing CQ example project versions.

jamesnetherton:
> I need to investigate further issues with syncing CQ example project versions.

Examples are building now. So problems seem to be resolved by the latest commit.

PR #8843, opened 2026-07-08
Use git shallow clone of depth 1 for Jenkins deploy job

apupier:
it allows to reduce the amount of space taken by the .git folder. it will decrease the size of it from 350M to 18M. Given that it is kept in workspace history of the last 6 builds, it will save 2G of disk space on Jenkins.
it also saves few seconds to do the git clone.

Not doing it for the Sonarcloud job as it requires the git history

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,9 +33,22 @@ pipeline {
             logRotator(artifactNumToKeepStr: '5', numToKeepStr: '10')
         )
         disableConcurrentBuilds()
+        skipDefaultCheckout()
     }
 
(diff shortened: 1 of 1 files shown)


apupier:
Note that I'm wondering if we could not even completely delete the workspace after the build as it is used only for Sonarcloud and snapshot deploy, I do not think that we have really needs to be able to investigate to the workspace after the build.
Even if not storing the workspace, this would gain few seconds and 330M on disk space

jamesnetherton (approved):

PR #8846, opened 2026-07-08
Do not store workspace on Jenkins in case of success

apupier:
It will save 8 to 10G on the different Jenkins nodes when the builds are in success

Keeping it only in other cases so that we can still investigate in case of failures/cancellation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -58,4 +58,17 @@ pipeline {
             }
         }
     }
+
+    post {
+        always {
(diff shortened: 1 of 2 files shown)


apupier:
build in progress to test the config: https://ci-builds.apache.org/job/Camel/job/Camel%20Quarkus%20Sonarcloud%20Deploy/job/main/1128/

EDIT: deletion worked fine

jamesnetherton (approved):

PR #8853, opened 2026-07-10
[3.33.x] Check for parent SocketException instead of ConnectException to make it work on different systems with different network settings

jamesnetherton:


jamesnetherton:
Test failure is not related and is a known random native flake.

PR #8868, opened 2026-07-10
Add sleep after Splunk container has started

llowinge:
To avoid getting 
```
Caused by: java.net.SocketException: Network is unreachable
```

i'm adding possibility to increase time after splunk is started (to give network configuration to take effect).


llowinge:
@ppalaga Could you check it?

PR #8913, opened 2026-07-24
Restrict Dev UI JSON-RPC bridge to declared console IDs, allowed options and values

jamesnetherton:


oscerd (approved):

extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/devui/CamelCoreDevConsoleProcessor.java
```diff
@@ -35,47 +44,58 @@ void createDevUICards(BuildProducer<CardPageBuildItem> cardsProducer) {
...
+                .componentLink("qwc-camel-core-blocked-exchanges.js")
+                .metadata(CamelDevUIConstants.CONSOLE_ID_METADATA_KEY, "blocked"));
 
         cardPageBuildItem.addPage(Page.webComponentPageBuilder()
                 .title("Browse")
                 .icon("font-awesome-solid:magnifying-glass")
-                .componentLink("qwc-camel-core-browse.js"));
+                .componentLink("qwc-camel-core-browse.js")
+                .metadata(CamelDevUIConstants.CONSOLE_ID_METADATA_KEY, "browse")
+                .metadata(CamelDevUIConstants.ALLOWED_OPTIONS_METADATA_KEY, "dump=false"));
```
apupier, line 55: If we can provide a single value, shouldn't we remove the option and hardcode it in qwc-camel-core-browse.js file directly?
Or is it that we want user to be able to create their own dev UI which can activate this option?

jamesnetherton: It is [hard coded](https://github.com/apache/camel-quarkus/blob/a86a69f626dfd98e062945af55136680e6e1aab3/extensions-core/core/deployment/src/main/resources/dev-ui/qwc-camel-core-browse.js#L32) in `qwc-camel-core-browse.js`.

The purpose of `ALLOWED_OPTIONS_METADATA_KEY` is to prevent an external entity from providing different values (e.g `true`).

apupier (approved):

PR #8940, opened 2026-07-31
Exclude .oss-ai-helper-rules from license header check

jamesnetherton:


apupier (approved):

jamesnetherton:
The initial-mvn-install step passed which is good enough for verification.

Merging this to unblock other PRs.

PR #8946, opened 2026-08-03
Fixes #8836: Don't silently use memory on agents with ql4j

### LINKED ISSUE #8836: Quarkus LangChain4j ChatMemoryProvider silently leaks into AgentWithoutMemory when default model is configured
## Description

When Quarkus LangChain4j (QL4J) is on the classpath and a default `ChatModel` is configured via `quarkus.langchain4j.*` properties, QL4J's `ChatMemoryProvider` silently leaks into Camel's `AgentWithoutMemory`, making it stateful despite the user's explicit intent for stateless behavior.

## Root Cause

QL4J replaces `AiServices.builder()` via SPI (`AiServicesFactory`) in JVM mode and via GraalVM `@Substitute` in native mode. When Camel calls `AiServices.builder(AiAgentWithoutMemoryService.class)`:

1. QL4J's `AiServicesProcessor.determinedImpliedRegisterAiService()` discovers `AiAgentWithoutMemoryService` because it has `@UserMessage` and `@SystemMessage` annotations. The `AGENTIC_PACKAGE_PREFIX` exclusion (`dev.langchain4j.agentic.*`) does not match Camel's package (`org.apache.camel.component.langchain4j.agent.api`).

2. QL4J creates a synthetic `QuarkusAiServiceContext` CDI bean for this interface. Since no explicit `chatMemoryProviderSupplier` is set on the implied registration, it defaults to `BEAN_CHAT_MEMORY_PROVIDER_SUPPLIER` — meaning "inject `ChatMemoryProvider` from CDI."

3. `ChatMemoryProcessor` always creates a default `@ApplicationScoped` `ChatMemoryProvider` (`MessageWindowChatMemory`). This gets injected into the synthetic context.

4. `QuarkusAiServiceContextFactory.create()` returns this pre-configured context (with `ChatMemoryProvider` already set) to Camel's builder.

5. Camel's `AgentWithoutMemory.createAiAgentService()` only calls `.chatModel(config.getChatModel())` — it never resets `chatMemoryProvider`. The QL4J-injected `ChatMemoryProvider` survives into the built proxy.

## Impact

Since `AiAgentWithoutMemoryService` has no `@MemoryId` parameter, the memory ID falls back to either a request-scope hash (HTTP routes) or the literal string `"default"` (non-HTTP routes like timer, file, kafka). This causes:

- **Non-HTTP routes**: All `AgentWithoutMemory` instances across the application share a single global chat memory. Messages from unrelated conversations contaminate each other.
- **HTTP routes**: Per-request isolation, but `InMemoryChatMemoryStore` entries are never evicted (memory leak).
- **All routes**: Each call sends up to 9 prior unrelated messages to the LLM (default window size = 10), increasing token cost, latency, and producing incorrect responses.

No error or warning is logged. The agent silently behaves differently than its name and contract imply.

## Steps to Reproduce

1. Add `quarkus-langchain4j-ollama` (or another QL4J model provider) as a dependency
2. Configure a default model:
   ```properties
   quarkus.langchain4j.ollama.base-url=http://localhost:11434
   quarkus.langchain4j.ollama.chat-model.model-id=llama3.2
   ```
3. Create an `AgentWithoutMemory` with a Camel-managed `ChatModel`:
   ```java
   new AgentWithoutMemory(new AgentConfiguration().withChatModel(chatModel))
   ```
4. Send multiple unrelated messages through the agent
5. Observe that responses reference content from previous, unrelated messages

## Expected Behavior

`AgentWithoutMemory` should be stateless — each call processed independently with no conversation history, regardless of whether QL4J is on the classpath.

## Why Current Tests Don't Catch This

The `langchain4j-agent-ql4j` integration test uses `@Identifier`-qualified `ChatModel` beans and does not configure a default QL4J model via properties. Without a default `ChatModel`, the CDI context bean creation fails and `QuarkusAiServiceContextFactory` falls back to a clean empty context. The `ChatMemoryProvider` doesn't leak — but only by accident, not by design.

## Additional Context

The `langchain4j-agent-ql4j` test README acknowledges related interference: `RetrievalAugmentor` had to be disabled (`cq-test.retrieval.augmentor.disabled=true`) because "Testing with quarkus-langchain4j does not work when the RetrievalAugmentor bean is present with scope <default>."

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/8836
requires https://github.com/quarkiverse/quarkus-langchain4j/pull/2722 and https://github.com/quarkiverse/quarkus-langchain4j/pull/2725 

When Quarkus LangChain4j is on the classpath with a default ChatModel configured, QL4J's AiServicesProcessor discovers AiAgentWithoutMemoryService and creates an implied AI service registration that injects a ChatMemoryProvider, silently making the stateless agent stateful.

Add a @BuildStep that produces ExcludeFromImpliedAiServiceBuildItem for AiAgentWithoutMemoryService, preventing QL4J from registering it as an implied AI service. The QL4J side is handled by quarkiverse/quarkus-langchain4j#2722.

### DIFF
--- a/extensions-support/langchain4j/deployment/pom.xml
+++ b/extensions-support/langchain4j/deployment/pom.xml
@@ -42,6 +42,11 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-support-langchain4j</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.langchain4j</groupId>
+            <artifactId>quarkus-langchain4j-core-deployment</artifactId>
+            <optional>true</optional>
+        </dependency>
     </dependencies>
 
     <build>
--- a/extensions-support/langchain4j/deployment/pom.xml
+++ b/extensions-support/langchain4j/deployment/pom.xml
@@ -52,6 +52,11 @@
             <artifactId>quarkus-junit-internal</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.langchain4j</groupId>
+            <artifactId>quarkus-langchain4j-core-deployment</artifactId>
+            <optional>true</optional>
+        </dependency>
     </dependencies>
 
     <build>
--- a/extensions-support/langchain4j/deployment/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/deployment/SupportQuarkusLangchain4jProcessor.java
+++ b/extensions-support/langchain4j/deployment/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/deployment/SupportQuarkusLangchain4jProcessor.java
@@ -31,6 +31,7 @@ import dev.langchain4j.guardrail.OutputGuardrail;
 import dev.langchain4j.model.embedding.EmbeddingModel;
 import dev.langchain4j.rag.RetrievalAugmentor;
 import dev.langchain4j.store.embedding.EmbeddingStore;
+import io.quarkiverse.langchain4j.deployment.ExcludeFromImpliedAiServiceBuildItem;
 import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
 import io.quarkus.arc.deployment.BeanDiscoveryFinishedBuildItem;
 import io.quarkus.arc.deployment.GeneratedBeanBuildItem;
@@ -251,6 +252,16 @@ class SupportQuarkusLangchain4jProcessor {
                 .addBeanClasses(CamelAiToolsInterceptor.class)
                 .setUnremovable()
                 .build());
+    }
+
+    @BuildStep
+    void excludeCamelAgentInterfacesFromImpliedRegistration(
+            CombinedIndexBuildItem combinedIndex,
+            BuildProducer<ExcludeFromImpliedAiServiceBuildItem> excludedFromImplied) {
+        String agentWithoutMemory = "org.apache.camel.component.langchain4j.agent.api.AiAgentWithoutMemoryService";
+        if (combinedIndex.getIndex().getClassByName(DotName.createSimple(agentWithoutMemory)) != null) {
+            excludedFromImplied.produce(new ExcludeFromImpliedAiServiceBuildItem(agentWithoutMemory));
+        }
     }
 
     // The retrieval filter supplier is only resolved programmatically at augmentor creation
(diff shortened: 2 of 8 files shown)


jamesnetherton:
Will close / reopen the PR to try and kickstart CI.

extensions-support/langchain4j/deployment/pom.xml
```diff
@@ -52,6 +52,11 @@
             <artifactId>quarkus-junit-internal</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.langchain4j</groupId>
+            <artifactId>quarkus-langchain4j-core-deployment</artifactId>
+            <optional>true</optional>
+        </dependency>
```
jamesnetherton, line 59: `<optional>true</optional>` can't work here.

If you run a test where QL4J is not present like `langchain4j-agent`, then you get:

```
ClassNotFoundException: io.quarkiverse.langchain4j.deployment.ExcludeFromImpliedAiServiceBuildItem
```

The only way around it is to make the dependency non-optional. But then you'd need to also make the runtime dependency also non-optional. That'd mean QL4J is always on the runtime classpath whether you want it or not.

extensions-support/langchain4j/runtime/pom.xml
```diff
@@ -70,6 +70,11 @@
             <artifactId>nativeimage</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.langchain4j</groupId>
+            <artifactId>quarkus-langchain4j-core</artifactId>
+            <optional>true</optional>
+        </dependency>
```
jamesnetherton, line 77: Can be removed because it's already declared.

extensions-support/langchain4j/pom.xml
```diff
@@ -30,6 +30,18 @@
...
+        <dependencies>
+            <dependency>
+                <groupId>io.quarkiverse.langchain4j</groupId>
+                <artifactId>quarkus-langchain4j-bom</artifactId>
+                <version>${quarkiverse-langchain4j.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
jamesnetherton, line 43: This is probably to get the `quarkus-langchain4j-core-deployment` artifact. We could remove this BOM import and just add the dependency into the CQ bom like we did for the core runtime dependency.

extensions-support/langchain4j/deployment/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/deployment/SupportQuarkusLangchain4jProcessor.java
```diff
@@ -253,6 +254,16 @@ void configureCamelAiToolTags(
                 .build());
     }
 
+    @BuildStep
+    void excludeCamelAgentInterfacesFromImpliedRegistration(
+            CombinedIndexBuildItem combinedIndex,
+            BuildProducer<ExcludeFromImpliedAiServiceBuildItem> excludedFromImplied) {
+        String agentWithoutMemory = "org.apache.camel.component.langchain4j.agent.api.AiAgentWithoutMemoryService";
```
jamesnetherton, line 261: Maybe we'd also want to include `AiAgentWithMemoryService`. Not sure what the consequences would be.

jamesnetherton (changes requested):
I think we should maybe hold this until after 3.39. It needs some more thought.

jamesnetherton:
Superseded by https://github.com/apache/camel-quarkus/pull/9161.

PR #8950, opened 2026-08-03
Add camel-quarkus-mcp-server extension

### LINKED ISSUE #8949: camel-quarkus-mcp-server - expose ai-tool routes as MCP tools via quarkus-mcp-server
Quarkus runtime of [CAMEL-24308](https://issues.apache.org/jira/browse/CAMEL-24308) (camel-mcp-server): a JVM-only extension providing an `McpServerEngine` over the quarkiverse [quarkus-mcp-server](https://github.com/quarkiverse/quarkus-mcp-server) programmatic `ToolManager` API, so `ai-tool` routes are exposed as MCP tools over streamable HTTP.

- Depends on `camel-mcp-server-api` (the runtime-agnostic bridge + SPI) — not the `camel-mcp-server` Vert.x engine used on Camel Main/JBang, so the MCP Java SDK and Reactor never reach the Quarkus classpath.
- Bridge-owned options as `quarkus.camel.mcp-server.*` (enabled/tags/tool-timeout); serving concerns (path, transports, dev UI, auth, guardrails) stay fully owned by `quarkus.mcp.server.*`, per the precedence rule recorded on [CAMEL-24311](https://issues.apache.org/jira/browse/CAMEL-24311).
- Requires camel 4.22 (`camel-mcp-server-api` is new in 4.22), so this targets the `camel-main` branch.

_Claude Code on behalf of Federico Mariani_

Croway:
Fixes #8949. The Quarkus runtime of [CAMEL-24308](https://issues.apache.org/jira/browse/CAMEL-24308) (camel-mcp-server).

> Targets the **camel-main** branch: `camel-mcp-server-api` is new in Camel 4.22. The camel side ([apache/camel#25306](https://github.com/apache/camel/pull/25306)) is merged; CI needs the 4.22.0-SNAPSHOT artifacts published to the Apache snapshots repository — draft until then. Verified locally against the installed 4.22.0-SNAPSHOT artifacts (JVM and native).

## What this adds

A `camel-quarkus-mcp-server` extension (**JVM and native**) exposing `ai-tool` routes as MCP tools over streamable HTTP, served through the quarkiverse [quarkus-mcp-server](https://github.com/quarkiverse/quarkus-mcp-server) (1.13.1) — per the design recorded on CAMEL-24308: Quarkus reuses its native MCP ecosystem (dev UI, guardrails, CDI, config) instead of the MCP Java SDK.

- **Runtime**: depends on `camel-mcp-server-api` (the runtime-agnostic bridge + `McpServerEngine` SPI) + `quarkus-mcp-server-http`. `QuarkusMcpServerEngine` maps `toolAdded`/`toolRemoved` onto the programmatic `ToolManager` API (`newTool(...).addArgument(...).setHandler(...).register()` / `removeTool`), with tool arguments derived from the ai-tool `ParameterDef` metadata and sanitized results mapped to `ToolResponse.success/error`. The MCP Java SDK and Reactor never reach the classpath.
- **Deployment**: `RuntimeCamelContextCustomizerBuildItem` (gated on `quarkus.camel.mcp-server.enabled`) binds the engine into the Camel registry (the bridge resolves registry-first) and adds the `McpServerBridge` as a CamelContext-managed service.
- **Native support**: the extension lives under `extensions/` with `nativeSince` declared — the engine needed no reflection or resource registration, the native image builds clean and all conformance scenarios pass against the native binary (`McpServerIT`).
- **Config**: `quarkus.camel.mcp-server.enabled` (default true) / `.tags` (unset ⇒ nothing exposed — the security default) / `.tool-timeout` (20s). Serving concerns (path, transports, dev UI, auth, guardrails) stay fully owned by `quarkus.mcp.server.*`, per the CAMEL-24311 precedence rule.
- **Docs**: generated extension page with hand-written intro/usage — combined `quarkus.camel.mcp-server.*` / `quarkus.mcp.server.*` configuration example with the ownership split, a link to the [quarkus-mcp-server configuration reference](https://docs.quarkiverse.io/quarkus-mcp-server/dev/index.html#configuration-reference), YAML DSL tool definition with the full parameter syntax, MCP client configuration examples (camel-openai agent + `mcp.json`-style), a **Serving over stdio** section (add `quarkus-mcp-server-stdio`, disable the default server's HTTP transport with `quarkus.mcp.server."<default>".http.enabled=false`, launch as a subprocess — typically a native executable), dynamic tools and error-handling semantics, and Camel Main/Spring Boot pointers.
- **`@Tool` coexistence proven by test**: quarkus-mcp-server annotated business methods and Camel `ai-tool` routes serve from the same MCP server and the same `tools/list` (annotated tools register at build time, Camel tools follow the route lifecycle).
- Housekeeping done: `cq:update-quarkus-metadata`, docs regen, `process-resources -Pformat`, BOM entries for `quarkus-mcp-server-http(-deployment)` under `quarkiverse-mcp-server.version`, `mcp-server` assigned to the `group-12` test category, dependabot entries.

## Testing

`integration-tests/mcp-server` runs the engine conformance scenarios (CAMEL-24313) driven by `McpAssured` from `quarkus-mcp-server-test` over streamable HTTP (no MCP Java SDK on the test classpath): tag-filtered `tools/list` (untagged default pool excluded), call success, execution error sanitized to a generic message, per-call timeout, the tools list following route stop/start (`tools/list_changed` path), and `@Tool` coexistence. Route control goes through a JAX-RS resource so the same tests run in JVM mode (`McpServerTest`) and against the native binary (`McpServerIT`). All 6 scenarios green in both modes.

---
_This PR was written by Claude Code on behalf of Federico Mariani (@Croway)._


integration-tests-jvm/mcp-server/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+        <version>3.39.0-SNAPSHOT</version>
+        <relativePath>../../poms/build-parent-it/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-integration-test-mcp-server</artifactId>
+    <name>Camel Quarkus :: Integration Tests :: MCP Server</name>
+    <description>Integration tests for Camel Quarkus MCP Server extension</description>
+
+    <properties>
+        <mcp-java-sdk.version>2.0.0</mcp-java-sdk.version>
```
jamesnetherton: Please move this property to the root pom.xml with the other test version properties:

https://github.com/apache/camel-quarkus/blob/main/pom.xml#L187

integration-tests/mcp-server/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+            <artifactId>mcp-core</artifactId>
+            <version>${mcp-java-sdk.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.modelcontextprotocol.sdk</groupId>
+            <artifactId>mcp-json-jackson2</artifactId>
+            <version>${mcp-java-sdk.version}</version>
+            <scope>test</scope>
+        </dependency>
```
jamesnetherton, line 54: Please add these to the dependabot config with the other test dependencies:

https://github.com/apache/camel-quarkus/blob/main/.github/dependabot.yml#L49

integration-tests-jvm/mcp-server/src/test/java/org/apache/camel/quarkus/component/mcp/server/it/McpServerTest.java
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * The engine conformance scenarios (CAMEL-24313) against the quarkus-mcp-server engine, driven by the official MCP SDK
+ * client over streamable HTTP. Mirrors {@code McpServerConformanceTestSupport} from camel-mcp-server-api, which cannot
+ * be reused as-is here because it manages its own CamelContext outside Quarkus.
+ */
+@QuarkusTest
+class McpServerTest {
+
+    @Inject
+    CamelContext camelContext;
```
jamesnetherton: Instead of @Inject into the test class, can we add a JAX-RS resource class that interacts with the Camel context for the tests to interrogate the route status etc. It'll make life easier if we promote the extension to native mode support. You can get examples from other integration tests.

pom.xml
```diff
@@ -56,6 +56,7 @@
         <quarkiverse-freemarker.version>1.3.0</quarkiverse-freemarker.version><!-- https://repo1.maven.org/maven2/io/quarkiverse/freemarker/quarkus-freemarker-parent/ -->
         <quarkiverse-fory.version>1.4.0</quarkiverse-fory.version><!-- https://repo1.maven.org/maven2/io/quarkiverse/fory/quarkus-fory-parent/ -->
         <quarkiverse-groovy.version>3.38.0</quarkiverse-groovy.version><!-- https://repo1.maven.org/maven2/io/quarkiverse/groovy/quarkus-groovy-parent/ -->
+        <quarkiverse-mcp-server.version>1.13.1</quarkiverse-mcp-server.version><!-- https://repo1.maven.org/maven2/io/quarkiverse/mcp/quarkus-mcp-server-parent/ -->
```
jamesnetherton: Nitpick - can we keep alphabetical order and move it above the `quarkiverse-micrometer.version` property.

extensions-jvm/mcp-server/runtime/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-core</artifactId>
+        </dependency>
+        <!-- the runtime-agnostic MCP bridge and engine SPI; the serving engine is provided
+             by the quarkiverse quarkus-mcp-server extension below (NOT the camel-mcp-server
+             Vert.x engine used on Camel Main / JBang) -->
```
jamesnetherton: Can probably remove this.

jamesnetherton (commented):
Some initial feedback before the Camel PR is merged. Otherwise it all looks good to me.

integration-tests-jvm/mcp-server/src/test/java/org/apache/camel/quarkus/component/mcp/server/it/McpServerTest.java
```diff
@@ -0,0 +1,155 @@
...
+ */
+@QuarkusTest
+class McpServerTest {
+
+    private McpSyncClient client;
+
+    private McpSyncClient client() {
+        if (client == null) {
+            client = McpClient
+                    .sync(HttpClientStreamableHttpTransport.builder("http://localhost:8081").build())
```
JiriOndrusek: Hardcoded 8081 can cause troubles if user changes port. Better is to get value from the property  (`quarkus.http.test-port.`)

JiriOndrusek (commented):
So far looks good to me - but I need to review this PR in more details. Here is  1 small test issue (which caught my eye)

JiriOndrusek:
@Croway I'm curious, what is blocking the native support? to my understanding the `camel-quarkus-mcp-servre` worksin native (https://docs.quarkiverse.io/quarkus-mcp-server/dev/)

Croway:
> @Croway I'm curious, what is blocking the native support? to my understanding the `camel-quarkus-mcp-servre` worksin native (https://docs.quarkiverse.io/quarkus-mcp-server/dev/)

good point, in theory nothing, I am adding the native profile and I'm going to test it.

integration-tests/mcp-server/pom.xml
zbendhiba: Could we use [McpAssured](https://docs.quarkiverse.io/quarkus-mcp-server/dev/guides-testing.html) from `quarkus-mcp-server-test `instead of the raw MCP SDK? It would replace both `mcp-core `and `mcp-json-jackson2 `test dependencies and their dependabot/BOM entries.

zbendhiba:
The engine only uses ToolManager APIs from the core module. Would it be possible to depend on just the core and let users choose the transport (`quarkus-mcp-server-http` or `quarkus-mcp-server-stdio`) in their own pom? That would support both HTTP and stdio use cases without hardcoding the transport.

JiriOndrusek:
@croway, FYI the recent `camel-main` branch's CI is almost green (only 1 example fails)therefore if you plan to  push something, consider rebase

Croway:
> The engine only uses ToolManager APIs from the core module. Would it be possible to depend on just the core and let users choose the transport (`quarkus-mcp-server-http` or `quarkus-mcp-server-stdio`) in their own pom? That would support both HTTP and stdio use cases without hardcoding the transport.

mhhhh I would like to achieve the following, OOB Camel MCP will use the http transport, but somehow, the user can configure stdio if needed. As of now this is not possible, I'll do a research if this is feasible somehow on all the runtimes. WDYT @zbendhiba ?

zbendhiba:
> > The engine only uses ToolManager APIs from the core module. Would it be possible to depend on just the core and let users choose the transport (`quarkus-mcp-server-http` or `quarkus-mcp-server-stdio`) in their own pom? That would support both HTTP and stdio use cases without hardcoding the transport.
> 
> mhhhh I would like to achieve the following, OOB Camel MCP will use the http transport, but somehow, the user can configure stdio if needed. As of now this is not possible, I'll do a research if this is feasible somehow on all the runtimes. WDYT @zbendhiba ?

What I suggest is:  instead of  depending on `quarkus-mcp-server-http`, depend only  on `quarkus-mcp-server-core-parent`. Then on the tests, users can pull either `quarkus-mcp-server-http` or `quarkus-mcp-server-stdio` (which probably means 2 flavor of integration tests)
Or have 2 camel quarkus mcp server. 

I don't understand what's limiting us to HTTP.

I don't know. WDYT @jamesnetherton @JiriOndrusek

Croway:
> > > The engine only uses ToolManager APIs from the core module. Would it be possible to depend on just the core and let users choose the transport (`quarkus-mcp-server-http` or `quarkus-mcp-server-stdio`) in their own pom? That would support both HTTP and stdio use cases without hardcoding the transport.
> > 
> > 
> > mhhhh I would like to achieve the following, OOB Camel MCP will use the http transport, but somehow, the user can configure stdio if needed. As of now this is not possible, I'll do a research if this is feasible somehow on all the runtimes. WDYT @zbendhiba ?
> 
> What I suggest is: instead of depending on `quarkus-mcp-server-http`, depend only on `quarkus-mcp-server-core-parent`. Then on the tests, users can pull either `quarkus-mcp-server-http` or `quarkus-mcp-server-stdio` (which probably means 2 flavor of integration tests) Or have 2 camel quarkus mcp server.
> 
> I don't understand what's limiting us to HTTP.
> 
> I don't know. WDYT @jamesnetherton @JiriOndrusek

but this way the user should configure the http or the stdio when using quarkus, right? it won't work by adding the extension only

jamesnetherton:
> WDYT @jamesnetherton @JiriOndrusek

If it is just literally a case of switching between `quarkus-mcp-server-http` & `quarkus-mcp-server-stdio`, then yes, we could just have the extension depend on `quarkus-mcp-server-core` and users choose the server impl.

If it's not that simple, then maybe we just go with HTTP initially?

Croway:
> > WDYT @jamesnetherton @JiriOndrusek
> 
> If it is just literally a case of switching between `quarkus-mcp-server-http` & `quarkus-mcp-server-stdio`, then yes, we could just have the extension depend on `quarkus-mcp-server-core` and users choose the server impl.
> 
> If it's not that simple, then maybe we just go with HTTP initially?

I'd go with HTTP, in particular, right now the following is possible:
* add the `quarkus-mcp-server-stdio` on the pom.xml
* enable stdio via properties `quarkus.mcp.server.stdio.enabled=true`
* disable http `quarkus.mcp.server."<default>".http.enabled=false`

This works with the current code, and can be documented, it might be a bit hacky, but this way the component is battery included, OOB, http works.

jamesnetherton:
> This works with the current code, and can be documented, it might be a bit hacky, but this way the component is battery included, OOB, http works

I'm ok with that. Maybe @zbendhiba has more thoughts about it?

I guess we can always adjust things as needed later.

zbendhiba:
ok

Thanks for the discussion

jamesnetherton:
This looks good to me. To avoid PR builds failing endlessly due to changes on the Camel `main` branch, I can merge this manually onto `camel-main`. I tested this locally and the tests pass fine.

PR #8965, opened 2026-08-05
Add openapi automated test in camel quarkus

JinyuChen97:
added automate test in camel quarkus to validate request body generated in openapi schema with no consume attribute provided. (Jira related)


aldettinger:
@JinyuChen97 Are we testing that a rest route without "consume" statement is working ? Other routes in the tests have this, so are we already covered ? Or maybe we are testing something else.

JinyuChen97:
> @JinyuChen97 Are we testing that a rest route without "consume" statement is working ? Other routes in the tests have this, so are we already covered ? Or maybe we are testing something else.

Hi @aldettinger, thanks for the feedback, the existing POST routes all explicitly define the consumes attribute. This test covers a different scenario from a downstream jira when consumes is not set, the generated OpenAPI spec was missing the requestBody entirely. The upstream camel fix defaults to application/json in that case, and this test verifies that behavior works in camel quarkus.

aldettinger (approved):

PR #8975, opened 2026-08-09
fix integration cassandra-source kamelet pod status is in running state but not ready

JinyuChen97:
This fix related to a downstream issue that tested camel quarkus integrated with cassandra-source kamelet in openshift/k8s found pod status is in running status but not ready.

**root cause found**:
cassandra-quarkus-client auto registered a health check that defaults to 127.0.0.1:9042 when quarkus.cassandra.contact-points isn't configured as camel managed it's own cqlsession, the health check could not find the cassandra target and caused openshift/k8s readiness probe failures.

**update**:
provide workaround in guide with quarkus.cassandra.health.enabled set to false to fix this issue.

**Before fix**:
jinyuchen@Jinyus-MacBook-Pro camel-quarkus % kubectl get pods                                          
NAME                                            READY   STATUS    RESTARTS   AGE
cassandra-5f7c94bf8c-g7lzp                      1/1     Running   0          3m15s
cassandra-source-kamelet-test-8448d69bd-cpfnd   **0/1**     Running   0          114s

**After fix**:
jinyuchen@Jinyus-MacBook-Pro camel-quarkus % kubectl get pods
NAME                                            READY   STATUS    RESTARTS   AGE
cassandra-5f7c94bf8c-g7lzp                      1/1     Running   0          25m
cassandra-source-kamelet-test-8cfd57cbc-9cqgh   **1/1**     Running   0          22s

### DIFF
--- a/docs/modules/ROOT/pages/migration-guide/3.39.0.adoc
+++ b/docs/modules/ROOT/pages/migration-guide/3.39.0.adoc
@@ -35,3 +35,18 @@ quarkus.camel.jolokia.remote-access-allowed=true
 ----
 
 This is safe when combined with SSL client authentication (enabled by default in Kubernetes environments), which ensures only clients presenting a valid certificate can connect. Refer to the xref:reference/extensions/jolokia.adoc[Jolokia extension documentation] for full details.
+
+== CassandraQL extension
+
+=== Cassandra health check may cause pod readiness failures
+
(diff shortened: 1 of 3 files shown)


jamesnetherton:
This is a breaking change for anyone using cassandra-quarkus managed client beans.

That said, this is a tricky problem to solve because cassandra-quarkus will always produce a 'default' client bean for `localhost: 9042`, if none of the `quarkus.cassandra.*` config options are set.

As a minimum we should:

* Document this change in behaviour in the 3.39.0 upgrade guide
* Add a note into our cassandra extension documentation about the need to explicitly enable cassandra-quarkus health checks when using its client beans with Camel

However...

A better solution might be to propose adding a new build time config property like `quarkus.cassandra.enabled` to cassandra-quarkus. We could set it to `false` in our cassandra extension to suppress creation of the client beans, health checks etc.

jamesnetherton (approved):

PR #8980, opened 2026-08-10
adding fix and update guide to describe the fixing integration cassandra-source kamelet to log on openshift is in running state but pod is not ready

JinyuChen97:
set quarkus.cassandra.health.enabled to false by default to prevent openshift/k8s readness probes issue caused pod is running but not ready status issue. Update docs to show case the command to use if user would like to set up quarkus cassandra client in camel quarkus.


JinyuChen97:
created proposal in quarkus-cassandra to have a better fix: https://github.com/datastax/cassandra-quarkus/issues/275

jamesnetherton (approved):

PR #9000, opened 2026-08-13
 feat(extensions): micrometer support log on shutdown format

squakez:
Ref https://issues.apache.org/jira/browse/CAMEL-24356


squakez:
It probably requires the upgrade to `4.23.0-SNAPSHOT` on `camel-main`

PR #9006, opened 2026-08-14
Fixes #9003. Add an ingestion tracker SPI with a JDBC impl for AI document ingestion

### LINKED ISSUE #9003: Add a sync ledger SPI for keeping RAG vector stores in sync with source documents
Vector stores cannot be enumerated for "what did I already ingest", and LangChain4j has no equivalent of LangChain-Python's indexing API (`RecordManager`/`cleanup`), so today any ingestion pipeline is either append-only (restarts re-embed the whole corpus, edits join their previous vectors instead of replacing them) or wipe-and-reload (loses everything between passes).

Upstream context showing this gap is real and repeatedly asked for, never fully closed:

- langchain4j/langchain4j#1085 — "Does langchain4j have indexing API?" (closed after only the removal-support half shipped; the record-manager/indexing API itself never followed)
- langchain4j/langchain4j#2931 — "Incremental update of the knowledge base" (open; maintainer confirms "there is currently no such functionality in LangChain4j, you need to track that yourself")
- quarkiverse/quarkus-langchain4j#473 — "Easy RAG: strategies to control whether data should be ingested or not" (closed; a checksum/diff re-ingestion idea was floated and called "awesome if we could do" but never built)

A `SyncLedger` SPI + `JdbcSyncLedger` JDBC implementation now exist on the `feature/sql-ladger` branch (`extensions-support/langchain4j`), with a behavioural contract test (`SyncLedgerContract`) verified against both H2 and real PostgreSQL (`integration-tests/langchain4j-sync-ledger`, JVM + native). It's currently experimental and internal (not yet wired into a user-facing extension) — this issue tracks turning it into a real PR and, longer term, wiring it into a `langchain4j-ingest` extension.

Note: this is deliberately a stopgap. If langchain4j ever ships its own record-manager/indexing API upstream (per langchain4j/langchain4j#1085 / langchain4j/langchain4j#2931 above), our `SyncLedger`/`JdbcSyncLedger` are candidates for removal in favor of an adapter over the upstream mechanism — `SyncLedger` is designed as a swappable SPI for exactly this reason, so this feature carries no long-term compatibility promise.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/9003

_This work is not plugged into any existing code: no current extension or code path changes behaviour, the module is purely additive and exists as groundwork for an upcoming declarative ingestion feature (a future `langchain4j-ingest` extension's `sync` mode)._

Adds the bookkeeping layer for keeping a vector store in sync with a changing document source,  tracked by #9003.

  - `IngestionTracker` SPI (`extensions-support/langchain4j`): one row per document —  fingerprint, content hash, committed/intended segment counts, two-phase `in_progress`/`done`/`failed` status, `tombstone`/`pinned` flags. The module README documents the full document-state machine.
  - `JdbcIngestionTracker`: reference implementation with dialect-free SQL (PostgreSQL and H2), JDK types only, owning its `cq_ingestion_tracker` schema.
  - `JdbcIngestionTrackerTest`: the behavioural contract on H2 — its test methods are written against the SPI only and are meant to be extracted into a shared contract base class once a second implementation exists.
  - `integration-tests/langchain4j-ingestion-tracker`: the same guarantees over REST against a real PostgreSQL (Dev Services), with a native twin.

  Status: **experimental, internal SPI** — not consumed by any user-facing extension yet; it is the foundation for a future `langchain4j-ingest` extension's `sync` mode. LangChain4j has no equivalent of LangChain-Python's `RecordManager` (langchain4j/langchain4j#2931), so this fills the gap deliberately as a stopgap: if an upstream record manager lands, the SPI is designed to be replaced wholesale by an adapter over it — no long-term compatibility promise is made.

### DIFF
--- a/extensions-support/langchain4j/README.adoc
+++ b/extensions-support/langchain4j/README.adoc
@@ -84,9 +84,18 @@ transition except `deleteRow`, which removes the row whole.
 reference implementation: deliberately dialect-free SQL (upserts are update-then-insert inside
 application code rather than vendor `MERGE`/`ON CONFLICT`), using only JDK types, so the same
 code runs unmodified against PostgreSQL and H2. It owns its own schema
-(`cq_ingestion_tracker`, created on first use via `ensureSchema()`) and expects to be handed a
+(`camel_quarkus_ingestion_tracker`, created on first use via `ensureSchema()`; document ids up
+to 1024 characters, the S3 object-key maximum) and expects to be handed a
 `javax.sql.DataSource` — in a Quarkus application that's typically an Agroal-managed datasource,
-with Dev Services providing one automatically in dev/test when none is configured.
+with Dev Services providing one automatically in dev/test when none is configured. Operators who
+run the application's DB user without DDL privileges can pre-provision the table and construct
+the tracker with `createTableIfNotExists = false`, in which case `ensureSchema()` only probes
+that the table exists (the same escape hatch camel-sql's `JdbcMessageIdRepository` offers).
+
+NOTE: the tracker table is a durable inventory of the corpus. Document content never lands in
+it, but `doc_id` values are typically paths, URLs or object keys, and those routinely encode
+sensitive structure (`/customers/12345/contract.pdf`). Document ids also appear in exception
+messages, so they reach logs and stack traces.
 
 `JdbcIngestionTrackerTest` (in this module) pins the behavioural guarantees every implementation
 must satisfy (durability of in-progress intents, shrink-bound tracking across re-intents,
--- a/extensions-support/langchain4j/README.adoc
+++ b/extensions-support/langchain4j/README.adoc
@@ -1,0 +1,107 @@
+= LangChain4j Support
+
+This module provides common support code shared by Camel Quarkus LangChain4j extensions.
+
+== Ingestion tracker
+
+Status: *Experimental*.
+
+`org.apache.camel.quarkus.component.support.langchain4j.tracker.IngestionTracker` is the bookkeeping
+SPI behind keeping a vector store in sync with a changing document source (files, buckets, feeds,
+...). Vector stores cannot be enumerated for "what did I already ingest", and LangChain4j has no
+equivalent of LangChain-Python's `RecordManager`
+(https://github.com/langchain4j/langchain4j/issues/2931[langchain4j#2931]), so without external
+bookkeeping any ingestion pipeline is either append-only (restarts re-embed the whole corpus,
+edited documents join their previous vectors instead of replacing them) or wipe-and-reload
+(loses everything between passes). `IngestionTracker` closes that gap: it tracks one row per document
+— fingerprint, content hash, segment count, a two-phase `in_progress`/`done`/`failed` status,
+and `tombstone`/`pinned` flags — as the single authority on what was ingested. The vector store
+itself stays a disposable projection that is never asked questions; losing the tracker only costs
+re-ingestion, never correctness, because segment ids are deterministic.
+
+The two-phase `writeIntent`/`commit` protocol is what makes the tracker crash-safe: a row left
+`in_progress` by a crash (killed process, OOM, ...) is never mistaken for "already ingested", so
+the next delivery of the same document converges the store instead of skipping it.
+
+=== Document states
+
+Each document is one row that moves through the states below. Transitions are the SPI methods;
+the annotations in parentheses describe when the consumer (the ingestion pipeline) invokes them.
+
+----
+ (no row)
+    │  writeIntent                 durable BEFORE the store is touched; a crash strands
+    ▼                              the row here, and an in_progress row is never trusted
+ in_progress                       as "already ingested" — the next delivery of the
+    │  commit                      document converges the store
+    ▼
+  done ◄──────────────────────┐
+    │                         │
+    ├─ (unchanged) ──► done   │  no transition: the skip that makes restarts free
+    │                         │
+    ├─ refreshFingerprint ────┘  fingerprint renewed after a content-hash match
+    │                            (tier-2), so the cheap tier-1 check works next pass
+    │
+    ├─ writeIntent ──► in_progress    (edited content: replace; the shrink bound
+    │                                  max(committed, intended) rides along in the row)
+    │
+    ├─ markFailed ──► failed          (processing failure — also fired straight from
+    │      │                           in_progress; previously committed segments, if
+    │      │                           any — a stale but valid version — keep serving)
+    │      │
+    │      ├─ (fingerprint unchanged) ──► failed   dead-lettered: skipped on every
+    │      │                                       pass until the content changes
+    │      │
+    │      └─ writeIntent ──► in_progress          (changed content: retry)
+    │
+    ├─ tombstone() ──► done(0) + tombstone    (explicit delete: the consumer removes
+    │      │                                   the vectors and records commit(0) with
+    │      │                                   the flag) — "deleted stays deleted":
+    │      │                                   every future ingest of the document is
+    │      │                                   suppressed, even though the source
+    │      │                                   still has it
+    │      │
+    │      └─ unsuppress() ──► done(0)         suppression lifted: the next pass
+    │                                          re-ingests the document
+    │
+    └─ pin() ──► done + pinned        (an API write corrected a source-owned document)
+           │                           — "the correction wins": source-origin updates
+           │                           are suppressed; API-origin writes still pass
+           │
+           └─ unpin() ──► done         the source owns the document again
+
+ any state ── deleteRow ──► (no row)   reconciliation: the source no longer lists the
+                                       document (guarded by the consumer's pass
+                                       interlock, never fired by the tracker itself)
+----
+
+`tombstone` and `pinned` are columns orthogonal to `status`: drawn above off `done` because
+that is how the consumer composes them, but they can accompany any state (a tombstone can even
+be recorded for a never-ingested document, as a pure suppression record) and they survive every
+transition except `deleteRow`, which removes the row whole.
+
+`org.apache.camel.quarkus.component.support.langchain4j.tracker.jdbc.JdbcIngestionTracker` is the
+reference implementation: deliberately dialect-free SQL (upserts are update-then-insert inside
+application code rather than vendor `MERGE`/`ON CONFLICT`), using only JDK types, so the same
+code runs unmodified against PostgreSQL and H2. It owns its own schema
+(`cq_ingestion_tracker`, created on first use via `ensureSchema()`) and expects to be handed a
+`javax.sql.DataSource` — in a Quarkus application that's typically an Agroal-managed datasource,
+with Dev Services providing one automatically in dev/test when none is configured.
+
+`JdbcIngestionTrackerTest` (in this module) pins the behavioural guarantees every implementation
+must satisfy (durability of in-progress intents, shrink-bound tracking across re-intents,
+tombstone/pin lifecycle, dead-letter semantics on failure, ...) against H2 for a fast unit loop.
+Its test methods are written against the SPI only, so they double as the contract: the day a
+second implementation exists — for example an adapter over a future upstream LangChain4j record
+manager — they are meant to be extracted into an abstract `IngestionTrackerContract` base class
+with one `*Test` subclass per implementation, and a replacement is a drop-in exactly when its
+subclass passes. The guarantees are verified twice:
+`integration-tests/langchain4j-ingestion-tracker` runs the same set over REST against a real
+PostgreSQL instance (Dev Services), because H2 alone does not prove the "works on PostgreSQL"
+half of the contract.
+
+CAUTION: `IngestionTracker` and its `tracker.*` package are experimental and an internal SPI, not a
+public API — no compatibility guarantees between releases, and applications must not implement or
+call it directly; it may change or be removed without a deprecation cycle. It exists so the
+implementation can be replaced wholesale later. It is not yet wired into a user-facing extension;
+it is consumed by the upcoming `langchain4j-ingest` extension's `sync` mode.
--- a/extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/IngestionTracker.java
+++ b/extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/IngestionTracker.java
@@ -1,0 +1,128 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.support.langchain4j.tracker;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * The ingestion tracker: one row per document, the authority on what was ingested. The vector store is
+ * a projection that is never asked questions — losing the tracker costs re-ingestion (which
+ * converges thanks to deterministic segment ids), never correctness. Reconciliation is
+ * tracker-versus-source; the store is never enumerated.
+ *
+ * <p>
+ * <strong>Status: Experimental.</strong> Internal SPI — not a public API. No compatibility
+ * guarantees between releases; applications must not implement or call this interface, and it
+ * may change or be removed without a deprecation cycle. It exists so the implementation can be
+ * replaced wholesale: LangChain4j has no equivalent of LangChain-Python's {@code RecordManager} yet
+ * (<a href="https://github.com/langchain4j/langchain4j/issues/2931">langchain4j#2931</a>), and
+ * once one lands upstream, an adapter implementing this interface replaces the
+ * {@code tracker.jdbc} package while every consumer stays untouched. New implementations must
+ * pass the behavioural tests of {@code JdbcIngestionTrackerTest}, whose test methods are written
+ * against this SPI only and are meant to be extracted into a shared contract base class the day
+ * a second implementation exists.
+ *
+ * <p>
+ * Two method groups, by replaceability:
+ * <ul>
+ * <li><em>Tracker subset</em> — {@link #ensureSchema}, {@link #read}, {@link #listDocuments},
+ * {@link #refreshFingerprint}, {@link #deleteRow}: mirrors what an upstream record manager
+ * provides and would delegate to it directly.</li>
+ * <li><em>Camel Quarkus extensions</em> — the two-phase {@link #writeIntent}/{@link #commit}
+ * protocol, {@link #tombstone}/{@link #unsuppress}, {@link #pin}/{@link #unpin},
+ * {@link #markFailed}: product semantics an upstream tracker will not carry; an adapter keeps
+ * these in side storage keyed the same way.</li>
+ * </ul>
+ */
+public interface IngestionTracker {
+
+    String ORIGIN_SOURCE = "source";
+    String ORIGIN_API = "api";
+
+    /** Creates or migrates the backing schema. Called once before first use. */
+    void ensureSchema();
+
+    Optional<TrackerRow> read(String pipeline, String documentId);
+
+    /** All rows of a pipeline — the reconciliation input. */
+    List<TrackerRow> listDocuments(String pipeline);
+
+    /**
+     * Durably records the intent to (re)write a document <em>before</em> the store is touched.
+     * A row left {@code in_progress} by a crash is never skipped by change detection, so the
+     * next delivery of the document converges the store.
+     */
+    void writeIntent(String pipeline, String documentId, String fingerprint, String contentHash,
+            int committedCount, int intendedCount, String origin);
+
+    /** Marks the write complete. Only rows in {@code done} status participate in skip decisions. */
+    void commit(String pipeline, String documentId, String fingerprint, String contentHash, int segmentCount);
+
+    /**
+     * Refreshes the stored fingerprint of an already-{@code done} row whose content turned out
+     * unchanged (tier-2 hash match after a tier-1 fingerprint miss), so the cheaper tier-1 check
+     * skips the document next time.
+     */
+    void refreshFingerprint(String pipeline, String documentId, String fingerprint);
+
+    /**
+     * Marks an explicit delete: the row survives as a suppression record, so the document stays
+     * deleted even while its source file still exists. Lifted with {@link #unsuppress}.
+     */
+    void tombstone(String pipeline, String documentId);
+
+    void unsuppress(String pipeline, String documentId);
+
+    /** Pins a document: source updates are suppressed until {@link #unpin}. */
+    void pin(String pipeline, String documentId);
+
+    void unpin(String pipeline, String documentId);
+
+    /** Removes a row entirely — used by deletion-by-disappearance, where nothing needs remembering. */
+    void deleteRow(String pipeline, String documentId);
+
+    /**
+     * Dead-letters a document after a processing failure: records the fingerprint of the failed
+     * attempt so subsequent passes skip the poison document until its content changes, instead
+     * of failing identically forever at cost. Previous committed segments (a stale version that
+     * correctly keeps serving) stay untouched.
+     */
+    void markFailed(String pipeline, String documentId, String fingerprint);
+
+    /**
+     * @param status {@code done}, {@code in_progress} or {@code failed}
+     * @param origin {@link #ORIGIN_SOURCE} or {@link #ORIGIN_API}
+     */
+    record TrackerRow(String pipeline, String documentId, String fingerprint, String contentHash,
+            int segmentCount, int intendedCount, String status, String origin, boolean tombstone,
+            boolean pinned) {
+
+        public boolean done() {
+            return "done".equals(status);
+        }
+
+        public boolean failed() {
+            return "failed".equals(status);
+        }
+
+        /** The shrink bound: the largest segment index that may exist in the store, ever intended. */
+        public int maxKnownCount() {
+            return Math.max(segmentCount, intendedCount);
+        }
+    }
+}
--- a/extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/IngestionTracker.java
+++ b/extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/IngestionTracker.java
@@ -65,32 +65,48 @@ public interface IngestionTracker {
     /**
      * Durably records the intent to (re)write a document <em>before</em> the store is touched.
      * A row left {@code in_progress} by a crash is never skipped by change detection, so the
-     * next delivery of the document converges the store.
+     * next delivery of the document converges the store. The recorded intended count is
+     * <em>monotone until commit</em>: a re-intent never lowers it, because once an intent for N
+     * is durable, up to N segments may exist in the store and only a {@link #commit} proves the
+     * sweep happened.
      */
     void writeIntent(String pipeline, String documentId, String fingerprint, String contentHash,
-            int committedCount, int intendedCount, String origin);
+            int intendedCount, String origin);
 
-    /** Marks the write complete. Only rows in {@code done} status participate in skip decisions. */
+    /**
+     * Marks the write complete. Only rows in {@code done} status participate in skip decisions.
+     * Throws {@link IllegalStateException} when no row exists for the key — a commit without a
+     * preceding intent is a protocol violation, never a no-op.
+     */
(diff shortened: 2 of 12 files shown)


extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
```diff
@@ -0,0 +1,288 @@
...
+                resultSet.getInt(3), resultSet.getInt(4), resultSet.getString(5),
+                resultSet.getString(6), resultSet.getBoolean(7), resultSet.getBoolean(8));
+    }
+
+    @Override
+    public void writeIntent(String pipeline, String documentId, String fingerprint, String contentHash,
+            int committedCount, int intendedCount, String origin) {
+        try (Connection connection = dataSource.getConnection()) {
+            String update = "UPDATE " + TABLE + " SET fingerprint = ?, content_hash = ?, "
+                    + "intended_count = ?, status = 'in_progress', origin = ?, updated_at = ? "
```
jamesnetherton: `intended_count = ?` overwrites unconditionally, so the shrink bound documented in the README ("the shrink bound `max(committed, intended)` rides along in the row") isn't actually held across an uncommitted intent — see the reproduction in the review body.

Once an intent for N is durable, up to N segments may exist in the store, so the bound has to stay >= N until a `commit` proves the sweep happened. `commit` lowering it is correct; this path lowering it isn't. Still dialect-free:

```sql
intended_count = CASE WHEN intended_count > ? THEN intended_count ELSE ? END
```

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/IngestionTracker.java
```diff
@@ -0,0 +1,128 @@
...
+    /** All rows of a pipeline — the reconciliation input. */
+    List<TrackerRow> listDocuments(String pipeline);
+
+    /**
+     * Durably records the intent to (re)write a document <em>before</em> the store is touched.
+     * A row left {@code in_progress} by a crash is never skipped by change detection, so the
+     * next delivery of the document converges the store.
+     */
+    void writeIntent(String pipeline, String documentId, String fingerprint, String contentHash,
+            int committedCount, int intendedCount, String origin);
```
jamesnetherton: `committedCount` is only read on the INSERT branch of the JDBC impl — passing `row.maxKnownCount()` back in on a re-intent has no effect at all (I checked). As it stands it's a parameter that looks like it preserves the bound and doesn't. Either honour it on the update path or drop it from the signature.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
```diff
@@ -0,0 +1,288 @@
...
+    public void pin(String pipeline, String documentId) {
+        setFlag(pipeline, documentId, "pinned", true);
+    }
+
+    @Override
+    public void unpin(String pipeline, String documentId) {
+        setFlag(pipeline, documentId, "pinned", false);
+    }
+
+    private void setFlag(String pipeline, String documentId, String column, boolean value) {
```
jamesnetherton: `setFlag` is UPDATE-only, so `tombstone`/`pin`/`unsuppress`/`unpin` silently no-op when no row exists — `tombstone("p", "ghost")` leaves no row and throws nothing. The README (line 80) explicitly promises the opposite: "a tombstone can even be recorded for a never-ingested document, as a pure suppression record".

So "deleted stays deleted" doesn't hold for a document deleted before its first successful ingest — the next pass re-ingests it. Either upsert a suppression row here or correct the README.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
```diff
@@ -0,0 +1,288 @@
...
+        } catch (SQLException e) {
+            throw new IllegalStateException("Tracker row deletion failed for '" + documentId + "'", e);
+        }
+    }
+
+    @Override
+    public void commit(String pipeline, String documentId, String fingerprint, String contentHash,
+            int segmentCount) {
+        String sql = "UPDATE " + TABLE + " SET fingerprint = ?, content_hash = ?, segment_count = ?, "
+                + "intended_count = ?, status = 'done', updated_at = ? WHERE pipeline = ? AND doc_id = ?";
```
jamesnetherton, line 317: `executeUpdate()`'s row count is never checked, so `commit` on a key with no preceding intent writes nothing and reports success. Given the two-phase protocol is the whole point of the design, a commit that quietly evaporates is a rough thing to debug — worth failing loudly.

Same shape at line 274 — `refreshFingerprint` silently no-ops when the row isn't `done`. That one may well be intentional, but it's undocumented.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
```diff
@@ -0,0 +1,288 @@
...
+                + "segment_count INT DEFAULT 0 NOT NULL, "
+                + "intended_count INT DEFAULT 0 NOT NULL, "
+                + "status VARCHAR(16) NOT NULL, "
+                + "origin VARCHAR(16) DEFAULT 'source' NOT NULL, "
+                + "tombstone BOOLEAN DEFAULT FALSE NOT NULL, "
+                + "pinned BOOLEAN DEFAULT FALSE NOT NULL, "
+                + "updated_at TIMESTAMP, "
+                + "PRIMARY KEY (pipeline, doc_id))";
+        try (Connection connection = dataSource.getConnection()) {
+            connection.createStatement().execute(ddl);
```
jamesnetherton: `CREATE TABLE IF NOT EXISTS` against the application's datasource forces the production DB user to hold DDL rights, with no opt-out. Worth mirroring `JdbcMessageIdRepository` in camel-sql: a `createTableIfNotExists` flag (default true) plus a `tableExists` probe, so operators can pre-provision and run least-privilege. When that becomes a config option it should be `ConfigPhase.RUN_TIME` per our security model doc.

Also minor: `connection.createStatement()` isn't in the try-with-resources.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
```diff
@@ -0,0 +1,288 @@
...
+ * remove this race without either {@code SERIALIZABLE} isolation or a dialect-specific upsert,
+ * both at odds with staying dialect-free; callers (the sync pass runner processes one document
+ * at a time per pipeline) are relied upon to hold this invariant instead.
+ *
+ * <p>
+ * <strong>Status: Experimental.</strong> See {@link IngestionTracker}.
+ */
+public class JdbcIngestionTracker implements IngestionTracker {
+
+    static final String TABLE = "cq_ingestion_tracker";
```
jamesnetherton: Everything caller-supplied goes through `PreparedStatement` parameters and `TABLE` is a constant, so there's no injection today. But a `sync` mode will almost certainly grow a configurable table/schema name, and this concatenation becomes injectable the moment it does. Worth a comment here saying it must never become caller-supplied without identifier validation.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
```diff
@@ -0,0 +1,288 @@
...
+
+    public JdbcIngestionTracker(DataSource dataSource) {
+        this.dataSource = dataSource;
+    }
+
+    @Override
+    public void ensureSchema() {
+        String ddl = "CREATE TABLE IF NOT EXISTS " + TABLE + " ("
+                + "pipeline VARCHAR(128) NOT NULL, "
+                + "doc_id VARCHAR(512) NOT NULL, "
```
jamesnetherton: `VARCHAR(512)` hard-fails rather than truncating on longer keys (confirmed: `JdbcSQLDataException`). S3 keys go to 1024 bytes and file URIs can exceed 512 too.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
```diff
@@ -0,0 +1,288 @@
...
+                statement.setString(1, fingerprint);
+                statement.setTimestamp(2, Timestamp.from(Instant.now()));
+                statement.setString(3, pipeline);
+                statement.setString(4, documentId);
+                updated = statement.executeUpdate();
+            }
+            if (updated == 0) {
+                String insert = "INSERT INTO " + TABLE
+                        + " (pipeline, doc_id, fingerprint, status, origin, updated_at) "
+                        + "VALUES (?, ?, ?, 'failed', 'source', ?)";
```
jamesnetherton, line 286: The INSERT branch hardcodes `origin = 'source'`, so a document that only ever fails records the wrong origin.

extensions-support/langchain4j/runtime/src/test/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTrackerTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+        IngestionTracker.TrackerRow row = tracker.read("p", "doc").orElseThrow();
+        assertTrue(row.done());
+        assertEquals("fp1", row.fingerprint());
+        assertEquals("hash1", row.contentHash());
+        assertEquals(3, row.segmentCount());
+    }
+
+    @Test
+    void reintentKeepsTheLargestKnownCount() {
```
jamesnetherton, line 91: This only covers re-intent *after* a commit — the path where `segment_count` carries the bound, which works. The intent -> intent path is the broken one and isn't covered.

Also missing: `markFailed` straight from `in_progress`, `commit` without a preceding intent, and `refreshFingerprint` on a non-`done` row.

integration-tests/langchain4j-ingestion-tracker/src/test/java/org/apache/camel/quarkus/component/langchain4j/ingestiontracker/it/Langchain4jIngestionTrackerTest.java
```diff
@@ -0,0 +1,163 @@
...
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasSize;
+
+/**
+ * Runs the {@code IngestionTracker} behavioural guarantees (mirrored from {@code JdbcIngestionTrackerTest})
+ * against a real PostgreSQL server, provisioned by Quarkus Dev Services, through
+ * {@link IngestionTrackerResource}.
+ */
+@QuarkusTest
+class Langchain4jIngestionTrackerTest {
```
jamesnetherton, line 33: Every method here shares `pipeline="p"` / `docId="doc"` with no cleanup and no DB reset. It passes today only because each one happens to rewrite the row before asserting — `deleteRowForgetsTheDocument` asserting `hasSize(0)` on pipeline `p` is one added test away from being order-dependent. Unique ids per test, or a truncate in `@BeforeEach`, would remove the trap.

integration-tests/langchain4j-ingestion-tracker/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingestiontracker/it/IngestionTrackerResource.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+@Path("/ingestion-tracker")
+public class IngestionTrackerResource {
+
+    @Inject
+    DataSource dataSource;
+
+    private IngestionTracker tracker;
+
+    @PostConstruct
```
jamesnetherton: DDL in `@PostConstruct` on an unscoped JAX-RS resource. `@ApplicationScoped` with a `StartupEvent` observer would be more predictable.

extensions-support/langchain4j/README.adoc
```diff
@@ -0,0 +1,107 @@
...
+           └─ unpin() ──► done         the source owns the document again
+
+ any state ── deleteRow ──► (no row)   reconciliation: the source no longer lists the
+                                       document (guarded by the consumer's pass
+                                       interlock, never fired by the tracker itself)
+----
+
+`tombstone` and `pinned` are columns orthogonal to `status`: drawn above off `done` because
+that is how the consumer composes them, but they can accompany any state (a tombstone can even
+be recorded for a never-ingested document, as a pure suppression record) and they survive every
```
jamesnetherton, line 80: This claim isn't implemented — see the comment on `setFlag`.

While you're here: worth an operator note that this table is a durable inventory of the corpus. Content correctly never lands in it, but `doc_id` is typically a path, URL or S3 key, and those routinely encode sensitive structure (`/customers/12345/contract.pdf`). Doc ids also go into every exception message, so they reach logs and stack traces.

integration-tests/langchain4j-ingestion-tracker/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-build-parent-it</artifactId>
+        <version>3.39.0-SNAPSHOT</version>
+        <relativePath>../../poms/build-parent-it/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-integration-test-langchain4j-ingestion-tracker</artifactId>
+    <name>Camel Quarkus :: Integration Tests :: LangChain4j Sync Tracker</name>
```
jamesnetherton: Leftover from the `sync-ledger` naming — should be "Ingestion Tracker".

jamesnetherton (changes requested):
Thanks for this — the state machine is well documented, and storing only a content hash rather than the content itself is the right call.

I built and ran the branch locally: the unit tests, the JVM IT against PostgreSQL Dev Services, and the native IT all pass, and `mvn process-resources -Pformat` is a no-op. So it's green — my comments are about the SPI contract rather than the build.

The main issue is that `writeIntent` doesn't hold the shrink bound the README says it does. `intended_count` is overwritten unconditionally on the update path, and `segment_count` only ever moves on `commit`, so a crash between intent and commit followed by a smaller re-intent drops the bound below what may already be in the store:

```
writeIntent(doc, fp1, intended=5)   -> in_progress, segment_count=0, intended=5
  ... consumer writes segments 0..4, process dies before commit ...
writeIntent(doc, fp2, intended=2)   -> intended_count=2
read(doc).maxKnownCount()           -> 2      (expected >= 5)
```

Segments 2-4 are then never swept. That matters beyond bookkeeping: the sweep bound is what the deletion path will use, so a document edited to strip sensitive content — or deleted because access was revoked or a retention request landed — can under-delete while the tracker reports it clean. Worth fixing before a consumer encodes the current behaviour, since the consumer *is* the deletion path.

Two smaller contract gaps in the same family: the flag setters silently do nothing when there's no row (which contradicts the README's "a tombstone can even be recorded for a never-ingested document"), and `commit` on an unknown key writes nothing and throws nothing.

Separately, `ensureSchema()` forces the runtime DB user to hold DDL privileges with no way to opt out. Camel's own `JdbcMessageIdRepository` has the escape hatch worth mirroring — `createTableIfNotExists`, default true, plus a `tableExists` probe — so operators can pre-provision the table and run DML-only.

Details inline.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
jamesnetherton: I know its like an internal thing, but the abbreviation 'cq' might be confusing for users when browsing their DB schemas. Might be better to use a clearer naming:

```suggestion
    static final String TABLE = "camel_quarkus_ingestion_tracker";
```

JiriOndrusek:
Thanks for the thorough review! All points are addressed in c0a0e17:
- `writeIntent` now holds the sweep bound across uncommitted re-intents (dialect-free `CASE`); the dead `committedCount` parameter is gone
- `tombstone` upserts a pure suppression row for never-ingested documents; `commit` without an intent now throws
- `ensureSchema()` gained the `createTableIfNotExists` escape hatch with a `tableExists` probe
- table renamed to `camel_quarkus_ingestion_tracker`, `doc_id` widened to 1024, README notes on `doc_id` sensitivity and pre-provisioning

Each fixed path is pinned by a new test (unit + PostgreSQL IT). Ready for another look.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
JiriOndrusek: Done — dialect-free `CASE` keeps `intended_count` monotone until commit; pinned by `reintentBeforeCommitKeepsTheLargestIntendedCount`.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/IngestionTracker.java
JiriOndrusek: Done — dropped from the SPI signature.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
JiriOndrusek: Done — `tombstone()` now upserts a pure suppression row; pin/unpin/unsuppress documented as row-less no-ops.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
JiriOndrusek: Done — commit without a preceding intent now throws; `refreshFingerprint`'s done-only guard documented and pinned by a test.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
JiriOndrusek: Done — `createTableIfNotExists` flag + `tableExists` probe, mirroring `JdbcMessageIdRepository`; DDL `Statement` in try-with-resources.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
JiriOndrusek: Done — guard comment added on the `TABLE` constant.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
JiriOndrusek: Done — widened to `VARCHAR(1024)`, the S3 object-key maximum.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
JiriOndrusek: Done — the `'source'` origin is now explained in a comment.

extensions-support/langchain4j/runtime/src/test/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTrackerTest.java
JiriOndrusek: Done — added intent→intent, commit-without-intent, markFailed-from-in_progress and refreshFingerprint-on-non-done tests.

integration-tests/langchain4j-ingestion-tracker/src/test/java/org/apache/camel/quarkus/component/langchain4j/ingestiontracker/it/Langchain4jIngestionTrackerTest.java
JiriOndrusek: Done — unique pipeline id per test.

integration-tests/langchain4j-ingestion-tracker/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingestiontracker/it/IngestionTrackerResource.java
JiriOndrusek: Done — `@ApplicationScoped` with a `StartupEvent` observer.

extensions-support/langchain4j/README.adoc
JiriOndrusek: Done — the claim is now implemented, and an operator note on `doc_id` sensitivity was added.

integration-tests/langchain4j-ingestion-tracker/pom.xml
JiriOndrusek: Done — renamed to "Ingestion Tracker".

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
JiriOndrusek: Done — renamed to `camel_quarkus_ingestion_tracker`.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
```diff
@@ -0,0 +1,352 @@
...
+                throw new IllegalStateException(
+                        "The ingestion tracker table '" + TABLE + "' does not exist and table creation is "
+                                + "disabled. Pre-provision the table or enable createTableIfNotExists.");
+            }
+            return;
+        }
+        String ddl = "CREATE TABLE IF NOT EXISTS " + TABLE + " ("
+                + "pipeline VARCHAR(128) NOT NULL, "
+                // 1024: the S3 object-key maximum; file paths and URLs used as ids fit as well
+                + "doc_id VARCHAR(1024) NOT NULL, "
```
jamesnetherton, line 90: `VARCHAR(1024)` inside the primary key is wider than PostgreSQL's btree index can hold for multibyte values — reproduced against `postgres:17.5`:

```
ERROR:  index row size 3088 exceeds btree version 4 maximum 2704 for index "t_pkey"
HINT:  Consider a function index of an MD5 hash of the value, or use full text indexing.
```

ASCII is fine at 1024, and compressible multibyte text passes because PostgreSQL compresses index values, but ~900+ incompressible CJK characters fail at insert. H2 has no such limit, so this can only ever show up on PostgreSQL and only with non-ASCII ids.

Either document the practical byte ceiling (the character count is misleading on its own), or key on `(pipeline, doc_id_hash)` and keep `doc_id` as a plain column — the latter fits a design that already computes hashes, and removes the ceiling entirely.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
```diff
@@ -0,0 +1,352 @@
...
+        } catch (SQLException e) {
+            throw new IllegalStateException(
+                    "Cannot create the ingestion tracker table '" + TABLE + "'. "
+                            + "mode=sync needs a working datasource — configure quarkus.datasource "
+                            + "(Dev Services provides one in dev mode) or use mode=append.",
+                    e);
+        }
+    }
+
+    private boolean tableExists() {
```
jamesnetherton, line 113: `tableExists()` swallows every `SQLException` and returns `false`, so an unreachable host, bad credentials or a permission error all surface as "table does not exist and table creation is disabled" with a null cause — pointing the operator at the wrong problem, and precisely in the least-privilege setup where connection and grant issues are most likely.

Worth letting connection-level failures propagate and only treating "relation does not exist" as `false`. `DatabaseMetaData.getTables()` is the more conventional probe here and distinguishes the two (it's what camel-sql uses).

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
```diff
@@ -0,0 +1,352 @@
...
+                + "segment_count INT DEFAULT 0 NOT NULL, "
+                + "intended_count INT DEFAULT 0 NOT NULL, "
+                + "status VARCHAR(16) NOT NULL, "
+                + "origin VARCHAR(16) DEFAULT 'source' NOT NULL, "
+                + "tombstone BOOLEAN DEFAULT FALSE NOT NULL, "
+                + "pinned BOOLEAN DEFAULT FALSE NOT NULL, "
+                + "updated_at TIMESTAMP, "
+                + "PRIMARY KEY (pipeline, doc_id))";
+        try (Connection connection = dataSource.getConnection();
+                java.sql.Statement statement = connection.createStatement()) {
```
jamesnetherton, line 102: `java.sql.Statement` as an inline FQN, while every other `java.sql` type in the file is imported.

jamesnetherton (approved):
Thanks — that's a thorough pass, and I like that you pushed back with a justifying comment on the `origin = 'source'` point rather than just changing it. The reasoning holds.

I re-ran the original findings against the fixed code directly rather than relying on the new tests, plus the full suites. Everything checks out:

- the sweep bound holds at 5 across descending re-intents, still collapses to 2 on a commit, and rises again on a larger intent
- `tombstone` on a never-ingested document creates the suppression row, and is idempotent on a repeat call
- `commit` without an intent throws, and `refreshFingerprint` correctly leaves a `failed` row's dead-letter gate alone
- pre-provisioned mode refuses a missing table and passes once it exists

Unit tests 16/16, JVM IT 13/13, native IT 13/13, all green.

One thing came out of the `doc_id` fix, and it's on my suggestion rather than yours: `VARCHAR(1024)` inside the primary key exceeds PostgreSQL's btree index limit for multibyte ids. Verified against a real `postgres:17.5`:

```
ERROR:  index row size 3088 exceeds btree version 4 maximum 2704 for index "t_pkey"
HINT:  Consider a function index of an MD5 hash of the value, or use full text indexing.
```

1024 ASCII characters are fine, and compressible multibyte text squeaks through because PostgreSQL compresses index values, but 1024 incompressible CJK characters — a plausible filename or object key — fails at insert time, breaking around ~900 such characters. H2 accepts all of it, so the unit tests can't catch this, and the IT would only catch it with a non-ASCII id. So 512 was mostly under the limit and 1024 is reliably over it for non-ASCII: the declared column width now promises more than the index can actually store.

Not blocking from my side — it's a narrow edge in code nothing consumes yet — but worth either documenting the practical byte ceiling or moving the key to `(pipeline, doc_id_hash)` with `doc_id` as a plain column, which would fit a design that already computes hashes. Approving so this isn't held up; details and two nits inline.

extensions-support/langchain4j/runtime/src/main/java/org/apache/camel/quarkus/component/support/langchain4j/tracker/jdbc/JdbcIngestionTracker.java
jamesnetherton, line 102: `java.sql.Statement` as an inline FQN, while every other `java.sql` type in the file is imported.

PR #9014, opened 2026-08-17
Fixes #9013. Harden the RAG augmentor bridge

### LINKED ISSUE #9013: Harden the RAG augmentor bridge: designated default resolution and retrieval filter   hook
### Describe the feature here

There are two problems in the Quarkus LangChain4j RAG augmentor bridge of  `camel-quarkus-support-langchain4j`:

  1. **Ambiguous default augmentor silently disables RAG.** If more `RetrievalAugmentor`s are configured as `quarkus.camel.langchain4j.rag.augmentors.<name>`. The deafault search from `@RegisterAiService`  fails and no retrieval is registered at all. The build should fail with actionable guidance instead, and  `quarkus.camel.langchain4j.rag.augmentors.<name>.default=true` should mark the designated  default 

  2. **No retrieval-side filter hook.** Segment metadata written at ingestion time (for example  a tenant key) cannot be enforced at query time without replacing the whole augmentor. A  `RagRetrievalFilterSupplier` SPI should plug a per-request `Filter` into every produced  augmentor (`dynamicFilter`), turning such metadata into actual access control.s

JiriOndrusek:
Fixes #9013. Harden the RAG augmentor bridge: designated default and retrieval filter hook
    
 Two augmentors with none marked default used to silently disable RAG for every AI service (ambiguous unqualified CDI lookup); the build now fails with the fix in the message, and quarkus.camel.langchain4j.rag.augmentors.<name>.default=true marks the designated default (the @RagAugmentorName qualifier suppresses the implicit Default on the others, keeping them selectable by name).
    
  RagRetrievalFilterSupplier plugs a per-request retrieval Filter into every produced augmentor (dynamicFilter), turning segment metadata like a tenant key into an actual access control. Being an access control, its ways of failing open are closed: more than one implementation, or a @Dependent one, fail the build, and a supplier no produced augmentor will ever consult is reported as a warning. The one bean serves every augmentor, so it is told which augmentor retrieves and over which embedding store, and it takes the Query, so implementations can key on chatMemoryId instead of thread-local state. All three inputs are covered by the tenant isolation tests in the rag-bridge integration module.

### DIFF
--- a/docs/modules/ROOT/pages/reference/extensions/langchain4j-embeddingstore.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/langchain4j-embeddingstore.adoc
@@ -62,6 +62,79 @@ from("direct:ingest-products")
 
 A store only referenced from routes does not need to be injected anywhere in Java code; it is retained and instantiated lazily on first use. If the Camel registry already resolves a different `EmbeddingStore` under the same name (for example a `@Named` bean), that existing bean keeps winning lookups and a warning is logged at startup.
 
+[id="extensions-langchain4j-embeddingstore-usage-retrieval-augmentors-for-registeraiservice"]
+=== Retrieval augmentors for `@RegisterAiService`
+
+With Quarkus LangChain4j present, a `RetrievalAugmentor` is produced automatically from the `@Default` `EmbeddingStore` and `EmbeddingModel` beans, so an `@RegisterAiService` interface answers from the same store a Camel route ingests into. Nothing is produced when the application declares its own `RetrievalAugmentor`.
+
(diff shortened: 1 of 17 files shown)


jamesnetherton (approved):
Solid PR — the designated-default resolution and retrieval filter hook both look correct. Build-time validation covers every way the filter could silently fail open, the `resolveDesignatedDefault` logic is well-tested, and the `BeanCensus` refactor is a clean single-pass replacement. Test coverage is thorough (unit tests for resolution, integration tests for tenant isolation with both static and query-derived tenants, native IT). No issues found.

PR #9018, opened 2026-08-18
Fixes #9015. Add the langchain4j-ingest extension: declarative doc ingestion

### LINKED ISSUE #9015: Add a langchain4j-ingest extension: declarative document ingestion into an embedding store
Ingesting documents into a vector store for RAG currently means writing the plumbing by hand: a consumer route, splitting, embedding, and storing, repeated in every application. This proposes a Camel Quarkus extension that declares that pipeline instead of coding it.

A pipeline is configured with a directory:

```properties
quarkus.camel.ai.ingest.products.source.directory=/var/data/product-docs
quarkus.camel.ai.ingest.products.embedding-store=products
```

or with any Camel consumer at all, which is the general form — there is no hand-curated list of source types. The Endpoint DSL is the way to reach one, so a connector's options are typed and checked by the compiler:

```java
@Ingest("product-docs")
IngestPipeline productDocs() {
    return IngestPipeline.from(Source.endpoint(aws2S3("product-docs").region("eu-west-1"))
                    .documentId("CamelAwsS3Key"))
            .embeddingStore("products");
}
```

What ingestion needs beyond transport is a stable document id, and where that lives is the consumer's business: `document-id` names the header (`CamelAwsS3Key`, `CamelKafkaKey`) or gives an expression. Every segment written carries `camel_quarkus_pipeline` and `camel_quarkus_document_id`, so retrieval can cite the document an answer came from. Combined with the RAG augmentor bridge (#9013), an `@RegisterAiService` interface then answers from the store a pipeline filled, without glue code.

## Scope of this first increment

- file and endpoint sources, the `@Ingest` builder API with a configuration twin
- build-time validation: unknown component with an add-extension hint, `@Ingest` method shape, splitter bounds, conflicting source declarations
- integration tests covering a directory, a pushed endpoint, a builder-declared pipeline, plus S3 (MinIO) and Kafka against containers

## Deliberately out of scope, to follow separately

- the synchronising engine — this increment writes what it is given and keeps no record, so a document ingested twice leaves two copies and a restart re-reads the directory. Skipping unchanged documents, replacing edited ones and removing deleted ones needs the tracker (#9003) and the sync engine on top of it
- document parsing: content is read as text, so PDF and DOCX must be converted first

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/9015

Declares the RAG ingestion pipeline instead of coding it: documents are read, split, embedded in batches and written to an EmbeddingStore, each segment carrying camel_quarkus_pipeline and camel_quarkus_document_id so retrieval can cite its source. With the RAG bridge (#9013), an @RegisterAiService answers from the store a pipeline filled — no glue code.

Declaration. Properties or Java, as strict twins: @Ingest methods (discovered via Jandex, validated at build time, invoked once at startup) convert into the configuration model, so both paths share one route-building implementation.

Sources. A directory (file consumer with guarded defaults: noop, sized idempotent register, changed read lock, UTF-8), or any of the ~300 Camel consumers — source.uri in properties, Source.endpoint(dsl -> dsl.aws2S3(...)) in Java, where dsl. completes every component's typed options. No hand-curated source types. Identity is the consumer's business: document-id names the header (CamelAwsS3Key, CamelKafkaKey) or gives an expression.

Config split. source.uri is build-time-fixed — a runtime-overridable consumer URI would be arbitrary component invocation. Per-deployment values (directory, document-id, enabled) stay runtime; placeholders in the URI resolve at startup, so credentials do too. Logged URIs are sanitized.

Engine. Deliberately naive and shipped as preview: it writes what it is given and keeps no record, so re-ingestion duplicates and restarts re-read — documented, with the synchronising engine (on tracker #9003) following separately behind the same API.

Validation. Everything decidable at build time fails there with the fix in the message: unknown component scheme (plus the add-extension command), malformed @Ingest methods, splitter bounds, conflicting source declarations — produced as ValidationErrorBuildItem so dev/test mode check too.

Tests. 15 deployment tests lock in every validation message; 6 integration tests cover directory, pushed endpoint, builder pipeline, S3 (MinIO) and Kafka (container), both scoped per class; native CI category added.

Out of scope: synchronisation, document parsing (text only), retry/DLC hooks, automatic pipeline→augmentor pairing.


integration-tests/langchain4j-ingest/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+                    <exclusions>
+                        <exclusion>
+                            <groupId>*</groupId>
+                            <artifactId>*</artifactId>
+                        </exclusion>
+                    </exclusions>
+                </dependency>
+            </dependencies>
+        </profile>
+    </profiles>
```
jamesnetherton, line 208: This module needs a `skip-testcontainers-tests` profile. The `integration-tests-alternative-platform (windows-latest)` job runs `mvn clean verify -Dskip-testcontainers-tests`, and the S3 and Kafka tests here need Docker. 74 other IT modules declare:

```xml
<profile>
    <id>skip-testcontainers-tests</id>
    <activation>
        <property>
            <name>skip-testcontainers-tests</name>
        </property>
    </activation>
    <properties>
        <skipTests>true</skipTests>
    </properties>
</profile>
```

I believe this is why that check is currently failing.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestBuildTimeConfig.java
```diff
@@ -0,0 +1,102 @@
...
+import io.quarkus.runtime.annotations.ConfigRoot;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
+import io.smallrye.config.WithParentName;
+
+/**
+ * The shape of an ingestion pipeline: what it reads and where it writes. Locations that differ
+ * per deployment are runtime configuration, see {@link IngestRunTimeConfig}.
+ */
+@ConfigMapping(prefix = "quarkus.camel.ai.ingest")
```
jamesnetherton: `quarkus.camel.ai.ingest` sits oddly next to `quarkus.camel.langchain4j.rag`, which #9013 introduced a day ago for the same feature family — and next to this extension's own name, `camel-quarkus-langchain4j-ingest`.

Worth picking one namespace now. Config keys are public API and this is much cheaper to change before the preview ships than after.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRoutes.java
```diff
@@ -0,0 +1,276 @@
...
+    @Any
+    Instance<EmbeddingModel> modelCandidates;
+
+    @Override
+    public void configure() {
+        // a pipeline may be declared entirely through runtime properties - the documented
+        // minimum is a directory and nothing else - so the two config roots are unioned. Keying
+        // off the build-time map alone would make that configuration a silent no-op, since
+        // SmallRye only materialises a map key for the mapping whose structure a property matches
+        Set<String> builderDeclared = builderPipelines.entries().stream()
```
jamesnetherton, line 80: The union of the two config roots works, and the comment explains why it is needed. Worth calling out the consequence somewhere user-visible, though: a pipeline declared purely through runtime properties never appears in the build-time map, so it is invisible to every build-time validation — including the duplicate-name check in `discoverBuilderPipelines`, which tests `config.pipelines().containsKey(name)` on the build-time map only.

That case is caught later by `configureBuilderPipeline`, so it is not a hole, but the two paths report it very differently.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRoutes.java
```diff
@@ -0,0 +1,276 @@
...
+    private Expression documentIdExpression(IngestRunTimeConfig.PipelineRunTimeConfig runtime,
+            String defaultHeader) {
+        String configured = runtime == null ? null : runtime.source().documentId().orElse(null);
+        if (configured == null) {
+            return ExpressionBuilder.headerExpression(defaultHeader);
+        }
+        // a bare header name is read as a header directly rather than parsed: a dotted name such
+        // as a dotted header name sends the simple parser into OGNL, and ${...} in a properties file is
+        // MicroProfile Config expansion, which would consume an expression before Camel saw it
+        return configured.contains("${") ? simple(configured) : ExpressionBuilder.headerExpression(configured);
```
jamesnetherton: `simple(configured)` returns a `ValueBuilder` wrapping a `SimpleExpression`, which is an `ExpressionDefinition`. It is never `init(CamelContext)`-ed here, so `ExpressionDefinition.evaluate` reifies it lazily on the first exchange:

```java
public <T> T evaluate(Exchange exchange, Class<T> type) {
    if (expressionValue == null) {
        expressionValue = createExpression(exchange.getContext());
    }
    ...
}
```

`expressionValue` is not volatile, so two concurrent exchanges on the first poll race on it. Calling `expression.init(getContext())` while building the route removes both the race and the first-exchange reification cost.

(The `headerExpression` branch is fine — it degrades to a constant expression, which needs no init.)

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRoutes.java
```diff
@@ -0,0 +1,276 @@
...
+
+    private static String required(String name, String value, String property) {
+        if (value == null) {
+            throw new IllegalStateException("Ingestion pipeline '" + name + "' has no " + property
+                    + ". Set quarkus.camel.ai.ingest." + name + "." + property);
+        }
+        return value;
+    }
+
+    private EmbeddingStore<TextSegment> resolveStore(String name, String configured) {
```
jamesnetherton, line 253: Named lookup goes through the Camel registry here, while the unnamed path in `single()` goes through the CDI `Instance`. The two see different sets: a store bound only via `@BindToRegistry` resolves by name but is invisible to the "exactly one" count, and the reverse holds for a CDI bean that never reaches the registry.

Worth using one mechanism for both.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
```diff
@@ -0,0 +1,224 @@
...
+        for (AnnotationInstance annotation : combinedIndex.getIndex().getAnnotations(ingestAnnotation)) {
+            if (annotation.target().kind() != AnnotationTarget.Kind.METHOD) {
+                continue;
+            }
+            MethodInfo method = annotation.target().asMethod();
+            String name = annotation.value().asString();
+            String location = method.declaringClass().name() + "#" + method.name();
+
+            if (name.isBlank()) {
+                throw new ConfigurationException("@Ingest on " + location + " has a blank pipeline name");
```
jamesnetherton: These `@Ingest` shape checks throw `ConfigurationException` directly, while `validatePipelines` and `validateConnectorsPresent` produce `ValidationErrorBuildItem`.

The PR description says validation is produced as `ValidationErrorBuildItem` so dev and test mode check too — but these are the checks users are most likely to hit, and they take the other path.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
```diff
@@ -0,0 +1,224 @@
...
+                throw new ConfigurationException("@Ingest method " + location + " must return "
+                        + IngestPipeline.class.getSimpleName());
+            }
+            if (!method.parameters().isEmpty()) {
+                throw new ConfigurationException("@Ingest method " + location + " must take no parameters");
+            }
+            // the method is invoked on a CDI bean instance, which a static method would bypass
+            // and a private one would run against the client proxy, seeing null injected fields
+            if (Modifier.isPrivate(method.flags()) || Modifier.isStatic(method.flags())) {
+                throw new ConfigurationException("@Ingest method " + location + " must not be private or static");
```
jamesnetherton: `private` and `static` are rejected because the method is invoked on a client proxy — but a `final` method (or a `final` declaring class) on a normal-scoped bean fails in exactly the same way, and silently: the call is not delegated, so it runs against the proxy with null injected fields.

Worth extending the check so that case gets the same clear message.

Separately, `CDI.current().select(declaringClass).get()` in `IngestBuilderPipelines` creates an instance that is never destroyed if the declaring bean happens to be `@Dependent`.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
```diff
@@ -0,0 +1,224 @@
...
+            }
+            if (!components.contains(scheme)) {
+                // the URI is sanitized: a consumer URI may legitimately carry credentials, and a
+                // build log is no place for them. The artifact hint is a heuristic - multi-scheme
+                // components (smtp -> camel-quarkus-mail) name their extension differently
+                validationErrors.produce(new ValidationErrorBuildItem(new ConfigurationException(
+                        "Ingestion pipeline '" + entry.getKey() + "' consumes from '"
+                                + URISupport.sanitizeUri(uri) + "', but the Camel component '" + scheme
+                                + "' is not on the classpath.\nAdd the extension that provides it, usually:"
+                                + "  ./mvnw quarkus:add-extension -Dextensions=camel-quarkus-" + scheme)));
```
jamesnetherton: Three things about this hint (it also appears in `IngestComponentPresence`):

- `./mvnw` assumes a wrapper in the user's project. Every user-facing page under `docs/modules/ROOT/pages/user-guide/` uses plain `mvn`; `./mvnw` appears only in the contributor guide, i.e. for building Camel Quarkus itself. This is the first extension in the repo to emit such a hint, so it sets the precedent.
- It assumes Maven. Gradle and JBang are both first-class Quarkus paths and would get a command that cannot work.
- `camel-quarkus-<scheme>` is a heuristic, as the comment above acknowledges — `smtp`/`imaps` are `camel-quarkus-mail`, `https` is `camel-quarkus-http`.

Naming the dependency rather than prescribing a command avoids all three, e.g. "Add the extension that provides it, e.g. `org.apache.camel.quarkus:camel-quarkus-<scheme>`".

Also, the `\n` is on the wrong side of the concatenation — this currently renders as `usually:  ./mvnw ...` on one line.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestComponentPresence.java
```diff
@@ -0,0 +1,90 @@
...
+                require(context, entry.name(), definition.sourceUri());
+            }
+        }
+    }
+
+    private static void require(CamelContext context, String name, String uri) {
+        String resolved;
+        try {
+            resolved = context.resolvePropertyPlaceholders(uri);
+        } catch (Exception e) {
```
jamesnetherton, line 80: Swallowing this without a trace means a typo'd `{{...}}` in `source.uri` produces Camel's later error with no indication that the presence check bailed out early. A debug log here would make that much easier to diagnose.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRunTimeConfig.java
```diff
@@ -0,0 +1,85 @@
...
+            /**
+             * Whether subdirectories are ingested too, when reading a directory.
+             */
+            @WithDefault("true")
+            boolean recursive();
+
+            /**
+             * Where the document id lives in the exchange the consumer delivers: normally the
+             * name of a header, such as `CamelAwsS3Key` for an S3 consumer or `CamelKafkaKey` for
+             * a Kafka one. A value containing a dollar-brace placeholder is taken as a
```
jamesnetherton: This documents a syntax the option cannot actually accept: MicroProfile Config expands `${...}` before Camel ever sees the value, as the javadoc goes on to admit. So the simple-expression branch is only reachable from `Source.documentId(...)` in Java, not from a properties file at all.

Either drop it from the config option's contract, or use a marker Config will not expand (`simple:...`, or Camel's `$simple{...}`) and document `\${...}` escaping.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/core/IngestService.java
```diff
@@ -0,0 +1,93 @@
...
+ * Deliberately naive: it writes whatever it is given and remembers nothing, so ingesting a
+ * document twice leaves two copies. Keeping a store in step with a changing source — skipping
+ * unchanged documents, replacing changed ones, removing deleted ones — needs a record of what was
+ * written, and that engine replaces this class.
+ */
+public class IngestService {
+
+    private static final Logger LOG = Logger.getLogger(IngestService.class);
+
+    public static final String METADATA_PIPELINE = "camel_quarkus_pipeline";
```
jamesnetherton, line 45: `camel_quarkus_pipeline` and `camel_quarkus_document_id` are the feature the PR description leads with, and they are not asserted by any test — they appear only here and in `usage.adoc`.

An assertion on the segment metadata would also pin down the `documentId("CamelAwsS3Key")` / `documentId("CamelKafkaKey")` resolution, which is currently only tested to the extent that it produces something non-null (a wrong but non-null id passes).

integration-tests/langchain4j-ingest/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/it/IngestResource.java
```diff
@@ -0,0 +1,116 @@
...
+        case "custom" -> customStore;
+        case "built" -> builtStore;
+        case "s3" -> s3Store;
+        case "events" -> eventsStore;
+        default -> productsStore;
+        };
+        var result = store.search(EmbeddingSearchRequest.builder()
+                .queryEmbedding(model.embed(query).content())
+                .maxResults(20)
+                .minScore(0.0)
```
jamesnetherton, line 102: `DeterministicEmbeddingModel` produces hash-seeded random unit vectors, so a query embedding has no semantic relationship to the document embeddings. With `maxResults(20)` and `minScore(0.0)`, this search effectively returns the whole store.

That is fine as an ingestion assertion, but it means tests named `...AndRetrievable` are really asserting "the text reached the store" — and the hardcoded 20 will silently start dropping matches once a shared store holds more segments than that.

integration-tests/langchain4j-ingest/src/test/java/org/apache/camel/quarkus/component/langchain4j/ingest/it/MinioTestResource.java
```diff
@@ -0,0 +1,84 @@
...
+import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.CreateBucketRequest;
+
+/**
+ * S3-compatible store for the {@code s3docs} pipeline. Enables the pipeline (disabled by
+ * default) and provides the endpoint/credentials as runtime configuration.
+ */
+public class MinioTestResource implements QuarkusTestResourceLifecycleManager {
```
jamesnetherton, line 36: `integration-tests-support/aws2` already provides `Aws2TestResource` (LocalStack), `Aws2Client` and a `DockerSocketAvailable` guard, which every other aws2 integration test uses. Reusing it would keep this module consistent with the rest of the repo, and the Docker guard would have covered the Windows job too.

jamesnetherton (commented):
Thanks for this — it is a lot of carefully thought-through work, and the build-time validation with the fix in the message is a nice touch. Most of what follows is detail; two items are worth settling before merge, and one is a scope question rather than a code issue.

Line-specific findings are left as inline comments. Summarised here:

## Worth settling before merge

**1. Missing `skip-testcontainers-tests` profile — this looks like the red Windows job.** `integration-tests/langchain4j-ingest/pom.xml` does not declare it. `integration-tests-alternative-platform (windows-latest)` runs `mvn clean verify -Dskip-testcontainers-tests`, and 74 other container-using IT modules carry the profile that sets `<skipTests>true</skipTests>`. The MinIO and Kafka tests will run there and fail.

**2. Config namespace.** #9013 shipped `quarkus.camel.langchain4j.rag.*` a day before this; this uses `quarkus.camel.ai.ingest.*`, in an extension named `camel-quarkus-langchain4j-ingest`. That is three names for one feature family. Config keys are public API, so it is much cheaper to align now than after the preview ships.

## Scope question

**3. This introduces a user-facing programming model that does not exist in Apache Camel.** `@Ingest`, `IngestPipeline`, `Source`, `IngestService`, `IngestHeaders.CamelIngestDocumentId` are all Camel Quarkus-only public API. They cannot be used from plain Camel or Camel Spring Boot, and we inherit their compatibility burden from the first release. Given `camel-langchain4j-*` already exists upstream, Camel itself may be the more natural home, with CQ providing the extension over it.

I am not asking for that to be re-plumbed here — but it would be good to have the decision recorded explicitly on #9015 or dev@ rather than arriving with the implementation, because reversing it after a release is expensive.

## Minor

- `runtime/pom.xml` declares `langchain4j` and `langchain4j-core` explicitly; both are already transitive from `camel-quarkus-support-langchain4j`. Nothing in the extension references that support module's API either, so a comment on why the dependency is there would help.
- `IngestResult` becomes the exchange body only on the endpoint path, not the file path. That is deliberate and documented, but the file path also discards an `EMPTY` outcome with no trace.

## Checked and fine

- `EndpointConsumerBuilder.getRawUri()` exists, and `camel-endpointdsl` is a direct dependency of `camel-quarkus-core`, so `Source.endpoint(dsl -> ...)` has a real dependency behind it.
- The doc claim that the duplicate-protection key is only committed on success holds — `GenericFileOnCompletion` removes the idempotent key on failure.
- `noop=true` does not conflict with `readLock=changed` in the file component.
- `ExpressionBuilder.headerExpression("CamelFileName")` degrades to a constant expression, so the bare-header path needs no `init` (only the `simple()` path does — see the inline comment).
- `runtime` is never dereferenced while null in `configureFileSource`; `required()` throws first.
- The FreeMarker `[#if quarkusLangChain4jSupport??]` nesting in `extension-doc-page.adoc` is correct, and the override renders as intended.

---

tooling/scripts/test-categories.yaml
```diff
@@ -143,6 +143,7 @@ group-07:
   - ldap
   - langchain4j-embeddings
   - langchain4j-rag-bridge-ql4j
+  - langchain4j-ingest
```
jamesnetherton, line 145: group-07 already carries `langchain4j-embeddings` and `langchain4j-rag-bridge-ql4j`. This module builds one native image and restarts it three times against MinIO and Kafka containers, so it is worth checking what it does to that group's wall time.

Minor: the entry also breaks the local alphabetical order — it belongs before `langchain4j-rag-bridge-ql4j`.

integration-tests/langchain4j-ingest/pom.xml
JiriOndrusek: Done in 2541fc3a5b — profile added; only the two container-backed test classes are excluded, so the file and endpoint tests keep running on that job (the pom comment explains the deviation).

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestBuildTimeConfig.java
JiriOndrusek: Done — renamed to quarkus.camel.langchain4j.ingest.*, route ids follow

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRoutes.java
JiriOndrusek: Done — the expression is built via ExpressionBuilder.simpleExpression and init()-ed at route build time.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRoutes.java
JiriOndrusek: Done — the startup-only validation of runtime-only pipelines is now called out in usage.adoc.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRoutes.java
JiriOndrusek: Done — the unnamed lookup now goes through Registry.findByType, the same mechanism as the named one.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
JiriOndrusek: Done — the @Ingest shape checks now produce ValidationErrorBuildItems and report together.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
JiriOndrusek: Done — final methods and final declaring classes are rejected at build time (new test), and @Dependent beans are destroyed after the invocation via Instance.getHandle().

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
JiriOndrusek: Done — the message now names  org.apache.camel.quarkus:camel-quarkus-<scheme> and prescribes no command in both checks.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestComponentPresence.java
JiriOndrusek: Done — the skipped check is debug-logged with the pipeline name and the cause.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRunTimeConfig.java
JiriOndrusek: Done — $simple{...} is now detected and documented as the properties form  (exercised end-to-end); ${...} stays for the Java builder.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/core/IngestService.java
JiriOndrusek: Done — the search endpoint returns (text, pipeline, documentId) and all five pipelines assert the metadata, pinning the CamelAwsS3Key/CamelKafkaKey resolution.

integration-tests/langchain4j-ingest/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/it/IngestResource.java
JiriOndrusek: Done — search now returns the whole store, with a comment noting the test  model has no ranking semantics, and the test name no longer claims retrievability.

tooling/scripts/test-categories.yaml
JiriOndrusek: Done — order fixed; native wall time is not measurable locally, so I will watch CI and move the module to a lighter group if it drags

integration-tests/langchain4j-ingest/src/test/java/org/apache/camel/quarkus/component/langchain4j/ingest/it/MinioTestResource.java
JiriOndrusek: Tried it, but it doesn't fit this module: Aws2TestResource is final and module-global by design, and with restrictToAnnotatedClass it fails (classloader cast; a delegating wrapper then hits a SmallRye config ServiceConfigurationError inside start()). Using it unrestricted would start the container for every test class, losing the file-test coverage on the Windows job that the selective excludes keep — so MinIO stays for now. Happy to revisit if the support grows a restrict-friendly mode

JiriOndrusek:
Thanks for the thorough review. Everything is addressed in 2541fc3a5b — kept deliberately as a separate commit so the delta stays reviewable against the original; I'll squash it into the main commit on merge if you prefer a single-commit history.

The two settle-before-merge items:
- `skip-testcontainers-tests` is declared. I excluded only the two container-backed test classes instead of the module-wide `skipTests`, so the file and endpoint tests still run on the alternative-platform job — a pom comment explains the deviation from the convention.
- The config namespace is now `quarkus.camel.langchain4j.ingest.*`, aligned with #9013 and the extension name; route ids follow suit.

The inline findings are applied — short confirmation on each thread — with one decline: reusing the shared aws2 test support. `Aws2TestResource` is final and module-global by design and does not survive `restrictToAnnotatedClass`, which this module needs to keep its Docker-free tests running on Windows; details on that thread.

On the scope question: agreed it deserves an explicit record — I'll follow up on #9015 with the plan. Short version: the engine core is deliberately Quarkus-free, the intent is to propose a `camel-langchain4j-ingest` component upstream and have this extension delegate to it while keeping the declarative surface (properties, `@Ingest`, build-time validation) here; the metadata keys will be parameterized so upstream can use runtime-neutral names.

tooling/scripts/test-categories.yaml
jamesnetherton: If we need to rebalance groups, there's a maven mojo that does it.

I can run it in isolation later at some point. So no further action needed here.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRoutes.java
jamesnetherton: Reopening this one — the change works for the *named* path but breaks the unnamed one.

`getRegistry().findByType(EmbeddingStore.class)` reaches `RuntimeBeanRepository.findByType`, which is `Arc.container().listAll(type)`. That is a raw required type, and CDI treats a parameterized bean type as assignable to a raw one only when its type parameters are unbounded variables or `Object` — so a bean typed `EmbeddingStore<TextSegment>` never matches. On this branch:

```
registry.findByType(EmbeddingStore)     = 0
registry.findByTypeWithName(EmbeddingStore) = []
Arc.listAll(EmbeddingStore)             = 0
Arc.listAll(EmbeddingModel)             = 1     <- not generic, so it does match
```

with `@Named("store")` set and `quarkus.arc.remove-unused-beans=false`, so neither naming nor bean removal is the variable.

Separately, the `@Inject @Any Instance<...>` fields that went away were the injection points keeping unnamed store and model beans unremovable. Even for the non-generic `EmbeddingModel`, an unnamed producer is now removed as unused; it resolves again with `quarkus.arc.remove-unused-beans=false`.

Net effect: a configuration-declared pipeline with exactly one unnamed store and model — the arrangement `usage.adoc` documents — starts at `dd193707` and fails at `2541fc3a` with `needs an embedding store, but no bean of that type exists`.

Going the other way would fix both at once: keep the `Instance` fields and resolve the named case through them, `candidates.select(NamedLiteral.of(configured))`. If the registry is the mechanism you would rather keep, it needs an `UnremovableBeanBuildItem` for both types and a `TypeLiteral<EmbeddingStore<TextSegment>>` lookup for the store.

Worth a test either way — `IngestNoStoreBeanTest` asserts the "no bean of that type" message, so it currently passes whether or not resolution works.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRunTimeConfig.java
jamesnetherton: `$simple{...}` is the right call, and the end-to-end coverage on the `custom` pipeline is good.

One narrowing though: the check is `configured.startsWith("$simple{")`, while Camel's own `LanguageSupport.hasSimpleFunction` uses `contains` for both tokens (`SIMPLE_FUNCTION_START = {"${", "$simple{"}`). So `id-$simple{header.Foo}` matches neither branch here and is silently taken as a header name, even though Camel would happily parse it. The `${` branch already uses `contains`, so the two halves disagree with each other too — `contains("$simple{")` lines both up with Camel.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
jamesnetherton: Thanks — the check and the test do what I was after.

One small over-reach: it rejects `final` on pseudo-scoped beans too. `@Singleton` and `@Dependent` get no client proxy, so a final method there is harmless. Narrowing the check to normal-scoped declaring beans would keep the message for the case that actually breaks. Not a blocker, and erring strict is defensible — just noting the false positive.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
jamesnetherton: The message reads well now, in both checks.

The class javadoc on `IngestComponentPresence` still describes the old form though — "Fails a pipeline whose consumer component is not on the classpath with the same add-extension hint the build gives for configured URIs".

integration-tests/langchain4j-ingest/pom.xml
jamesnetherton: Confirmed locally: `mvn verify -Dskip-testcontainers-tests` on this module runs the four non-container tests, starts no container, and passes.

The surefire-only exclusion is fine as things stand — failsafe is bound only in the `native` profile, and no workflow combines `-Dnative` with `-Dskip-testcontainers-tests`, so the `*IT` classes never run on those jobs. Worth keeping in mind if one ever does, since the 74 modules using `<skipTests>true</skipTests>` would be covered there and this one would not.

jamesnetherton (commented):
Thanks for the fast turnaround — I went through `2541fc3a` by building the extension and running the tests rather than reading the diff alone. Thirteen of the fourteen items check out. One of the fixes introduced a regression that I think has to be settled before merge.

## The blocker

**Moving the unnamed store/model lookup onto the registry broke implicit resolution entirely.** `usage.adoc` still promises that `embedding-store` and `embedding-model` "may be omitted when the application has exactly one of each". I put a single unnamed `EmbeddingStore<TextSegment>` and `EmbeddingModel` producer behind a configuration-declared pipeline:

- at `dd193707` the application starts;
- at `2541fc3a` it fails with `Ingestion pipeline 'docs' needs an embedding store, but no bean of that type exists`.

Two independent causes, both measured against this branch:

1. `getRegistry().findByType(EmbeddingStore.class)` is a **raw**-type lookup, and `RuntimeBeanRepository.findByType` delegates to `Arc.container().listAll(type)`. A parameterized bean type is assignable to a raw required type only when all its type parameters are unbounded type variables or `Object`, so a bean typed `EmbeddingStore<TextSegment>` never matches. Measured on this branch: `Arc.listAll(EmbeddingStore) = 0` and `findByTypeWithName(EmbeddingStore) = []`, even with `@Named("store")` and `quarkus.arc.remove-unused-beans=false`. `EmbeddingModel` is not generic, so it returns `1`.
2. The deleted `@Inject @Any Instance<...>` fields were also the injection points that made unnamed store and model beans unremovable. Without them an unnamed `EmbeddingModel` producer is dropped as unused — it resolves again with `quarkus.arc.remove-unused-beans=false`.

CI does not catch it: every pipeline in the integration test names its store and model, and `IngestResource` injects them; the one deployment test on this path, `IngestNoStoreBeanTest`, asserts the failure message, so it passes for the wrong reason.

The cheapest fix is probably the other direction from the one taken: keep the `Instance` injection points — they handle the parameterized type and the unremovability at once — and move the *named* lookup onto them with `select(NamedLiteral.of(configured))`. If the registry is the preferred mechanism, it needs an `UnremovableBeanBuildItem` for both types plus a `TypeLiteral<EmbeddingStore<TextSegment>>`-based lookup. Either way, a test that resolves a single unnamed store and model would pin it.

## Verified

- The `skip-testcontainers-tests` profile does what the pom comment says. `mvn verify -Dskip-testcontainers-tests` on the module ran the four non-container tests only, started no container, and passed. The surefire-only exclusion is safe today because failsafe is bound solely in the `native` profile and no workflow combines `-Dnative` with `-Dskip-testcontainers-tests`; worth remembering if one ever does.
- The namespace rename is complete — no occurrence of `quarkus.camel.ai.ingest` or `camel-quarkus-ai-ingest` survives, route ids, `quarkus-extension.yaml` and the generated doc page included.
- `ExpressionBuilder.simpleExpression(...)` plus `init(getContext())` is exactly right: that adapter builds its delegate in `init`, so the lazy reification race is gone.
- `$simple{...}` is genuine — Camel's `SIMPLE_FUNCTION_START` is `{"${", "$simple{"}` — and it is now exercised end-to-end by the `custom` pipeline.
- The `@Ingest` shape checks now report together through `ValidationErrorBuildItem`; all 18 deployment tests pass, the new final-method test included.
- The metadata assertions, the `maxResults` change, the artifact hint, the placeholder debug log and the test-categories ordering all land as described. The reasoning for keeping `MinioTestResource` over `Aws2TestResource` is fair enough.

## Minor

Left inline. Also still open from the last round: the scope question about `@Ingest`, `IngestPipeline`, `Source` and `IngestService` being Camel Quarkus-only public API. I am not asking for it to be re-plumbed in this PR, but it would be good to have the decision recorded on #9015 or dev@ before the first release makes it expensive to reverse.

---

JiriOndrusek:
The issue is fixed. 
Regarding the scope question about @Ingest, IngestPipeline, Source and IngestService - only the ingestSergvice is "portable" into Camel (but brings a quite small benefit from doing it), the rest are working only for Camel-quarkus (they configures camel routes under the hood, without needing a knowledge of Camel)

jamesnetherton (approved):

PR #9034, opened 2026-08-20
Fix #2398: Add AWS2 EKS integration test

### LINKED ISSUE #2398: Test AWS 2 EKS

ramu11:
## Description

Adds an AWS2 EKS integration test covering the `camel-aws2-eks` component.

The test:
- Adds a dedicated `aws2-eks` integration test module.
- Covers EKS cluster create, describe, list, and delete operations.
- Uses WireMock mappings to mock AWS EKS API responses.
- Registers the module in the AWS2 integration test group.

Fixes #2398

## Testing

- `mvnw -pl integration-test-groups/aws2/aws2-eks process-resources -Pformat`
- `git diff --check`

Formatting completed successfully with no changes required.

### DIFF
--- a/integration-test-groups/aws2/aws2-eks/pom.xml
+++ b/integration-test-groups/aws2/aws2-eks/pom.xml
@@ -1,0 +1,146 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one or more
+    contributor license agreements.  See the NOTICE file distributed with
+    this work for additional information regarding copyright ownership.
+    The ASF licenses this file to You under the Apache License, Version 2.0
+    (the "License"); you may not use this file except in compliance with
+    the License.  You may obtain a copy of the License at
+
+         http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://www.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.camel.quarkus</groupId>
+        <artifactId>camel-quarkus-build-parent-it</artifactId>
+        <version>3.40.0-SNAPSHOT</version>
+        <relativePath>../../../poms/build-parent-it/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>camel-quarkus-integration-test-aws2-eks</artifactId>
+    <name>Camel Quarkus :: Integration Tests :: AWS 2 EKS</name>
+    <description>Integration tests for Camel Quarkus AWS 2 EKS extension</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
+            <artifactId>camel-quarkus-aws2-eks</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jackson</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>eks</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- test dependencies -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.camel.quarkus</groupId>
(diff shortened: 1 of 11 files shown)


jamesnetherton (commented):
Like most of the other AWS tests, Floci should be used for local testing. WireMock is a 'poor mans' alternative in cases where container based testing is not possible.

ramu11:
Updated the EKS integration test to use Floci, following the existing AWS2 test infrastructure. Removed the WireMock-based test resource and mappings, and added the EKS Aws2TestEnvCustomizer. Focused EKS tests pass successfully with Floci 1.6.0.

ramu11:
Could you please take another look and merge if everything looks good?

jamesnetherton (commented):
The Floci wiring looks right — `FLOCI_SERVICES_EKS_MOCK=true` is the correct choice, since Floci's default EKS mode starts a `rancher/k3s` container per cluster and `Aws2TestResource` only binds the Docker socket for `Service.LAMBDA`.

My main concern is the **"Running against real AWS"** section of the new README: as written, that flow cannot succeed, so we'd be documenting something that fails on the first call.

### 1. `CreateCluster` cannot work against real AWS

Two blockers in `Aws2EksResource`:

* `TEST_ROLE_ARN` is hardcoded to `arn:aws:iam::123456789012:role/cq-eks-test-role`. That's the AWS documentation placeholder account — real `CreateCluster` validates the role exists in the caller's account, so this fails with `InvalidParameterException` / `AccessDeniedException`.
* `VpcConfigRequest.builder().build()` is empty. `resourcesVpcConfig` is required by the real EKS API and needs at least two subnets in different Availability Zones.

Neither is overridable, so even someone with a properly configured EKS cluster role can't run this.

### 2. Cluster creation is asynchronous on real AWS

Even with the above fixed, `CreateCluster` returns immediately with status `CREATING` and takes ~10–15 minutes to reach `ACTIVE`. The `finally` block deletes right away, and `DeleteCluster` on a `CREATING` cluster is rejected with `ResourceInUseException` — so the test fails *and* leaks a control plane billed at ~$0.10/hr.

This passes against Floci only because mock mode transitions the cluster [directly to `ACTIVE` on creation](https://github.com/floci-io/floci/blob/main/docs/services/eks.md) and doesn't validate the role or subnets.

### Suggested approach — follow `aws2-msk`

`aws2-msk` is a close analogue (subnets required, slow async create/delete) and already solves all of this:

1. **Configurable prerequisites with mock-friendly defaults** — mirror `aws2-msk/src/main/resources/application.properties`:
   ```properties
   aws.eks.role.arn=${AWS_EKS_ROLE_ARN:arn:aws:iam::123456789012:role/cq-eks-test-role}
   aws.eks.subnets=${AWS_EKS_SUBNETS:subnet-1,subnet-2}
   ```
   Inject them with `@ConfigProperty` instead of the `TEST_ROLE_ARN` constant, and pass the subnets into `VpcConfigRequest`. Floci doesn't validate these, so the defaults keep the mock run unchanged.

2. **Wait for state transitions** — see `Aws2MskTest`: ordered test methods with Awaitility, waiting for `ACTIVE` before describe/delete, and a final check that the cluster has dropped out of `ListClusters` (deletion is async too). Using `.pollDelay(0, TimeUnit.SECONDS)` keeps this instant against the mock, so one code path covers both backends and CI is unaffected. This needs `org.awaitility:awaitility` as a test dependency (see `aws2-msk/pom.xml`) and `describeCluster` to also return `cluster.status()`.

3. **Document the prerequisites** — add a "Prerequisites for real AWS environment" section like MSK's, covering `AWS_EKS_ROLE_ARN` (a role with `AmazonEKSClusterPolicy` attached) and `AWS_EKS_SUBNETS`. Please also add `iam:PassRole` to the IAM permissions list — `CreateCluster` fails without it.

### Minor

* `hasItem(CLUSTER_NAME)` combined with `maxResults=10` is order-dependent against a real account with more than 10 clusters. Consider asserting membership against the un-paginated list and using the `maxResults=10` call only to assert the result size.
* The PR description says "Uses WireMock mappings to mock AWS EKS API responses", but there are no mappings in the diff and the module uses Floci. Worth updating so the description matches the implementation.

ramu11:
I've updated the AWS2 EKS integration test to support real AWS EKS execution by making the IAM role and subnet configuration configurable, while keeping the existing Floci/mock defaults.

The test now also handles the asynchronous EKS cluster lifecycle by waiting for the cluster to become `ACTIVE` before further operations and waiting for deletion to complete.

The changes have been amended into the existing commit and pushed to the PR. Local Floci validation passes with `BUILD SUCCESS` (4/4 tests passed).

Could you please review the latest changes and let me know if anything else needs to be addressed?

integration-test-groups/aws2/aws2-eks/pom.xml
```diff
@@ -17,51 +17,37 @@
...
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://www.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.apache.camel.quarkus</groupId>
         <artifactId>camel-quarkus-build-parent-it</artifactId>
-        <version>3.40.0-SNAPSHOT</version>
-        <relativePath>../../poms/build-parent-it/pom.xml</relativePath>
+        <version>3.39.0-SNAPSHOT</version>
```
jamesnetherton: The project version got bumped recently:

```suggestion
        <version>3.40.0-SNAPSHOT</version>
```

ramu11:
Updated the AWS2 EKS integration test based on the review feedback, including the real AWS configuration, asynchronous cluster lifecycle handling, prerequisites documentation, and the `3.40.0-SNAPSHOT` version update.

 The changes are amended into the existing commit and pushed to the PR. Local validation passes with `BUILD SUCCESS`.

 Please take a look at the latest changes when you get a chance. If everything looks good, the PR is ready to merge.

jamesnetherton (approved):

PR #9040, opened 2026-08-21
Fixes #9039. Add a configurable idempotent repo. to langchain4j-ingest

### LINKED ISSUE #9039: langchain4j-ingest: configurable idempotent repository for ingestion pipelines
The langchain4j-ingest extension currently forgets what it has ingested: every application restart re-reads and re-embeds the whole source, wasting embedding calls and piling duplicate content into the store. Pipelines fed by a component (Kafka, S3, ...) have no duplicate protection at all - a redelivered message is ingested again.

Until the full synchronising engine arrives, a pipeline should be able to remember what it already ingested - across restarts and, where the backing store allows it, across instances -  and skip duplicate deliveries. The register backing this should be pluggable, following what  Camel users already know: the idempotent repository.

JiriOndrusek:
fixes https://github.com/apache/camel-quarkus/issues/9039

Adds a configurable idempotent repository ("duplicate register") to camel-quarkus-langchain4j-ingest — the interim tier before the synchronising engine: remember what was ingested across restarts/instances and skip duplicate deliveries.

  Configuration surface
  - quarkus.camel.langchain4j.ingest.<pipeline>.source.idempotent-repository (runtime) — names an IdempotentRepository bean replacing the built-in in-memory register (100 000 keys,  per-boot).
  - …source.idempotent-repository-auto-create=true — binds an in-memory register under that name unless a bean with the name exists (existing bean wins).
  - Builder twins on Source: .idempotentRepository(String), .idempotentRepositoryAutoCreate(boolean); the external source.* override guard covers both.

  Mechanics
  - Three bean definition styles, all resolved the same way: auto-create, camel.beans.* (mind Camel's camelCase normalisation of dashed keys), CDI producer. Resolution is eager, by name, at route build — a missing bean fails startup on both source kinds; CamelContextAware gets the context injected.
  - Directory source: consumer-level idempotentKey = ${file:absolute.path}:${file:modified}:${file:size} — unchanged files skipped, edited files re-ingest (append; old segments remain).
  - Consumer-fed source: idempotentConsumer EIP (eager, removeOnFailure) keyed on the validated document id held in an exchange property (expression evaluated once); duplicates answered with the new IngestResult.Outcome.SKIPPED via a tail processor keyed on Exchange.DUPLICATE_MESSAGE. First write wins per id.

  Native
  ReflectiveClassBuildItem registers MemoryIdempotentRepository and FileIdempotentRepository  (#class: beans are instantiated reflectively); custom repository classes need their own registration (documented). Verified with a native run.

  Docs & tests
  usage.adoc gains a "duplicate register" chapter (definition styles, sizing — FileIdempotentRepository correctness is bounded by maxFileStoreSize, not cacheSize — no off-switch by design, first-write-wins caveat, link to the Idempotent Consumer EIP guide). 4  deployment tests (missing bean on both source kinds, auto-create without a name, builder
  override) and 5 integration tests incl. failed-delivery key release and an H2-backed JdbcMessageIdRepository via Quarkus Dev Services (no container).

  🤖 Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
+++ b/extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
@@ -74,6 +74,29 @@ class Langchain4jIngestProcessor {
     }
 
     /**
+     * {@code #class:} beans are instantiated reflectively, which native mode allows only for
+     * registered classes: the two camel-core repositories the documentation recommends, plus
+     * every {@code IdempotentRepository} implementation the Jandex index knows — application
+     * classes always, third-party ones when their jar carries an index.
+     */
+    @BuildStep
+    void repositoryReflection(CombinedIndexBuildItem combinedIndex,
+            BuildProducer<ReflectiveClassBuildItem> reflectiveClasses) {
+        Set<String> repositories = new HashSet<>(Set.of(
+                "org.apache.camel.support.processor.idempotent.MemoryIdempotentRepository",
+                "org.apache.camel.support.processor.idempotent.FileIdempotentRepository"));
+        combinedIndex.getIndex()
+                .getAllKnownImplementations(DotName.createSimple("org.apache.camel.spi.IdempotentRepository"))
+                .stream()
+                .filter(repository -> !Modifier.isAbstract(repository.flags()))
+                .map(repository -> repository.name().toString())
+                .forEach(repositories::add);
+        reflectiveClasses.produce(ReflectiveClassBuildItem.builder(repositories.toArray(new String[0]))
+                .methods()
+                .build());
+    }
+
+    /**
      * Discovers {@code @Ingest} builder methods: validated here (return type, no parameters,
      * unique names, no collision with configuration-declared pipelines), invoked reflectively once
      * at startup. Violations are reported as {@link ValidationErrorBuildItem}s — the channel every
--- a/extensions/langchain4j-ingest/deployment/src/test/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/IngestAutoCreateWithoutNameTest.java
+++ b/extensions/langchain4j-ingest/deployment/src/test/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/IngestAutoCreateWithoutNameTest.java
@@ -17,31 +17,27 @@
 package org.apache.camel.quarkus.component.langchain4j.ingest.deployment;
 
 import io.quarkus.test.QuarkusExtensionTest;
-import jakarta.enterprise.context.ApplicationScoped;
-import org.apache.camel.quarkus.component.langchain4j.ingest.Ingest;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
-/** An {@code @Ingest} method that does not return {@code IngestPipeline} fails the build. */
-class IngestBuilderMethodShapeTest {
+/** Auto-create without a repository name to create it under fails the start. */
+class IngestAutoCreateWithoutNameTest {
 
     @RegisterExtension
     static final QuarkusExtensionTest CONFIG = new QuarkusExtensionTest()
-            .withApplicationRoot(jar -> jar.addClasses(Pipelines.class))
-            .assertException(t -> ValidationTestSupport.assertFailure(t, "must return IngestPipeline"));
+            .withApplicationRoot(jar -> jar.addClasses(TestEmbeddingBeans.class))
+            .overrideConfigKey("quarkus.camel.langchain4j.ingest.docs.embedding-store", "store")
+            .overrideConfigKey("quarkus.camel.langchain4j.ingest.docs.embedding-model", "model")
+            .overrideConfigKey("quarkus.camel.langchain4j.ingest.docs.source.directory", "target/auto-docs")
+            .overrideConfigKey("quarkus.camel.langchain4j.ingest.docs.source.idempotent-repository-auto-create",
+                    "true")
+            .assertException(t -> ValidationTestSupport.assertFailure(t,
+                    "sets source.idempotent-repository-auto-create",
+                    "no source.idempotent-repository name"));
 
     @Test
-    void buildMustFail() {
-        Assertions.fail("The build was expected to fail");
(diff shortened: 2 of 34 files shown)


extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRoutes.java
```diff
@@ -204,20 +220,49 @@ private void configureFileSource(String name, IngestRunTimeConfig.PipelineRunTim
...
+                    });
+            return;
+        }
+        // duplicates skip the block and the tail processor answers SKIPPED; first write wins
+        // per id. The EIP keys on the validated id property, evaluating the expression once
         from(uri)
                 .routeId(routeId(name))
+                .process(exchange -> exchange.setProperty(DOCUMENT_ID_PROPERTY,
+                        requireDocumentId(name, documentId, exchange)))
+                .idempotentConsumer(exchangeProperty(DOCUMENT_ID_PROPERTY),
```
jamesnetherton: Question on the eager `idempotentConsumer` + first-write-wins interaction: the key is added eagerly and `removeOnFailure` only releases it on an *exception*. An `EMPTY`-outcome ingest (blank body) completes normally, so the id stays claimed — a later non-blank document delivered under the **same id** is then answered `SKIPPED` and never ingested.

This is consistent with the documented "first write wins per id", but here the first write wrote nothing. If a source can legitimately deliver an empty-then-populated document under a stable id, that content is lost silently. Is that intended? If not, releasing the key on `EMPTY` (or not registering it) would fix it. Either way, a one-line note in `usage.adoc` that an empty first delivery still claims the id would help users.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
```diff
@@ -73,6 +73,29 @@ AdditionalBeanBuildItem beans() {
...
     }
 
+    /**
+     * {@code #class:} beans are instantiated reflectively, which native mode allows only for
+     * registered classes: the two camel-core repositories the documentation recommends, plus
+     * every {@code IdempotentRepository} implementation the Jandex index knows — application
+     * classes always, third-party ones when their jar carries an index.
+     */
+    @BuildStep
+    void repositoryReflection(CombinedIndexBuildItem combinedIndex,
```
jamesnetherton, line 83: Just flagging for awareness (you note native was verified, so likely fine): `#class:` instantiation needs the no-arg constructor plus the property setters. `.methods()` covers the setters and the builder registers constructors by default, so this should be complete — correctness rests on those two facts holding in the pinned Quarkus version, and on third-party repos shipping a Jandex index (which the docs correctly call out). No change requested.

jamesnetherton (commented):
Thanks for this — a genuinely well-scoped, single-issue feature with unusually thorough docs and tests (deployment unit tests for the failure paths, plus an H2-backed `JdbcMessageIdRepository` via Dev Services and a failed-delivery key-release case). Conventions all check out: commit format `Fixes #9039. …`, single focused commit, `@QuarkusTest`↔`@QuarkusIntegrationTest` pairing present, and the generated `docs/modules/.../langchain4j-ingest.adoc` is regenerated and in sync. No new root-pom dependency versions; test-only deps confined to the IT module. No blocking issues — just two questions/edge-cases left inline, neither of which needs to hold up the merge.

---
Scope note: this is a conventions + diff review. It does not replace CodeRabbit, Sourcery, or SonarCloud, nor a full JVM/native test run.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
JiriOndrusek: Yes that is correct, the doc already contains that information:
```
For native mode the extension registers the camel-core repositories shown here and every `IdempotentRepository` implementation found in the Jandex index — application classes always, component-provided ones through their jar's index. A class from a jar without an index needs `@RegisterForReflection(targets = ...)` or a `quarkus.index-dependency.*` entry.
```

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestRoutes.java
JiriOndrusek: Thank you for spotting this! I fixed it.

extensions/langchain4j-ingest/deployment/src/test/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/IngestAutoCreateWithoutNameTest.java
```diff
@@ -0,0 +1,43 @@
...
+import io.quarkus.test.QuarkusUnitTest;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+/** Auto-create without a repository name to create it under fails the start. */
+class IngestAutoCreateWithoutNameTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest CONFIG = new QuarkusUnitTest()
```
apupier: Is it possible to use non-deprecated QuarkusExtensionTest class as mentioned in javadoc? https://github.com/quarkusio/quarkus/blob/2e48ca4985ebcc4696404cec3c040ed2043f8bca/test-framework/junit-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java#L17

apupier (approved):
non-blocking comment.

i do nto understand fully the related codebase but t was already approved by James

extensions/langchain4j-ingest/deployment/src/test/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/IngestAutoCreateWithoutNameTest.java
JiriOndrusek: Thanks for the  review, I changed all QuarkusUnitTests for io.quarkus.test.QuarkusExtensionTest.

PR #9064, opened 2026-08-26
Fixes #9061. Verify release signatures before publishing checksums

### LINKED ISSUE #9061: Release staging script publishes checksums for unverified downloads
`release-utils/scripts/upload-source.sh`

The script downloads the release artifacts and their detached signatures, then generates the checksums that get published from those downloads:

```sh
wget ${sourcesUrl}/camel-quarkus-${version}-src.zip -O apache-camel-quarkus-${version}-src.zip
wget ${sourcesUrl}/camel-quarkus-${version}-src.zip.asc -O apache-camel-quarkus-${version}-src.zip.asc
sha512sum -b apache-camel-quarkus-${version}-src.zip > apache-camel-quarkus-${version}-src.zip.sha512
```

The same pattern repeats for both CycloneDX SBOMs, and the directory is then `svn import`ed to `dist.apache.org/repos/dist/dev`. `gpg --verify` is never invoked, so the `.asc` files are copied alongside the artifacts without ever being checked, and the published `.sha512` files describe whatever was downloaded rather than something verified.

Separately, the staging probe on line 31 disables certificate validation:

```sh
if [[ "$(curl -k -L -s -o /dev/null -w "%{http_code}" ${sourcesUrl})" != "200" ]]; then
```

Only a status code is consumed there, so the impact of `-k` alone is minimal, but it is unnecessary.

**Suggested fix**

Run `gpg --verify` on each `.asc` against the checked-in `KEYS` file before generating any checksum or importing to `dist`, and drop `-k` from the probe.

oscerd:
Fixes #9061.

`release-utils/scripts/upload-source.sh` downloaded each artifact together with its detached signature, then generated the `.sha512` files that get published directly from those downloads and `svn import`ed the directory to `dist.apache.org/repos/dist/dev`. `gpg --verify` was never invoked, so the `.asc` files were copied along unchecked and the published checksums described whatever had been downloaded rather than something verified.

**Changes**

- Import the in-repo `KEYS` into a throwaway `--homedir` (via `mktemp -d`, removed by an `EXIT` trap) so verification is answered by the project release keys rather than by whatever is in the release manager's own keyring.
- Add a `fetch_verify_checksum` helper that downloads the artifact and its `.asc`, verifies the signature, and only then writes the `.sha512`. This collapses the three near-identical blocks into three calls.
- Drop `-k` from the staging probe. Only a status code is consumed there so the impact was minimal, but disabling certificate validation bought nothing.

`set -e` is already in force, so a failed verification aborts the release rather than publishing an unverified artifact.

**Verification**

- `bash -n` passes.
- `KEYS` imports cleanly into a throwaway keyring (25 public keys).
- `gpg --verify` against a bogus signature exits non-zero, confirming `set -e` will abort.
- The `location` used to resolve `KEYS` is captured before the script `cd`s into the version directory, so the relative path still resolves.

### DIFF
--- a/release-utils/scripts/upload-source.sh
+++ b/release-utils/scripts/upload-source.sh
@@ -28,25 +28,38 @@ version=$1
 stagingRepoId=$2
 sourcesUrl=https://repository.apache.org/content/repositories/orgapachecamel-${stagingRepoId}/org/apache/camel/quarkus/camel-quarkus/${version}
 
-if [[ "$(curl -k -L -s -o /dev/null -w "%{http_code}" ${sourcesUrl})" != "200" ]]; then
+if [[ "$(curl -L -s -o /dev/null -w "%{http_code}" ${sourcesUrl})" != "200" ]]; then
   echo "Failed to access ${sourcesUrl}. Is the ${version} staging repository closed?"
   exit 1
 fi
 
+# Import the release keys into a throwaway keyring, so the verification below is
+# answered by the project KEYS file rather than by whatever the release manager
+# happens to have in their own keyring
+gpgHome=$(mktemp -d)
+chmod 700 ${gpgHome}
+trap 'rm -rf ${gpgHome}' EXIT
+gpg --homedir ${gpgHome} --quiet --import ${location}/../../KEYS
+
+# Download an artifact with its detached signature, verify the signature, and only
+# then generate the checksum that gets published alongside it. set -e aborts the
+# release if any verification fails.
+fetch_verify_checksum() {
+  remoteName=$1
+  localName=$2
+
+  wget ${sourcesUrl}/${remoteName} -O ${localName}
+  wget ${sourcesUrl}/${remoteName}.asc -O ${localName}.asc
+  gpg --homedir ${gpgHome} --verify ${localName}.asc ${localName}
+  sha512sum -b ${localName} > ${localName}.sha512
+}
+
 mkdir ${version}/
 cd ${version}/
 
-wget ${sourcesUrl}/camel-quarkus-${version}-src.zip -O apache-camel-quarkus-${version}-src.zip
-wget ${sourcesUrl}/camel-quarkus-${version}-src.zip.asc -O apache-camel-quarkus-${version}-src.zip.asc
-sha512sum -b apache-camel-quarkus-${version}-src.zip > apache-camel-quarkus-${version}-src.zip.sha512
-
-wget ${sourcesUrl}/camel-quarkus-${version}-cyclonedx.json -O apache-camel-quarkus-${version}-sbom.json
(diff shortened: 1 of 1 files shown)


jamesnetherton (approved):
Nice improvement — worth noting this takes the script from no signature verification at all to actually verifying against our `KEYS`, and I checked the three `fetch_verify_checksum` call sites expand to exactly the same URLs and filenames as the blocks they replace. Two non-blocking notes:

**`gpg --verify` won't reject revoked or expired keys.** It exits 0 for both, printing only a warning (`WARNING: This key has been revoked by its owner!` / `Note: This key has expired!`). That matters here because `KEYS` legitimately contains expired keys — currently `08356B51E31AF533`, `1572E5CE94CB9113` and `DD7050A76B400F69` — since ASF policy says we never remove keys that signed past releases. So the realistic gap isn't a forged signature, it's an RM whose key quietly expired mid-cycle sailing through. If you want to close it: `--status-fd 1` and reject `EXPKEYSIG`/`REVKEYSIG`. Otherwise a note in the comment above the verify would do.

**A failed verification leaves `${version}/` behind.** Only `gpgHome` has an EXIT trap, so if verify fails on the second or third artifact, the re-run dies at `mkdir ${version}/` with `File exists` and stale partial downloads sitting in the directory. Either extend the trap or fail with an explicit "remove `${version}/` and re-run".

oscerd:
Thanks @jamesnetherton — both notes are right, and the second one reproduces exactly as you described.

**Leftover `${version}/` — fixed.** I stubbed out `curl`/`wget`/`gpg`/`svn` and forced the verification to fail on the *second* artifact, i.e. after the script has already `cd`'d into the staging directory. Against the previous revision the directory is left behind and the re-run dies at `mkdir` with `File exists`; with this change it is cleared and the re-run gets past `mkdir`. The trap now covers `stagingDir` as well as `gpgHome`, captured as an absolute path before the `cd` so it still resolves from inside the directory it is removing. The explicit `rm -rf ${version}/` at the end is gone, since the trap covers the success path too.

**Revoked and expired keys — documented, not enforced.** You are right that `gpg --verify` exits 0 for both and only warns, and that `KEYS` holds expired keys by design because ASF keeps whatever signed past releases. So the realistic gap really is an RM whose key lapsed mid-cycle rather than a forged signature. I have taken the option you offered and put a note above `fetch_verify_checksum` recording the gap and naming the fix (`--status-fd 1`, reject `EXPKEYSIG`/`REVKEYSIG`), rather than folding the enforcement into this PR — it changes which releases pass, so it seemed better as its own change. Happy to do it here instead if you would rather.

Also confirmed your reading of the three `fetch_verify_checksum` call sites: they expand to the same URLs and filenames as the blocks they replaced.

jamesnetherton (approved):
Thanks!

PR #9065, opened 2026-08-26
Fixes #9051. Resolve camel.debug.enabled instead of testing for key presence

### LINKED ISSUE #9051: Debug enablement gates test for property presence, not value
`CamelDebugProcessor` and `DebugProcessor` both decide whether debug support is active by testing for the *presence* of a property rather than its value.

**`extensions-core/core/.../CamelDebugProcessor.java:40-45`**

```java
return StreamSupport.stream(ConfigProvider.getConfig().getPropertyNames().spliterator(), false)
        .anyMatch(key -> key.startsWith("camel.debug"));
```

The whole class is `@BuildSteps(onlyIf = CamelDebugConfigurationPresent.class)` and unconditionally produces `AllowJNDIBuildItem`. Setting `camel.debug.enabled=false` — a natural way to pin debugging off for a production build — still matches, so JNDI support is still allowed in.

**`extensions/debug/.../DebugProcessor.java:70-79`**

`DebugEnabled` additionally reads the bare `camel.debug.enabled` property through `ConfigProvider` at augmentation time. That maps to the `CAMEL_DEBUG_ENABLED` environment variable, so an ambient value in the build environment enables the debugger and produces `NativeMonitoringBuildItem(JMXSERVER/JMXCLIENT)` without ever appearing in the typed `quarkus.camel.*` config that the build report records.

**Suggested fix**

- Resolve the boolean value in `CamelDebugConfigurationPresent.getAsBoolean()` instead of testing key presence.
- Gate `DebugEnabled` on the typed `quarkus.camel.*` build-time config so enablement is explicit and auditable.

oscerd:
Fixes #9051.

`CamelDebugProcessor.CamelDebugConfigurationPresent` decided whether the class's build steps ran by scanning for *any* property name starting with `camel.debug`:

```java
return StreamSupport.stream(ConfigProvider.getConfig().getPropertyNames().spliterator(), false)
        .anyMatch(key -> key.startsWith("camel.debug"));
```

Because that tests presence rather than value, `camel.debug.enabled=false` still matched and still produced `AllowJNDIBuildItem`. So did any unrelated `camel.debug.*` key. The supplier now resolves the boolean, and is renamed `CamelDebugEnabled` to match what it actually tests.

The value lookup lives in a package-private `isDebugEnabled(Config)`; `getAsBoolean()` passes `ConfigProvider.getConfig()` into it. That keeps runtime behaviour identical while letting the test supply its own `Config` without touching the one registered for the class loader.

**Scope — `DebugProcessor` is deliberately unchanged.**

The issue also raised `DebugProcessor.DebugEnabled` reading the bare `camel.debug.enabled` property through `ConfigProvider`. That read is intentional and must stay: it is the camel-main way to enable debugging, it has a dedicated test (`DebugEnabledFromCamelMainTest`), and `integration-tests/main` and `integration-tests/management` both rely on it. `ManagementProcessor` reads the same property. Removing it would break all of those.

**Tests**

`CamelDebugEnabledTest` covers the supplier directly. Each case builds its own `SmallRyeConfig`, so nothing here disturbs the configuration the rest of the JVM shares.

| Case | Expected |
|---|---|
| unconfigured | not enabled |
| `camel.debug.enabled=true` | enabled |
| `camel.debug.enabled=false` | not enabled |
| unrelated `camel.debug.*` key only | not enabled |
| `camel.debug.enabled=false` + another `camel.debug.*` key | not enabled |

Verified as a real regression test: 3 of the 5 fail against the previous presence-scan implementation with `expected: <false> but was: <true>`; all 5 pass with this change.

The full `camel-quarkus-core-deployment` suite passes — 78 tests, 0 failures, 0 errors — as does `./mvnw clean install -DskipTests` from the repository root, with no regenerated artifacts left uncommitted.

> An earlier revision of this PR drove the test through system properties and released the class loader's config after each case. That broke `CamelDevModeProfileTest` and `CamelDevModeSingletonBeanTest` with `SRCFG00015: No configuration is available for this class loader`, since surefire shares a JVM. The current revision holds no global state.

### DIFF
--- a/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelDebugProcessor.java
+++ b/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelDebugProcessor.java
@@ -21,6 +21,7 @@ import java.util.function.BooleanSupplier;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.BuildSteps;
 import io.quarkus.deployment.builditem.AllowJNDIBuildItem;
+import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigProvider;
 
 /**
@@ -38,11 +39,18 @@ public class CamelDebugProcessor {
     static final class CamelDebugEnabled implements BooleanSupplier {
         @Override
         public boolean getAsBoolean() {
-            // Resolve the value rather than testing whether any camel.debug.* key is present, so that
-            // pinning debugging off with camel.debug.enabled=false does not still allow JNDI
-            return ConfigProvider.getConfig()
-                    .getOptionalValue("camel.debug.enabled", boolean.class)
-                    .orElse(false);
+            return isDebugEnabled(ConfigProvider.getConfig());
+        }
+
+        /**
+         * Resolves the value rather than testing whether any {@code camel.debug.*} key is present, so that pinning
+         * debugging off with {@code camel.debug.enabled=false} does not still allow JNDI.
+         *
+         * Takes the {@link Config} so that it can be exercised without touching the configuration registered for the
+         * class loader, which is shared with every other test in the JVM.
+         */
+        static boolean isDebugEnabled(Config config) {
+            return config.getOptionalValue("camel.debug.enabled", boolean.class).orElse(false);
         }
     }
 }
--- a/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelDebugProcessor.java
+++ b/extensions-core/core/deployment/src/main/java/org/apache/camel/quarkus/core/deployment/CamelDebugProcessor.java
@@ -17,7 +17,6 @@
 package org.apache.camel.quarkus.core.deployment;
 
 import java.util.function.BooleanSupplier;
-import java.util.stream.StreamSupport;
 
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.BuildSteps;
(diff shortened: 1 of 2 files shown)


oscerd:
`functional-extension-tests` was failing on this PR and it was my fault, not a flake.

`CamelDebugEnabledTest` drove the supplier through system properties and released the config registered for the class loader after each case, so that SmallRye would pick the new values up. Surefire shares one JVM, so that left everything after it without a configuration:

```
java.lang.IllegalArgumentException: SRCFG00015: No configuration is available for this class loader
	at io.quarkus.test.QuarkusDevModeTest.beforeEach(QuarkusDevModeTest.java:302)
```

`CamelDevModeProfileTest` and `CamelDevModeSingletonBeanTest` both died on it. My own five assertions passed throughout, and running the class on its own passed too, which is exactly how it got through review locally.

Fixed by extracting `isDebugEnabled(Config)` from the supplier. `getAsBoolean()` still passes `ConfigProvider.getConfig()` into it, so runtime behaviour is unchanged, but each test case now builds its own `SmallRyeConfig` and touches no global state.

The whole `camel-quarkus-core-deployment` suite passes locally now — 78 tests, 0 failures, 0 errors, against `Errors: 2` before. `./mvnw clean install -DskipTests` from the root is also green with no regenerated artifacts left behind.

The test is still a genuine regression test, and hermetic now: 3 of the 5 cases fail against the previous presence-scan implementation. It used to be 4, because `notEnabledWhenUnconfigured` only failed thanks to `camel.debug.*` properties leaking in from sibling cases — the very problem this removes.

apupier (approved):

jamesnetherton (approved):

PR #9069, opened 2026-08-26
Fixes #9052. Do not enable the CLI connector in prod mode by default

### LINKED ISSUE #9052: cli-connector is enabled by default outside dev and test
`extensions-jvm/cli-connector/.../CamelCliConnectorConfig.java:32`

```java
@WithDefault("true")
boolean enabled();
```

`CliConnectorProcessor` applies no launch-mode condition — the only gate is this flag — and `CamelCliConnectorRecorder` unconditionally calls `factory.setEnabled(true)`. Only `sigterm()` handling is launch-mode aware. The connector's action-file IPC is therefore active in production builds.

**Suggested fix**

Gate it the way the console extension is gated — dev/test by default with an explicit opt-in for production — or flip the default to `false` outside dev/test.

oscerd:
Fixes #9052.

`CamelCliConnectorConfig.enabled()` carries `@WithDefault("true")` and `CliConnectorProcessor` applied no launch mode condition — the only gate was that flag — so the connector's action file IPC was active in production builds. `CamelCliConnectorRecorder` then calls `factory.setEnabled(true)` unconditionally.

**Change**

Adds `quarkus.camel.cli.exposure-mode`, following the same shape as `quarkus.camel.console.exposure-mode` (the issue suggested gating it like the console extension):

| Value | Effect |
|---|---|
| `dev-test` (default) | dev and test modes only |
| `all` | dev, test and prod |
| `none` | never |

`quarkus.camel.cli.enabled=false` still turns it off in every mode. Dev and test behaviour is unchanged, so this only affects prod builds, and a deployer opts back in with `quarkus.camel.cli.exposure-mode=all`. Migration guide entry added.

**Tests**

`CliConnectorEnabledSupplierTest` exercises the gate directly across launch modes — `QuarkusUnitTest` always runs in test mode, so it cannot cover the prod case. Against the previous gate, `devTestModeNotEnabledInProd` and `noneModeNeverEnabled` both fail with `expected: <false> but was: <true>`; both pass with this change.

The existing `CliConnectorEnabledTest` (no config override, expects the factory present) and `CliConnectorDisabledTest` still pass — 7/7 green. `@QuarkusTest`-based `CliConnectorTest` is unaffected, since that is test mode.

`./mvnw clean validate -pl docs` passes.

> Note: this adds `docs/modules/ROOT/pages/migration-guide/3.40.0.adoc`, as does #9066. Whichever merges second will need a trivial conflict resolution in that file and in `migration-guide/index.adoc`.

### DIFF
--- a/docs/modules/ROOT/pages/migration-guide/3.40.0.adoc
+++ b/docs/modules/ROOT/pages/migration-guide/3.40.0.adoc
@@ -18,3 +18,20 @@ To keep the previous behaviour for a context that relied on the default, set it 
 ----
 quarkus.camel.ldap.dir-contexts."my-context".security-authentication=none
 ----
+
+== CLI connector extension changes
+
+=== The connector is no longer active in prod mode by default
+
+`quarkus.camel.cli.enabled` defaults to `true` and had no launch mode condition, so the Camel CLI connector's action file IPC was active in production builds as well as in dev and test.
+
+A `quarkus.camel.cli.exposure-mode` option is added, following the same shape as `quarkus.camel.console.exposure-mode`. It defaults to `dev-test`, which enables the connector only in dev and test modes. Dev and test behaviour is unchanged.
+
+To keep the connector active in prod mode, ask for it explicitly.
+
+[source,properties]
+----
+quarkus.camel.cli.exposure-mode=all
+----
+
+`none` never enables the connector, and `quarkus.camel.cli.enabled=false` still turns it off in every mode.
--- a/docs/modules/ROOT/pages/reference/extensions/cli-connector.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/cli-connector.adoc
@@ -53,6 +53,14 @@ a|icon:lock[title=Fixed at build time] [[quarkus-camel-cli-enabled]]`link:#quark
 Sets whether to enable Camel CLI Connector support.
 | `boolean`
 | `true`
+
+a|icon:lock[title=Fixed at build time] [[quarkus-camel-cli-exposure-mode]]`link:#quarkus-camel-cli-exposure-mode[quarkus.camel.cli.exposure-mode]`
+
+The modes in which the Camel CLI connector is available. The default `dev-test` enables the connector only in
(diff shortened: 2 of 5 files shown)


apupier (approved):

jamesnetherton (approved):
Not sure if we'd need to eventually drop a note into the Camel CLI guide?

https://camel.apache.org/manual/camel-jbang-running.html#_running_with_spring_boot_or_quarkus

IIRC when you use `--runtime=quarkus`, it'll launch the app in `prod` mode via `mvn quarkus:run`. So if you don't have `exposure-mode=all`, then the CLI may not be able to interact with the running app.

I've not tested this BTW, it's just my assumption.

oscerd:
@jamesnetherton your assumption is correct, and it is worse than a docs gap — this PR would break that workflow. Closing it.

`Run.java` in camel-jbang builds the goal like this, in both `runQuarkus()` and `runExistingQuarkusProject()`:

```java
cmd.add("quarkus:" + (dev ? "dev" : "run"));
```

and the flag behind it is

```java
@Option(names = { "--reload", "--dev" },
        description = "Enables dev mode (live reload when source files are updated and saved)")
boolean dev;
```

which defaults to `false`. So a plain `camel run --runtime=quarkus` goes through `quarkus:run`, i.e. **prod mode**, and the gate this PR adds would switch the CLI connector off there. `camel ps`, `camel trace`, `camel top` and friends would stop seeing those applications. Only `--dev` / `--reload` takes the `quarkus:dev` path that still works.

That is a poor trade. The finding behind this was an insecure default, but the connector's action-file IPC is local-only and same-user: anyone who can use it already has filesystem access as the same user, so it is not a meaningful boundary. Breaking the primary jbang workflow to close it costs more than it buys, and a documentation note would not help — the failure is silent, with nothing pointing at `exposure-mode`.

Closing this PR and #9052. If it is worth revisiting, the right order is the other way round: have camel-jbang pass `-Dquarkus.camel.cli.exposure-mode=all` when it launches a Quarkus app, and only then change the default here. That needs an `apache/camel` change first, so I have not opened anything for it — happy to if you think it is worth pursuing.

Thanks for catching it; I had not considered the jbang path at all.

PR #9072, opened 2026-08-27
Fixes #9056. Stop swallowing JTA rollback and resume failures

### LINKED ISSUE #9056: JTA transaction policy swallows rollback and resume failures
`extensions/jta/.../TransactionalJtaTransactionPolicy.java`

Two sites swallow failures in the shared policy base class used by all six `PROPAGATION_*` beans registered by `JtaProcessor`.

**`rollback(boolean isNew)`, lines 72-82**

```java
try {
    if (isNew) {
        transactionManager.rollback();
    } else {
        transactionManager.setRollbackOnly();
    }
} catch (Throwable e) {
    LOG.warn("Could not rollback transaction!", e);
}
```

In the participating case (`isNew=false`, i.e. `PROPAGATION_REQUIRED`/`MANDATORY` joining an existing transaction) a failed `setRollbackOnly()` is logged and discarded, so the outer transaction continues as though it had been marked for rollback.

**`resumeTransaction(...)`, lines 88-98**

Same pattern — a failed `transactionManager.resume(...)` is logged and discarded, leaving subsequent work running outside the transaction the caller believes was resumed.

**Suggested fix**

Rethrow (or wrap and rethrow) in both places so a failed rollback, rollback-marking, or resume does not allow subsequent work to proceed as if transactional integrity were intact.

oscerd:
Fixes #9056.

`TransactionalJtaTransactionPolicy` is the base class of all six `PROPAGATION_*` beans registered by `JtaProcessor`. Two sites logged a failure and carried on:

```java
final protected void rollback(boolean isNew) throws Exception {
    try {
        if (isNew) { transactionManager.rollback(); } else { transactionManager.setRollbackOnly(); }
    } catch (Throwable e) {
        LOG.warn("Could not rollback transaction!", e);   // and likewise for resume
    }
}
```

In the participating case (`PROPAGATION_REQUIRED` / `MANDATORY` joining an existing transaction) a failed `setRollbackOnly()` meant the outer transaction was never marked, so a route that handled the exception could go on to commit. A failed `resume()` meant subsequent work ran outside the transaction the policy was expected to restore. Both were visible only as a log line.

**The care this needs is not to make it worse.** `rollback` is called from two places that already have an exception on its way out, and `resumeTransaction` is called from `finally` blocks in `RequiresNewJtaTransactionPolicy` and `NotSupportedJtaTransactionPolicy` — a naive rethrow would replace the original failure with the cleanup failure, which is a worse outcome than the current logging.

So:

- `rollback` now raises a `CamelException`. `runWithTransaction` and `commit` route it through a `rollbackSuppressing` helper that attaches it to the in-flight exception via `addSuppressed`. The original exception still propagates unchanged.
- `resumeTransaction(Transaction, Throwable)` takes the in-flight exception. When there is one the resume failure is attached to it; when the body succeeded, the resume failure is raised on its own. The two suspending policies capture the exception and pass it in.

**No signatures change.** `resumeTransaction(Transaction)` keeps its exact shape and wraps a failure in a `RuntimeCamelException`, so nothing outside the repository stops compiling. If you would rather it simply declared `throws Exception` and dropped the overload, that is a small change and I am happy to make it — I kept compatibility because the issue did not settle that point.

**Tests**

`JtaTransactionFailurePropagationTest` covers the four cases, using the existing `MockTransactionManagerProducer`:

| Case | Asserts |
|---|---|
| `setRollbackOnly()` fails while participating | original exception still thrown, marking failure suppressed |
| `rollback()` fails | original exception still thrown, rollback failure suppressed |
| `resume()` fails, body succeeded | resume failure surfaces |
| `resume()` fails, body threw | body's exception still thrown, resume failure suppressed |

All four fail against the previous behaviour, each reporting `got []` — nothing attached, which is the defect. The module is green at 18/18, with all six existing policy test classes unaffected. `./mvnw clean install -DskipTests` from the root passes with no regenerated artifacts left behind.

Migration guide entry added, since an application that handles exceptions and continues may now learn about a failure it previously never saw.

> Note: this adds `docs/modules/ROOT/pages/migration-guide/3.40.0.adoc`, as do #9066 and #9069. Whichever merge later need a trivial conflict resolution there and in `migration-guide/index.adoc`.

### DIFF
--- a/extensions/jta/deployment/src/test/java/org/apache/camel/quarkus/component/jta/JtaTransactionFailurePropagationTest.java
+++ b/extensions/jta/deployment/src/test/java/org/apache/camel/quarkus/component/jta/JtaTransactionFailurePropagationTest.java
@@ -1,0 +1,137 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.quarkus.component.jta;
+
+import java.util.Arrays;
+
+import io.quarkus.test.QuarkusUnitTest;
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+import jakarta.transaction.Status;
+import jakarta.transaction.SystemException;
+import jakarta.transaction.Transaction;
+import jakarta.transaction.TransactionManager;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.when;
+
+/**
+ * A transaction whose rollback, rollback marking or resumption fails must not let that failure disappear, and must not
+ * let it replace the exception that made the rollback necessary.
+ */
+public class JtaTransactionFailurePropagationTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest CONFIG = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(MockTransactionManagerProducer.class, MockTransaction.class));
+
+    @Inject
+    TransactionManager transactionManager;
+
+    @Inject
+    @Named("PROPAGATION_REQUIRED")
+    RequiredJtaTransactionPolicy requiredPolicy;
+
+    @Inject
+    @Named("PROPAGATION_REQUIRES_NEW")
+    RequiresNewJtaTransactionPolicy requiresNewPolicy;
+
+    @AfterEach
+    public void afterEach() {
+        reset(transactionManager);
+    }
+
+    @Test
+    public void failedRollbackMarkingIsReportedAndKeepsTheOriginalCause() throws Exception {
+        // Participating in an outer transaction, so the policy marks it rather than rolling it back
+        when(transactionManager.getStatus()).thenReturn(Status.STATUS_ACTIVE);
+        doThrow(new SystemException("mark failed")).when(transactionManager).setRollbackOnly();
+
+        Exception routeFailure = new Exception("route failed");
+        Throwable thrown = assertThrows(Throwable.class, () -> requiredPolicy.run(() -> {
+            throw routeFailure;
+        }));
+
+        // The original failure still surfaces
+        assertEquals(routeFailure, thrown);
+        // and the marking failure rides along rather than vanishing into a log line
+        assertTrue(Arrays.stream(thrown.getSuppressed())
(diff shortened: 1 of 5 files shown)


extensions/jta/deployment/src/test/java/org/apache/camel/quarkus/component/jta/JtaTransactionFailurePropagationTest.java
```diff
@@ -0,0 +1,137 @@
...
+import static org.mockito.Mockito.when;
+
+/**
+ * A transaction whose rollback, rollback marking or resumption fails must not let that failure disappear, and must not
+ * let it replace the exception that made the rollback necessary.
+ */
+public class JtaTransactionFailurePropagationTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest CONFIG = new QuarkusUnitTest()
```
apupier: is it possible to use the non-deprecated QuarkusExtensionTest class mentioned in javadoc deprecation notice?

https://github.com/quarkusio/quarkus/blob/2e48ca4985ebcc4696404cec3c040ed2043f8bca/test-framework/junit-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java#L17

apupier (approved):
not blocking fr the deprecated use of QuarkusUnitTest

oscerd:
Thanks @apupier — done, pushed.

It is a straight swap: `QuarkusUnitTest` and `QuarkusExtensionTest` both extend `AbstractQuarkusExtensionTest<S>`, and `setArchiveProducer` is declared on that shared base, so only the import and the field type changed. `quarkus-junit-internal` 3.39.0 ships `QuarkusExtensionTest`, and `QuarkusUnitTest` is indeed marked `Deprecated` in the same jar.

The module is still green at 18/18 — my 4 cases plus the 6 existing policy test classes.

For what it is worth, this is now the only file in the repository using the new class: `QuarkusUnitTest` is referenced by 180 test files, including the six sibling JTA tests right next to this one. I have deliberately not touched those here, since a bulk migration is a much bigger change than this PR should carry, but it is probably worth its own issue if the deprecation is going to bite at some point.

zhfeng (approved):
Good catch - I think I borrowed these operations from the camel-springboot-jta before. So can you check if there is the similar issue in camel-spring-boot as well?

apupier:
> For what it is worth, this is now the only file in the repository using the new class: `QuarkusUnitTest` is referenced by 180 test files, including the six sibling JTA tests right next to this one. I have deliberately not touched those here, since a bulk migration is a much bigger change than this PR should carry, but it is probably worth its own issue if the deprecation is going to bite at some point.

provided https://github.com/apache/camel-quarkus/pull/9075

jamesnetherton:
@oscerd can you fix up the `migration-guide/3.40.0.adoc` conflict?

oscerd:
@zhfeng I checked, and the pattern does not appear to exist outside camel-quarkus — so there is nothing to fix on the Spring Boot side.

**camel-spring-boot**: no JTA transaction policy code at all. No files matching `*jta*`, no references to `JtaTransactionPolicy` or `TransactionalJta`, and no direct `setRollbackOnly` / `transactionManager.rollback` calls.

**camel core `camel-jta`**: `JtaTransactionPolicy` is abstract and declares `public abstract void run(Runnable) throws Throwable`. It has no `rollback`, no `resumeTransaction` and no `catch (Throwable)` anywhere. `runWithTransaction` does not exist anywhere in camel core.

So the `runWithTransaction` / `rollback` / `resumeTransaction` implementation — and with it the swallowing — looks specific to camel-quarkus's own `TransactionalJtaTransactionPolicy`. If the operations were borrowed from something in the Spring Boot JTA support originally, whatever they came from is no longer there.

Worth saying that my check was a code search across the two checkouts rather than a deep review, so if you know of another place that grew from the same source it is worth a look — but the obvious candidates are clean.

zhfeng:
Hmm, I recall that they were all from `camel-jta` and the `camel-spring` use the different mechanism by TransactionTemplate. So it looks good and sorry for the over look and my bad memory.

jamesnetherton (approved):

oscerd:
No need to apologise @zhfeng — your recollection was closer than mine, and it explains *why* the pattern stayed local.

I checked and you are right about the mechanism: `SpringTransactionPolicy` (`camel-spring-parent/camel-spring/.../spi/SpringTransactionPolicy.java`) holds a Spring `TransactionTemplate` and hands the work to it, rather than driving a `TransactionManager` directly. Since it never calls `rollback()` / `setRollbackOnly()` / `resume()` itself, there is no place for this kind of swallowing to live — Spring owns that part. `camel-jta`'s `JtaTransactionPolicy` is abstract and leaves `run(Runnable)` to subclasses, so the concrete `runWithTransaction` / `rollback` / `resumeTransaction` really did only exist here.

So the two things together are why this was camel-quarkus-only: Spring delegates it, and camel-jta never implemented it. Worth having in the thread for whoever looks next.

Thanks for the review, and for prompting the cross-check — it was worth doing.

PR #9078, opened 2026-08-28
Fixes apache#9076. Add tika and docling parsers to the ingestion pipelines

### LINKED ISSUE #9076: langchain4j-ingest: document parsers for binary formats (tika, docling)
Ingestion pipelines read every payload as plain text, so the documents people actually want to ingest — PDFs, office files, HTML — come out as garbage instead of their content.

A pipeline should be able to name a parser that extracts the text first:

```properties
quarkus.camel.langchain4j.ingest.manuals.parser=tika
```

Two options, reusing the existing Camel components: `tika` extracts text in-process, `docling` converts to markdown through a Docling Serve instance. Works for both directory and consumer-fed pipelines, in JVM and native mode.

JiriOndrusek:
Fixes #9076. Add tika and docling parsers to the ingestion pipelines

* Adds an optional per-pipeline, build-time parser option (tika | docling) that turns binary payloads (PDF, office, HTML) into text before splitting — tika extracts  in-process, docling produces markdown via a Docling Serve instance. Stacked on #9039's PR.

* The PR also refactors the route generation around the new parse stage, splitting  the grown IngestRoutes into focused classes so each route topology reads as a single narrative chain.

  Architecture:

  - The parse step slots between the consumer and the ingestion engine, identically for both route topologies (watched directory and consumer-fed endpoint).
  - Route generation now has one creation path for both declaration styles (properties and @Ingest builder), producing a PipelineSpec and dispatching to one of two route topologies. IngestRoutes keeps only that narrative; step factories live in IngestSteps, the parse stage in IngestParsers, bean resolution in IngestBeanResolver.
  - A parsed document is treated as untrusted input: its identity is captured before the parse, and the payload cannot influence parser behavior via headers or charset heuristics.
  - Native support covers the Tika PDF recipe (scoped angus-activation feature exclusion, PDFBox runtime initialization + resources); tika-parser-pdf-module is BOM-managed.
  - Build-time validation rejects unknown parsers and missing parser components, naming the artifact to add; covered by deployment tests and JVM+native ITs for both parsers and both topologies.


extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
```diff
@@ -68,11 +74,93 @@ FeatureBuildItem feature() {
...
+
+    /**
+     * The same recipe brings PDFBox itself. Native image needs its AWT-touching classes
+     * initialized at run time and its font metrics, glyph lists and ICC profile embedded —
+     * mirroring the camel-quarkus-pdf extension's registrations, widened to the rendering and
+     * graphics packages that Tika's text extraction reaches and camel-pdf does not. Produced
+     * only when PDFBox is on the classpath.
+     */
+    @BuildStep
+    void pdfboxInNative(
```
jamesnetherton, line 112: **Layering question.** The PDFBox runtime-init/resource registrations here (and the `excludeAngusActivationFeature` step just above) are gated on PDFBox being on the classpath, so they only activate for apps that pull in `langchain4j-ingest`. An app that uses `camel-quarkus-tika` + `tika-parser-pdf-module` directly — without ingest — would hit the same native failures with no coverage, since `camel-quarkus-tika`'s deployment has no PDFBox handling today.

Is scoping this to the ingest extension deliberate (with the tika extension left as a follow-up), or would this native config be better homed in `camel-quarkus-tika` so any Tika+PDF user benefits? Happy either way — mainly want the decision to be explicit.

jamesnetherton (commented):
Thanks for this — it's a thorough and notably security-conscious addition. The pre-parse document-id capture (anti-spoofing) and the XXE hardening on the XHTML extraction are especially nice, and the JVM+native coverage across both parsers and both route topologies is great.

Just two non-blocking points, both around the native PDFBox wiring. Neither needs to hold up the merge.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
```diff
@@ -68,11 +74,93 @@ FeatureBuildItem feature() {
...
+                "org/apache/pdfbox/resources/afm/Times-BoldItalic.afm",
+                "org/apache/pdfbox/resources/afm/Times-Italic.afm",
+                "org/apache/pdfbox/resources/afm/Times-Roman.afm",
+                "org/apache/pdfbox/resources/afm/ZapfDingbats.afm",
+                "org/apache/pdfbox/resources/glyphlist/additional.txt",
+                "org/apache/pdfbox/resources/glyphlist/glyphlist.txt",
+                "org/apache/pdfbox/resources/glyphlist/zapfdingbats.txt",
+                "org/apache/pdfbox/resources/icc/ISOcoated_v2_300_bas.icc",
+                "org/apache/pdfbox/resources/text/BidiMirroring.txt",
+                "org/apache/pdfbox/resources/ttf/LiberationSans-Regular.ttf"));
```
jamesnetherton: **Maintainability.** This resource list mirrors `camel-quarkus-pdf`'s registrations, as the javadoc notes. On a PDFBox upgrade that adds or renames AFM/glyphlist/ICC/TTF resources, the two lists can drift apart silently. Could the comment cross-reference `PdfProcessor` (or the shared list be factored into a constant both extensions use), so the coupling is discoverable the next time PDFBox moves?

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
JiriOndrusek: Deliberate: camel-quarkus-tika doesn't provide PDF parsing out of the box (camel-tika ships only the html and text parser modules), nor any native config for it — while this extension documents the PDF recipe and its native ITs exercise it. Filed #9087 to home the native config in camel-quarkus-tika as a follow-up. Everything here is classpath-gated, so nothing activates without PDFBox.

extensions/langchain4j-ingest/deployment/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/deployment/Langchain4jIngestProcessor.java
JiriOndrusek: Thanks for pointing at this!
I've found a better way of doing the same, and I reported a follow-upticket https://github.com/apache/camel-quarkus/issues/9091 to do the same in pdf extension.

JiriOndrusek:
Superseded by https://github.com/apache/camel-quarkus/pull/9143

PR #9081, opened 2026-09-01
Document Twitter native locale requirements #2564

### LINKED ISSUE #2564: Twitter native test : Locale and Timezone
I've been testing the twitter extension in native mode. And as I have a Locale fr, I've got errors parsing the responses from Twitter.com.
The Twitter Java API expect to have the Locale.US installed and expect also Timezone UTC.
We need to address the issue or at least document it.

AzazelSensei:
Fixes #2564

Twitter4J parses API dates with Locale.US. Native images only embed the build-machine locale, so a French locale fails to parse the responses.

I documented `quarkus.default-locale=en-US` on the Twitter extension page, and set the same property on the Twitter integration tests.

### DIFF
--- a/docs/modules/ROOT/pages/reference/extensions/twitter.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/twitter.adoc
@@ -63,6 +63,7 @@ quarkus.default-locale=en-US
 
 See also xref:user-guide/native-mode.adoc#locale[Locale in native mode].
 
+
 [id="extensions-twitter-ssl-in-native-mode"]
 == SSL in native mode
 
--- a/docs/modules/ROOT/pages/reference/extensions/twitter.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/twitter.adoc
@@ -47,17 +47,10 @@ ifeval::[{doc-show-user-guide-link} == true]
 Check the xref:user-guide/index.adoc[User guide] for more information about writing Camel Quarkus applications.
 endif::[]
 
-[id="extensions-twitter-ssl-in-native-mode"]
-== SSL in native mode
-
-This extension auto-enables SSL support in native mode. Hence you do not need to add
-`quarkus.ssl.native=true` to your `application.properties` yourself. See also
-https://quarkus.io/guides/native-and-ssl[Quarkus SSL guide].
-
 [id="extensions-twitter-camel-quarkus-limitations"]
 == Camel Quarkus limitations
 
-The Twitter Java client parses timestamps with `Locale.ENGLISH` and expects timezone UTC.
+The Twitter Java client parses timestamps with `Locale.US`.
 In native mode, GraalVM includes only the build machine locale by default. A non-English locale
 (for example French) then fails to parse Twitter API responses.
 
@@ -68,19 +61,11 @@ Set the default locale to `en-US` in `application.properties`:
 quarkus.default-locale=en-US
 ----
 
-You can also pin language and country on the native image:
-
-[source,properties]
-----
-quarkus.native.user-language=en
-quarkus.native.user-country=US
-----
-
-Pin the timezone to UTC for the native build:
-
-[source,properties]
-----
-quarkus.native.additional-build-args=-J-Duser.timezone=UTC
-----
-
 See also xref:user-guide/native-mode.adoc#locale[Locale in native mode].
+
+[id="extensions-twitter-ssl-in-native-mode"]
+== SSL in native mode
+
+This extension auto-enables SSL support in native mode. Hence you do not need to add
+`quarkus.ssl.native=true` to your `application.properties` yourself. See also
+https://quarkus.io/guides/native-and-ssl[Quarkus SSL guide].
--- a/docs/modules/ROOT/pages/reference/extensions/twitter.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/twitter.adoc
@@ -53,3 +53,34 @@ endif::[]
 This extension auto-enables SSL support in native mode. Hence you do not need to add
 `quarkus.ssl.native=true` to your `application.properties` yourself. See also
 https://quarkus.io/guides/native-and-ssl[Quarkus SSL guide].
+
+[id="extensions-twitter-camel-quarkus-limitations"]
+== Camel Quarkus limitations
+
+The Twitter Java client parses timestamps with `Locale.ENGLISH` and expects timezone UTC.
+In native mode, GraalVM includes only the build machine locale by default. A non-English locale
+(for example French) then fails to parse Twitter API responses.
+
+Set the default locale to `en-US` in `application.properties`:
+
+[source,properties]
+----
+quarkus.default-locale=en-US
+----
+
+You can also pin language and country on the native image:
+
+[source,properties]
+----
+quarkus.native.user-language=en
+quarkus.native.user-country=US
+----
+
+Pin the timezone to UTC for the native build:
+
+[source,properties]
+----
+quarkus.native.additional-build-args=-J-Duser.timezone=UTC
+----
+
+See also xref:user-guide/native-mode.adoc#locale[Locale in native mode].
--- a/extensions/twitter/runtime/src/main/doc/limitations.adoc
+++ b/extensions/twitter/runtime/src/main/doc/limitations.adoc
@@ -1,0 +1,27 @@
+The Twitter Java client parses timestamps with `Locale.ENGLISH` and expects timezone UTC.
+In native mode, GraalVM includes only the build machine locale by default. A non-English locale
+(for example French) then fails to parse Twitter API responses.
+
+Set the default locale to `en-US` in `application.properties`:
+
+[source,properties]
+----
+quarkus.default-locale=en-US
+----
+
+You can also pin language and country on the native image:
+
+[source,properties]
+----
+quarkus.native.user-language=en
+quarkus.native.user-country=US
+----
+
+Pin the timezone to UTC for the native build:
+
+[source,properties]
+----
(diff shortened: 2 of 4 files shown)


docs/modules/ROOT/pages/reference/extensions/twitter.adoc
```diff
@@ -53,3 +53,34 @@ endif::[]
 This extension auto-enables SSL support in native mode. Hence you do not need to add
 `quarkus.ssl.native=true` to your `application.properties` yourself. See also
 https://quarkus.io/guides/native-and-ssl[Quarkus SSL guide].
+
+[id="extensions-twitter-camel-quarkus-limitations"]
+== Camel Quarkus limitations
```
jamesnetherton, line 51: This page is generated ("Do not edit directly!"), and this section is in the wrong position — the template emits `Camel Quarkus limitations` before `SSL in native mode`. Regenerating will move it and CI's uncommitted-changes check will fail. Run `./mvnw -pl extensions/twitter/deployment process-classes` and commit the output instead of editing this file directly.

extensions/twitter/runtime/src/main/doc/limitations.adoc
```diff
@@ -0,0 +1,27 @@
+The Twitter Java client parses timestamps with `Locale.ENGLISH` and expects timezone UTC.
```
jamesnetherton: twitter4j uses `Locale.US`, not `Locale.ENGLISH` (`ParseUtil` in both 4.0.7 and 4.1.2) — this is also how #2564 words it. Worth being precise, since `Locale.ENGLISH` is `en` whereas the property recommended just below is `en-US`.

Also "and expects timezone UTC" should go — see the note on the timezone block below.

extensions/twitter/runtime/src/main/doc/limitations.adoc
```diff
@@ -0,0 +1,27 @@
...
+----
+
+You can also pin language and country on the native image:
+
+[source,properties]
+----
+quarkus.native.user-language=en
+quarkus.native.user-country=US
+----
+
```
jamesnetherton:
```suggestion
```
These are from a 2021 comment on #2564 and were valid then, but they have since been deprecated in favour of `quarkus.default-locale` and removed — they are absent from `quarkus-core-deployment-3.39.0.jar` entirely (`pom.xml:68`). Anyone following this gets an "Unrecognized configuration key" warning and no behaviour change. The `quarkus.default-locale=en-US` block above is sufficient; `quarkus.locales` covers the multi-locale case if it ever comes up.

extensions/twitter/runtime/src/main/doc/limitations.adoc
```diff
@@ -0,0 +1,27 @@
...
+Pin the timezone to UTC for the native build:
+
+[source,properties]
+----
+quarkus.native.additional-build-args=-J-Duser.timezone=UTC
+----
+
```
jamesnetherton:
```suggestion
```
I would drop the timezone guidance rather than document it. It was never confirmed on #2564 — the tests could not be run for want of credentials — and it does not hold now: `-J-D...` sets the property on the image-builder JVM rather than the executable, and twitter4j parses with explicit-offset patterns (4.1.2) or an explicitly UTC-set formatter (4.0.7), so it never reads the JVM default zone.

jamesnetherton (commented):
Thanks for picking this one up — it has been open since 2021 and documenting it is exactly what #2564 asked for. I can see you followed the guidance in the issue thread closely, which was the right instinct; the problem is that most of that advice dates from May 2021 and has since been overtaken. Details below, but the short version is that this reduces to `quarkus.default-locale=en-US` plus a regenerated doc page.

**1. `quarkus.native.user-language` / `quarkus.native.user-country` no longer exist.**
These come from a comment on #2564, and they were correct advice at the time. They have since been deprecated in favour of the global `quarkus.default-locale` and removed: I confirmed the strings are present in `NativeConfig` in 3.9.4 and absent from the whole of `quarkus-core-deployment-3.39.0.jar`, which is what `pom.xml:68` pins. As they stand they would produce "Unrecognized configuration key ... it will be ignored" and no locale change. This is more or less what a later comment in the thread was pointing at, noting that locale handling had moved to GraalVM level — `quarkus.default-locale` (and `quarkus.locales` for the multi-locale case the issue hoped for) is the modern form, and it is already documented in `user-guide/native-mode.adoc#locale`.

**2. The UTC part was never actually confirmed, and does not hold today.**
Reading back through #2564, the timezone half was always provisional — the reporter wanted the integration tests run to check whether there was a TimeZone problem at all, but could not, for want of developer credentials. Nobody ever confirmed one. Two things now argue against it: `-J-D...` passes an option to the image-*builder* JVM rather than the produced executable, so it would not pin the runtime timezone even if it were needed; and camel-twitter 4.22 pulls twitter4j-core 4.1.2, whose `ParseUtil` builds `DateTimeFormatter.ofPattern(pattern, Locale.US)` with explicit-offset patterns (`EEE MMM d HH:mm:ss Z yyyy`, `EEE, d MMM yyyy HH:mm:ss Z`, `yyyy-MM-dd'T'HH:mm:ss'Z'`). Even 4.0.7, current when the issue was filed, called `setTimeZone("UTC")` on the formatter itself. Neither path consults the JVM default zone. I would drop the timezone material entirely rather than carry an unverified 2021 assumption into the docs. Dropping it also avoids claiming `quarkus.native.additional-build-args` for the module — if a build arg is ever needed here, the `-append` variant is the one to use (cf. `integration-tests/milo/pom.xml:85`).

**3. `docs/.../twitter.adoc` needs regenerating, not hand-editing.**
`limitations.adoc` is correctly included — thank you, that is the right home for this — but the generated page was edited by hand and the new section landed after `== SSL in native mode`. The template emits the `limitations` block *before* the `activatesNativeSsl` block, which is the order every other extension has (activemq 64/71, activemq6 63/70, jolokia 398/409, xmlsecurity 50/56). Running `./mvnw -pl extensions/twitter/deployment process-classes` will move it, and the "Fail if there are uncommitted changes" step in `ci-build.yaml` will then fail. Please regenerate and commit the result.

Minor: the client uses `Locale.US`, not `Locale.ENGLISH` — that is what `ParseUtil` does in both 4.0.7 and 4.1.2, and it is also how the issue itself is worded ("expect to have the Locale.US installed"). Worth correcting, since `Locale.ENGLISH` is `en` whereas the property being recommended is `en-US`.

One question on scope: #2564 asked to "address the issue or at least document it". This PR takes the documenting route, which is a legitimate reading, but it is worth deciding explicitly whether setting `quarkus.default-locale=en-US` in the integration test is the actual fix and the issue can be closed, or whether it stays open for the multi-locale support the issue also asks for. Since the description says "Fixes #2564", merging as-is will close it either way.

integration-tests/twitter/src/main/resources/application.properties
```diff
@@ -43,3 +43,11 @@ camel.component.twitter-directmessage.consumerKey={{env:TWITTER_CONSUMER_KEY}}
 camel.component.twitter-directmessage.consumerSecret={{env:TWITTER_CONSUMER_SECRET}}
 camel.component.twitter-directmessage.accessToken={{env:TWITTER_ACCESS_TOKEN}}
 camel.component.twitter-directmessage.accessTokenSecret={{env:TWITTER_ACCESS_TOKEN_SECRET}}
+
+#
+# Native image: Twitter4J parses dates with Locale.ENGLISH and timezone UTC
+#
+quarkus.default-locale=en-US
+quarkus.native.user-language=en
+quarkus.native.user-country=US
```
jamesnetherton:
```suggestion
```
Same as the doc — unrecognized under Quarkus 3.39, so these only add build warnings. `quarkus.default-locale=en-US` on line 50 is the whole fix. Compare `integration-tests/minio/src/main/resources/application.properties:19`, which sets only the default locale.

integration-tests/twitter/src/main/resources/application.properties
```diff
@@ -43,3 +43,11 @@ camel.component.twitter-directmessage.consumerKey={{env:TWITTER_CONSUMER_KEY}}
...
+quarkus.native.additional-build-args=-J-Duser.timezone=UTC
```
jamesnetherton:
```suggestion
```
This does not affect the native executable's default timezone (`-J-D` targets the builder JVM), and twitter4j does not depend on the JVM default zone. Setting `quarkus.native.additional-build-args` also claims the property for this whole module — if a build arg is genuinely needed later, use the `-append` variant as `integration-tests/milo/pom.xml:85` does.

integration-tests/twitter/README.adoc
```diff
@@ -17,3 +17,6 @@ export TWITTER_USER_NAME=my-twitter-username
 ----
 
 
+
+Native tests need `Locale.ENGLISH` and timezone UTC, otherwise Twitter API date parsing fails.
+The test `application.properties` already sets `quarkus.default-locale=en-US` and UTC for the native image.
```
jamesnetherton: "and UTC for the native image" is not accurate — the timezone arg does not reach the native executable. Once the UTC line is dropped from `application.properties`, this sentence should say only that the default locale is set. Also `Locale.ENGLISH` on line 21 should be `Locale.US`.

AzazelSensei:
Dropped `quarkus.native.user-language` / `user-country` and the UTC build-arg. Those keys are gone in 3.39, and `-J-Duser.timezone=UTC` does not reach the native image. Tests now only set `quarkus.default-locale=en-US`. Regenerated `twitter.adoc` from `limitations.adoc`.

Docs now say `Locale.US`, matching twitter4j `ParseUtil`.

Keeping `Fixes #2564`. The issue asked to document this; extra locales are already `quarkus.locales`.

jamesnetherton (approved):

Issue #9054, opened 2026-08-26
LDAP securityAuthentication default contradicts its javadoc


oscerd:
`extensions/ldap/.../CamelLdapConfig.java:60-68`

The javadoc states:

> The security level to use. [...] If this property is unspecified, the behaviour is determined by the service provider.

but the option carries `@WithDefault("none")`, so it is never unspecified.

`CamelLdapRecorder.createDirContexts` then does:

```java
env.put(Context.SECURITY_AUTHENTICATION, dirConfig.securityAuthentication());
...
env.putAll(dirConfig.additionalOptions());
```

`additionalOptions` is merged afterwards, so setting `java.naming.security.authentication` there does take effect. But an operator who supplies only `java.naming.security.principal` and `java.naming.security.credentials` through `additional-options` — the only way to pass credentials — silently gets an anonymous bind, because `none` is already in the map and nothing overrides it.

**Suggested fix**

Make `securityAuthentication()` an `Optional<String>` set only when configured (matching the javadoc), or derive `simple` automatically when a principal/credentials are present. Either way the javadoc and the behaviour should agree.

PR #9084, opened 2026-09-02
Backport #9066 and #9072 to 3.39.x

oscerd:
Backports two fixes from `main` to `3.39.x`. Both defects are present on this branch.

| Original | Fix |
|---|---|
| #9066 (issue #9054) | LDAP `security-authentication` is only put into the JNDI environment when configured, so credentials supplied through `additional-options` are no longer silently ignored on an anonymous bind |
| #9072 (issue #9056) | JTA rollback and resume failures are raised instead of logged and discarded, attached via `addSuppressed` so they never replace the exception that made the rollback necessary |

Both cherry-picks were taken with `-x`, so each commit records the `main` SHA it came from.

**One adaptation commit on top**, for the thing a cherry-pick cannot know: the changes carry a `3.40.0` migration guide, which belongs to `main` rather than this release line. It is replaced with a `3.39.1` guide holding the same two notes, and `migration-guide/index.adoc` points at that instead. This follows the `3.20.2.adoc` precedent — a patch level guide added only on its maintenance branch for a behaviour change backported into it.

**Verification on this branch**

- `extensions/jta/deployment` — 18 tests, 0 failures (the 4 new cases plus the 6 existing policy test classes)
- `integration-tests/ldap` — 6 tests, 0 failures
- `./mvnw clean validate -pl docs` passes

https://claude.ai/code/session_018YmgydQvapQHWoJfotF8tK

### DIFF
--- a/docs/modules/ROOT/pages/migration-guide/3.39.1.adoc
+++ b/docs/modules/ROOT/pages/migration-guide/3.39.1.adoc
@@ -1,6 +1,6 @@
-= Camel Quarkus 3.40.0 Migration Guide
+= Camel Quarkus 3.39.1 Migration Guide
 
-The following guide outlines how to adapt your code to changes that were made in Camel Quarkus 3.40.0.
+The following guide outlines how to adapt your code to changes that were made in Camel Quarkus 3.39.1.
 
 == LDAP extension changes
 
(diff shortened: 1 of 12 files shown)


jamesnetherton:
I think we can live without this since 3.39.x is not an LTS stream (3.40.x will be).

PR #9086, opened 2026-09-02
Backport #9066 and #9072 to 3.27.x

oscerd:
Backports two fixes from `main` to the `3.27.x` LTS line. Both defects are present on this branch.

| Original | Fix |
|---|---|
| #9066 (issue #9054) | LDAP `security-authentication` is only put into the JNDI environment when configured, so credentials supplied through `additional-options` are no longer silently ignored on an anonymous bind |
| #9072 (issue #9056) | JTA rollback and resume failures are raised instead of logged and discarded, attached via `addSuppressed` so they never replace the exception that made the rollback necessary |

Both cherry-picks were taken with `-x`, so each commit records the `main` SHA it came from.

**One adaptation commit on top**, covering two things a cherry-pick cannot know:

1. The changes carry a `3.40.0` migration guide, which belongs to `main` rather than this release line. Replaced with a `3.27.6` guide holding the same two notes, with `migration-guide/index.adoc` pointing at that instead. This follows the `3.20.2.adoc` precedent — a patch level guide added only on its maintenance branch for a behaviour change backported into it.

2. **`QuarkusExtensionTest` does not exist in Quarkus 3.27.x.** This line uses `quarkus-junit5-internal`, which ships only `QuarkusDevModeTest`, `QuarkusProdModeTest` and `QuarkusUnitTest`, so the new test uses `QuarkusUnitTest` here, as its six sibling JTA tests already do. The swap to the non-deprecated class on `main` came from #9075, which is not part of this backport. Worth noting because the cherry-pick applies cleanly and only fails at compile time.

**Verification on this branch**

- `extensions/jta/deployment` — 18 tests, 0 failures (the 4 new cases plus the 6 existing policy test classes)
- `integration-tests/ldap` — 6 tests, 0 failures
- `./mvnw clean validate -pl docs` passes

https://claude.ai/code/session_018YmgydQvapQHWoJfotF8tK

### DIFF
--- a/docs/modules/ROOT/pages/migration-guide/3.27.6.adoc
+++ b/docs/modules/ROOT/pages/migration-guide/3.27.6.adoc
@@ -1,6 +1,6 @@
-= Camel Quarkus 3.40.0 Migration Guide
+= Camel Quarkus 3.27.6 Migration Guide
 
-The following guide outlines how to adapt your code to changes that were made in Camel Quarkus 3.40.0.
+The following guide outlines how to adapt your code to changes that were made in Camel Quarkus 3.27.6.
 
 == LDAP extension changes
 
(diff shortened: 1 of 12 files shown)


jamesnetherton:
Pubsub failures are unrelated to these changes.

Something weird is happening when pulling `gcr.io/google.com/cloudsdktool/cloud-sdk:492.0.0-emulators`. Seems it does not exist. It can be fixed up seperately.

jamesnetherton (approved):

PR #9096, opened 2026-09-02
Use a floating tag for the Google Cloud emulators test container image

jamesnetherton:
Relates to failing google-pubsub tests from #9086. 

Google [enforces](https://github.com/GoogleCloudPlatform/cloud-sdk-docker/blob/master/README.md#package-retention-policy) a 365 day retention policy on the Google Cloud CLI Docker image repositories, so pinned version tags are deleted after a year.

Thus I think we should use a floating container image tag like the Camel core and Quarkus Google Cloud services is doing.


aldettinger (approved):
Indeed, no better idea.

apupier (approved):

PR #9101, opened 2026-09-03
Document Camel Quarkus build items #2900

### LINKED ISSUE #2900: Document Camel Quarkus build items
It could be useful for newcomers to the project and potentially also for folks who want to write their own Camel component extensions, if we had a summary page of build items like [Quarkus](https://quarkus.io/guides/all-builditems).

AzazelSensei:
Adds a contributor-guide page that lists the core and Camel Main build items, in the same spirit as the Quarkus all-builditems catalog.

I also linked it from the custom components guide and the contributor nav.

Fixes #2900

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -18,6 +18,7 @@
 ** xref:migration-guide/3.2.0.adoc[Camel Quarkus 3.2.0 migration guide]
 * xref:contributor-guide/index.adoc[Contributor guide]
 ** xref:contributor-guide/create-new-extension.adoc[Create new extension]
+** xref:contributor-guide/build-items.adoc[Build items]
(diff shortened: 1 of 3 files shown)


jamesnetherton:
Thanks @AzazelSensei. I probably should have added more info to the issue.

This task is somewhat more complex than just adding a static doc page.

* There are other extensions outside of the extensions-core/core that provide their own build items. So they also need to be listed
* The page needs to be built dynamically so that it stays fresh. Not sure what the best way to do that it is. But it needs to be unobtrusive and not impact project build. Maybe something added to `tooling/maven-plugin` where we have various mojos for manipulating docs.

AzazelSensei:
Dropped the static core-only list. The page is now generated from the *BuildItem classes in every deployment module. The mojo lives in tooling/maven-plugin and runs from the docs module, so the main build is not involved.

PR #9102, opened 2026-09-03
chore: backport targets are the LTS lines only

oscerd:
Corrects the backport-targets list added in #9074.

That list named `3.39.x` as a target alongside the LTS lines. It is not an LTS stream, and @jamesnetherton declined a backport to it on exactly those grounds when closing #9084:

> I think we can live without this since 3.39.x is not an LTS stream (3.40.x will be).

So the rule as merged points the next reader at a PR that gets closed.

**Changes**

- Targets are now `main` plus the **LTS lines only** — currently `3.33.x` and `3.27.x` — with a note that `3.40.x` becomes the next LTS.
- States explicitly that non-LTS release lines are not targets however recently they were released, citing #9084.
- Replaces the "a branch existing in `git branch -r` is not evidence it is supported" warning with a stronger one: **neither its presence nor its recency tells you it is a target**, because the newest active branch is usually the current non-LTS release line.

That last point is the actual trap, and the one I fell into: I included `3.39.x` precisely *because* it was the most recently committed-to branch. The committer-date command is still useful for telling a live branch from a dormant one, so it stays — but it is now framed as answering only that narrower question.

Documentation only; no build or code impact.

https://claude.ai/code/session_01Nvs8UTajTMYsXa2rcHdiRz


apupier (approved):

jamesnetherton (approved):
Not sure if maybe referencing this could help a bit?

https://github.com/quarkusio/quarkus/wiki/Release-Planning

Might give some hints to the agent about which streams are LTS?

Or maybe better:

https://registry.quarkus.io/client/platforms

The current active LTS streams are flagged by `"lts" : true` in the JSON response.

PR #9116, opened 2026-09-04
Apply Camel's JAXP external access restrictions in the xslt extension

jamesnetherton:
Camel restricts access to external DTDs and stylesheets on the TransformerFactory it creates. The xslt extension supplies its own factory, so those restrictions were not applied to the transformations it performs. The factory is also registered as the JAXP default, so this covers any code obtaining one through TransformerFactory.newInstance(), including applications using camel-quarkus-tika and camel-quarkus-xmlsecurity, which depend on the same support extension without using XSLT themselves.

XalanTransformerFactory now applies the restrictions itself:

* Documents being transformed are parsed with an XMLReader that does not resolve external general entities, matching what Camel's XmlConverter does for the bodies camel-xslt converts to a SAXSource itself. A SAXSource carrying a caller configured XMLReader is used as it is, and DOMSource and StAXSource are already parsed.
* Resources fetched at transform time by document() are denied unless the application's own URIResolver resolves them. Xalan does not propagate the factory resolver onto everything it hands out, so it is installed on each entry point; an application setting its own resolver, which camel-xslt does on every exchange, keeps overriding it as before.

xsl:import and xsl:include are unchanged. Xalan dereferences those hrefs itself while compiling, so they cannot be restricted here, which matches what plain Camel does on the component path. A test pins that behaviour so a Xalan upgrade changing it does not go unnoticed.

CamelXsltRecorder applies quarkus.camel.xslt.features to every template instead of only to those compiled to a translet at build time, where it previously had no effect on templates loaded at runtime, and warns once when secure processing is disabled.

### DIFF
--- a/extensions-support/xalan/runtime/src/main/java/org/apache/camel/quarkus/support/xalan/XalanTransformerFactory.java
+++ b/extensions-support/xalan/runtime/src/main/java/org/apache/camel/quarkus/support/xalan/XalanTransformerFactory.java
@@ -24,6 +24,7 @@ import javax.xml.XMLConstants;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.parsers.SAXParserFactory;
 import javax.xml.transform.ErrorListener;
+import javax.xml.transform.Result;
 import javax.xml.transform.Source;
 import javax.xml.transform.Templates;
 import javax.xml.transform.Transformer;
@@ -37,10 +38,13 @@ import javax.xml.transform.sax.TemplatesHandler;
 import javax.xml.transform.sax.TransformerHandler;
 import javax.xml.transform.stream.StreamSource;
 
+import org.xml.sax.Attributes;
 import org.xml.sax.InputSource;
+import org.xml.sax.Locator;
 import org.xml.sax.SAXException;
 import org.xml.sax.XMLFilter;
 import org.xml.sax.XMLReader;
+import org.xml.sax.ext.DeclHandler;
 
 import org.apache.xalan.xsltc.trax.TrAXFilter;
 import org.slf4j.Logger;
@@ -198,12 +202,13 @@ public final class XalanTransformerFactory extends SAXTransformerFactory {
 
     @Override
     public XMLFilter newXMLFilter(Source source) throws TransformerConfigurationException {
-        return secure(delegate.newXMLFilter(source));
+        final Templates templates = delegate.newTemplates(source);
+        return templates == null ? null : newXMLFilter(templates);
     }
 
     @Override
     public XMLFilter newXMLFilter(Templates templates) throws TransformerConfigurationException {
-        return secure(delegate.newXMLFilter(unwrap(templates)));
+        return new SecuredTrAXFilter(unwrap(templates), restrictingUriResolver);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


aldettinger (approved):

extensions-support/xalan/runtime/src/main/java/org/apache/camel/quarkus/support/xalan/XalanTransformerFactory.java
```diff
@@ -138,12 +198,224 @@ public TemplatesHandler newTemplatesHandler() throws TransformerConfigurationExc
...
+     * the bodies camel-xslt converts itself, so that {@link Source}-shaped bodies get the same treatment.
+     */
+    private static Source secureInputSource(Source source) throws TransformerException {
+        if (source instanceof StreamSource) {
+            final StreamSource streamSource = (StreamSource) source;
+            final InputStream inputStream = streamSource.getInputStream();
+            final Reader reader = streamSource.getReader();
+            if (inputStream == null && reader == null) {
+                // Nothing but a systemId; let the delegate resolve it as before
+                return source;
```
zhfeng, line 257: A system-ID-only `StreamSource` bypasses the hardened parser here. For example, `new StreamSource(input.toUri().toString())` has neither an input stream nor a reader, so this returns the source unchanged and lets Xalan open and parse the document with its own parser. If that document contains a `SYSTEM` entity, the `XMLReader` configured above with external general entities disabled is never used, and the entity may still be resolved.

Could this branch instead construct an `InputSource` with the system ID and return `new SAXSource(createSecureXmlReader(), inputSource)`? That preserves loading of the primary document and its base URI while ensuring it is parsed by the secured reader. A regression test using a system-ID-only `StreamSource` containing the external entity would cover this path.

extensions-support/xalan/runtime/src/main/java/org/apache/camel/quarkus/support/xalan/XalanTransformerFactory.java
```diff
@@ -138,12 +198,224 @@ public TemplatesHandler newTemplatesHandler() throws TransformerConfigurationExc
...
+        }
+
+        @Override
+        public ErrorListener getErrorListener() {
+            return delegate.getErrorListener();
+        }
+
+        @Override
+        public void reset() {
+            delegate.reset();
```
zhfeng, line 418: Could we add a regression test for the behavior after `reset()`? The restricting URI resolver is installed on the delegate after the Xalan transformer is created, while JAXP defines `reset()` as restoring the transformer's original configuration. It would be useful to verify that calling `reset()` here does not discard `RestrictingUriResolver` and allow a subsequent transformation using `document()` to resolve an external resource. If the test exposes that behavior, this override could reinstall the restricting resolver after `delegate.reset()`.

extensions-support/xalan/runtime/src/main/java/org/apache/camel/quarkus/support/xalan/XalanTransformerFactory.java
jamesnetherton: Good catch, confirmed and fixed. I reproduced it with a system-ID-only `StreamSource` over a document carrying a `SYSTEM` entity: the entity was resolved and the fetch went out, so this was a real hole rather than a theoretical one.

Fixed as you suggested. The branch now builds an `InputSource` with the system ID and returns `new SAXSource(createSecureXmlReader(), inputSource)`, so the document is parsed by the secured reader instead of being opened by Xalan. I also moved `setSystemId` out of the stream/reader branches so it is carried in every case, which keeps relative references resolving against the document. The early return survives only when there is no stream, no reader and no system id, where there is nothing to parse and the delegate reports it as before.

Two regression tests cover it: `externalEntityInSystemIdOnlyStreamSourceIsNotResolved` for the entity, and `systemIdOnlyStreamSourceIsStillTransformed` to pin that loading by system id still works. Both fail on the previous code and pass now.

extensions-support/xalan/runtime/src/main/java/org/apache/camel/quarkus/support/xalan/XalanTransformerFactory.java
jamesnetherton: You were right to ask, and the test exposed it. After `reset()` the resolver was gone and `document()` fetched and returned the external resource, so `reset()` did discard the restriction.

Fixed by reinstalling it: `SecuredTransformer.reset()` now calls `delegate.reset()` and puts the restricting resolver back. Chasing the same idea turned up `setURIResolver(null)` as a second way to lose it, so a null now falls back to the restriction rather than lifting it. The JDK is immune to both because it enforces `ACCESS_EXTERNAL_STYLESHEET` through the factory rather than through a `URIResolver`, so this was a parity gap either way.

That led to a third case in a follow-up commit. `TransformerHandler.getTransformer()` and `TrAXFilter.getTransformer()` hand out the transformer the handler then uses itself, so the same two calls took the restriction off from underneath it. Both are now wrapped so `getTransformer()` returns the secured transformer. Xalan reads its own transformer from a field and never calls `getTransformer()`, so overriding it does not disturb the transformation.

Tests: `documentFunctionStaysDeniedAfterTransformerReset`, `documentFunctionStaysDeniedWhenUriResolverIsCleared`, and for the SAX paths `documentFunctionStaysDeniedAfterResettingTheHandlersTransformer`, `documentFunctionStaysDeniedWhenTheHandlersResolverIsCleared` and `documentFunctionStaysDeniedAfterResettingTheFiltersTransformer`. Each one fails without its fix. `transformerHandlerStillTransforms` and `xmlFilterStillTransforms` pin that the new wrappers still delegate the SAX events correctly.

The migration guide now states both semantics, since they differ from a plain JAXP factory. Native mode is verified for the component path (`xml-grouped` native ITs pass); the SAX entry points are covered by the JVM tests only, as nothing in the repo drives that API.

zhfeng (approved):
NIce work!

PR #9117, opened 2026-09-04
Fix Azure Storage Blob tests failing when blob versioning is enabled

JinyuChen97:
Use separate blob names for different blob types (block, append, page) to avoid InvalidBlobType conflicts when versioning retains type history. Guard blobVersions test with BlobVersioningEnabled condition and sort versions by versionId to handle accumulated version history.

_Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>_

### DIFF
--- a/integration-test-groups/azure/README.adoc
+++ b/integration-test-groups/azure/README.adoc
@@ -68,6 +68,13 @@ To clean up, run
 $ ./azure-resources.sh delete
 ----
 
+To run blob versioning tests, https://learn.microsoft.com/en-us/azure/storage/blobs/versioning-enable?tabs=portal[enable blob versioning] on your storage account and set the following (requires `CAMEL_QUARKUS_START_MOCK_BACKEND=false`):
+
+[source,shell]
+----
+export AZURE_BLOB_VERSIONING_ENABLED=true
+----
+
 You may want to `export CAMEL_QUARKUS_START_MOCK_BACKEND=false` to avoid starting the local Azurite container and make sure that you test against the real remote Azure API.
 
 You may want to disable identity tests except key-vault (which is mandatory)  `export CAMEL_QUARKUS_DISABLE_IDENTITY_EXCEPT_KEY_VAULT=true` to avoid starting test using identity with other extension then key-vault. This option might be helpful in case of testing with client without permissions for eventhubs.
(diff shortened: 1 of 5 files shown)


jamesnetherton:
Thanks for the fix - splitting the append/page blobs onto their own names is the right call, and the header override works as expected since `BlobConfigurationOptionsProxy.getBlobName()` checks `BlobConstants.BLOB_NAME` in the headers before falling back to the endpoint configuration.

A few things on the `blobVersions` test:

**1. Version selection isn't scoped to the blob under test**

`AzureStorageBlobTest.java:805-813` picks `currentVersion` / `oldVersion` from every entry returned by `/blob/versions/list`. That listing is container-wide: `direct:listBlobVersions` sets neither `blobName` nor `regex`, so `BlobContainerOperations.listBlobVersions` only applies `BlobListDetails.setRetrieveVersions(true)`.

This PR makes that more likely to bite, since `test-append` and `test-page` now live in the same container. If any blob other than `test` has versions at that point, the descending sort can select a foreign `versionId`, which is then read back against `blobName=test` and will 404 or return unexpected content.

The resource already emits `name` for each entry, so filtering on it should be enough:

```java
.filter(v -> BLOB_NAME.equals(v.get("name")))
```

on both streams.

**2. The new comparator can NPE**

`AzureStorageBlobTest.java:812` dereferences `versionId` unconditionally, but `AzureStorageBlobResource` only adds that key when it is non-null:

```java
if (blobItem.getVersionId() != null) {
    versionBuilder.add("versionId", blobItem.getVersionId());
}
versionBuilder.add("isCurrentVersion", blobItem.isCurrentVersion() != null && blobItem.isCurrentVersion());
```

`isCurrentVersion` is always present and defaults to `false`, so an entry with no version id (a blob written before versioning was enabled on the account, for example) passes the `!isCurrentVersion` filter and then NPEs inside `compareTo`. That surfaces as a bare NPE rather than a useful assertion failure. Either filter out null ids first, or use `Comparator.comparing(v -> (String) v.get("versionId"), Comparator.nullsFirst(Comparator.naturalOrder())).reversed()`.

**3. README wording**

`BlobVersioningEnabled` also requires `!MockBackendUtils.startMockBackend()`, and the mock backend is on by default. The next paragraph does mention `CAMEL_QUARKUS_START_MOCK_BACKEND=false`, but as "You may want to" - for these tests it is required, otherwise they are silently skipped. Worth saying so in the new section.

Nothing here breaks the intended flow, and the rest looks good to me - the descending sort does select v1 for the create/update sequence, so dropping the `assertEquals(2, versions.size())` assertion is sound.

---
_This review was AI-generated, on behalf of James Netherton._

jamesnetherton (approved):
Thanks for the quick turnaround - all three points are addressed.

Filtering out the null `versionId` entries before the sort is nicer than the `nullsFirst` comparator I suggested: a version-less entry can no longer be selected at all, so if one ever shows up the existing `assertNotNull(oldVersion, ...)` fails with a readable message instead of an NPE.

CI is green. LGTM.

---
_This review was AI-generated, on behalf of James Netherton._

PR #9126, opened 2026-09-07
Let jgroups components and tests using quarkus-artemis use the same version of jgroups

ppalaga:


ppalaga:
A new attempt. @jamesnetherton could you please review? I am not quite sure introducing a new support extension is the best way forward, while https://redhat.atlassian.net/browse/JGRP-3041 is open.

jamesnetherton:
> A new attempt. @jamesnetherton could you please review? I am not quite sure introducing a new support extension is the best way forward, while https://redhat.atlassian.net/browse/JGRP-3041 is open.

@ppalaga I guess it's probably all we can do for now and 3.40.x.

For the version property, not sure if we want to use `${jgroups-version}` from `camel-dependencies` or `@sync` it from somewhere else?

ppalaga:
5.5.7.Final is the newest. I tought, we are more flexible with that one, when https://redhat.atlassian.net/browse/JGRP-3041 is fixed, or when there are CVEs. But I am open to change it to `${jgroups-version}` which is 5.5.6.Final ATM.

jamesnetherton:
@ppalaga Actually now I think more. I'm not sure adding a jgroups support extension is a good move from a UX perspective. Anyone wanting to use the Artemis JMS client will need to add it manually to their apps. Which I don't much like TBH.

ppalaga:
> Anyone wanting to use the Artemis JMS client will need to add it manually to their apps. Which I don't much like TBH.

Yeah, I agree. What else can we do? Have it in camel-quarkus-core and activate it anytime jgroups is in class path?

jamesnetherton:
> What else can we do? Have it in camel-quarkus-core and activate it anytime jgroups is in class path?

Yes, I think that would be a fair compromise.

ppalaga:
555f144:

* Uses `<jgroups.version>${jgroups-version}</jgroups.version>`
* JGroups native config moved to cq-core

jamesnetherton (approved):

ppalaga:
Thanks, I need to port this to 3.33

PR #9132, opened 2026-09-08
Move AI tool specific code into the ai-tool extension

jamesnetherton:


extensions/ai-tool/deployment/pom.xml
```diff
@@ -38,6 +38,10 @@
             <groupId>org.apache.camel.quarkus</groupId>
             <artifactId>camel-quarkus-ai-tool</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
```
JiriOndrusek: non-optional quarkus-jackson drags Jackson into every ai-tool app (extensions/ai-tool/runtime/pom.xml:48-51, plus quarkus-jackson-deployment in the deployment pom). I verified camel-ai-tool 4.22.0 depends only on camel-support — no Jackson anywhere. The injected ObjectMapper is used solely in CamelAiToolProvider.parseArguments(), i.e. only when the bridge is active, and the bridge can only activate when quarkus-langchain4j-core is present — which itself has a hard quarkus-jackson dependency (verified in its 1.13.x pom). So marking quarkus-jackson <optional>true</optional> (and dropping quarkus-jackson-deployment) is safe in every scenario and keeps the extension true to its own description, "framework-agnostic".

JiriOndrusek (commented):
I found some small issues and one nitpick:

```
@CamelAiTools present without quarkus-langchain4j now produces no diagnostic at all (the whole step is skipped); previously the support module warned.

```

extensions/ai-tool/deployment/src/main/java/org/apache/camel/quarkus/component/ai/tool/deployment/AiToolProcessor.java
```diff
@@ -16,15 +16,105 @@
...
+            BuildProducer<AdditionalBeanBuildItem> additionalBeans,
+            AiToolRecorder recorder) {
+
+        IndexView index = combinedIndex.getIndex();
+        Map<String, String> tagMap = new HashMap<>();
+        for (AnnotationInstance annotation : index.getAnnotations(CAMEL_AI_TOOLS_DOTNAME)) {
+            if (annotation.target().kind() == AnnotationTarget.Kind.CLASS) {
+                String className = annotation.target().asClass().name().toString();
+                if (annotation.value() == null) {
+                    LOG.warnf("@CamelAiTools on %s has no value — skipping", className);
```
JiriOndrusek: @CamelAiTools with no value (value() has default "") or a blank value is skipped with only a build-time warning, and the service then receives all registered tools — the opposite of the annotation's restrictive intent.

extensions/ai-tool/deployment/pom.xml
jamesnetherton: Agreed, and confirmed the reasoning end to end: `camel-ai-tool` pulls only `camel-support`, the `ObjectMapper` is used solely in `CamelAiToolProvider.parseArguments()`, that bean is registered only when Quarkus LangChain4j is present, and `quarkus-langchain4j-core` declares a non-optional `quarkus-jackson`.

Fixed in 50a5464: `quarkus-jackson` is now `<optional>true</optional>` in the runtime pom and `quarkus-jackson-deployment` is dropped from the deployment pom. Verified with `integration-tests/ai-tool` (no Quarkus LangChain4j, so Jackson genuinely absent) in both JVM and native mode, 12/12 each. Also checked `integration-tests/mcp-server`, the other module depending on `camel-quarkus-ai-tool` without declaring Jackson itself — it gets `quarkus-jackson` from `quarkus-mcp-server`, so it is unaffected.

extensions/ai-tool/deployment/src/main/java/org/apache/camel/quarkus/component/ai/tool/deployment/AiToolProcessor.java
jamesnetherton: Good catch — a blank tag left `CURRENT_TAG` unset, so `provideTools()` fell through to `getAllTools()` and the service received every registered tool, which is the opposite of what the annotation asks for.

Fixed in 50a5464: it is now a build failure rather than a warning, produced as a `ValidationErrorBuildItem` with a `ConfigurationException` to match `validateRagRetrievalFilterSupplier` in the support module:

```
@CamelAiTools on ...AdminAiService has no tag. An empty tag would expose every registered
Camel AI tool to the service instead of the intended subset. Give it the tag used by the
ai-tool: routes it should see, or remove the annotation to receive all tools deliberately.
```

Verified by temporarily setting `@CamelAiTools("")` on a test service — augmentation fails with that message. `value()` keeps its `""` default because the interceptor binding declaration on `CamelAiToolsInterceptor` needs it, so validation is where the check belongs.

jamesnetherton:
Thanks for the review — all three points are addressed in 50a5464, along with the CI failure.

**On `@CamelAiTools` without Quarkus LangChain4j:** the premise isn't quite right, so noting it for the record. The old warning lived in `validateAndRegisterAiServices` inside a class annotated `@BuildSteps(onlyIf = QuarkusLangchain4jPresent.class)` (line 86 of the pre-PR `SupportQuarkusLangchain4jProcessor`), so without Quarkus LangChain4j there was no diagnostic before this PR either. Nothing regressed there.

The gap you're pointing at is real, though, and it is cheap to close now that `@CamelAiTools` lives in this extension. Diagnostics moved into a `validateCamelAiToolsUsage` step with **no** `onlyIf` gate — gating on the very things that may be missing is what silenced the report — so:

* `@CamelAiTools` without Quarkus LangChain4j now warns that the annotation has no effect
* the existing "`camel-langchain4j-agent` missing" warning still fires as before
* `registerCamelAiToolProvider` is reduced to just registering the bean

**Separately, the CI failure** was a genuine regression from this PR: moving the bridge into `camel-quarkus-ai-tool` means it is registered only when that extension is a dependency, and `integration-tests/langchain4j-agent-ql4j` was relying on getting it transitively via `camel-quarkus-support-langchain4j`. It silently lost its Camel tools, so the request went out with `"tools": []` and WireMock rejected it. The module now declares `camel-quarkus-ai-tool` (which it should, since it uses `ai-tool:` routes directly), and a startup warning names the affected tools when tools are registered without the bridge, so this cannot fail silently again. The migration note is corrected accordingly.

**Verification:** `ai-tool` and `langchain4j-agent-ql4j` pass in both JVM and native mode; `ai-tool-langchain4j`, `langchain4j-agent`, `langchain4j-agent-bean-binding-ql4j` and `mcp-server` pass in JVM mode; full reactor `clean install -Dquickly` is green. The startup warning was confirmed firing from a native binary by temporarily removing the dependency again.

JiriOndrusek (approved):
Verified all three points on 50a5464:

- **Missing diagnostic** — the unconditional `validateCamelAiToolsUsage` step now covers both absence cases, and the extra runtime warning in langchain4j-agent plus the migration guide note cover the moved bridge registration. More thorough than I asked for.
- **Jackson** — confirmed `quarkus-jackson` optional in the runtime pom and `quarkus-jackson-deployment` dropped; re-ran `integration-tests/ai-tool` in JVM mode, 12/12.
- **Blank tag** — confirmed behaviorally: `@CamelAiTools("")` on `AdminAiService` fails augmentation with the documented `ConfigurationException` message; a valid tag builds fine.

One non-blocking follow-up idea: a functional extension test asserting the blank-tag `ConfigurationException`, so that behavior stays locked in.

Thanks! LGTM.

PR #9133, opened 2026-09-08
Move the RAG augmentor bridge into the langchain4j-embeddingstore extension

jamesnetherton:
The bridge producing RetrievalAugmentor beans and binding @EmbeddingStoreName CDI beans into the Camel registry lived in camel-quarkus-support-langchain4j, so it was active for all six extensions depending on that artifact. Its default augmentor was produced from an EmbeddingStore and an EmbeddingModel bean alone, so it could switch RAG on for every @RegisterAiService in applications where no Camel component writes to a store. It is now registered when camel-quarkus-langchain4j-embeddingstore is a dependency, which camel-quarkus-langchain4j-ingest now depends on: the two extensions that fill stores.

* Move RagRetrievalFilterSupplier, RagAugmentorName, RagBridgeConfig and DefaultRetrievalAugmentorSupplier to the extension runtime, and split the RAG half of QuarkusLangchain4jRecorder into Langchain4jEmbeddingstoreRecorder
* Move the six RAG build steps, BeanCensus and AugmentorDefinition into Langchain4jEmbeddingstoreRagProcessor, along with EasyRagPresent and its unit test. SupportQuarkusLangchain4jProcessor drops from 435 to 152 lines
* Make camel-quarkus-langchain4j-ingest depend on camel-quarkus-langchain4j-embeddingstore, since ingestion pipelines fill stores through CDI beans without the embeddingstore component, and keep the bridge for them as in 3.39.0
* Document the per-service opt-out, RegisterAiService.NoRetrievalAugmentorSupplier
* Document the narrowed activation in the 3.40.0 migration guide, covering configured augmentors, the default augmentor and AgentConfiguration.withRetrievalAugmentor, along with the RagRetrievalFilterSupplier package change


JiriOndrusek (changes requested):
Nice split — the moved build steps, the optional `quarkus-langchain4j-core`, and the migration  guide entry all look right, and I ran the rag-bridge ITs against the branch: 20/20.

  My one real concern is that the upgrade failure mode is silent. Two groups of users lose RAG without any signal:

  1. Apps configuring `quarkus.camel.langchain4j.rag.augmentors.*` — the properties just go inert (Quarkus's "unrecognized key" warning names no remedy, and nobody reads it).
  2. Apps relying on the zero-config default augmentor: quarkus-langchain4j + a store + a model bean + `@RegisterAiService`. After the upgrade the AI service keeps answering, just without the knowledge base. No error, no warning — the bot just gets quietly dumber, and the migration guide doesn't explicitly name this case.

  Could we make the loss loud, the same way you did for `@CamelAiTools` in #9132?

  - In support-langchain4j (which all six extensions still have): fail the build with a
  `ValidationErrorBuildItem` when `quarkus.camel.langchain4j.rag.*` keys are present but the bridge classes aren't on the classpath, naming the dependency to add.
  - Same place: log a warning when the old auto-activation conditions hold (ql4j + AI services + store + model beans, no user augmentor) and the bridge is absent — every app matching that was getting an augmentor before this PR.
  - One added sentence in the migration guide naming the default-augmentor case explicitly, and a matching update to the langchain4j-ingest usage doc, which still promises the bridge "without glue code".

  Small nit: the PR text says no Camel component consumes a `RetrievalAugmentor` — the langchain4j-agent API does accept one programmatically (`AgentConfiguration`), so agent users injecting a named bridge augmentor are also in group 1.

  _This review was generated with AI assistance and may contain inaccuracies. Please verify all  suggestions before applying._

  _Claude Code on behalf of Jiří Ondrušek_

jamesnetherton:
Thanks @JiriOndrusek. Digging into your points turned up more than a missing warning, so here's a proposal before I push anything.

What changed my view:

- Before this PR, the zero-config augmentor fired on "≥1 `EmbeddingStore` + ≥1 `EmbeddingModel` bean" in any app using one of the six extensions that depend on `support-langchain4j`. Apart from ingest, none of them fill a store, so RAG could switch on without being asked for.
- `langchain4j-ingest` is a legitimate bridge user. It fills stores through CDI beans without the embeddingstore component, and its docs promise the bridge. You're right on the agent nit too: `AgentConfiguration.withRetrievalAugmentor` exists. I'll fix the PR text.

Proposal:

1. **All RAG code stays in `langchain4j-embeddingstore`, auto-detection included and unchanged.** Scoped to extensions that fill stores, zero-config is what a user adding them expects, the same model as Easy RAG. An AI service can opt out with `@RegisterAiService(retrievalAugmentor = RegisterAiService.NoRetrievalAugmentorSupplier.class)`, which I'll document.
2. **`langchain4j-ingest` depends on `langchain4j-embeddingstore`**, so ingest apps keep the bridge exactly as in 3.39.0. The cost is `camel-langchain4j-embeddingstore` and `langchain4j-embeddings` on ingest's classpath.
3. **Nothing RAG-related goes back into `support-langchain4j`.** Detecting the bridge's absence there needs a copy of the bean census, the Easy RAG check and the config prefix, which puts back the code this PR removes.
4. **Who changes:** apps on chat/agent/tools/web-search/embeddings without embeddingstore or ingest. Zero-config ones stop receiving an augmentor over a store no Camel component wrote to. Ones configuring `rag.*` get Quarkus's unrecognized-key warning, and the migration guide tells them to add embeddingstore.

Does that address your concern? The part I'd most like your view on is 4: whether the unrecognized-key warning plus the guide is enough there.

jamesnetherton:
The other thing to consider is that most of this functionality is new and somewhat experimental. So any migration path are somewhat easier given there's likely to be limited use of this functionality in our user base.

JiriOndrusek:
I think that the mentioned approach covers all issues.

> The other thing to consider is that most of this functionality is new and somewhat experimental. So any migration path are somewhat easier given there's likely to be limited use of this functionality in our user base.

I agree

jamesnetherton:
Thanks @JiriOndrusek. The proposal is implemented and the branch is rebased on main:

- `9ccef98356d` moves the bridge, unchanged apart from the rebase.
- `b42c13ae778` makes `langchain4j-ingest` depend on `langchain4j-embeddingstore`, documents the `NoRetrievalAugmentorSupplier` opt-out, and names the affected applications in the migration guide.

Locally, `langchain4j-rag-bridge-ql4j` (20/20), `langchain4j-embeddingstore-ql4j` and `langchain4j-ingest` (JVM and native) pass. Ready for another look.

JiriOndrusek (approved):
Verified against the agreed proposal:

- `9ccef98356d` is unchanged apart from the rebase (range-diff shows only migration-guide context drift from #9144). The moved classes are verbatim: package renames, the javadoc-to-asciidoclet conversion in `RagBridgeConfig`, and renamed processor/recorder references are the only deltas — the `quarkus.camel.langchain4j.rag` config prefix and all logic are untouched. `QuarkusLangchain4jPresent` is a faithful copy (inlined class-name constant), with the original staying in support for its remaining steps.
- `b42c13ae778` covers everything agreed: the ingest -> embeddingstore dependency (runtime + deployment), the `NoRetrievalAugmentorSupplier` opt-out in the docs, the migration guide naming all three affected groups (`rag.augmentors.*`, the default augmentor, `AgentConfiguration.withRetrievalAugmentor`), and the ingest docs linking the bridge. The PR text nit is fixed too.

Ran locally: rag-bridge-ql4j 20/20, ingest deployment tests 23/23, ingest JVM ITs 10/10 (Kafka/S3 excluded — my container runtime is unwell today, so embeddingstore-ql4j and native are trusted to your run and CI).

Optional follow-up, not blocking: a small `@RegisterAiService` check in a ql4j IT would assert that an ingest-only app keeps the bridge transitively, guarding the new dependency against accidental removal.

LGTM, thanks!

_This review was generated with AI assistance and may contain inaccuracies. Please verify all suggestions before applying._

_Claude Code on behalf of Jiří Ondrušek_

Issue #9091, opened 2026-09-02
camel-quarkus-pdf: register PDFBox's resource directory instead of a hand-listed constant


JiriOndrusek:
`camel-quarkus-pdf` registers PDFBox's native resources from a hand-listed constant, `PdfProcessor.RUNTIME_RESOURCES`. It has drifted: it names `icc/ISOcoated_v2_300_bas.icc`, which PDFBox 3.0.8 replaced with `icc/CGATS001Compat-v2-micro.icc` — so it registers a resource not in the jar and misses the real ICC profile (latent native failure for PDFs that load the default output-intent profile).

Fix: replace the list with `NativeImageResourceDirectoryBuildItem("org/apache/pdfbox/resources")`, so native image embeds whatever the jar ships and the set self-heals on upgrades. See #9078 for the same change in `langchain4j-ingest` as an example.

Issue #9087, opened 2026-09-02
Move PDFBox/Tika PDF native-image configuration from langchain4j-ingest to camel-quarkus-tika


JiriOndrusek:
PR #9078 added PDFBox runtime-init/resource registrations and the angus-activation feature exclusion to the langchain4j-ingest deployment processor, classpath-gated on PDFBox. Any camel-quarkus-tika + tika-parser-pdf-module application would benefit from the same configuration, so it should be homed in camel-quarkus-tika (with native coverage using tika-parser-pdf-module in its integration tests) and the ingest processor slimmed accordingly. See the review discussion on #9078.

JiriOndrusek:
One more gap this should close:
  neither camel-quarkus-tika's reference page nor Camel's tika component page currently documents that Tika parser modules are user-supplied per format (camel-tika itself ships only the html and text modules). When the native PDFBox configuration moves here, the extension should also gain a usage.adoc stating that contract and which parser modules the native mode covers.

PR #9143, opened 2026-09-10
[camel-main] Fixes #9142. Delegate langchain4j-ingest to the camel component

### LINKED ISSUE #9076: langchain4j-ingest: document parsers for binary formats (tika, docling)

### LINKED ISSUE #9142: Delegate the langchain4j-ingest extension to the camel-langchain4j-ingest component
### Description

  Apache Camel 4.23.0 introduces the `camel-langchain4j-ingest` component (CAMEL-24621), which provides the same document-ingestion engine this extension has been carrying internally since 3.39.0: splitting, embedding, storing and per-document-id deduplication behind an idempotent repository.

  With the component available upstream, the extension should stop maintaining its own copy of that engine and drive the component instead. The extension remains the Quarkus-side declaration surface — the `quarkus.camel.langchain4j.ingest.*` configuration model, the `@Ingest` builder API, build-time validation, CDI bean resolution and native-image support — while each generated pipeline route ends in the component's producer.

JiriOndrusek:
Fixes #9142
Fixes #9076
Supersedes #9078

Delegates the extension to the `camel-langchain4j-ingest` component introduced in Camel 4.23.0 (CAMEL-24621): the extension's own engine (`core/IngestService`) is deleted, each pipeline route now ends in the component's producer, and the Quarkus declaration surface (`quarkus.camel.langchain4j.ingest.*`, `@Ingest`, CDI resolution, build-time validation, native support) stays. The route topology is generated by a package-private pipeline assembly
  (`IngestPipelineDefinition`, `IngestPipelineRouteBuilder`) around the producer — planned to be replaced by the langchain4j-ingest Kamelets in a follow-up once camel-kamelets ships them.

  Also fixes #9076 and **supersedes #9078**: the tika/docling parser support is implemented here against the delegated engine (`parser` option, pre-parse document-id capture, raw-byte size guard), together with `embedding-batch-size`, `document-splitter` and `max-document-size`. #9078 will be closed in favour of this PR.

  Behaviour changes, documented in the new 4.0.0 migration guide:
  - route ids: `camel-quarkus-langchain4j-ingest-<name>` → `langchain4j-ingest-<name>`
  - the missing-document-id failure is worded by the producer now (it names the documented `CamelIngestDocumentId` header — the extension points the producer's `documentIdHeader` option at it)
  - the reply type is the component's `IngestResult` (compile-time import rename)
  - the metadata keys already moved to `camel_ingest_*` on main (#9144); nothing further changes there

  The `CamelIngestDocumentId` header default stays pinned, so existing routes keep working.

  Known follow-ups: the PDFBox/angus native-image build steps belong in camel-quarkus-tika (#9087) with camel-quarkus-pdf aligned to them (#9091) — marked with TODOs.

  ⚠️  Merge gate: targets `camel-main` and requires `camel-langchain4j-ingest` from the Camel 4.23.0-SNAPSHOT — mergeable only once Camel Quarkus is on Camel 4.23.

  Verified: 30 deployment test suites, 18 JVM and 18 native integration tests green (Kafka, S3/MinIO, WireMock-stubbed Docling Serve, pre-parse size-guard test).

  One honesty note before you paste it: the "Verified" line reflects the last full test pass — several later rounds (dead-overload removal, bounded stream guard, documentIdHeader, guide move) are compile-verified only. Either rerun the suites before pushing or let PR CI stand behind that line.


jamesnetherton:
We probably need to start a migration guide for 4.0.0. That'd be the correct place for migration notes relating to this change.

JiriOndrusek:
> We probably need to start a migration guide for 4.0.0. That'd be the correct place for migration notes relating to this change.

I agree, I'll start new migration guide doc for 4.0.0

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestPipelineRouteBuilder.java
```diff
@@ -0,0 +1,390 @@
...
+    /** The route: consume → resolve the id → (parse) → split, embed, store; the reply is the result. */
+    private void consumerRoute(IngestPipelineDefinition pipeline, String storeRef, String modelRef, String splitterRef) {
+        String scheme = StringHelper.before(pipeline.uri(), ":");
+        requireComponent(scheme, pipeline, "its source");
+        String registerRef = repositoryRef(pipeline, false);
+        Expression documentId = documentIdExpression(pipeline.documentId(), LangChain4jIngestHeaders.DOCUMENT_ID);
+
+        ProcessorDefinition<?> route = from(pipeline.uri())
+                .routeId(routeId(pipeline))
+                .setProperty(LangChain4jIngest.DOCUMENT_ID_PROPERTY, documentId);
```
jamesnetherton, line 157: (non-blocking) If `documentId` evaluates to null here, `setProperty` removes the property instead of setting it. The producer's `resolveDocumentId` then falls back to the id header, but by then Tika has copied document metadata into headers.

Example: a consumer pipeline with `parser=tika` gets a delivery with no id header (a keyless Kafka record with `documentId=CamelKafkaKey`, or a `direct:` call without `CamelIngestDocumentId`). If the payload contains `<meta name="CamelIngestDocumentId" content="victim">`, it is ingested under `victim`. The previous engine failed that delivery. The spoofing test only covers the case where the header is present.

Directory pipelines whose `document-id` resolves to null, and simple-expression pipelines that fall back to `CamelLangChain4jIngestDocumentId`, work the same way. Could the route fail the exchange before the parse when the captured id is null or blank?

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestPipelineRouteBuilder.java
```diff
@@ -0,0 +1,390 @@
...
+        route = parseSteps(route, pipeline);
```
jamesnetherton, line 158: (non-blocking) On consumer pipelines with `idempotent-repository`, the Tika or Docling parse runs before the `langchain4j-ingest` producer claims the id. So every duplicate is parsed in full, including a remote Docling Serve conversion with OCR, only to be answered `SKIPPED`. That goes against the producer's note that a skipped duplicate never pays for materialising a large payload.

Example: an `aws2-s3` source with `deleteAfterRead(false)`, a register and `parser=docling` re-lists every object on each poll, so every object is downloaded and sent to docling-serve on every poll. Could the id be checked or claimed before `parseSteps`?

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestPipelineRouteBuilder.java
```diff
@@ -0,0 +1,390 @@
...
+                // bytes instead of being materialized whole in the heap just to be measured;
+                // an accepted stream is consumed here, so the bytes replace it as the body
+                int limit = pipeline.maxDocumentSize() == Integer.MAX_VALUE
+                        ? Integer.MAX_VALUE
+                        : pipeline.maxDocumentSize() + 1;
+                bounded = stream.readNBytes(limit);
+                size = bounded.length;
+            } else {
+                // a null body carries no bytes to guard; it flows on and becomes the EMPTY outcome
+                byte[] body = exchange.getMessage().getBody(byte[].class);
```
jamesnetherton, line 223: (non-blocking) Only `InputStream` bodies get the bounded read. Other body types go through `getBody(byte[].class)`, which loads the whole payload into memory just to measure it, and the parser then reads it again.

Example: the documented `source.uri=file:/inbox?delete=true` pipeline with `parser=tika` and `max-document-size=10000000`. It is a consumer pipeline, so `CamelFileLength` isn't trusted and the `GenericFile` body is converted in full. A multi-GB file can exhaust the heap (or go past the 2 GB array limit) before the guard rejects it, which is what the guard is meant to prevent. Converting to `InputStream` first and reusing the bounded path would avoid that.

jamesnetherton (commented):
A few non-blocking observations on `IngestPipelineRouteBuilder`. None of them need to hold up the merge; they could be handled here or in a follow-up.

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestPipelineRouteBuilder.java
JiriOndrusek: Fixed in the amended commit: a parser pipeline now fails the exchange before the parse when the captured id resolved to nothing — same behaviour as the previous engine — with a test (`consumerDeliveryWithoutIdFailsBeforeTheParse`). Worth noting camel-tika filters `Camel*` metadata names since CAMEL-24423, so the `CamelIngestDocumentId` forgery specifically is blocked either way — but a custom non-Camel header name was genuinely forgeable, and the null-id delivery silently ingesting was a regression. Both closed by the pre-parse guard.

_Claude Code on behalf of @JiriOndrusek_

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestPipelineRouteBuilder.java
JiriOndrusek: Fixed in the amended commit: consumer pipelines with a register now answer a known duplicate `SKIPPED` before the parse via an advisory `repository.contains(id)` check — the producer's eager claim stays authoritative, so a duplicate racing the check is still caught there; the check only saves the parse cost, which covers the S3 re-listing scenario.

_Claude Code on behalf of @JiriOndrusek_

extensions/langchain4j-ingest/runtime/src/main/java/org/apache/camel/quarkus/component/langchain4j/ingest/IngestPipelineRouteBuilder.java
JiriOndrusek: Fixed in the amended commit: the guard now measures every body through `getBody(InputStream.class)` and the bounded read — a `GenericFile` streams from disk and a `byte[]` merely wraps, so nothing is materialized beyond `maxDocumentSize + 1` bytes; null bodies keep flowing to the EMPTY outcome.

_Claude Code on behalf of @JiriOndrusek_

JiriOndrusek:
All issues are fixed (fix is amended) and the branch is rebased

jamesnetherton (approved):

JiriOndrusek:
Hand pushed into camel-main - see https://github.com/apache/camel-quarkus/commit/0d00739a01eb03f4c73012a5f7085730517c0cb8

PR #9154, opened 2026-09-11
Fixes #9153. Native build fails on Kamelets because of Jackson

### LINKED ISSUE #9153: Native build fails when the kamelet DelegatingSchemaResolver becomes reachable without camel-jackson
`camel-kamelet` ships `org.apache.camel.component.kamelet.utils.format.schema.DelegatingSchemaResolver`, which hard-references the schema resolvers of the *optional* `camel-jackson`, `camel-jackson-avro` and `camel-jackson-protobuf` dependencies.

When that class becomes reachable in a native build — e.g. an application combining `camel-quarkus-kamelet` with a scheduled batch polling consumer such as `file` — and any of those dependencies is absent, the GraalVM analysis fails:

```
UnresolvedElementException: Error loading a referenced type:
java.lang.NoClassDefFoundError: org/apache/camel/component/jackson/transform/JsonSchemaResolver
```

The existing kamelet integration tests never make the class reachable, so CI does not catch it.

_Claude Code on behalf of @JiriOndrusek_

JiriOndrusek:
Fixes #9153. 

`camel-kamelet` ships `DelegatingSchemaResolver`, which hard-references the schema resolvers of the *optional* `camel-jackson`, `camel-jackson-avro` and `camel-jackson-protobuf` dependencies. The class is the `#class:` template bean of the catalog's `resolve-pojo-schema-action` kamelet — whenever that kamelet is embedded into the native image (any application with the default `quarkus.camel.kamelet.identifiers=*`), the deployment registers the class reflectively and the GraalVM analysis fails parsing it if any of the optional dependencies is absent: `UnresolvedElementException: NoClassDefFoundError: org/apache/camel/component/jackson/transform/JsonSchemaResolver`.

  `KameletProcessor` now removes the class from the `camel-kamelet` artifact for native builds unless all three resolver classes are on the classpath (`RemovedResourceBuildItem`, the `OpenaiProcessor` idiom). JVM mode is unaffected — the class was never loadable without the optional dependencies anyway.

  Regression guard: the kamelet integration tests now embed `resolve-pojo-schema-action` — without the fix their native build fails — and gain a file-backed source kamelet test along the way.

  Verified: kamelet integration tests green in JVM and native mode; native fails without the fix.

  _Claude Code on behalf of @JiriOndrusek_


jamesnetherton (commented):
Sorry to nitpick here - I think it'd be more correct to fix the root cause. Which is registering `DelegatingSchemaResolver` for reflection when it does not need to be.

So instead of using `RemovedResourceBuildItem`, we have a check in the reflective class registration like:

```
Set<String> kameletBeanClasses = resolveKameletBeanClasses(kameletResources);
if (!isClassPresentAtRuntime(JSON_SCHEMA_RESOLVER)
        || !isClassPresentAtRuntime(AVRO_SCHEMA_RESOLVER)
        || !isClassPresentAtRuntime(PROTOBUF_SCHEMA_RESOLVER)) {
    kameletBeanClasses.remove(DELEGATING_SCHEMA_RESOLVER);
}
```

JiriOndrusek:
Reworked as suggested — the RemovedResourceBuildItem step is gone; KameletProcessor now drops  DelegatingSchemaResolver from the reflective kamelet bean registration when any of the three optional resolver classes is missing. Re-verified: kamelet integration tests green in JVM and native mode.

jamesnetherton (approved):

Issue #9124, opened 2026-09-07
Add a dedicated support extension for Quarkus LangChain4j build time integration


jamesnetherton:
Currently the Quarkus LangChain4j build time integration is added to camel-quarkus-support-langchain4j. While this works, it causes issues if you want to depend on types that are only found within QL4J. See [here](https://github.com/apache/camel-quarkus/pull/8946#discussion_r3813088573) for an example of this.

We could add a `camel-quarkus-support-langchain4j-ql4j` support extension carrying a [dependency condition](https://quarkus.io/guides/conditional-extension-dependencies) on `io.quarkiverse.langchain4j:quarkus-langchain4j-core`, registered as a conditional dependency of `camel-quarkus-support-langchain4j`. Its deployment module can then depend on`quarkus-langchain4j-core-deployment` directly, and applications get it only when Quarkus LangChain4j is on the classpath.

PR #9161, opened 2026-09-14
Add camel-quarkus-support-langchain4j-ql4j extension

### LINKED ISSUE #8836: Quarkus LangChain4j ChatMemoryProvider silently leaks into AgentWithoutMemory when default model is configured

jamesnetherton:
Relates to #9124.

This change opens the ability to fix #8836. So I pulled the commit from #8946.


JiriOndrusek (approved):

jamesnetherton:
Setting to draft while I investigate what the consequences are of `ExcludeFromImpliedAiServiceBuildItem`. Currently the agents without memory are excluded. But stateful agents are not. So there's potentially a bit of a mismatch in terms of behaviour with Camel & QL4J.

jamesnetherton:
I'll merge this as-is.

I created https://github.com/apache/camel-quarkus/issues/9172 as something we might want to follow up and think about.

JiriOndrusek (commented):
sorry, wrong ticket

PR #9201, opened 2026-09-21
Fixes #9200. Build Camel with JDK 25 in the camel-main CI workflows

### LINKED ISSUE #9200: [camel-main] SedaVirtualThreadsIT fails in native mode: CI-built Camel lacks the Java 21 multi-release classes
### Bug description

`SedaVirtualThreadsIT.sedaExecutesOnVirtualThread` (foundation-grouped, native group-03) fails on the `camel-main` branch since the tests were re-enabled by #9189.

```
Expected: a string containing "java.lang.VirtualThread"
  Actual: java.lang.Thread
```

CI run: https://github.com/apache/camel-quarkus/actions/runs/35573093554

The same test passes on `main`. The difference is the Camel artifact, not the Camel code: `main` uses the released Camel jars (built on JDK 25, `Multi-Release: true`, containing `META-INF/versions/21/.../CamelThreadFactory.class`), while the camel-main CI builds Camel from source on JDK 17. Camel compiles its Java 21/25 sources only through JDK-activated profiles (`java-21-sources`, `java-25-sources` in [core/camel-util/pom.xml](https://github.com/apache/camel/blob/main/core/camel-util/pom.xml)), so the JDK 17 build produces a jar with the base [CamelThreadFactory](https://github.com/apache/camel/blob/main/core/camel-util/src/main/java/org/apache/camel/util/concurrent/CamelThreadFactory.java) only, which always creates platform threads. The virtual-thread variant lives in [src/main/java21](https://github.com/apache/camel/blob/main/core/camel-util/src/main/java21/org/apache/camel/util/concurrent/CamelThreadFactory.java).

Reproduced locally: with camel-util built on JDK 25 the native IT passes; with the two profiles disabled (equivalent to a JDK 17 build) it fails with the CI message. The nightly `camel-master-cron.yaml` builds Camel on JDK 17 as well.

jiriOndrusek-agent:
Fixes #9200

**Failed run:** https://github.com/apache/camel-quarkus/actions/runs/35573093554 (`camel-main`, Native Tests - group-03)

### Problem
`SedaVirtualThreadsIT` fails on `camel-main` because the CI builds Camel from source on JDK 17. Camel compiles its Java 21/25 multi-release sources (e.g. the virtual thread aware `CamelThreadFactory`) only when built on JDK 21+ (`java-21-sources` / `java-25-sources` profiles in `core/camel-util/pom.xml`), so the CI-built jars lack those classes, unlike the released Camel jars used on `main`.

### Change
Set up JDK 25 only for the "Build Camel" step and switch back to JDK 17 right after it, in both `ci-build.yaml` (guarded by the existing camel-main condition, so inert on `main` and PR builds) and `camel-master-cron.yaml`. The JDK 17 baseline of camel-quarkus is unchanged.

### Verification
Locally the native `SedaVirtualThreadsIT` passes with camel-util built on JDK 25 and fails with the CI message when the Java 21/25 profiles are disabled (equivalent to a JDK 17 build). actionlint reports nothing new for either workflow.


JiriOndrusek:
wrong PR

